PR #3, opened 2015-09-09
Vagrant config and Makefile improvements

brousch:
These are changes to allow a dev instance to be run via Vagrant. some new Makefile commands are included and the README has been updated to reflect the changes. Some changes that may raise eyebrows are:

.env.vagrant:
I'm not sure about all of the settings in the .env file, but this is sufficient for running a dev instance using SQLite.
Line 3: The DB needs to be kept out of the repo due to Windows issues in Vagrant if it's in the repo.

Makefile:
Line 5: honcho run ./manage.py is not reliable when manage.py is not executable, which can happen when using Vagrant on Windows. Using the full honcho run python ./manage.py everywhere avoids this issue.
Line 9: If they are running Vagrant on Windows, line endings can be a problem here. The new code here strips off the troublesome ^M should it appear.

Procfile.dev:
Line 3: We cannot expose only 127.0.0.1:5000 because the port will not be forwarded to the host in Vagrant, so we run on 0.0.0.0:5000

Vagrantfile:
Line 25: Port 80 is forwarded in anticipation of production working, thought this is not yet complete.

opencraft/settings.py:
Line 39: This is really a bug fix. The instructions mention that you can use a .env file, but the necessary code for reading that file was not present.


antoviaque:
@brousch Thank you for the PR! That looks like a very useful contribution to run this, especially for people not using Ubuntu or Debian. I'll schedule a review for it in our next sprint, ie likely next week.

The build seem to fail because of missing auth, I think this is because the PR is from a fork - I'll need to look into this, not sure if there is a secure way to handle these cases without making the auth potentially publicly available.

Makefile
```diff
@@ -1,6 +1,16 @@
 # Config
 WORKERS = 4
 SHELL = /bin/bash
+MANAGE := python ./manage.py
```
antoviaque: @brousch There isn't really a need to call python explicitly here - it duplicates the shebang from `manage.py`. I would only keep `HONCHO_MANAGE := honcho run ./manage.py`

Makefile
```diff
@@ -1,6 +1,16 @@
...
+HONCHO_MANAGE := honcho run $(MANAGE)
+
+.PHONY: install_system_dependencies
```
antoviaque: @brousch Good point, all of the target are actually phony targets. Could you add it on the other ones too, for consistency?

README.md
```diff
@@ -6,18 +6,26 @@ Install
...
-Install the system package dependencies & virtualenv:
+If you're installing to your local Ubuntu 14.04:
 
 ```
-$ sudo apt-get install `cat debian_packages.lst`
-$ pip3 install --user virtualenv && pip3 install --user virtualenvwrapper
+$ make install_system_dependencies
 ```
 
-Ensure you load virtualenv with Python 3 in `~/.bashrc`:
```
antoviaque: @brousch You need to keep mentioning the need to add this to `.bashrc` for local (non-Vagrant) installs, this information is currently lost with your changes.

Also, for Vagrant, why not go all the way, and do this change in `.bashrc` on the Vagrant VM automatically, rather thatn asking the user to go do it manually himself?

Vagrantfile
```diff
@@ -0,0 +1,75 @@
...
+
+  # Disable automatic box update checking. If you disable this, then
+  # boxes will only be checked for updates when the user runs
+  # `vagrant box outdated`. This is not recommended.
+  # config.vm.box_check_update = false
+
+  # Create a forwarded port mapping which allows access to a specific port
+  # within the machine from a port on the host machine. In the example below,
+  # accessing "localhost:8080" will access port 80 on the guest machine.
+  config.vm.network "forwarded_port", guest: 80, host: 8080
```
antoviaque: @brousch The port 80 isn't used - on the other side, it's currently missing forwarding the websocket port, 2001 (see `DRAGON_SERVER_ADDRESS_PORT`)

Vagrantfile
```diff
@@ -0,0 +1,75 @@
...
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Define a Vagrant Push strategy for pushing to Atlas. Other push strategies
+  # such as FTP and Heroku are also available. See the documentation at
+  # https://docs.vagrantup.com/v2/push/atlas.html for more information.
+  # config.push.define "atlas" do |push|
+  #   push.app = "YOUR_ATLAS_USERNAME/YOUR_APPLICATION_NAME"
+  # end
```
antoviaque: @brousch Can you remove the commented lines?

Vagrantfile
```diff
@@ -0,0 +1,75 @@
...
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+   config.vm.provision "shell", inline: <<-SHELL
+     ln -s /vagrant /home/vagrant/opencraft
+     cd /home/vagrant/opencraft
+     sudo apt-get update
+     make install_system_dependencies
+     cp .env.vagrant .env
```
antoviaque: @brousch This will not be working out of the box, as it won't be able to authenticate to the services like GitHub or OpenStack, which are necessary to get anything working. I would also like to limit the number of `.env` files to maintain, as changes in configuration options will require to edit all of them every time. Would it be possible instead to use the `.env.test` file on the vagrant VM, since it will need to be edited in any case?

opencraft/settings.py
```diff
@@ -36,6 +36,7 @@
 
 env = environ.Env()
 root = environ.Path(os.path.dirname(__file__), os.pardir)
+environ.Env.read_env('.env')
```
antoviaque: @brousch Why is this needed? It should already be done by `environ.Env()`, no? Were you getting issues without it?

README.md
```diff
@@ -34,7 +42,8 @@ Configure
 ---------
 
 Create an `.env` file at the root of the repository or set environment variables, customizing the
-settings from `opencraft/settings.py` which are loaded via `env()`.
+settings from `opencraft/settings.py` which are loaded via `env()`. An example development .env is
+already installed if you're using Vagrant.
```
antoviaque: @brousch Cf my remark about using the `.env.test` file as an example, instead.

Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: ./manage.py runserver_plus
-websocket: ./websocket.py
-worker: ./manage.py run_huey --no-periodic
+web: python ./manage.py runserver_plus 0.0.0.0:5000
+websocket: python ./websocket.py
+worker: python ./manage.py run_huey --no-periodic
```
antoviaque: @brousch The `python` part is redundant, and specifying `0.0.0.0:5000` here would override the `RUNSERVERPLUS_SERVER_ADDRESS_PORT` config option. Could you change this in the config instead?

.env.vagrant
```diff
@@ -0,0 +1,16 @@
+DEBUG=true
+SECRET_KEY='dev'
+DATABASE_URL='sqlite:////home/vagrant/opencraft.db'
```
antoviaque: @brousch We should not use sqlite - there are too many differences between it and a proper postgresql DB. If people use it to develop, they will miss issues (like enforcing character limits on fields) because sqlite gnores them, while a proper DB doesn't.

Vagrantfile
```diff
@@ -0,0 +1,75 @@
...
+  # https://docs.vagrantup.com/v2/push/atlas.html for more information.
+  # config.push.define "atlas" do |push|
+  #   push.app = "YOUR_ATLAS_USERNAME/YOUR_APPLICATION_NAME"
+  # end
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+   config.vm.provision "shell", inline: <<-SHELL
+     ln -s /vagrant /home/vagrant/opencraft
```
antoviaque: @brousch Use `config.vm.synced_folder` rather than symlinking to the default share, to make this more explicit?

Vagrantfile
```diff
@@ -0,0 +1,75 @@
...
+     cd /home/vagrant/opencraft
+     sudo apt-get update
```
antoviaque: @brousch Maybe put this in `install_system_dependencies`?

Makefile
antoviaque: Oh, I missed your comment on this:

>Line 5: honcho run ./manage.py is not reliable when manage.py is not executable, which can happen when using Vagrant on Windows. Using the full honcho run python ./manage.py everywhere avoids this issue.

Sounds good then. Maybe add a comment in the Makefile to ensure people understand why this is necessary?

Vagrantfile
antoviaque: Also missed your comment there:

>Line 25: Port 80 is forwarded in anticipation of production working, thought this is not yet complete.

What do you mean by "production working"?

README.md
```diff
@@ -6,18 +6,26 @@ Install
...
+
+Install virtualenv and virtualenvwrapper for Python3
 
 ```
-export PATH="$PATH:$HOME/.local/bin" VIRTUALENVWRAPPER_PYTHON=/usr/bin/python3
-source $HOME/.local/bin/virtualenvwrapper.sh
+$ make install_virtualenvwrapper_py3
+$ export PATH="$PATH:$HOME/.local/bin" VIRTUALENVWRAPPER_PYTHON=/usr/bin/python3
+$ source $HOME/.local/bin/virtualenvwrapper.sh
 ```
```
antoviaque: @brousch Could also be worth adding a few lines about what to do to connect to the vagrant box (`vagrant ssh`) and get to the right directory (`cd opencraft`)

README.md
```diff
@@ -6,18 +6,26 @@ Install
...
 
 Then reload `~/.bashrc`, create the virtual env and install the Python requirements:
```
antoviaque: @brousch Could also be worth doing this as part of the vagrant provisioning.

antoviaque:
@brousch Thanks again for this PR! It's going to be very useful to add vagrant support.

antoviaque:
Re-run the tests from a branch from the main repo to get the auth - it passed correctly. @brousch when you will push new commits to your branch, it will fail again due to the missing auth on your fork, but I'll run them again when I re-review later.

Vagrantfile
antoviaque: @brousch Also, the copy should be conditional - testing if a `.env` file already exists, to avoid overriding an existing configuration (which happened to my configuration file when testing this PR : ) ).

PR #4, opened 2015-09-11
Add PR number to Github instances and show PR link in the Web UI.

smarnach:


instance/static/html/instance/index.html
```diff
@@ -50,7 +50,13 @@
...
                 <span>Studio: <a href="{{ selected.instance.studio_url }}" target="_new">{{ selected.instance.studio_url }}</a></span>
               </li>
-              <li><a href="{{ selected.instance.github_branch_url }}" target="_new">{{ selected.instance.github_base_url }}</a> | {{ selected.instance.branch_name }}</li>
+              <li>
+                <a href="{{ selected.instance.github_branch_url }}" target="_new">{{ selected.instance.github_base_url }}</a> {{ selected.instance.branch_name }}
+                {% if selected.instance.github_pr_url %}
+                | <a href="{{ selected.instance.github_pr_url }}" target="_new">PR#{{ selected.instance.github_pr_number }}</a>
+                {% endif %}
+              </li>
+              {% endif %}
```
antoviaque: @smarnach Is that showing up correctly on your development environment? This template isn't interpreted by the Django templating language, it uses the angular syntax, so `{% if ... %}` won't work - it shows this for me:

screenshot-localhost 5000-2015-09-12-08-53-20

Instead you need to use the `ng-if` directive, like in https://github.com/open-craft/opencraft/blob/smarnach/add-pr-link/instance/static/html/instance/index.html#L33

instance/tasks.py
```diff
@@ -66,6 +66,7 @@ def watch_pr():
             instance, created = OpenEdXInstance.objects.get_or_create(
                 sub_domain=pr_sub_domain,
                 fork_name=pr.fork_name,
+                github_pr_number=pr.number,
```
antoviaque: @smarnach Since this is in a `get_or_create()` call, adding this will cause all existing PR sandbox to be duplicated when pushing this... Maybe add setting the PR number on the existing instances in the migration?

instance/static/html/instance/index.html
smarnach: Oh dear, apparently I was quite tired on Friday night. :)  As mentioned in the PR description, I had some trouble getting the development server running.  I'll give it another try and send an update.

instance/tasks.py
smarnach: @antoviaque Good point, but I'm not sure about adding the PR number in the migration code.  I somehow don't like calling out to the Github API in migration code.  This feels too fragile to me.

I resorted to the very simple solution of setting the PR number a few lines further down.  The keys that are currently in the `get_or_create()` call are enough to uniquely identify the instance, so we can just unconditionally overwrite the PR number on the instance.

smarnach:
@antoviaque I pushed a fix for the issues you mentioned.  This time I verified it myself, since I now have a working development server. :)

instance/static/html/instance/index.html
antoviaque: @smarnach : )

Out of curiosity, what were the issues you had to setup the dev environment? If there are some gotchas other might run into, it could be worth adding a note or fixing them.

instance/tasks.py
antoviaque: @smarnach True, it would have to be done carefully in the migrations, to handle potential API call failures. But here your approach sounds good to me - it means that we wouldn't be able to have several sandboxes from a single branch, even if they have multiple PRs (even if they are successive) - but that might actually turn out to be a feature. We'll see.

antoviaque:
@smarnach Thanks for the fixes! Looks good now to me :+1:

instance/static/html/instance/index.html
smarnach: @antoviaque The problems were mostly with myself, not with the instructions in the README.  The README could be a bit more detailed on what to put in the `.env` file and which settings actually need useful values to get the development server running.  (I simply copied `.env.test`, removed the `-test` suffix from the database name and ran `createdb opencraft`, which is probably the fastest way of getting a somewhat working setup.)

instance/tasks.py
smarnach: @antoviaque Since the PR number is also included in the subdomain, which is one of the keys in the `get_or_create()` call, it would actually be possible to have multiple PRs for the same branch. (I haven't ever found this useful in practice, but who knows.)

instance/static/html/instance/index.html
antoviaque: @smarnach Yup, that seem like good instructions to add to the README - starting from the test environment can get things to start. For having everything working, there is not really a way around going through the configuration file, and setting the right values/creating the right accounts. Configuration seem a recurring pain point, you're not the first one to mention it - but not sure how to do much better than this.

instance/tasks.py
antoviaque: Ah, true for the subdomain - that's actually good then, perfect.

instance/models/instance.py
```diff
@@ -234,6 +235,15 @@ def github_base_url(self):
         return 'https://github.com/{0.fork_name}'.format(self)
 
     @property
+    def github_pr_url(self):
+        """
+        Web URL of the Github PR, or None if the PR number is not set.
+        """
+        if self.github_pr_number is None:
+            return None
+        return '{0.github_base_url}/pull/{0.github_pr_number}'.format(self)
```
antoviaque: @smarnach I just realized, testing on the real set of sandboxes, that there is a bug here - `github_base_url` is the base URL of the repo being deployed - not the repo the PR is opened on. This works well when they are the same, but not when the PR is opened from a fork. 

This should be using `settings.WATCH_FORK` or - likely better since we'll eventually watch mutiple forks - store the full PR URL instead of just the number. Since this is already merged and the ticket closed, I'll add a ticket in the backlog about this.

smarnach: Oh, darn, of course.  Yes, adding the full URL seems like the best solution until we have some more complete PR model.

PR #5, opened 2015-09-13
JS Unit Tests & JSHint, using Jasmine

antoviaque:
Adds unit tests for the JS code, and a linter check (JSHint). Run via jasmine in the browser (see `make test_js_web` and then go to http://localhost:8888/ ) for debugging, or in a CI manner via selenium and `jasmine-ci` (see `make test_js`).


instance/tests/js/helpers/restangular.js
```diff
@@ -0,0 +1,47 @@
...
+                  "restangularized", "several", "oneUrl", "allUrl", "fromServer",
+                  "getParentList", "save");
+};
+
+
+// Apply "sanitizeRestangularOne" function to an array of items
+jasmine.sanitizeRestangularAll = function(items) {
+    return _.map(items, function(value, index) {
+        return jasmine.sanitizeRestangularOne(value);
+    });
```
smarnach: Nit: This looks like it is equivalent to 

    return _.map(items, jasmine.sanitizeRestangularOne);

instance/static/js/src/instance.js
```diff
@@ -57,48 +56,30 @@ app.factory('OpenCraftAPI', function(Restangular) {
...
 app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
     function ($scope, Restangular, OpenCraftAPI, $q) {
-        // Display loading message
-        $scope.loading = true;
 
-        // Selection
-        $scope.selected = Array();
+        $scope.init = function() {
+            $scope.loading = true;
+            $scope.selected = Array();
```
smarnach: Is there a reason to use an array here?  It seems to be used like an object in the rest of the code, so `{}` would make more sense.

instance/tests/js/instance_app_spec.js
```diff
@@ -0,0 +1,121 @@
...
+        OpenCraftAPI,
+        $scope;
+
+    beforeEach(function() {
+        window.swampdragon = {
+            onChannelMessage: jasmine.createSpy(),
+            ready: jasmine.createSpy()
+        };
+        angular.mock.module('restangular');
+        module('InstanceApp');
```
smarnach: Out of curiosity: What namespace is the second `module` function coming from?  I would have expected that `module` refers to `angular.mock.module`, but I don't know since I'm learning about Angular while doing this review.

instance/tests/js/instance_app_spec.js
```diff
@@ -0,0 +1,121 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(function(){
+"use strict";
+
+// Tests //////////////////////////////////////////////////////////////////////
+
+describe('Instance app', function () {
+    var $controller,
```
smarnach: This variable `$controller` isn't used.

instance/tests/js/instance_app_spec.js
```diff
@@ -0,0 +1,121 @@
...
+
+        describe('$scope.select', function() {
+            it('select the instance', function() {
+                $scope.select('a', 'b');
+                expect($scope.selected.a).toEqual('b');
+            });
+        });
+
+        describe('$scope.updateInstanceList', function() {
+            it('loads the instance list from the API on init', inject(function ($httpBackend) {
```
smarnach: Do we need to inject `$httpBackend` here?  It's not used in the function body, and it is available via `httpBackend` set in `beforeEach()` anyway.

instance/tests/js/instance_app_spec.js
```diff
@@ -0,0 +1,121 @@
...
+                expect(jasmine.sanitizeRestangularAll($scope.instanceList)).toEqual(instanceList);
+            }));
+
+            it('sets $scope.loading while making the ajax call', inject(function ($httpBackend) {
+                expect($scope.loading).not.toBeTruthy();
```
smarnach: Same here, and for the other tests as well.

smarnach:
@antoviaque Looks good to me, :+1:.  I had to learn about Angular while doing this review, so some of the comments I added are probably unqualified. :)

Can the JS dependencies be installed via bower or npm or something?  Including the whole files in the repository feels a bit cumbersome to me.

There are outstanding model changes that don't have a migration yet (not introduced with this PR).

instance/tests/js/helpers/restangular.js
antoviaque: @smarnach Changed.

instance/static/js/src/instance.js
antoviaque: @smarnach Changed.

instance/tests/js/instance_app_spec.js
antoviaque: @smarnach Thanks for asking - I thought they were different functions, but you're right, they are the same. Changed.

https://docs.angularjs.org/api/ngMock/function/angular.mock.module

And I'm learning about angular too, so glad you're catching some of my beginner's mistakes. : )

instance/tests/js/instance_app_spec.js
antoviaque: @smarnach Good point, removed.

instance/tests/js/instance_app_spec.js
antoviaque: @smarnach True, it's a remnant from a refactoring - removed.

antoviaque:
@smarnach Thanks for the review! I'm learning Angular too, so my code is also unqualified : ) Thanks for catching these issues.

For the JS dependencies, I agree that it isn't optimal to have them in the repository, they pollute diffs & PRs. However, I would like to avoid node/npm dependencies - I've always found the packaging messy, and it makes the application environment more complex. What is possible to do without any additional dependency, though, is to retrieve the files via Makefile rules - I've done that: https://github.com/open-craft/opencraft/commit/0b90a97dd5feb14fcca0a4781f83976c42bb9c91 .

For the outstanding model changes, could that be because of https://github.com/open-craft/opencraft/pull/7#discussion_r40174892 ? I'll take care of it in that PR, since Matjaz already commented about it there.

static/js/external/Makefile
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Files #######################################################################
+
+all: angular.min.js angular.min.js.map angular-mocks.js angular-route.min.js angular-route.min.js.map \
+	 angular-ui-router.min.js jquery.min.js jshint.js restangular.min.js underscore-min.js \
+	 underscore-min.map
+
+angular.min.js:
+	$(GET) $@ $(CDNJS)/angular.js/1.4.6/angular.min.js
```
smarnach: These rules have the disadvantage that they won't reload the files when the version was changed in the Makefile.  So if upstream bumps the version, you pull and run `make <whatever>`, you will still have the old versions of the JS files, since make can't detect they are not up to date.

A proper solution to this problem would be to cache the version (or URL) used for each file, and redownload it if the URL changed, but this would require a bit of coding.  A simpler solution that doesn't get all cases right would be to use

    GET = wget -qN

above, remove `$@` in all rules, create a `.FORCE` target

    .PHONY: .FORCE

and add it as a dependency to all rules.  Make will run the rules every time now, and wget will compare the timestamp of the local files with the remote file, and redownload it if the remote file is newer (so downgrading will likely fail, while upgrading to newer versions should work correctly).

(You could also use `wget -qNO` and leave `$@` in.  However, `-N` disables the numbered backups anyway, so the output filename will be set correctly automatically.)

antoviaque: @smarnach It will reload them - the files are removed as part of the `make clean` rule, which is invoked before any execution. So it should always be up to date:

https://github.com/open-craft/opencraft/pull/5/files#diff-b67911656ef5d18c4ae36cb6741b7965R43

PR #6, opened 2015-09-14
Update: Python 3.5, Django 1.9, Ubuntu 16.04

antoviaque:
**Includes:**
* Python 3.5
* Django 1.9
* Ubuntu 16.04
* Pylint 1.5
* All other Python dependencies updated to their latest version.

**Exclusions of scope, to be addressed post-merge:**
* Upstream warnings raised by the upgrade (OC-1571)
* pylint exclusions cleanup (OC-1572, eg. [1](https://github.com/open-craft/opencraft/pull/6/files#diff-135cf0e4fd560de4eb8bd0b4192740e5R69) [2](https://github.com/open-craft/opencraft/pull/6/files#diff-b8a4de340708f89b8ea67cae413ff7f2R22) [3](https://github.com/open-craft/opencraft/pull/6/files#diff-7eb80b6d620bede230009c3350e9c11aR371))

**Local test instructions:**

1. Make a copy of your current box, in case you need to revert to it: https://www.vagrantup.com/docs/virtualbox/boxes.html
1. After switching to the current branch: `vagrant destroy && vagrant up`
1. Test the most important features (including provisioning a sandbox from the shell, with the worker processes working separately)

**Todo:**
* [x] _Get tests to pass (green build)_
* [x] _Get server & shell commands to work_
* [x] _Add `OC-*` tickets in backlog for exclusions of scope_


antoviaque:
Currently fails due to pylint not supporting python 3.5 just yet. Seems to be in progress, for the next pylint release: https://bitbucket.org/logilab/pylint/issues/643/regression-attributeerror-call-object-has

antoviaque:
Rebased on top of #10 which should get merged before.

omarkhan:
@antoviaque pylint should support python 3.5 now.

antoviaque:
@omarkhan Yup, the dependencies need an update - Django 1.9 has been released too, for example.

smarnach:
Django 1.9 isn't an LTS version, so if we do upgrade we will be forced to upgrade for every minor version they release.  (I'm just pointing this out and not implying we shouldn't do this.)

antoviaque:
@smarnach True - although waiting for the next LTS release also means accumulating quite a lot of technical debt in the meantime, as you still need to address the individual version upgrades/incompatibilities at once when the next LTS does come, instead of doing it progressively - you likely end up with more work overall, no?

smarnach:
@antoviaque I don't think it's more work overall, but it's most definitely less pleasant to have to do it all at once, and you miss out on any Django improvements in the meantime.  So yeah, we should probably do all the minor version upgrades.

Vagrantfile
```diff
@@ -2,11 +2,18 @@
 # vi: set ft=ruby :
 
 Vagrant.configure(2) do |config|
-  config.vm.box = 'ubuntu/trusty64'
+  # TODO: Switch back to the official box once it is released
```
itsjeyd: @antoviaque They seem to have an official box available for this now: https://vagrantcloud.com/ubuntu/boxes/xenial64. Might be worth trying to get it into this PR?

bin/bootstrap
```diff
@@ -1,13 +1,13 @@
 #!/bin/bash
 # This script provisions a development environment with local postgres and
-# redis servers for development and testing. Assumes ubuntu 14.04
+# redis servers for development and testing. Assumes Ubuntu 16.04
 
 set -e
 
-# If this is a vagrant VM, cd to /vagrant on login
+# If this is a vagrant VM, cd to ~/vagrant on login
```
itsjeyd: @antoviaque "cd to ~/opencraft on login"

instance/gandi.py
```diff
@@ -45,14 +45,14 @@ def __init__(self, api_url='https://rpc.gandi.net/xmlrpc/'):
         self.client = xmlrpc.client.ServerProxy(api_url)
 
     @property
-    def api_key(self): #pylint: disable=no-self-use
```
itsjeyd: @antoviaque Nice to see that we can remove these now :)

instance/tasks.py
```diff
@@ -22,8 +22,9 @@
 
 # Imports #####################################################################
 
-from huey.djhuey import crontab, db_periodic_task, db_task
+import logging
 
+from huey.contrib.djhuey import crontab, db_periodic_task, db_task
```
itsjeyd: @antoviaque Nit: Any reason to list `huey` before `django` here (instead of sticking to alphabetical order)? Just wondering since both of them are third-party libraries.

requirements.txt
```diff
@@ -1,110 +1,124 @@
...
+logilab-common==1.2.0
+Markdown==2.6.6
 MarkupSafe==0.23
-mccabe==0.3.1
-mock==1.3.0
-monotonic==0.3
-msgpack-python==0.4.6
-mysqlclient==1.3.6
+mccabe==0.4.0
+mock==2.0.0
```
itsjeyd: @antoviaque Since we are using Python >= 3.3 I'm wondering if there is a special reason for keeping this as an external dependency (instead of using the built-in `unittest.mock`)?

Vagrantfile
antoviaque: @itsjeyd I had tried it and was actually hoping to use it, but it looks like it still has the same issues it had before the xenial release, and doesn't appear to be correctly prepared to work with vagrant - ie uses the default ubuntu user instead of vagrant, and fails during provisioning (it doesn't seem to have the guest additions installed either):

```
antoviaque@antoviaque-laptop:~/prog/opencraft$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'ubuntu/xenial64'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'ubuntu/xenial64' is up to date...
==> default: Setting the name of the VM: ubuntu-xenial-16.04-cloudimg
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 2001 => 2001 (adapter 1)
    default: 5000 => 5000 (adapter 1)
    default: 8888 => 8888 (adapter 1)
    default: 22 => 2222 (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: ubuntu
    default: SSH auth method: password
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Connection timeout. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Authentication failure. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
    default: 
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
==> default: Mounting shared folders...
    default: /home/vagrant/opencraft => /home/antoviaque/prog/opencraft
Failed to mount folders in Linux guest. This is usually because
the "vboxsf" file system is not available. Please verify that
the guest additions are properly installed in the guest and
can work properly. The command attempted was:

mount -t vboxsf -o uid=`id -u ubuntu`,gid=`getent group ubuntu | cut -d: -f3` home_vagrant_opencraft /home/vagrant/opencraft
mount -t vboxsf -o uid=`id -u ubuntu`,gid=`id -g ubuntu` home_vagrant_opencraft /home/vagrant/opencraft

The error output from the last command was:

sudo: unable to resolve host ubuntu-xenial
mesg: ttyname failed: Inappropriate ioctl for device
mount: unknown filesystem type 'vboxsf'
```

instance/tasks.py
antoviaque: @itsjeyd I just got my alphabet mixed up : ) Thanks for catching it!

requirements.txt
antoviaque: @itsjeyd Good point - fixed in https://github.com/open-craft/opencraft/pull/6/commits/970d37ddc61db873eaa0516c748036c9a9524c0a

instance/tests/models/test_instance.py
```diff
@@ -22,13 +22,13 @@
 
 # Imports #####################################################################
 
+from unittest.mock import call, patch, Mock
 from urllib.parse import urlparse
 
 import re
 import yaml
```
itsjeyd: @antoviaque Seems like these imports aren't in the right place - `re` should go above `from unittest.mock ...`, and `yaml` should probably go below the `django` imports.

instance/models/server.py
```diff
@@ -23,21 +23,19 @@
 # Imports #####################################################################
 
 import logging
-import novaclient
 import time
 
-from swampdragon.pubsub_providers.data_publisher import publish_data
-
+import novaclient
```
itsjeyd: @antoviaque Does this come before the `django` and `swampdragon` imports to avoid mixing `import ...` and `from ... import ...` patterns? Looking at the [file](https://github.com/open-craft/opencraft/blob/python3.5/instance/models/server.py) as a whole, the imports section does look nice :) But the ordering is inconsistent with the ordering in other files, where you seemed to be pretty strict about alphabetical ordering. That's why I'm pointing it out :)

instance/models/mixins/ansible.py
```diff
@@ -19,7 +19,11 @@
 """
 Instance app model mixins - Ansible
 """
+
+# Imports #####################################################################
+
```
antoviaque: @e-kolpakov FYI, you forgot these in the new files you created : )

instance/tests/models/test_instance.py
antoviaque: @itsjeyd Good catch - fixed in https://github.com/open-craft/opencraft/pull/6/commits/1216ddaec678259f01de679b561031f246851f24

instance/models/server.py
antoviaque: @itsjeyd Good catch too. The change pylint introduced (requiring to import core libs before external dependencies) threw me a bit off balance and it took me a few tries to find a new ordering that was both happy with and that was consistent. I originally tried to keep `import ...` before `from ... import ...` for aesthetic, but that just adds too much constraints, and alphabetical ordering is more practical. So I switched in the middle the the changes, and it looks like I didn't properly re-apply it everywhere at the end. Just did a pass in https://github.com/open-craft/opencraft/pull/6/commits/1216ddaec678259f01de679b561031f246851f24 - see if that's better : )

For reference, the ordering is:

1. Core libs
2. External dependencies
3. Internal dependencies

Each of them ordered alphabetically, from the top-level module name, regardless of the import format.

instance/repo.py
```diff
@@ -22,16 +22,16 @@
 
 # Imports #####################################################################
 
-import git
+from contextlib import contextmanager
+import logging
 import tempfile
 import shutil
```
itsjeyd: @antoviaque Nit: Switch `tempfile` and `shutil`

itsjeyd:
@antoviaque :+1:

antoviaque:
@itsjeyd Thank you for the thorough review and testing! I'm merging now.

@open-craft/core FYI, you will need to reprovision your vagrant environment to get the upgrade - ensure you don't have any data you care about in the VM, and:

```$ vagrant destroy -f && vagrant up```

Also, to avoid issues with certain versions of Virtualbox/Vagrant, it's recommended that you update to the following versions:

* Vagrant 1.8.1 (from https://www.vagrantup.com/downloads.html )
* Virtualbox 5.0.18 (I have 5.0.18-dfsg-2build1 - the default with xenial; 5.0.20 seem to have been released upstream of Ubuntu, which might work too, but hasn't been tested: https://www.virtualbox.org/wiki/Downloads )

See OC-1570 for the details.

**Comment by bradenmacdonald** [comment]
@antoviaque Did you run this in "production mode" or deploy it? Because the `Procfile` here no longer works:
```
01:35:31 periodic.1  | Usage: manage.py run_huey [options] 
01:35:31 periodic.1  | 
01:35:31 periodic.1  | Run the queue consumer
01:35:31 periodic.1  | 
01:35:31 periodic.1  | manage.py: error: no such option: --periodic
```

The fix is simply to remove `--periodic` from the Procfile, since that option is implied whenever `--no-periodic` is not specified.

antoviaque:
@bradenmacdonald Ah damn - I think I only ever ran it with `make rundev`, sorry for having missed that. The deployment itself is done in OC-1533, so we haven't seen it yet. (FYI @smarnach ). I'll have a look and open a PR for the change.

antoviaque:
@bradenmacdonald Done: https://github.com/open-craft/opencraft/pull/70

bradenmacdonald:
I noticed in my VM that since this upgrade (and new vagrant setup), the bash history of the `vagrant` user was not being saved (which was annoying), because `~/.bash_history` was owned by `root`. Any idea why? I don't see any reference to it in the Vagrantfile or bootstrap script.

antoviaque:
@bradenmacdonald That is likely coming from the box I used as a base for building `opencraft/xenial64` - likely a way for the upstream to ensure he doesn't forget to clear his history before publishing his box : ) If you want I can edit the base box to change the owner back to `vagrant`. Editing, publishing and testing the base box takes time though, so alternatively we could put this in a ticket for the next update? The plan is eventually to switch to the official `ubuntu/xenial64` box anyway, once they have solved the issues it currently has.

bradenmacdonald:
@antoviaque We could also just put a fix into the `Vagrantfile` provisioning section, especially if/when we move to the official upstream box.

antoviaque:
@bradenmacdonald I think the upstream box should not have this - or did you had to do that in the previous box, before the upgrade? My hope is that we'll not have to do anything then - but we'll need to check. If the upstream box also has this, then yep the `Vagrantfile` would work (or the ansible playbook, actually.)

bradenmacdonald:
OK, hopefully that is the case :)

PR #7, opened 2015-09-21
Logging Management - Use Custom Logging Handlers

antoviaque:
Use LoggerAdapters to attach standard logger functions to given objects, and a dedicated LoggingHandler to route the incoming log messages to the database & client display.


instance/migrations/0025_auto_20150920_0907.py
```diff
@@ -0,0 +1,36 @@
...
+        ('instance', '0024_auto_20150911_2304'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='GeneralLogEntry',
+            fields=[
+                ('id', models.AutoField(serialize=False, auto_created=True, verbose_name='ID', primary_key=True)),
+                ('created', django_extensions.db.fields.CreationDateTimeField(default=django.db.models.fields.NOT_PROVIDED, verbose_name='created', auto_now_add=True)),
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, default=django.db.models.fields.NOT_PROVIDED, verbose_name='modified')),
```
mtyaka: @antoviaque All LogEntry models have a `modified` field. Do we ever expect log entries to be modified after they are created? I would expect log entries to be immutable.

instance/migrations/0025_auto_20150920_0907.py
```diff
@@ -0,0 +1,36 @@
...
+            ],
+            options={
+                'verbose_name_plural': 'General Log Entries',
+            },
+            bases=(instance.models.utils.ValidateModelMixin, models.Model),
+        ),
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='base_domain',
+            field=models.CharField(default='plebia.net', max_length=50),
```
mtyaka: @antoviaque Why is the `plebia.net` default set at the database level?

instance/models/instance.py
```diff
@@ -134,6 +152,68 @@ def status(self):
...
+        return {'instance_id': self.pk}
+
+    @staticmethod
+    def on_post_save(sender, instance, created, **kwargs):
+        """
+        Triggered by the post_save event, with `created` a boolean indicating if the object was just
+        created in the DB
+        """
+        self = instance
+        self.update_logger()
```
mtyaka: @antoviaque `self = instance; self.update_logger()` is a bit confusing. `instance.update_logger()` would make more sense to me.

instance/models/instance.py
```diff
@@ -134,6 +152,68 @@ def status(self):
...
+
+    def update_logger(self):
+        """
+        Start referencing the instance in logs once the DB entry exist
+        """
+        if self.pk is not None:
+            self.logger = InstanceLoggerAdapter(logger, {'obj': self})
```
mtyaka: @antoviaque Having to call `update_logger` feels dirty to me. I would prefer if `self.logger` was always created in the `__init__` method, with `'obj'` set to `self`, regardless of whether the object has been saved to the database yet. The `if self.pk is not None` checks could be moved to the logger adapters instead (`DBHandler` already does that).

instance/models/instance.py
```diff
@@ -134,6 +152,68 @@ def status(self):
...
+
+        while instance_log_entry is not None:
+            log_text += '{}\n'.format(instance_log_entry)
+            instance_log_entry = next_instance_log_entry()
+
+        while server_log_entry is not None:
+            log_text += '{}\n'.format(server_log_entry)
+            server_log_entry = next_server_log_entry()
+
+        return log_text
```
mtyaka: @antoviaque This method makes me wonder whether spreading the logs over multiple tables is the best approach. It might be worth considering having a single `log_entries(created, level, text, instance, server)` table where `server` and `instance` could be `NULL`.

instance/logger_adapter.py
```diff
@@ -0,0 +1,56 @@
...
+
+class InstanceLoggerAdapter(logging.LoggerAdapter):
+    """
+    Custom LoggerAdapter for Instance objects
+    Include the instance name in the output
+    """
+    def process(self, msg, kwargs):
+        msg, kwargs = super().process(msg, kwargs)
+
+        if self.extra.get('obj', None):
```
mtyaka: @antoviaque As mentioned in the model class, I would prefer if `obj` was always defined. Here we would check for `obj.pk` instead, or perhaps even better only check whether `obj.sub_domain` is present - logging the subdomain could be helpful even if  the object hasn't been saved to the database yet.

instance/migrations/0025_auto_20150920_0907.py
antoviaque: @mtyaka This comes from the `TimeStampedModel` mixin from [Django extensions](http://django-extensions.readthedocs.org/en/latest/model_extensions.html), it's present on all models. Log entries aren't expected to change, no, but the field doesn't seem harmful either - creating a new mixin without the modified field to replace https://github.com/django-extensions/django-extensions/blob/199e659fd8339d09e809d257f485eeda1b7f9e93/django_extensions/db/models.py#L19 might be more trouble that it's worth.

instance/migrations/0025_auto_20150920_0907.py
antoviaque: @mtyaka Hmm, true - this seem to be because of https://github.com/open-craft/opencraft/blob/8a4ce6f5ac23f05ba668991c7b2885f5c5550d3c/instance/models/instance.py#L91 . I think that still does the right thing (load the default from local settings on each instance), but this does confuse the autogenerated migrations. Not sure what would be a better approach here - setting the default to `None` could work too, but then that requires to handle that case to replace it with the default setting when null every time the field is accessed.

instance/models/instance.py
antoviaque: @mtyaka Except that `instance` here doesn't have the meaning that it appears to have - the `on_post_save()` args need to match https://docs.djangoproject.com/en/1.8/ref/signals/#post-save - ie `instance` refers to being an instance of the model, not an `OpenEdXInstance` or `Instance` from the current file. This gets really confusing for the `Server` model, which had https://github.com/open-craft/opencraft/blob/8a4ce6f5ac23f05ba668991c7b2885f5c5550d3c/instance/models/server.py#L148 before this PR. I thought assigning it to `self` made things more obvious, and applied the pattern here too for consistency. Do you think it's a bad idea?

instance/models/instance.py
antoviaque: @mtyaka Good point, that's better - will do.

instance/models/instance.py
antoviaque: @mtyaka True, that might be better - we'll still need different types of log entries, and that might make it harder to have a clean pattern for matching an object to its corresponding log entry object though. I'll have a closer look.

instance/migrations/0025_auto_20150920_0907.py
mtyaka: @antoviaque Oh, I see... It's not worth it then, the field doesn't do any harm.

instance/migrations/0025_auto_20150920_0907.py
mtyaka: @antoviaque Setting to `None` would be my preferred solution. Rather then checking for null every time the field is accessed, I would set the field to the default value (if not already set to something else) right before saving. I am not sure how easy or idiomatic doing that is in django (perhaps using a `pre_save` signal?).

instance/models/instance.py
mtyaka: @antoviaque Ah, I didn't think of that! I checked django docs before reading further, so I know what `instance` means, but yeah I can definitely see why the name `instance` can be confusing :)
The `self = instance` bit is also confusing, though. In the server class you could maybe just name the parameter `server` instead of `instance`:

    def on_post_save(sender, server, created, **kwargs):
        ...

antoviaque: @mtyaka Yup, that was what I first tried, but there seem to be some argument manipulation going on, Django was choking on a missing instance attribute when I changed it there. If you see something else I can try to avoid the `self = instance` line, I'd be happy to - I'm not particularly happy about it either.

instance/migrations/0025_auto_20150920_0907.py
antoviaque: @mtyaka Good idea - I've done that: https://github.com/open-craft/opencraft/commit/e1eb7dc06ee9b72dfa6f22012e59653d4798e38c

instance/logger_adapter.py
antoviaque: @mtyaka Done - you're right about checking `obj.sub_domain` instead, that's better.

instance/models/instance.py
antoviaque: @mtyaka I'm not sure about this change. On on side it looks better, having a unique table makes things simpler - it avoids the type of logic we're seeing here, and would allow to group 3 different types of objects into one. But on the other side, it makes some parts harder to evolve - for example, if we add a new (different) type of server, like say an `AWSServer`, we have to create a brand new table, or add a superfluous column. Same thing with adding different types of instances. Or simply if we want to add DB logging for another object - it will be harder to share the internal plumbing.

You might end up being right, but I'd like to see how we are going to deal with things like https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9937191-self-deployment or https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9942105-add-support-for-aws-servers before deciding to change.

antoviaque:
@mtyaka I've also rebased to stay in sync with the other recent changes on master. This should be ready for another pass.

instance/models/instance.py
mtyaka: @antoviaque I agree it makes sense to wait for the code to evolve a bit before deciding whether the proposed change is actually the way to go.

mtyaka:
@antoviaque Looking good :+1:

PR #8, opened 2015-09-22
Full branch url

Kelketek:
Stores the full URL of a pull request in the instance model instead of just the number.

Also tweaks the makefile to fix issues I ran into when trying to use it.

### DIFF
--- a/instance/github.py
+++ b/instance/github.py
@@ -162,3 +162,10 @@ class PR:
         Extra settings contained in the PR body
         """
         return get_settings_from_pr_body(self.body)
+
+    @property
+    def github_pr_url(self):
+        """
+        Construct the URL for the pull request
+        """
+        return 'https://github.com/{fork_name}/pull/{number}'.format(fork_name=self.fork_name, number=self.number)
--- a/instance/migrations/0031_openedxinstance_github_pr_url.py
+++ b/instance/migrations/0031_openedxinstance_github_pr_url.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import models, migrations
+
+
+def migrate_urls(apps, schema_editor):
+    Instance = apps.get_model("instance", "OpenEdXInstance")
+    instances = Instance.objects.exclude(github_pr_number__isnull=True)
+    for instance in instances:
+        instance.github_pr_url = 'https://github.com/{fork}/pull/{number}'.format(
+            fork=getattr(settings, 'WATCH_FORK', 'edx/edx-platform'),
+            number=instance.github_pr_number
+        )
+        instance.save()
+
+
+def reverse_migrate_urls(apps, schema_editor):
+    Instance = apps.get_model("instance", "OpenEdXInstance")
+    instances = Instance.objects.exclude(github_pr_url='')
+    for instance in instances:
+        instance.github_pr_number = int(instance.github_pr_url.split('/')[-1])
+        instance.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0030_auto_20150927_1045'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='github_pr_url',
+            field=models.URLField(blank=True),
+        ),
+        migrations.RunPython(migrate_urls, reverse_code=reverse_migrate_urls),
+        migrations.RemoveField(
+            model_name='openedxinstance',
+            name='github_pr_number',
+        ),
+    ]
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -279,7 +279,7 @@ class GitHubInstanceMixin(VersionControlInstanceMixin):
     """
     github_organization_name = models.CharField(max_length=200, db_index=True)
     github_repository_name = models.CharField(max_length=200, db_index=True)
-    github_pr_number = models.IntegerField(blank=True, null=True)
+    github_pr_url = models.URLField(blank=True)
     github_admin_organization_name = models.CharField(max_length=200, blank=True,
                                                       default=settings.DEFAULT_ADMIN_ORGANIZATION)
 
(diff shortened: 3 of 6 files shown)


Kelketek:
@antoviaque This is ready for review.

Makefile
```diff
@@ -9,9 +9,9 @@ ifeq (test_one,$(firstword $(MAKECMDGOALS)))
 endif
 
 clean:
-	find -name '*.pyc' -delete
-	find -name '*~' -delete
-	find -name '__pycache__' -type d -delete
+	find . -name '*.pyc' -delete
+	find . -name '*~' -delete
+	find . -name '__pycache__' -type d -delete
```
antoviaque: @Kelketek Why do we need the `.` here?

instance/migrations/0026_pr_url.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0025_openedxinstance_github_pr_url'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_urls, reverse_code=reverse_migrate_urls)
```
antoviaque: @Kelketek Nit: Why not grouping the 3 operations in a single migration? To make it easier to resume if one of them fails? That must be a common pattern, I'm curious how Django handles resuming failing migrations with several operations.

instance/models/instance.py
```diff
@@ -204,7 +204,7 @@ class GitHubInstanceMixin(VersionControlInstanceMixin):
     """
     github_organization_name = models.CharField(max_length=200, db_index=True)
     github_repository_name = models.CharField(max_length=200, db_index=True)
-    github_pr_number = models.IntegerField(blank=True, null=True)
+    github_pr_url = models.CharField(max_length=200, blank=True, default='')
```
antoviaque: @Kelketek Shouldn't this be a URL field? Also the `default=''` is superflous I think - Django might ask for a default value when creating the migrations, but you can put `''` there rather than here?

Makefile
smarnach: Just happened to see this comment – for some non-GNU versions of find, the path is mandatory, and it looks like OSX ships with such a version of find.

antoviaque: I see - does that mean that if we include this change, the rest can work natively on MacOS? I'm happy to include it regardless, but for the future I wonder if it's worth the effort to support multiple environments to run it, especially now that we have Vagrant integration.

instance/migrations/0026_pr_url.py
Kelketek: @antoviaque I know that, at least for PostgreSQL and South, data migrations and schema migrations cannot be done in the same migration. But maybe this is not an issue for the built in Django migrations. I'll take a look.

Makefile
Kelketek: @antoviaque There's actually more changes that need doing to make it work properly on Mac OS X. I was just confused on how this ever worked, since I didn't think this syntax worked in Linux either.

Where's the vagrant info? I'd be happy to drop this change and just use that instead.

instance/models/instance.py
Kelketek: @antoviaque I don't think the default='' is superflous. blank=True means that the admin/form generation won't complain if the field isn't filled, but unless null=True is also set, it will still fail validation. But you don't want string fields to be null because that causes ambiguous situations when looking up unset values (is it '' or None?)

I wasn't aware of URLField. I could do that.

Kelketek:
@antoviaque This is ready for a second look.

Makefile
antoviaque: @Kelketek You seem to have dropped it already so you might have found it in the meantime, but the latest master has the Vagrantfile: https://github.com/open-craft/opencraft/blob/master/Vagrantfile (cf the README also).

instance/models/instance.py
antoviaque: @Kelketek I thought that the empty string was the default for string-based fields? From a quick look at the code, https://github.com/django/django/blob/stable/1.8.x/django/db/models/fields/__init__.py#L796-L799 seem to show this behavior - `empty_strings_allowed` is the default value for a `Field` https://github.com/django/django/blob/stable/1.8.x/django/db/models/fields/__init__.py#L97 and doesn't seem to be redefined for `CharField` or `URLField`. Did you get a validation error when trying to save an `Instance` object without `default=''`?

instance/migrations/0031_openedxinstance_github_pr_url.py
```diff
@@ -0,0 +1,44 @@
...
+        instance.github_pr_url = 'https://github.com/{fork}/pull/{number}'.format(
+            fork=getattr(settings, 'WATCH_FORK', 'edx/edx-platform'),
+            number=instance.github_pr_number
+        )
+        instance.save()
+
+
+def reverse_migrate_urls(apps, schema_editor):
+    Instance = apps.get_model("instance", "OpenEdXInstance")
+    instances = Instance.objects.exclude(github_pr_url='')
```
antoviaque: Since the integer field on the current model is nullable, would the empty string be correctly converted to a null value? Not sure how the django migrations behave in this case.

antoviaque:
@Kelketek Thanks for the changes! LGTM - once we have resolved https://github.com/open-craft/opencraft/pull/8#discussion-diff-40491416 I'll give it a try locally to give you the +1.

instance/models/instance.py
Kelketek: @antoviaque It appears you're correct. I'm thinking of behavior I ran into when working with Django 1.4.

antoviaque:
@Kelketek Thanks for checking, and for the changes! It worked well when I tested locally, both with preexisting cross-fork PRs, and new ones. The only issue I had was that my vagrant VM ran out of memory when trying to run `make run`, so I posted https://github.com/open-craft/opencraft/pull/17

This will need a rebase, but otherwise LGTM. :+1:

instance/github.py
```diff
@@ -162,3 +162,10 @@ def extra_settings(self):
         Extra settings contained in the PR body
         """
         return get_settings_from_pr_body(self.body)
+
+    @property
+    def github_pr_url(self):
+        """
+        Construct the URL for the pull request
+        """
+        return 'https://github.com/{fork_name}/pull/{number}'.format(fork_name=self.fork_name, number=self.number)
```
antoviaque: @Kelketek I pushed this to production, and this actually still seem wrong - the migrations do the right thing, so this was hiding the issue when I tested locally, but this doesnt, cf https://github.com/open-craft/opencraft/blob/34782a79d1b4a08b0d59f93ef91ff7611c0ffb7e/instance/github.py#L97 - `fork_name` is again the repo from which the PR is sent, not the destination repo of the PR.

PR #9, opened 2015-09-25
Provision local development environment with vagrant

omarkhan:
This pull request is similar to https://github.com/open-craft/opencraft/pull/3, with the following differences:

- Uses postgresql instead of sqlite
- Does not change the Makefile
- Does not require any manual changes to `~/.bashrc`


omarkhan:
Not sure why the integration test failed, looks like scp was expecting a password.

antoviaque:
@omarkhan Thank you for this! I'll have a look at it over the week-end. Given that this partially duplicates #3 would it make sense for you and @brousch to collaborate, and consolidate your submissions on a single PR? I haven't looked at the current PR yet, so not sure if that makes sense, but I'd like to know what you think about it. 

Note that this isn't a requirement or part of the recruitment process - the additional contribution and/or collaboration is a bonus, but it won't be held against you if you don't do it.

antoviaque:
Also - for the build, that's because the commits come from a fork, so CircleCI doesn't load the auth. I've pushed the branch to the main repo to trigger a rebuild with the right auth.

omarkhan:
@brousch I had already pretty much finished this by the time I saw your changes, so I made a separate pull request. I had a quick look through yours - it looks like you made a few changes to make it easier for people running windows. I'm not a windows user so I don't really know what that entails, what extra things do you have to do to get it working?

brousch:
For Windows, I had to assume that .py files were not executable (for example, always use `python manage.py` instead of just `manage.py`, and use `honcho run python manage.py`). I haven't looked at your DB setup, but try not to create files like SQLite DBs in the project dir as these can run into Windows file locking issues. They should live entirely within the VM.

omarkhan:
Thanks. This uses postgres instead of sqlite so that shouldn't be an issue.

For the executable issues, will pulling in your Makefile changes be enough, or is there more to it? Would you be able to test this on windows?

brousch:
I think we should use your Vagrantfile PR and then I could make a separate PR for the Makefile changes so I can test it properly on Windows.

omarkhan:
Good idea :+1:

README.md
```diff
@@ -141,9 +176,9 @@ $ make test_unit
 $ make test_integration
 ```
 
-Note that the integration tests aren't run by default, as they require a working OpenStack cluster
-configured. To run them, create a `.env.test` file - your development environment is likely a good
-starting point:
+Note that the integration tests aren't run by default, as they require a working
+OpenStack cluster configured. To run them, create a `.env.integration` file -
```
antoviaque: @omarkhan Good catch - thanks!

Vagrantfile
```diff
@@ -0,0 +1,61 @@
...
+PROVISION = <<SCRIPT
+set -e
+
+# cd to /vagrant on login
+cd /vagrant
+grep -q 'cd /vagrant' ~/.bashrc || echo 'cd /vagrant' >> ~/.bashrc
+
+# Install system packages
+sudo apt-get update --quiet
+sudo apt-get install -y $(cat debian_packages.lst) postgresql
```
antoviaque: @omarkhan I liked how @brousch put this type of steps in the `Makefile` - it would allow to factorize duplications between here and the `README` file. You would still have to also install `postresql`, which I agree is better left out of the dependencies, but that can be kept separate (or with two targets, like `install_dependencies` and `install_db_dependencies`).

This is a nit though, or it might be something @brousch could add -- I don't feel too strongly about it.

Vagrantfile
```diff
@@ -0,0 +1,61 @@
...
+
+# Upgrade pip
+sudo pip3 install --upgrade pip
```
antoviaque: @omarkhan It's tempting to use global packages given that the application is isolated on the VM, but that also means that if we do need to install a second app on the VM later on, we'll have to migrate to virtualenvs then. I'd rather apply good practices from the start, and use a venv. the effort is minimal and the commands are already listed in the README.

That has also the advantage of not having to handle the differences between people using a virtualenv or not, or having to redefine the location of the `src` dir, right below.

Vagrantfile
```diff
@@ -0,0 +1,61 @@
...
+sudo -u postgres createuser -d vagrant
+
+# Allow access to postgres from localhost without password
+cat << EOF | sudo tee /etc/postgresql/9.3/main/pg_hba.conf
+local   all             postgres                                peer
+local   all             all                                     trust
+host    all             all             127.0.0.1/32            trust
+host    all             all             ::1/128                 trust
+EOF
+sudo service postgresql restart
```
antoviaque: @omarkhan It's not the case by default? Honest question, I don't remember having to relax the default auth on my Ubuntu - but the defaults might have changed since I installed it (or I might incorrectly remember it).

Vagrantfile
```diff
@@ -0,0 +1,61 @@
...
+
+# Create postgres database
+createdb --encoding utf-8 --template template0 vagrant
```
antoviaque: @omarkhan Can we name the database `opencraft` rather than `vagrant`?

Vagrantfile
```diff
@@ -0,0 +1,61 @@
...
+
+# Copy configuration for local development
+cp .env.vagrant .env
```
antoviaque: @omarkhan Cf the comments I made on @brousch 's PR about this: https://github.com/open-craft/opencraft/pull/3#discussion_r39335426

opencraft/settings.py
```diff
@@ -209,7 +209,7 @@
 SWAMP_DRAGON_REDIS_PORT = REDIS_URL_OBJ.port
 SWAMP_DRAGON_CONNECTION = ('swampdragon.connections.sockjs_connection.DjangoSubscriberConnection', '/data')
 DRAGON_SERVER_ADDRESS_PORT = env('DRAGON_SERVER_ADDRESS_PORT', default='localhost:2001')
-DRAGON_URL = env('DRAGON_URL', default='http://{}/'.format(DRAGON_SERVER_ADDRESS_PORT))
+DRAGON_URL = env('DRAGON_URL', default='http://localhost:2001/')
```
antoviaque: @omarkhan Why was this needed?

antoviaque:
@omarkhan @brousch Thanks again to both of you, it's definitely going to be useful to have Vagrant support : )

@omarkhan I made a few comments, but they are minor points - good job on this PR.

Vagrantfile
omarkhan: Fair enough, will do

Vagrantfile
omarkhan: Default auth on ubuntu is md5 encrypted passwords for everyone except the postgres user over a unix socket, which uses peer auth.

Vagrantfile
omarkhan: Sure, I named it vagrant so that you can simply type `psql` and get a shell. Your call

opencraft/settings.py
omarkhan: The current implementation uses the `DRAGON_SERVER_ADDRESS_PORT`, which is actually a bind address. So if I set that to `0.0.0.0:2001` to get it working with vagrant, the default `DRAGON_URL` becomes `http://0.0.0.0:2001/`, which is clearly wrong.

Vagrantfile
omarkhan: Done: https://github.com/omarkhan/opencraft/commit/b77c1a08cc7f18c7a0781dff3239bbbd0c5d3c72#diff-23b6f443c01ea2efcb4f36eedfea9089R44

Vagrantfile
omarkhan: See https://github.com/omarkhan/opencraft/commit/b77c1a08cc7f18c7a0781dff3239bbbd0c5d3c72#diff-23b6f443c01ea2efcb4f36eedfea9089R48

Vagrantfile
omarkhan: @brousch is working on Makefile changes in a separate PR

Vagrantfile
omarkhan: Done: https://github.com/omarkhan/opencraft/commit/beb54b9ee8c498bdfdfa00dd80c860f595fc3c01

opencraft/settings.py
antoviaque: @omarkhan Good point - sounds good.

Vagrantfile
antoviaque: @omarkhan Thanks! And good catch for the loggers from the test env.

Vagrantfile
antoviaque: @omarkhan Ok, thanks for the details.

Vagrantfile
antoviaque: @omarkhan Ok - I'm merge as is then, and let @brousch port this part of his PR, if he wants to.

antoviaque:
@omarkhan Thank you again for this - works like a charm here. :+1: 

I'll merge if the integration tests pass.

PR #11, opened 2015-09-28
Persistent sandbox databases

omarkhan:
This pull request adds support for instances to use external mysql and mongo databases. The goal is to [preserve databases during sandbox reprovisioning](http://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9541143-preserve-databases-during-sandbox-reprovisioning).

Default databases for new instances can be set through the `INSTANCE_MYSQL_URL` and `INSTANCE_MONGO_URL` settings.

If the default is to use external databases but the PR submitter would like to use local databases, she can include `(ephemeral databases)` in the pull request body on the same line as the sandbox url to override the default.

See also: https://github.com/edx/configuration/pull/2310

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,6 +1,8 @@
 DEBUG=true
 SECRET_KEY='tests'
 DATABASE_URL='postgres://localhost/opencraft'
+INSTANCE_MYSQL_URL='mysql://root@localhost'
+INSTANCE_MONGO_URL='mongodb://localhost'
 HUEY_ALWAYS_EAGER=true
 OPENSTACK_USER='test'
 OPENSTACK_PASSWORD='pass'
--- a/instance/github.py
+++ b/instance/github.py
@@ -29,6 +29,7 @@ import requests
 import yaml
 
 from django.conf import settings
+from django.template.defaultfilters import truncatewords
 
 
 # Logging #####################################################################
@@ -87,6 +88,19 @@ def get_settings_from_pr_body(pr_body):
         return m.groups()[0]
     else:
         return ''
+
+
+def is_pr_body_requesting_ephemeral_databases(pr_body, domain):
+    """
+    Return True if the PR body specified that the sandbox should use ephemeral
+    databases, False if it specifies persistent databases, or None otherwise
+    """
+    escaped_domain = re.escape(domain)
+    if re.search(r'{0}.*ephemeral databases?'.format(escaped_domain), pr_body):
+        return True
+    if re.search(r'{0}.*persistent databases?'.format(escaped_domain), pr_body):
+        return False
+    return None
 
 
 def get_pr_by_number(pr_target_fork_name, pr_number):
@@ -163,6 +177,13 @@ class PR:
         self.body = body
 
     @property
+    def truncated_title(self):
+        """
+        This PR's title truncated to 4 words
+        """
+        return truncatewords(self.title, 4)
+
+    @property
     def extra_settings(self):
         """
         Extra settings contained in the PR body
@@ -187,3 +208,9 @@ class PR:
         Construct the URL for the pull request
         """
         return 'https://github.com/{repo_name}/pull/{number}'.format(repo_name=self.repo_name, number=self.number)
+
+    def use_ephemeral_databases(self, domain):
+        """
+        Does this PR request ephemeral databases?
+        """
+        return is_pr_body_requesting_ephemeral_databases(self.body, domain)
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -24,15 +24,21 @@ Instance app models - Instance
 
 import logging
 import os
+import string
 
 from functools import partial
 
 from django.conf import settings
 from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models
+from django.db.backends.utils import truncate_name
 from django.template import loader
 from django.utils import timezone
+from django.utils.crypto import get_random_string
 from django_extensions.db.models import TimeStampedModel
+
+import MySQLdb as mysql
+import pymongo
 
 from instance import ansible, github
 from instance.gandi import GandiAPI
@@ -264,6 +270,19 @@ class GitHubInstanceQuerySet(models.QuerySet):
     Additional methods for instance querysets
     Also used as the standard manager for the GitHubInstance model
     """
+    def update_or_create_from_pr(self, pr, sub_domain):
+        """
+        Create or update an instance for the given pull request
+        """
+        instance, created = self.get_or_create(
+            sub_domain=sub_domain,
+            fork_name=pr.fork_name,
+            branch_name=pr.branch_name,
+        )
+        instance.update_from_pr(pr)
+        instance.save()
+        return instance, created
+
     def create(self, *args, **kwargs):
         """
         Augmented `create()` method:
@@ -350,6 +369,13 @@ class GitHubInstanceMixin(VersionControlInstanceMixin):
         return '{0.github_organization_name}/{0.github_repository_name}'.format(self)
 
     @property
+    def reference_name(self):
+        """
+        A descriptive name for the instance, which includes meaningful attributes
+        """
+        return '{0.github_organization_name}/{0.branch_name} ({0.commit_short_id})'.format(self)
+
+    @property
     def github_base_url(self):
         """
         Base GitHub URL of the fork (eg. 'https://github.com/open-craft/edx-platform')
@@ -420,8 +446,7 @@ class GitHubInstanceMixin(VersionControlInstanceMixin):
 
             # Update the hash in the instance title if it is present there
             # TODO: Find a better way to handle this - include the hash dynamically?
-            # TODO: Figure out why the warnings aren't suppressed despite the fact that it's a mixin
-            if self.name and old_commit_short_id: #pylint: disable=access-member-before-definition
+            if self.name and old_commit_short_id:
                 #pylint: disable=attribute-defined-outside-init
                 self.name = self.name.replace(old_commit_short_id, self.commit_short_id)
 
@@ -444,6 +469,14 @@ class GitHubInstanceMixin(VersionControlInstanceMixin):
         self.github_repository_name = fork_repo
         if commit:
             self.save()
+
+    def update_from_pr(self, pr):
+        """
+        Update this instance with settings from the given pull request
+        """
+        self.name = ('PR#{pr.number}: {pr.truncated_title}' +  #pylint: disable=attribute-defined-outside-init
+                     ' ({pr.username}) - {i.reference_name}').format(pr=pr, i=self)
+        self.github_pr_url = pr.github_pr_url
 
 
 # Ansible #####################################################################
@@ -586,9 +619,86 @@ class AnsibleInstanceMixin(models.Model):
         return (log, returncode)
 
 
+# Databases ###################################################################
+
+class MySQLInstanceMixin(models.Model):
+    """
+    An instance that uses mysql databases
+    """
+    mysql_user = models.CharField(max_length=16, blank=True)  # 16 chars is mysql maximum
+    mysql_pass = models.CharField(max_length=32, blank=True)
+    mysql_provisioned = models.BooleanField(default=False)
+
+    class Meta:
+        abstract = True
+
+    @property
+    def mysql_database_names(self): #pylint: disable=no-self-use
+        """
+        An iterable of database names
+        """
+        return NotImplementedError
+
+    def provision_mysql(self):
+        """
+        Create mysql user and databases
+        """
+        if settings.INSTANCE_MYSQL_URL_OBJ and not self.mysql_provisioned:
+            connection = mysql.connect(
+                host=settings.INSTANCE_MYSQL_URL_OBJ.hostname,
+                user=settings.INSTANCE_MYSQL_URL_OBJ.username,
+                passwd=settings.INSTANCE_MYSQL_URL_OBJ.password or '',
+                port=settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+            )
+            cursor = connection.cursor()
+
+            for database in self.mysql_database_names:
+                # We can't use the database name in a parameterized query, the
+                # driver doesn't escape it properly. Se we escape it here instead
+                database_name = connection.escape_string(database).decode()
+                cursor.execute('CREATE DATABASE {0} DEFAULT CHARACTER SET utf8'.format(database_name))
+                cursor.execute('GRANT ALL ON {0}.* TO %s IDENTIFIED BY %s'.format(database_name),
+                               (self.mysql_user, self.mysql_pass))
+
+            self.mysql_provisioned = True
+            self.save()
+
+
+class MongoDBInstanceMixin(models.Model):
+    """
+    An instance that uses mongo databases
+    """
+    mongo_user = models.CharField(max_length=16, blank=True)
+    mongo_pass = models.CharField(max_length=32, blank=True)
+    mongo_provisioned = models.BooleanField(default=False)
+
+    class Meta:
+        abstract = True
+
+    @property
+    def mongo_database_names(self): #pylint: disable=no-self-use
+        """
+        An iterable of database names
+        """
+        return NotImplementedError
+
+    def provision_mongo(self):
+        """
+        Create mongo user and databases
+        """
+        if settings.INSTANCE_MONGO_URL and not self.mongo_provisioned:
+            mongo = pymongo.MongoClient(settings.INSTANCE_MONGO_URL)
+            for database in self.mongo_database_names:
+                mongo[database].add_user(self.mongo_user, self.mongo_pass)
+
+            self.mongo_provisioned = True
+            self.save()
+
+
 # Open edX ####################################################################
 
-class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
+# pylint: disable=too-many-instance-attributes
+class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
     """
     A single instance running a set of Open edX services
     """
@@ -601,7 +711,13 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
     s3_secret_access_key = models.CharField(max_length=50, blank=True)
     s3_bucket_name = models.CharField(max_length=50, blank=True)
 
-    ANSIBLE_SETTINGS = AnsibleInstanceMixin.ANSIBLE_SETTINGS + ['ansible_s3_settings']
+    use_ephemeral_databases = models.BooleanField()
+
+    ANSIBLE_SETTINGS = AnsibleInstanceMixin.ANSIBLE_SETTINGS + [
+        'ansible_s3_settings',
+        'ansible_mysql_settings',
+        'ansible_mongo_settings',
+    ]
 
     class Meta:
         verbose_name = 'Open edX Instance'
@@ -615,13 +731,6 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         return settings.DEFAULT_FORK
 
     @property
-    def reference_name(self):
-        """
-        A descriptive name for the instance, which includes meaningful attributes
-        """
-        return '{s.github_organization_name}/{s.branch_name} ({s.commit_short_id})'.format(s=self)
-
-    @property
     def ansible_s3_settings(self):
         """
         Ansible settings for the S3 bucket
@@ -631,6 +740,37 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
 
         template = loader.get_template('instance/ansible/s3.yml')
         return template.render({'instance': self})
+
+    @property
+    def ansible_mysql_settings(self):
+        """
+        Ansible settings for the external mysql database
+        """
+        if self.use_ephemeral_databases or not settings.INSTANCE_MYSQL_URL_OBJ:
+            return ''
+
+        template = loader.get_template('instance/ansible/mysql.yml')
+        return template.render({'user': self.mysql_user,
+                                'pass': self.mysql_pass,
+                                'host': settings.INSTANCE_MYSQL_URL_OBJ.hostname,
+                                'port': settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+                                'database': self.mysql_database_name})
+
+    @property
+    def ansible_mongo_settings(self):
+        """
+        Ansible settings for the external mongo database
+        """
+        if self.use_ephemeral_databases or not settings.INSTANCE_MONGO_URL_OBJ:
+            return ''
+
+        template = loader.get_template('instance/ansible/mongo.yml')
+        return template.render({'user': self.mongo_user,
+                                'pass': self.mongo_pass,
+                                'host': settings.INSTANCE_MONGO_URL_OBJ.hostname,
+                                'port': settings.INSTANCE_MONGO_URL_OBJ.port or 27017,
+                                'database': self.mongo_database_name,
+                                'forum_database': self.forum_database_name})
 
     @property
     def studio_sub_domain(self):
@@ -653,6 +793,70 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         """
         return u'{0.protocol}://{0.studio_domain}/'.format(self)
 
+    @property
+    def database_name(self):
+        """
+        The database name used for external databases. Escape all non-ascii characters and truncate to 64 chars, the
+        maximum for mysql
+        """
+        name = self.domain.replace('.', '_')
+        allowed = string.ascii_letters + string.digits + '_'
+        escaped = ''.join(char for char in name if char in allowed)
+        return truncate_name(escaped, length=64)
+
+    @property
+    def mysql_database_name(self):
+        """
+        The mysql database name for this instance
+        """
+        return self.database_name
+
+    @property
+    def mysql_database_names(self):
+        """
+        List of mysql database names
+        """
+        return [self.mysql_database_name]
+
+    @property
+    def mongo_database_name(self):
+        """
+        The name of the main external mongo database
+        """
+        return self.database_name
+
+    @property
+    def forum_database_name(self):
+        """
+        The name of the external database used for forums
+        """
+        return '{0}_forum'.format(self.database_name)
+
+    @property
+    def mongo_database_names(self):
+        """
+        List of mongo database names
+        """
+        return [self.mongo_database_name, self.forum_database_name]
+
+    def save(self, **kwargs):
+        """
+        Set this instance's default field values
+        """
+        if self.use_ephemeral_databases is None:
+            self.use_ephemeral_databases = settings.INSTANCE_EPHEMERAL_DATABASES
+        super().save(**kwargs)
+
+    def update_from_pr(self, pr):
+        """
+        Update this instance with settings from the given pull request
+        """
+        super().update_from_pr(pr)
+        self.ansible_extra_settings = pr.extra_settings
+        self.use_ephemeral_databases = pr.use_ephemeral_databases(self.domain)
+        self.ansible_source_repo_url = pr.get_extra_setting('edx_ansible_source_repo')
+        self.configuration_version = pr.get_extra_setting('configuration_version')
+
     @log_exception
     def provision(self):
         """
@@ -661,7 +865,6 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         Returns: (server, log)
         """
         self.last_provisioning_started = timezone.now()
-        self.reset_ansible_settings(commit=True)
 
         # Server
         self.logger.info('Terminate servers')
@@ -678,7 +881,14 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         self.logger.info('Updating DNS: Studio at %s...', self.studio_domain)
         gandi.set_dns_record(type='CNAME', name=self.studio_sub_domain, value=self.sub_domain)
 
+        # Provisioning (external databases)
+        if not self.use_ephemeral_databases:
+            self.logger.info('Provisioning external databases...')
+            self.provision_mysql()
+            self.provision_mongo()
+
         # Provisioning (ansible)
+        self.reset_ansible_settings(commit=True)
         log, exit_code = self.deploy()
         if exit_code != 0:
             server.update_status(provisioning=True, failed=True)
@@ -693,3 +903,21 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         self.logger.info('Provisioning completed')
 
         return (server, log)
+
+    def provision_mysql(self):
+        """
+        Set mysql credentials and provision the database.
+        """
+        if not self.mysql_provisioned:
+            self.mysql_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
+            self.mysql_pass = get_random_string(length=32)
+        return super().provision_mysql()
+
+    def provision_mongo(self):
+        """
+        Set mongo credentials and provision the database.
+        """
+        if not self.mongo_provisioned:
+            self.mongo_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
+            self.mongo_pass = get_random_string(length=32)
+        return super().provision_mongo()
--- a/instance/tests/models/test_instance.py
+++ b/instance/tests/models/test_instance.py
@@ -24,11 +24,20 @@ OpenEdXInstance model - Tests
 
 import os
 import re
+import subprocess
 from mock import call, patch
+from urllib.parse import urlparse
+
+from django.conf import settings
+from django.test import override_settings
+
+import pymongo
+import yaml
 
 from instance.models.server import OpenStackServer, Server
 from instance.models.instance import InconsistentInstanceState, OpenEdXInstance
 from instance.tests.base import TestCase
+from instance.tests.factories.pr import PRFactory
 from instance.tests.models.factories.instance import OpenEdXInstanceFactory
 from instance.tests.models.factories.server import (
     StartedOpenStackServerFactory, BootedOpenStackServerFactory, ProvisioningOpenStackServerFactory,
@@ -370,10 +379,135 @@ class AnsibleInstanceTestCase(TestCase):
             self.assertEqual(returncode, 0)
 
 
+class MySQLInstanceTestCase(TestCase):
+    """
+    Test cases for MySQLInstanceMixin models
+    """
+    def check_mysql(self, instance):
+        """
+        Check that the mysql databases and user have been created, then remove them
+        """
+        self.assertIs(instance.mysql_provisioned, True)
+        self.assertTrue(instance.mysql_user)
+        self.assertTrue(instance.mysql_pass)
+        databases = subprocess.check_output("mysql -u root -e 'SHOW DATABASES'", shell=True).decode()
+        try:
+            for database in instance.mysql_database_names:
+                self.assertIn(database, databases)
+                mysql_cmd = "mysql -u {user} --password={password} -e 'SHOW TABLES' {db_name}".format(
+                    user=instance.mysql_user,
+                    password=instance.mysql_pass,
+                    db_name=database,
+                )
+                tables = subprocess.call(mysql_cmd, shell=True)
+                self.assertEqual(tables, 0)
+        finally:
+            for database in instance.mysql_database_names:
+                #pylint: disable=undefined-loop-variable
+                subprocess.check_call("mysql -u root -e 'DROP DATABASE IF EXISTS {0}'".format(database), shell=True)
+            subprocess.check_call("mysql -u root -e 'DROP USER {0}'".format(instance.mysql_user), shell=True)
+
+    def test_provision_mysql(self):
+        """
+        Provision mysql database
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_mysql()
+        self.check_mysql(instance)
+
+    @override_settings(INSTANCE_MYSQL_URL_OBJ=None)
+    def test_provision_mysql_no_url(self):
+        """
+        Don't provision a mysql database if INSTANCE_MYSQL_URL is not set
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_mysql()
+        databases = subprocess.check_output("mysql -u root -e 'SHOW DATABASES'", shell=True).decode()
+        for database in instance.mysql_database_names:
+            self.assertNotIn(database, databases)
+
+    def test_provision_mysql_weird_domain(self):
+        """
+        Make sure that database names are escaped correctly
+        """
+        sub_domain = 'really.really.really.really.long.subdomain'
+        base_domain = 'this-is-a-really-long-unusual-domain-แปลกมาก.com'
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False,
+                                          sub_domain=sub_domain,
+                                          base_domain=base_domain)
+        instance.provision_mysql()
+        self.check_mysql(instance)
+
+    def test_provision_mysql_again(self):
+        """
+        Only create the database once
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_mysql()
+        self.assertIs(instance.mysql_provisioned, True)
+
+        mysql_user = instance.mysql_user
+        mysql_pass = instance.mysql_pass
+        instance.provision_mysql()
+        self.assertEqual(instance.mysql_user, mysql_user)
+        self.assertEqual(instance.mysql_pass, mysql_pass)
+        self.check_mysql(instance)
+
+
+class MongoDBInstanceTestCase(TestCase):
+    """
+    Test cases for MongoDBInstanceMixin models
+    """
+    def check_mongo(self, instance):
+        """
+        Check that the instance mongo user has access to the external mongo database
+        """
+        mongo = pymongo.MongoClient(settings.INSTANCE_MONGO_URL)
+        for database in instance.mongo_database_names:
+            self.assertTrue(mongo[database].authenticate(instance.mongo_user, instance.mongo_pass))
+            mongo[database].remove_user(instance.mongo_user)
+
+    def test_provision_mongo(self):
+        """
+        Provision mongo databases
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_mongo()
+        self.check_mongo(instance)
+
+    def test_provision_mongo_no_url(self):
+        """
+        Don't provision any mongo databases if INSTANCE_MONGO_URL is not set
+        """
+        mongo = pymongo.MongoClient(settings.INSTANCE_MONGO_URL)
+        with override_settings(INSTANCE_MONGO_URL=None):
+            instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+            instance.provision_mongo()
+            databases = mongo.database_names()
+            for database in instance.mongo_database_names:
+                self.assertNotIn(database, databases)
+
+    def test_provision_mongo_again(self):
+        """
+        Only create the databases once
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_mongo()
+        self.assertIs(instance.mongo_provisioned, True)
+
+        mongo_user = instance.mongo_user
+        mongo_pass = instance.mongo_pass
+        instance.provision_mongo()
+        self.assertEqual(instance.mongo_user, mongo_user)
+        self.assertEqual(instance.mongo_pass, mongo_pass)
+        self.check_mongo(instance)
+
+
 class OpenEdXInstanceTestCase(TestCase):
     """
     Test cases for OpenEdXInstanceMixin models
     """
+    @override_settings(INSTANCE_EPHEMERAL_DATABASES=False)
     @patch('instance.models.instance.github.get_commit_id_from_ref')
     def test_create_defaults(self, mock_get_commit_id_from_ref):
         """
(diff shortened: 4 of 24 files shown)


omarkhan:
@antoviaque am I on the right track here?

I still need to update the `watch_pr` task to check for ephemeral databases. From what I can tell reading the code any settings included in the PR are included as extra ansible variables, is that correct? Is there any end-user documentation for this?

Also, what about elasticsearch? It should be possible to connect to an external elasticsearch cluster, right?

antoviaque:
@omarkhan Thanks for the WIP PR! I had a brief look, and yep, it looks like the right path to me. @smarnach will be your main reviewer (although he might only get around to review later in the week, he's currently busy on some tight deadlines), so I'll let him comment too, but imho you can keep going in that direction in the meantime.

antoviaque:
@omarkhan FYI, I re-run the build with the right auth, you have some pylint validation errors: https://circleci.com/gh/open-craft/opencraft/143

omarkhan:
Thanks @antoviaque, fixed in f12fcad3e8dfaa02e8ff3dd41a49e8f02e4557c9

omarkhan:
@antoviaque @smarnach this is done. That last commit is a minor refactor, I can revert it if you don't like it.

Where is the end-user documentation for this? It will have to be updated with the new settings.

antoviaque:
@omarkhan Thanks! We'll review it. For the documentation, that made me realize what we have currently is on a private repo - I'll move it to this repo. In the meantime, you can edit/add things in the README, I'll integrate it in the documentation on my side.

instance/github.py
```diff
@@ -148,6 +149,12 @@ class PR:
     """
     Representation of a GitHub Pull Request
     """
+    SETTINGS = {
+        'DATABASE_URL': 'DATABASE_URL',
+        'MONGO_URL': 'MONGO_URL',
+        'EPHEMERAL_DATABASES': 'EPHEMERAL_DATABASES',
+    }
```
smarnach: I think you don't need this map.  The setting names only appear a single time in the properties below.  The properties already form an abstraction layer, so we don't need another level of abstraction.  (If you want to keep the dict anyway for some reason, make sure to update the tests to use this dict instead of hard-coding the setting names there.)

instance/github.py
```diff
@@ -157,8 +164,39 @@ def __init__(self, number, fork_name, branch_name, title, username, body=''):
...
     def extra_settings(self):
         """
-        Extra settings contained in the PR body
+        Extra ansible settings from the PR body, as a yaml string
         """
-        return get_settings_from_pr_body(self.body)
+        if not self.settings:
+            return ''
+        return yaml.dump({key: value for key, value in self.settings.items()
+                          if key not in self.SETTINGS.values()})
```
smarnach: What is the rationale for this change?  Normalising the format?

Note that the new version will parse the body twice – one time for each invocation of `self.settings`.  It might be worthwhile to cache the settings once they got evaluated to avoid parsing them in every invocation of the property.

instance/models/instance.py
```diff
@@ -373,8 +398,7 @@ def set_to_branch_tip(self, branch_name=None, ref_type=None, commit=True):
 
             # Update the hash in the instance title if it is present there
             # TODO: Find a better way to handle this - include the hash dynamically?
-            # TODO: Figure out why the warnings aren't suppressed despite the fact that it's a mixin
-            if self.name and old_commit_short_id: #pylint: disable=access-member-before-definition
+            if self.name and old_commit_short_id:
```
smarnach: Just out of curiosity – any idea what changed so that we don't need to suppress this warning anymore?

instance/models/instance.py
```diff
@@ -509,11 +550,26 @@ def default_fork(self):
...
+        Set this to True to unset any external database settings
+        """
+        if value is True:
+            self.database_url = ''
+            self.mongo_url = ''
+        elif value is False:
+            if self.ephemeral_databases is not False:
+                raise ValueError('Set database_url or mongo_url')
+        else:
+            raise TypeError('Expected bool')
```
smarnach: Making a dependent property writable strikes me as a somewhat surprising interface.  Given that this setting can only be meaningfully changed from `False` to `True`, I think it would be clearer to simply set the attributes this depends on, i.e. `database_url` and `mongo_url`, directly.

instance/models/instance.py
```diff
@@ -547,6 +639,19 @@ def studio_url(self):
...
+        """
+        Update this instance with settings from the given pull request
+        """
+        super().update_from_pr(pr)
+        if pr.database_url:
+            self.database_url = pr.database_url
+        if pr.mongo_url:
+            self.mongo_url = pr.mongo_url
+        if pr.ephemeral_databases is not None:
+            self.ephemeral_databases = pr.ephemeral_databases
```
smarnach: This could throw an exception if the user changed this from `True` to `False` without also changing one of the database URLs.

smarnach:
@omarkhan I had a look at this PR, and the code is looking clean and well-tested.  I added comments for a few details – they are suggestions, so if you don't like anything just reply.

This pull request allows to set a default database for all instances.  My understanding of the feature we wanted to implement is that we can configure a default database _server_, and the instance manager would create a new database on this server for each PR and automatically point the instance to this newly created database, which would be automatically reused when the instance is reprovisioned.  With the old code we were already able to manually point to external databases, but we would have to manually create them.

This might be a misunderstanding on my side.  @antoviaque can you please comment?

instance/github.py
```diff
@@ -157,8 +164,39 @@ def __init__(self, number, fork_name, branch_name, title, username, body=''):
...
+        Settings contained in the PR body
+        """
+        return yaml.load(get_settings_from_pr_body(self.body)) or {}
+
+    @property
+    def database_url(self):
+        """
+        An optional database url, from the DATABASE_URL setting in the PR body
+        """
+        return self.settings.get(self.SETTINGS['DATABASE_URL'])
```
smarnach: I'm not entirely sure I like conflating the Ansible settings with other non-Ansible settings.  It might be better to keep them separate.  (Namespace conflation always starts with small steps like this one, and one day you end up with GNU make. :) )

antoviaque:
> This pull request allows to set a default database for all instances. My understanding of the feature we wanted to implement is that we can configure a default database server, and the instance manager would create a new database on this server for each PR and automatically point the instance to this newly created database, which would be automatically reused when the instance is reprovisioned. With the old code we were already able to manually point to external databases, but we would have to manually create them.
> 
> This might be a misunderstanding on my side. @antoviaque can you please comment?

@smarnach You understood correctly - separate instances would need to use different databases, and the instance manager needs to create them for each instance, using the DB server credentials from the settings.

instance/github.py
antoviaque: +1

instance/github.py
```diff
@@ -157,8 +164,39 @@ def __init__(self, number, fork_name, branch_name, title, username, body=''):
...
 
     @property
+    def settings(self):
+        """
+        Settings contained in the PR body
+        """
+        return yaml.load(get_settings_from_pr_body(self.body)) or {}
+
+    @property
+    def database_url(self):
```
antoviaque: Could we improve the consistency of the configuration variable names? mongodb is a database too. Maybe `RELATIONAL_DATABASE_URL` and `DOCUMENT_DATABASE_URL`? Or simply `MYSQL_URL` and `MONGO_URL`.

instance/github.py
antoviaque: +1

instance/github.py
```diff
@@ -157,8 +164,39 @@ def __init__(self, number, fork_name, branch_name, title, username, body=''):
...
+        """
+        An optional database url, from the DATABASE_URL setting in the PR body
+        """
+        return self.settings.get(self.SETTINGS['DATABASE_URL'])
+
+    @property
+    def mongo_url(self):
```
antoviaque: Also, there are two mongo DBs for each instance. One for edxapp/modulestore, and one for the forums.

instance/github.py
antoviaque: Also, could you split this into several lines? Conciseness is good, but not when it gets into the way of readability - here you're packing several distinct operations into a single line.

instance/models/instance.py
```diff
@@ -217,6 +220,21 @@ class GitHubInstanceQuerySet(models.QuerySet):
     Additional methods for instance querysets
     Also used as the standard manager for the GitHubInstance model
     """
+    def update_or_create_from_pr(self, pr):
+        """
+        Create or update an instance for the given pull request
+        """
+        pr_sub_domain = 'pr{number}.sandbox'.format(number=pr.number)
```
antoviaque: Is there a way to avoid logic specific to sandboxes in instances models/managers? (ie the domain name, here) Sandboxes are a specific use case of instances, there might be PRs that aren't related to sandboxes in the future; like a PR against a release branch used for a production instance, for example.

instance/models/instance.py
```diff
@@ -398,6 +422,16 @@ def set_fork_name(self, fork_name, commit=True):
         if commit:
             self.save()
 
+    def update_from_pr(self, pr):
+        """
+        Update this instance with settings from the given pull request
+        """
+        truncated_title = truncatewords(pr.title, 4)
+        #pylint: disable=attribute-defined-outside-init
```
antoviaque: It's only for the self.name attribute, right? Can this be limited to a single line, to avoid disabling it also for the next one?

instance/models/instance.py
```diff
@@ -495,7 +529,14 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
     s3_secret_access_key = models.CharField(max_length=50, blank=True)
     s3_bucket_name = models.CharField(max_length=50, blank=True)
 
-    ANSIBLE_SETTINGS = AnsibleInstanceMixin.ANSIBLE_SETTINGS + ['ansible_s3_settings']
+    database_url = models.CharField(max_length=255, blank=True, default=settings.INSTANCE_DATABASE_URL or '')
```
antoviaque: Using an application setting as the default causes migrations to be generated whenever the current setting differs. See https://github.com/open-craft/opencraft/pull/7#discussion_r40174892

instance/models/instance.py
antoviaque: +1

instance/models/instance.py
```diff
@@ -527,6 +583,42 @@ def ansible_s3_settings(self):
...
+
+        mongo = urlparse(self.mongo_url)
+        ansible_settings = {
+            'EDXAPP_MONGO_USER': mongo.username,
+            'EDXAPP_MONGO_PASSWORD': mongo.password,
+            'EDXAPP_MONGO_HOSTS': [mongo.hostname] if mongo.hostname else None,
+            'EDXAPP_MONGO_PORT': mongo.port,
+            'EDXAPP_MONGO_DB_NAME': mongo.path.lstrip('/'),
+        }
+        return yaml.dump({key: value for key, value in ansible_settings.items() if value})
```
antoviaque: Duplicated code - to factorize into a separate function?

instance/tests/test_tasks.py
```diff
@@ -59,13 +63,14 @@ def test_watch_pr_new(self, mock_get_username_list, mock_get_pr_list_from_userna
         New PR created on the watched repo
         """
         mock_get_username_list.return_value = ['itsjeyd']
-        pr = github.PR(
+        settings = 'WATCH: true\r\nDATABASE_URL: mysql://db.opencraft.com\r\nMONGO_URL: mongo://mongo.opencraft.com\r\n'
```
antoviaque: To avoid putting too much into this test, can you keep it to the "fake" `WATCH: true` content, and create separate tests for additional logic?

instance/tests/test_tasks.py
```diff
@@ -77,7 +82,38 @@ def test_watch_pr_new(self, mock_get_username_list, mock_get_pr_list_from_userna
         self.assertEqual(instance.fork_name, 'watched/fork')
         self.assertEqual(instance.github_pr_number, 234)
         self.assertEqual(instance.branch_name, 'watch-branch')
-        self.assertEqual(instance.ansible_extra_settings, 'WATCH: true\r\n')
+        self.assertEqual(instance.database_url, 'mysql://db.opencraft.com')
+        self.assertEqual(instance.mongo_url, 'mongo://mongo.opencraft.com')
+        self.assertEqual(yaml.load(instance.ansible_extra_settings), {'WATCH': True})
```
antoviaque: That answers @smarnach 's question in https://github.com/open-craft/opencraft/pull/11/files#r41064245 - I'm not sure I like this additional magic, of removing variables from the user defined `ansible_extra_settings`. Can we instead use the order of precedence that comes from `ANSIBLE_SETTINGS` to determine which value we end up passing to ansible? https://github.com/open-craft/opencraft/pull/11/files#diff-848aab5153020ca3817d6db9d4d3a144R535

instance/tests/test_tasks.py
```diff
@@ -77,7 +82,38 @@ def test_watch_pr_new(self, mock_get_username_list, mock_get_pr_list_from_userna
...
+    def test_watch_pr_ephemeral_databases(self, mock_get_username_list,
+                                          mock_get_pr_list_from_username,
+                                          mock_provision_instance,
+                                          mock_get_commit_id_from_ref):
+        """
+        Asking for ephemeral databases in the PR should unset external
+        database settings, even if the default is to use external databases
+        """
+        mock_get_username_list.return_value = ['itsjeyd']
+        settings = 'EPHEMERAL_DATABASES: yes\r\n'
```
antoviaque: Doesn't that differ from the specs?

https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9541143-preserve-databases-during-sandbox-reprovisioning

> It should be possible to disable it from a PR description, by adding the string "ephemeral database" on a line also containing the sandbox domain name - ie it would for example be enabled by:
> 
> `* [LMS](http://sandbox.example.com) (ephemeral database)`

omarkhan: Yes I went off spec a bit there. Could you send me some of the end-user docs for this, so I can get more of an understanding of how it is meant to be used?

omarkhan:
Thanks for the feedback @antoviaque @smarnach. I think I misunderstood the requirements here. I will fix all this up during the week.

instance/tests/test_tasks.py
antoviaque: @omarkhan Yup - it will need to be reformatted a little to be included in the repo, to be made more generic, but in the meantime I copied the current doc there: https://gist.github.com/antoviaque/8f989565138f05d08aaf

instance/github.py
omarkhan: Got it, thanks. Are there any other DBs in there I should know about?

What about elasticsearch? As far as I can tell it is used by the forum and the notes api.

antoviaque: I think you would be good with these 3 - for elasticsearch we currently don't enable it on the sandboxes, so you can ignore it for now.

omarkhan:
@antoviaque the spec mentions mongolab, but as far as I know they do not provide a way to create new databases programmatically. This is the case with most hosted database services in my experience.

Can I assume that the instance manager will have ssh access to the database servers to create new persistent databases? If so I can use the existing ansible roles to provision the databases. If not I will access the databases directly.

smarnach:
@omarkhan Doesn't MongoDB automatically create a new database the first time you try to connect to it?  Of course this would require sufficient permissions for the connecting user, so I'm not sure whether this works on the shared instances on Mongolab, though.

antoviaque:
@omarkhan @smarnach I was thinking of using the dedicated plans, which allow multiple databases - but good point, checking the doc it seem that even in that case, they don't allow to create databases programmatically - they don't provide a user with rights to create DBs, and their API doesn't seem to provide a way to do this either - unless I missed something, they actually require to go on their interface and click on buttons to do this. The mongo world never cease to amaze me.

So, yes, it looks like we'll have to run our own here. : ( I wouldn't want to assume SSH access, but you could assume a global mongo user with rights to create new DBs (which would only be used to create the local permission collection & the mongo user for a specific instance.

Sounds good?

omarkhan:
@antoviaque that sounds fine, but if we have to host these dbs ourselves anyway, then why not just use the existing ansible roles to provision them? Otherwise we end up duplicating logic that's already in the playbooks.

See https://github.com/omarkhan/opencraft/commit/fd9fc6ee8f142bdfaf84761a0f8a357405f1628c

omarkhan:
@antoviaque also, can you clarify this from the spec:

> It should be possible to disable it from a PR description, by adding the string "ephemeral database" on > a line also containing the sandbox domain name - ie it would for example be enabled by:
> 
> * [LMS]\(http://sandbox.example.com) (ephemeral database)

Is that line with the sandbox domain name something that contributors are already using? I can't find any mention of it in the docs you sent me, or in the code.

omarkhan:
@antoviaque one more question:

I am trying to write integration tests for this. I would like to actually make requests to the sandbox once it has been provisioned, to check that things are working correctly. I am trying to do that with the current integration test. I have provisioned it, it lives at <uid>.integration.plebia.net. When I try to make requests to that url, it can't resolve dns. When I try to hit the server's public ip directly, the connection is reset by peer.

What am I doing wrong here?

antoviaque:
@omarkhan I would be fine with using existing roles, and I agree that it would be good to be able to use them. However, from a security standpoint I'd rather limit the rights granted to the instance manager to the minimum, giving ssh access (and likely requiring sudo rights?) increases the potential attack vectors. I haven't looked at how the existing roles do the work though - maybe see if there is a way to use them (or refactor/split them) to allow to create the DBs using them, but remotely without SSH access? Not sure how much sense that would make.

For the domain name, it isn't in the doc, as this is currently not used for any configuration - but it's being added to PRs that we create, to point reviewers to the sandbox - see for example https://github.com/edx/edx-platform/pull/9903 or https://github.com/edx/edx-platform/pull/9848 - the idea is to add the option on the same line.

For the DNS, note that they can take time to propagate - once the logs show that they have been set, you still need to wait ~20-30min before they start resolving (sometimes more if you try to resolve it before it is in place on the Gandi NS servers, since your DNS will cache the result). The public IP should answer correctly though - are you sure you waited for the instance to be booted? Or, if you do this at the end, that the server hasn't been deprovisioned? In any case, if ansible is able to access it, you should be able to access it too - maybe also check that you're using the right private key? Although if that was the issue, the message from SSH should be more explicit.

smarnach:
@omarkhan I deployed a MongoDB instance for testing for you – it's available at 104.196.6.206, and you should be able to ssh in to it using the username "omarkhan" and any of the identities corresponding to your Github public keys.  It's deployed on Google Compute Engine using [Click-to-deploy](https://cloud.google.com/solutions/mongodb/click-to-deploy) – feel free to do reconfigure it in whatever way you want.  Let me know when you are done, I'll delete the instance again.

omarkhan:
Thanks @smarnach.

@antoviaque thanks for clarifying, that makes sense. I was trying to connect to the sandbox over http not ssh. From reading the playbooks I assumed the sandbox would be serving the LMS. I will try to ssh in and see what's going on.

omarkhan:
@antoviaque @smarnach this is ready for another look now.

I can't ssh into the instances I create, because I don't have the right ssh key. The ansible-playbook run fails silently because of this, but the integration test still passes (see https://github.com/open-craft/opencraft/pull/18).

Without access to the instances I can't really do any integration testing, but I have tried to keep good unit test coverage.

antoviaque:
@omarkhan Thanks! I'll let @smarnach review, and on my side I'll likely do another pass this week-end once Sven is good with it (one of the rare moments where I get uninterrupted time to look at this calmly : ) ).

For the SSH key, this is actually a setting in the env file. If you are running the integration tests yourself (vs relying on the circleci build), you can edit your `.env.integration` to use another key (to create beforehand in the RunAbove Horizons panel, or via the OpenStack commands).

instance/models/instance.py
```diff
@@ -586,4 +669,19 @@ def provision(self):
...
+        if settings.INSTANCE_MYSQL_URL_OBJ:
+            connection = mysql.connect(
+                host=settings.INSTANCE_MYSQL_URL_OBJ.hostname,
+                user=settings.INSTANCE_MYSQL_URL_OBJ.username,
+                passwd=settings.INSTANCE_MYSQL_URL_OBJ.password or '',
+                port=settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+            )
+            cursor = connection.cursor()
+            database_name = connection.escape_string(self.database_name).decode()
+            cursor.execute('CREATE DATABASE IF NOT EXISTS {0}'.format(database_name))
```
smarnach: I think you can leave escaping the DB name to the database driver – something like

    cursor.execute('CREATE DATABASE IF NOT EXISTS %s', [self.database_name])

or similar.  Not sure whether that works in this particular case, and I haven't checked the details.

.env.test
```diff
@@ -1,6 +1,7 @@
 DEBUG=true
 SECRET_KEY='tests'
 DATABASE_URL='postgres://localhost/opencraft'
+INSTANCE_MYSQL_URL='mysql://root@localhost'
```
smarnach: Do we also need INSTANCE_MONGO_URL here?

smarnach:
@omarkhan Thanks for the update.  This looks good now!  I haven't actually tested the changes manually, but couldn't find anything to complain about by looking at the code.

I think there is one thing still outstanding – creating a new MongoDB user for each database (see [Xavier's comment above](https://github.com/open-craft/opencraft/pull/11#issuecomment-145763739).  The user configured in INSTANCE_MONGO_URL could be used to create a new user for each instance, with permissions to only access the databases belonging to that instance.  I think creating the user is all that's needed – the databases and collections will be implicitly created by the LMS and the forums.

instance/models/instance.py
omarkhan: I actually tried that first, it doesn't work for database names. The driver escapes it incorrectly. I will add a comment explaining this

.env.test
omarkhan: It's not currently needed for the tests. But I will add it when I write the logic for creating mongo users

omarkhan:
Thanks @smarnach - I will finish this tomorrow

omarkhan:
@smarnach this is done. If you are happy with it I will rebase it

smarnach:
@omarkhan Very nice, the additional changes are looking good.  I didn't even find a minor thing to complain about, so :+1: from my side.  @antoviaque Will do a final pass on this in the weekend.

smarnach:
@omarkhan I just noticed the tests didn't pass, but the error is a connection reset while downloading the requirements.  I restarted the tests now in the hope that this was a transient issue – if they fail again please check whether you may have caused the failure.

omarkhan:
@smarnach the tests all pass at my end

Vagrantfile
```diff
@@ -11,8 +11,9 @@ cd /vagrant
 grep -Fq 'cd /vagrant' ~/.bashrc || echo 'cd /vagrant' >> ~/.bashrc
 
 # Install system packages
+export DEBIAN_FRONTEND=noninteractive
 sudo apt-get update --quiet
-sudo apt-get install -y $(cat debian_packages.lst) postgresql
+sudo -E apt-get install -y $(cat debian_packages.lst) postgresql mysql-server mongodb-server
```
antoviaque: @omarkhan These additional dependencies would be missing for installations not using Vagrant.

instance/github.py
```diff
@@ -86,6 +87,18 @@ def get_settings_from_pr_body(pr_body):
         return ''
 
 
+def get_ephemeral_databases_from_pr_body(pr_body):
```
antoviaque: The name of this method makes it look like it returns databases objects - maybe something like `is_pr_body_requesting_ephemeral_databases()`?

instance/github.py
```diff
@@ -86,6 +87,18 @@ def get_settings_from_pr_body(pr_body):
...
+    """
+    Return True if the PR body specified that the sandbox should use ephemeral
+    databases, False if it specifies persistent databases, or None otherwise
+    """
+    if re.search(r'..Sandbox:?...*\(ephemeral databases?\)', pr_body):
```
antoviaque: @omarkhan This still doesn't match the specs:

> It should be possible to disable it from a PR description, by adding the string "ephemeral database" on a line also containing the **sandbox domain name** - ie it would for example be enabled by:
> 
> * [LMS](http://sandbox.example.com) (ephemeral database)

https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9541143-preserve-databases-during-sandbox-reprovisioning

The addition of "persistent databases" is a good idea; on the other side, checking against "sandbox" wouldn't match the line from the example above - plus it adds logic specific to sandboxes again. I know you likely did this because you don't have the instance domain name on the PR object, but this could be passed on as an argument.

instance/github.py
```diff
@@ -86,6 +87,18 @@ def get_settings_from_pr_body(pr_body):
...
+        return True
+    if re.search(r'..Sandbox:?...*\(persistent databases?\)', pr_body):
+        return False
```
antoviaque: @omarkhan Also, can you add instructions in the README about this? I know currently the rest of the end-user documentation is missing, so you don't need to integrate the rest of the gist I was giving, but this would still need to be documented somewhere.

instance/github.py
```diff
@@ -157,6 +170,20 @@ def __init__(self, number, fork_name, branch_name, title, username, body=''):
...
+        This PR's title truncated to 4 words
+        """
+        return truncatewords(self.title, 4)
+
+    @property
+    def sub_domain(self):
+        """
+        The subdomain for this PR
+        """
+        return 'pr{number}.sandbox'.format(number=self.number)
```
antoviaque: @omarkhan Similar remark to https://github.com/open-craft/opencraft/pull/11#discussion_r41085163 - PRs might not all be for sandboxes, so I'd rather stay clear of adding logic specific to sandboxes there too, at least for now. Can we keep the choice of domain name in the task that creates the instance?

instance/github.py
```diff
@@ -164,6 +191,13 @@ def extra_settings(self):
         return get_settings_from_pr_body(self.body)
 
     @property
+    def ephemeral_databases(self):
```
antoviaque: Same remark about the name as in the function above. `use_ephemeral_databases()`?

instance/models/instance.py
```diff
@@ -217,6 +222,19 @@ class GitHubInstanceQuerySet(models.QuerySet):
     Additional methods for instance querysets
     Also used as the standard manager for the GitHubInstance model
     """
+    def update_or_create_from_pr(self, pr):
```
antoviaque: @omarkhan To preserve your refactor here despite my remarks about keeping the choice of domain name in the task, you could add the domain name to this method:

```
def update_or_create_from_pr(self, pr, sub_domain):
```

instance/models/instance.py
```diff
@@ -495,7 +527,15 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
     s3_secret_access_key = models.CharField(max_length=50, blank=True)
     s3_bucket_name = models.CharField(max_length=50, blank=True)
 
-    ANSIBLE_SETTINGS = AnsibleInstanceMixin.ANSIBLE_SETTINGS + ['ansible_s3_settings']
+    ephemeral_databases = models.BooleanField()
```
antoviaque: @omarkhan `use_ephemeral_databases`, to show that it's a boolean rather than DB objects?

instance/models/instance.py
```diff
@@ -547,6 +611,48 @@ def studio_url(self):
         """
         return u'{0.protocol}://{0.studio_domain}/'.format(self)
 
+    @property
+    def database_name(self):
+        """
+        The database name used for external databases
+        """
+        return self.domain.replace('.', '_')
```
antoviaque: @omarkhan Wouldn't that require more sanitizing? For example, to take into account the maximum size of the database name (64 for MySQL for example)? And are there other characters than '.' that might be allowed in domain names, but not in database names?

instance/models/instance.py
```diff
@@ -586,4 +697,41 @@ def provision(self):
...
         return (server, ansible_log)
 
+    def provision_external_databases(self):
+        """
+        Provision external databases for this instance
+        """
+        self.provision_external_mysql()
+        self.provision_external_mongo()
+
+    def provision_external_mysql(self):
```
antoviaque: @omarkhan Would it be worth moving some of these methods to mixin classes? This doesn't seem specific to Open edX instances.

instance/models/instance.py
```diff
@@ -586,4 +697,41 @@ def provision(self):
...
+                host=settings.INSTANCE_MYSQL_URL_OBJ.hostname,
+                user=settings.INSTANCE_MYSQL_URL_OBJ.username,
+                passwd=settings.INSTANCE_MYSQL_URL_OBJ.password or '',
+                port=settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+            )
+            cursor = connection.cursor()
+
+            # We can't use the database name in a parameterized query, the
+            # driver doesn't escape it properly. Se we escape it here instead
+            database_name = connection.escape_string(self.database_name).decode()
```
antoviaque: @omarkhan Is there a way to not need this escaping, to be able to use the `database_name` directly everywhere? For example by only chosing characters that don't need escaping when chosing the base database name, initially?

Also, since there are several databases, it's not obvious from the attribute name that `self.database_name` is the name of the MySQL database. The way you build DB names out of a base DB name by appending suffixes is a good approach imho, but to be consistent the MySQL DB name should also be built the same way - ie having a `self.mysql_database_name` attribute.

instance/models/instance.py
```diff
@@ -586,4 +697,41 @@ def provision(self):
...
+
+            cursor.execute('CREATE DATABASE IF NOT EXISTS {0}'.format(database_name))
```
antoviaque: @omarkhan What if the DB exists, but is already used by another instance? Could we check for DB existence before creating it the first time?

instance/models/instance.py
```diff
@@ -586,4 +697,41 @@ def provision(self):
...
+            cursor.execute('GRANT ALL ON {0}.* TO %s IDENTIFIED BY %s'.format(database_name),
+                           (self.database_user, self.database_pass))
+
+    def provision_external_mongo(self):
+        """
+        Provision external mongo databases for this instance
+        """
+        if settings.INSTANCE_MONGO_URL:
+            mongo = pymongo.MongoClient(settings.INSTANCE_MONGO_URL)
+            for database in (self.database_name, self.forum_database_name):
```
antoviaque: @omarkhan Ah, I see, you actually reuse the same DB name for both MySQL and mongo - just using a suffix for the 2nd DB for a given driver. That makes sense, actually - I would be fine keeping it, if you find a way to address my comment about sanitizing the db name in the `self.database_name` property to not need escaping, in a way that works across DB.

instance/tests/models/test_instance.py
```diff
@@ -336,10 +344,12 @@ def test_run_playbook(self, mock_open_repo, mock_run_playbook, mock_inventory, m
         ), mock_run_playbook.mock_calls)
 
 
+#pylint: disable=too-many-public-methods
```
antoviaque: @omarkhan Would it be worth splitting this `TestCase` class into several ones, rather than silencing the warning?

.env.test
```diff
@@ -1,6 +1,8 @@
...
+INSTANCE_MONGO_URL='mongodb://localhost'
```
antoviaque: @omarkhan This causes to require MySQL and Mongo to be installed locally on the instance VM for the tests to pass - while it seem right to require the instance manager's own DB to be accessible for tests, the DBs used by instances themselves are a different type, and wouldn't be expected to be on the instance manager's own VM. If we add other instance types in the future, which could themselves require other DB types, that could make the instance manager VM environment pretty unwieldy.

Looking at the tests that use this below, the ones that require this look more like integration tests - would it make sense to move the tests that require these DBs to be setup to integration tests?

instance/tests/models/test_instance.py
```diff
@@ -436,3 +577,96 @@ def update_status(self, provisioned=False, rebooting=False):
...
+        instance.provision_external_databases()
+        self.assertEqual(mock_provision_mysql.call_count, 1)
+        self.assertEqual(mock_provision_mongo.call_count, 1)
+
+    def test_provision_external_mysql(self):
+        """
+        Provision external mysql database
+        """
+        instance = OpenEdXInstanceFactory()
+        instance.provision_external_mysql()
```
antoviaque: @omarkhan Cf my comment above about moving tests that require mysql and mongo to be installed to integration tests. The unit tests above that mock the actual calls to the DBs are fine to keep here - and you could also keep this one and the others below by mocking the actual calls to the mysql/mongo drivers.

instance/tests/models/test_instance.py
```diff
@@ -436,3 +577,96 @@ def update_status(self, provisioned=False, rebooting=False):
...
+
+        # Has the database been created?
+        databases = subprocess.check_output("mysql -u root -e 'SHOW DATABASES'", shell=True).decode()
+        self.assertIn(instance.database_name, databases)
+
+        try:
+            # Can the database user access it?
+            #pylint: disable=missing-format-attribute
```
antoviaque: @omarkhan To avoid this, you could split the next line into two:

```python
mysql_cmd = 'mysql -u {user} --password={password} -e 'SHOW TABLES' {db_name}".format(
    user=instance.database_user,
    password=instance.database_pass,
    db_name=instance.database_name
)
tables = subprocess.call(mysql_cmd, shell=True)
```

instance/utils.py
```diff
@@ -70,3 +76,10 @@ def get_requests_retry(total=10, connect=10, read=10, redirect=10, backoff_facto
         redirect=redirect,
         backoff_factor=backoff_factor
     )
+
+
+def random_string(alphabet, length):
+    """
+    Returns a random string of the given length
+    """
+    return ''.join(random.choice(alphabet) for i in range(length))
```
antoviaque: @omarkhan Could you write a test for this?

antoviaque:
@omarkhan Thanks for all the changes? It's looking good overall. I still had a fair number of comments, but they are more on details, and I do feel we're getting close to a merge-able state.

instance/utils.py
omarkhan: Done in a52f27a

.env.test
omarkhan: They are technically integration tests, but in terms how easy and quick they are to run they are much closer to the existing unit tests than the existing integration tests:

- They run in well under a second, vs several minutes for the integration tests
- No need for any kind of credentials (openstack, etc)
- No need for any manual setup when using vagrant

Installing the databases on the instance VM is just for convenience. I assume that the instance manager's own postgres database is not running on the same VM in production either.

If we add new instance types with other databases in the future, the setup can easily be automated with vagrant. This automation should keep things from getting unwieldy.

omarkhan:
@antoviaque I have made the changes you suggested, except for the one about making the external db tests integration tests - reasons given inline

.env.test
omarkhan: Reading your comment again I realise you might be talking about CI. If I remember correctly CircleCI provides mysql and mongodb on its standard containers, so it should be enough just to add the `INSTANCE_MYSQL_URL` and `INSTANCE_MONGO_URL` environment variables to the test environment.

instance/models/instance.py
```diff
@@ -547,6 +685,80 @@ def studio_url(self):
...
+    @property
+    def database_name(self):
+        """
+        The database name used for external databases. Escape all non-ascii characters and truncate to 64 chars, the
+        maximum for mysql
+        """
+        name = self.domain.replace('.', '_')
+        allowed = string.ascii_letters + string.digits + '_'
+        escaped = ''.join(char for char in name if char in allowed)
+        return escaped[:64]
```
smarnach: You could also use `django.db.backends.utils.truncate_name()` here, which includes a hash in the name in case it gets shortened, thereby reducing the chances of collisions (though I guess that this functionality will only be needed in rather arcane cases, which will come with problems we didn't think about anyway).

.env.test
smarnach: Using `localhost` here is a bit confusing, since it will provision the databases on the machine that runs the instance manager, but then tell the instances the database is at `localhost`, i.e. on the host of the instances themselves.  These configuration options will have to point to a hostname that resolves from both the instance manager and the instances to the same database server (though `localhost` might work fine for testing).

smarnach:
You addition of

    export DEBIAN_FRONTEND=noninteractive

to the Vagrantfile seems to have got lost in some merge, resulting in the Vagrant provisioning to fail.

.env.test
omarkhan: I assumed that the hostname can be resolved from both the instance manager and the instances. If that is not the case I can add another setting. `localhost` here is just for testing.

instance/models/instance.py
omarkhan: That's a useful function, I didn't know about it. I have made the change, although I think collisions here are pretty unlikely in practice.

omarkhan:
@smarnach good catch in the Vagrantfile. Fixed in d133f7e

.env.test
antoviaque: @omarkhan Ok for me to keep them on the vagrant instance. I'm still a bit wary of the long term effects of that decision though, as even if it does work relatively well in the current case, it still blurs the boundaries between the instances environment and the instance manager's one. But we can wait and see how that goes.

One thing that I would ask though, is to add at least one integration test for this feature - when we add features, or bugfixes around it, if there is no existing integration test for this, it will make things harder for the person who has to add it then, because there would be no existing infrastructure for it. It will also ensure that we keep deciding, on a case by case basis, whether changes around this feature are better tested by a unit or an integration test.

instance/github.py
```diff
@@ -86,6 +87,19 @@ def get_settings_from_pr_body(pr_body):
...
 
+def is_pr_body_requesting_ephemeral_databases(pr_body, domain):
+    """
+    Return True if the PR body specified that the sandbox should use ephemeral
+    databases, False if it specifies persistent databases, or None otherwise
+    """
+    escaped_domain = re.escape(domain)
+    if re.search(r'{0}.*\(ephemeral databases?\)'.format(escaped_domain), pr_body):
+        return True
+    if re.search(r'{0}.*\(persistent databases?\)'.format(escaped_domain), pr_body):
```
antoviaque: Can you remove the parenthesis in the regex? This will allow to have several options on the same line, like `... (ephemeral databases, other option)`

antoviaque:
@omarkhan Thanks for all the changes! I posted a couple of comments above, but the code looks all good for me now, I'm :+1: on it once these two comments are addressed. Thanks for all the good work and sticking through the review, the result is great.

You will also need to rebase your branch against the current master, and squash your commits into one/several atomic commits. We'll also need to test it once it is rebased, but assuming we don't find issues, it should be good to merge I think.

instance/models/instance.py
```diff
@@ -547,6 +686,78 @@ def studio_url(self):
...
+        self = instance
+
+        if self.use_ephemeral_databases is None:
+            self.use_ephemeral_databases = settings.INSTANCE_EPHEMERAL_DATABASES
+        for attr in ('mysql_user', 'mongo_user'):
+            if not getattr(self, attr, None):
+                setattr(self, attr, get_random_string(length=16, allowed_chars=string.ascii_lowercase))
+        for attr in ('mysql_pass', 'mongo_pass'):
+            if not getattr(self, attr, None):
+                setattr(self, attr, get_random_string(length=32))
```
smarnach: We should refrain from filling in the database users and passwords in the case that `use_ephemeral_databases` is `True`.  Otherwise, we will have values for these fields in the database that are completely meaningless and different from the values that are actually used on the deployed instance, which is a bit confusing.

smarnach: Thinking about it another time, the most natural thing to do might be to generate the usernames and passwords when the databases are first provisioned.  At that point, they start having a meaning, so it makes sense to keep them in the DB.  Even when the user temporarily switches to ephemeral databases, the usernames and passwords would be kept so they are still available when switching back to the external databases.  We would also leave the `mysql_provisiones`/`mongo_provisioned` flags set to indicate the databases already exist.

smarnach:
@omarkhan I've manually tested this PR rather thoroughly, and it worked fine.  I added one further comment for you to consider, but I'm fine with the PR either way.  It just needs to be rebased again.

omarkhan:
@antoviaque @smarnach thanks for the review. I will be away next week, I will add integration tests and rebase when I get back.

instance/models/instance.py
```diff
@@ -480,9 +513,85 @@ def run_playbook(self):
...
+                passwd=settings.INSTANCE_MYSQL_URL_OBJ.password or '',
+                port=settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+            )
+            cursor = connection.cursor()
+
+            for database in self.mysql_database_names:
+                # We can't use the database name in a parameterized query, the
+                # driver doesn't escape it properly. Se we escape it here instead
+                database_name = connection.escape_string(database).decode()
+                cursor.execute('CREATE DATABASE {0}'.format(database_name))
```
mtyaka: Can we make this `'CREATE DATABASE {0} DEFAULT CHARACTER SET utf8'`? It depends on MySQL settings, but by default the character set is not unicode friendly and I had issues when not specifying the default character set on RDS.

mtyaka:
Nice work @omarkhan. When using the local mysql database, there are some tasks in the ansible edxlocal role (https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/tasks/main.yml) that create the databases and MySQL users. It only works for local MySQL installations, though.

I think it would make sense to modify the existing tasks from the `configuration` repository (and probably move them out from `edxlocal` into a separate role) so that they also work with external databases instead of adding code to create the databases to the `opencraft` codebase. What do you think?

omarkhan:
@mtyaka that is what I tried initially - see https://github.com/open-craft/opencraft/pull/11#issuecomment-145763739. There are a few downsides:

- Getting anything merged in the `configuration` repository takes a long time. I made a pull request there weeks ago and nobody has reviewed it yet.
- Using those roles to provision external databases makes the code more complex, while tightly coupling it to the playbooks.
- The instance manager would need ssh access to the database servers.

mtyaka:
Getting stuff merged upstream can take a long time, that's true. We might be able to speed up the process a bit by pinging some people on the PR. I think it would be beneficial to the community and it would be useful for us as well - I currently manually create databases in RDS when deploying to AWS, it would be nice if it was done automatically :)

You don't need ssh access to create the tables. You only need MySQL credentials with enough privileges, and you would have to modify the tasks to pass mysql host/port. I used to do that in the past, but never got to submitting a proper upstream PR. See https://github.com/aquenteno/configuration/commit/2e077c7b39df6b668faedeec7a7b2366bc0564a6 for example.

omarkhan:
@mtyaka ok - can we do that in a separate PR?

It looks like they have split mysql from the edxlocal role in `configuration` now, that will make things easier.

mtyaka:
@omarkhan Yup, doing it in a separate PR makes sense, thanks!

antoviaque:
Agreed that it would be nice to have a standard way to do this from a role in the configuration repo. Note that @bradenmacdonald is currently working on a discovery story on OpenStack production setup support, which includes looking at using the mysql & mongo_3_0 roles from the configuration repo to manage the DBs (and replace RDS/mongolab).

omarkhan:
@antoviaque can we have an external database server for integration tests or do you want to set that up and tear it down as part of the tests?

antoviaque:
@omarkhan For now we could setup a DB server to run the integration tests. There isn't one at the moment, but that could be setup and added to the integration environment.

omarkhan:
@antoviaque I am having some real trouble running the integration tests. They take forever, and never seem to finish when I run them on my machine. I just killed a test run after 3 hours.

I have set up an RDS mysql db and a mongodb server on digitalocean - can you try running the integration tests on this branch for me? I will send you the database credentials in a PM.

smarnach:
@omarkhan I can run the tests for you.  You can put the credentials in a Google document and share it with sven.marnach@gmail.com.

omarkhan:
Shared, thanks @smarnach

antoviaque:
@smarnach Cool, thank you! : )

smarnach:
@omarkhan @antoviaque The integration tests indeed take forever!  I've started a run about two hours ago, and it hasn't finished yet.  I'm travelling starting in about twenty minutes' time, so even if it finishes now, I won't be able to look at the results anymore.  Xavier, I guess it will be up to you then, or it has to wait until Monday.

omarkhan:
@antoviaque @smarnach this seems to solve my problem: https://github.com/open-craft/configuration/pull/1

I haven't had time to let the tests run all the way through yet, as they take so long. I will try again tonight before going to bed.

omarkhan:
Just a thought - would it be worth setting up a CI pipeline for the opencraft fork of the configuration repo that would build sandbox VM images? That way we could just provision a base image with all the dependencies and just use the ansible playbooks for launching the services with the right configuration.

omarkhan:
I have set up an instance on EC2 to run these tests now, so I can leave them unattended. Let's see what happens :smile:

antoviaque:
@omarkhan Switching to building images instead of running ansible on the the VMs directly is something we'd like to get to - although that would still need to be done individually for each instance. The second part in what you're mentioning, running ansible on top of an image that has already the dependencies would be a good way to speed things up yes - that needs to be done carefully though, as it's easy to get ansible failures due to the existing state, or if the requirements don't exactly match. But yep, eventually it would be good to have something like that.

antoviaque:
@omarkhan Added the extra settings you gave to the integration tests env, and started the integration tests here - the result will be in https://circleci.com/gh/open-craft/opencraft/296

Sorry that it took so long to do this - hope it runs well! If you need another run let me know.

antoviaque:
@omarkhan Started another run, based on your latest commits.

omarkhan:
Thanks @antoviaque. I have set up an EC2 instance to run the integration tests now, which cuts the time down to about 1 hour per test. Still a pretty long feedback loop but manageable. Ironing out some issues, should be done soon.

antoviaque:
@omarkhan Sounds good - if you need anything let me know. Thanks!

omarkhan:
@antoviaque @smarnach the integration test is working now. You will need to set up mysql and mongodb servers and set the `INSTANCE_MYSQL_URL` and `INSTANCE_MONGO_URL` environment variables. If either of these variables is unset the test will be skipped.

antoviaque:
@omarkhan Thanks! I had set the variables that you had sent earlier, so I'm rerunning the tests now - let's see if it passes, if not I'll check the variables.

omarkhan:
Those settings won't work anymore as I have terminated the instances they
point to

antoviaque:
@omarkhan Ok, good to know - in that case we'll need to create the MySQL & Mongo instances that we'll keep using for the integration tests, as well as for the sandboxes - but that's fine, we'll need to do this anyway. That's a bit more involved though, so I'll schedule a task for the next sprint/week on this, as well as checking if everything is in order in this PR & merge it.

Thanks for the work!

omarkhan:
Has anyone had a chance to review this?

antoviaque:
@omarkhan Yes - @smarnach has actually spent a good deal of time over the last few days to setup the integration tests, and even rebase this for you to avoid having to bother you. Sven will likely post an update here when he sees this, but note that you can also follow the ticket about this now that you have access: https://tasks.opencraft.com/browse/OC-1168

smarnach:
@omarkhan I'm sorry things have taken longer than they should.

I've looked through your latest changes and everything is looking good.  I'm still having trouble with the test to check whether the instance succesfully came up at the moment, but this might be related to some upstream bug in the configuration repository.

The rebased version of this branch is available in #33.  Feel free to update your branch based on it, so we continue the discussion in this PR.  Once I manage to get the integration tests passing, we should be ready to finally merge this.

smarnach:
@omarkhan It took a while, but now I managed to get a green build for this PR; see #33.  I had to parallelize the tests to fit into CirclCI's time limit, and I had to revert a small part of your integration test commit.  Could you please update your branch for this PR with the source branch of #33 (`open-craft:external-databases`)?  I'd prefer to merge this PR, since all the history and discussion is here.

The `assert_instance_up()` you added still seems to be a bit flaky, and sometimes gives spurious errors.  Maybe we should wait a while and try again if it fails?

omarkhan:
@smarnach I have updated this PR with your changes and rebased onto master again

smarnach:
@omarkhan Thanks!  I made `assert_instance_up()` retry if it fails, since sometimes the machine reboot wasn't complete when the test tried to send a request.  I also squashed my fixups into your commits, and managed to get the tests passing again.  They still don't pass every time, but some of the failures seem to be infrastructure failures on the part of CircleCI, so there isn't much we can do about that.  Could you please reset your branch to open-craft/external-databases again?  No need to rebase, it is already up to date with current master, so a simple hard reset should be enough.  I promise you that this will be the final time I ask you to reset your branch. :)  I will then merge this PR.

omarkhan:
@smarnach done!

smarnach:
@omarkhan Thanks again for following through!  This turned out to be way more involved than we anticipated.  Good work!

PR #12, opened 2015-09-28
Vagrant for windows

brousch:
Changes mainly to the Makefile to help OpenCraft run on Windows under Vagrant. The README has also been updated to reflect the new Makefile commands.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,10 +12,11 @@ grep -Fq 'cd /vagrant' ~/.bashrc || echo 'cd /vagrant' >> ~/.bashrc
 
 # Install system packages
 sudo apt-get update --quiet
-sudo apt-get install -y $(cat debian_packages.lst) postgresql
+make install_system_dependencies
+make install_system_db_dependencies
 
 # Set up a virtualenv
-sudo pip3 install virtualenv
+make install_virtualenv_system
 mkdir -p ~/.virtualenvs
 virtualenv -p python3 ~/.virtualenvs/opencraft
 source ~/.virtualenvs/opencraft/bin/activate
@@ -25,7 +26,7 @@ grep -Fq 'source ~/.virtualenvs/opencraft/bin/activate' ~/.bashrc ||
   echo 'source ~/.virtualenvs/opencraft/bin/activate' >> ~/.bashrc
 
 # Install python dependencies
-pip install -r requirements.txt
+pip3 install -r requirements.txt
 
 # Create postgres user
 sudo -u postgres createuser -d vagrant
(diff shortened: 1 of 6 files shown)


debian_db_packages.lst
```diff
@@ -0,0 +1 @@
+postgresql
```
antoviaque: @brousch What about redis?

Vagrantfile
```diff
@@ -12,10 +12,11 @@ grep -Fq 'cd /vagrant' ~/.bashrc || echo 'cd /vagrant' >> ~/.bashrc
 
 # Install system packages
 sudo apt-get update --quiet
```
antoviaque: @brousch Shouldn't that get into `make install_system_*dependencies`?

Also, to avoid having to run the update twice, you could make the rule `make install_system_db_dependencies` depend on `make install_dependencies` (maybe renaming it in the process).

antoviaque:
@brousch Thanks for the PR! I've had a quick look, but only had a couple of minor comments so far. I'll review more thoroughly, likely over next week-end -- I'll also need to test it under Windows.

debian_db_packages.lst
brousch: Good question. It looks like Redis was not installed or configured in the Vagrantfile from @omarkhan . All of the tests pass, so there must also be no test for Redis.

omarkhan: Redis is in `debian_packages.lst`

Vagrantfile
brousch: Good idea

debian_db_packages.lst
brousch: So it is! Excellent.

antoviaque: @brousch I meant that redis could probably belong to the current dependency file - it's not exactly a conventional DB, and isn't used to actually store data at the moment, but what do you think?

antoviaque:
@brousch Thanks for the change for `apt-get update`. There is my other pending comment, then this will need a rebase.

debian_db_packages.lst
brousch: I don't have a strong opinion about this.

brousch:
@antoviaque I think this is ready to go.

antoviaque:
@brousch Thanks! But I was actually asking for a rebase - a merge leaves traces in the history. That's fine - I'll take care of it when I test it on Windows.

antoviaque:
@brousch Tested on windows, it worked well! :+1: 

Rebased and squashed your commits in https://github.com/open-craft/opencraft/pull/21

PR #13, opened 2015-09-30
Error states handling

chromano:
Handle properly error states during deployment and display in the UI the errors that happened.  Also enable retries (default: 3). /cc @antoviaque


antoviaque:
@chromano Thanks for the PR! Just restarted the build with the proper auth, so you can see the result. 

FYI @smarnach will be your main reviewer, although he might review a bit later this week. I'll also have a look at the PRs over the week-end.

antoviaque:
@chromano FYI you have a build failure - apparently the test coverage got too low. You should be able to reproduce locally with "make test".

antoviaque:
@chromano Just rerun the build with auth - you still have pylint build failures: https://circleci.com/gh/open-craft/opencraft/156 . You're not seeing them with "make test"?

chromano:
@antoviaque Sorry about that! I was more focused on integration tests and so was running "test_integration" directly rather than make test. Lesson learned.

antoviaque:
@chromano No worries : ) Restarted another build.

instance/migrations/0031_merge.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
smarnach: It might be worthwhile to recreate the migrations on top of current master, to keep a linear list of migrations, which is simpler to manage.

instance/models/instance.py
```diff
@@ -455,29 +461,53 @@ def run_playbook(self):
...
+                        self.logger.error(
+                            'Playbook failed for instance {}'.format(self))
+                        attempt += 1
+                        continue
+                    else:
+                        break
+
+        if processus.returncode == 0:
+            self.logger.info('Playbook completed for instance {}'.format(self))
+        return (log_lines, processus.returncode)
```
smarnach: This function is getting a bit unwieldy, with many nesting levels, so I think it could do with a bit of refactoring.

instance/models/instance.py
```diff
@@ -455,29 +461,53 @@ def run_playbook(self):
...
-            ) as processus:
-                for line in processus.stdout:
-                    line = line.decode('utf-8').rstrip()
-                    self.logger.info(line)
-                    log_lines.append([line.rstrip()])
-
-        self.logger.info('Playbook run completed')
-        return log_lines
+        attempt = 0
+        while attempt < self.attempts:
```
smarnach: How about simply using a `for` loop here?

    for attempt in range(self.attempts):

instance/models/instance.py
```diff
@@ -455,29 +461,53 @@ def run_playbook(self):
...
+                    self.inventory_str,
+                    self.vars_str,
+                    playbook_path,
+                    self.ansible_playbook_filename,
+                    username=settings.OPENSTACK_SANDBOX_SSH_USERNAME,
+                ) as processus:
+                    output = itertools.chain(
+                        zip(itertools.repeat(logging.INFO), processus.stdout),
+                        zip(itertools.repeat(logging.ERROR), processus.stderr)
+                    )
```
smarnach: Iterating over both stdout and stderr at the same time is notoriously prone to deadlocks.  This code will iterate over stdout first, so if the output buffer for stderr runs full while the process is still running, this will simply block.  It's probably unlikely to actually happen in this particular case, but we should code defensively against this anyway.  One way to do so is using the `communicate()` method of the process, but this doesn't allow forwarding the output in real time.  So some asynchronous mechanism is needed, be it `select.poll()`, threads or using Python 3.4's new `asyncio` module.

instance/models/server.py
```diff
@@ -236,8 +238,11 @@ def update_status(self, provisioned=False, rebooting=False):
...
 
-        elif self.status == self.BOOTED and provisioned:
-            self._set_status(self.PROVISIONED)
+        elif self.status == self.BOOTED:
+            if provisioned:
+                self._set_status(self.PROVISIONED)
+            elif provisioning_failure:
+                self._set_status(self.PROVISIONING_FAILURE)
 
         elif self.status in (self.PROVISIONED, self.READY) and rebooting:
```
smarnach: I think `PROVISIONING_FAILURE` should be included in this list.

smarnach:
@chromano Thanks for your work on this!  I had a first look over the PR and added a few comments.  I didn't get completely through the code yet, so I will take another look on Sunday.  (I don't expect you to address the comments until then. :) )

instance/models/instance.py
chromano: Thanks, I think an alternative is to iterate over both stdout and stderr at the same time, what do you think? Something like:

    for line in processus.stdout:
        self.logger.info(line)
        log_lines.append(line)
    for line in processus.stderr:
        self.logger.error(line)
        log_lines.append(line)

I don't like the repetition, so maybe a nested function would also work?

smarnach: @chromano The problem is the use of blocking I/O.  While you wait for the next line from either stdout or stderr with blocking I/O, the buffer for the other stream might fill up, blocking the process writing to the stream.  The process will block waiting for the buffer of one stream to become available, but your code is blocking waiting for output on the other stream, and none of the two things will ever happen.  As I said before, this is notoriously tricky, and you need some asynchronos mechanism to solve this problem (or `select.poll()`, which isn't strictly asynchronous).

smarnach: @chromano One more thought on this issue.  All I said so far was assuming that we actually need to read from two different streams.  I'm not sure we really need to distinguish between the two streams in this case.  We could simply set `stderr=subprocess.STDOUT` in the `Popen()` call to run the playbook and would only have to read from a single stream.  We wouldn't be able to set the log levels as you did anymore, but it would save us the trobule mentioned earlier.

chromano: Ouch, I was thinking about generators and ended up writing that code, sorry -- below is what I was thinking originally:

    def f(*fds):
        while fds:
            for fd in fds[:]:
                try:
                    yield next(fd)
                except StopIteration:
                    fds.remove(fd)

    for line in f(process.stdout, process.stderr):
        print(line)

But this code still suffers from the problem you reported anyways, which is blocking on either stdout or stderr reads.

chromano: I don't think a truly async approach would be available for files, but maybe it is just my lack of knowledge. Anyways, I've come to a simple approach:

    def f(*fds):
        poll = select.poll()
        fd_map = {}
        for fd in fds:
            poll.register(fd.fileno(), select.POLLIN)
            fd_map[fd.fileno()] = fd
        while fd_map:
            available = poll.poll()
            for fileno, event in available:
                line = fd_map[fileno].readline()
                if line:
                    yield line
                else:
                    poll.unregister(fileno)
                    del fd_map[fileno]

It is working well for me, specially because we only have two descriptors. Do you think this solution is acceptable?

instance/api/instance.py
```diff
@@ -47,7 +47,9 @@ def provision(self, request, pk=None):
         Start the (re-)provisioning of an instance
         """
         instance = self.get_object()
-        if instance.status not in (instance.EMPTY, instance.READY):
+        if instance.status not in (instance.EMPTY,
+                                   instance.READY,
+                                   instance.PROVISIONING_FAILURE):
```
antoviaque: Is there a way to improve this? To not have to maintain this list when we add new states, and avoid duplicating the test in other places - like the client for example.

instance/models/instance.py
```diff
@@ -87,6 +88,7 @@ class Instance(ValidateModelMixin, TimeStampedModel):
     ACTIVE = 'active'
     BOOTED = 'booted'
     PROVISIONED = 'provisioned'
+    PROVISIONING_FAILURE = 'provisioning_failure'
```
antoviaque: Use of a noun instead of an adjective/adverb

Also, would this be the best way to define an error state? To add error handling for the other states, we would then need to duplicate each individual state, to add a `..._FAILED` state for each of them. Maybe a separate field would be better? Ie something that could have the values `RUNNING`, `DONE`, `FAILED` could be useful. It would also allow to have a more rubust and maintainable test in https://github.com/open-craft/opencraft/pull/13/files#diff-772e1d190b3c56d74ec801abe130ba8cR50

instance/models/instance.py
```diff
@@ -455,29 +461,53 @@ def run_playbook(self):
...
+        attempt = 0
+        while attempt < self.attempts:
+            with open_repository(self.ansible_source_repo_url,
+                                 ref=self.configuration_version) as configuration_repo:
+                playbook_path = os.path.join(configuration_repo.working_dir,
+                                             'playbooks')
+                requirements_path = os.path.join(configuration_repo.working_dir,
+                                                 'requirements.txt')
+
+                log = ('Running playbook "{path}/{name}" for instance {instance}, '
```
antoviaque: The instance name should already be added by the logger.

instance/static/html/instance/index.html
```diff
@@ -24,6 +24,11 @@
       <li ng-repeat="instance in instanceList"
           ng-class="{active: instance == selected.instance}">
         <a ng-click="select('instance', instance)">{{ instance.name }}</a>
+        <div class="status-icon" ng-switch="instance.status">
+          <i class="fa fa-ellipsis-h" tooltip-placement="right" tooltip="Provisioning in progress" ng-switch-default></i>
+          <i class="fa fa-times" tooltip-placement="right" tooltip="Provisioning failure" ng-switch-when="provisioning_failure"></i>
+          <i class="fa fa-check" tooltip-placement="right" tooltip="Ready" ng-switch-when="ready"></i>
+        </div>
```
antoviaque: Cf my comment for https://github.com/open-craft/opencraft/pull/13/files#diff-772e1d190b3c56d74ec801abe130ba8cR50 - maybe the approach from https://github.com/open-craft/opencraft/pull/13/files#diff-848aab5153020ca3817d6db9d4d3a144R91 would help with this too.

instance/static/html/instance/index.html
```diff
@@ -63,7 +68,13 @@
...
               Reprovision
             </button>
-            <img src="/static/img/running.png" tooltip-placement="bottom" tooltip="Provisioning in progress..." alt="Provisioning in progress..." ng-switch-default />
+            <button class="button reprovision" ng-switch-when="provisioning_failure"
+                                               ng-click="provision(selected.instance)">
+              Reprovision
+            </button>
+            <i class="fa fa-2x fa-ellipsis-h" tooltip-placement="right" tooltip="Provisioning in progress" ng-switch-default></i>
+            <i class="fa fa-2x fa-times" tooltip-placement="right" tooltip="Provisioning failure" ng-switch-when="provisioning_failure"></i>
+            <i class="fa fa-2x fa-check" tooltip-placement="right" tooltip="Ready" ng-switch-when="ready"></i>
```
antoviaque: This duplicates https://github.com/open-craft/opencraft/pull/13/files#diff-a8e64e20aeb5487bab76b72ae22e9aefR28 . The same comment applies, plus refactoring the test within the client, to only define it once - this is worth adding as a method to the controller.

instance/static/html/instance/index.html
```diff
@@ -71,7 +82,8 @@
     <div class="row">
       <tabset>
         <tab heading="Log">
-          <pre class="instance-log">{{ selected.instance.log_text }}</pre>
+          <pre ng-if="selected.instance.status == 'provisioning_failure'" class="error-log" ng-repeat="line in error_log() track by $index">{{ line }}</pre><br/>
+          <pre class="instance-log" ng-repeat="line in selected.instance.log_entries track by $index">{{ line }}</pre>
```
antoviaque: Nice to have switched to a list, and used `track by`.

instance/static/js/src/instance.js
```diff
@@ -122,12 +122,37 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
...
 
+        $scope.error_log = function() {
+            if (!$scope.selected.instance) {
+                return;
+            }
+            var log_entries = $scope.selected.instance.log_entries;
+            var error = [];
+            var counting = false;
+            for (var i = log_entries.length; i > 0; i--) {
+                if (!/ERROR \|/.test(log_entries[i - 1])) {
```
antoviaque: Rather than using regex here, now that the log lines are a list, it might be worth getting further in that direction and include the log level along with the log message. Then, add it dynamically to each line, similarly to the way it's done when printing out to the shell console: https://github.com/open-craft/opencraft/blob/b58fabf98301bdc4129e4e3222c131f6235eab4b/opencraft/settings.py#L297

instance/static/js/src/instance.js
```diff
@@ -122,12 +122,37 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
...
+                    counting = true;
+                }
+            }
+            return error.reverse();
+        };
+
+        swampdragon.ready(function() {
+            swampdragon.subscribe('notifier', 'notification', null);
+            swampdragon.subscribe('notifier', 'log', null);
+        });
```
antoviaque: Why do you need to repeat this? It's already in the `init()`? https://github.com/chromano/opencraft/blob/master/instance/static/js/src/instance.js#L72-L75

instance/static/scss/instance.scss
```diff
@@ -22,9 +33,19 @@
...
+    .status-error, .status-success {
+        margin-top: 12px;
+    }
+
     .configuration,
-    .instance-log {
+    .instance-log,
+    .error-log {
         font-size: 0.8em;
+        white-space: pre-wrap;
```
antoviaque: Neat, I didn't know about the `white-space` property

instance/tests/integration/integration_instance.py
```diff
@@ -56,3 +58,52 @@ def test_provision_instance(self):
...
+    """
+    Make sure failures to the deployment process are handled
+    """
+    def test_ansible_failure(self):
+        """
+        Ensure failures in the ansible flow are reflected in the instance
+        """
+        with mock.patch('git.refs.head.Head.checkout'):
+            with mock.patch('git.Git.working_dir', new_callable=mock.PropertyMock) as wd:
+                wd.return_value = os.path.join(os.path.dirname(__file__), "ansible")
```
antoviaque: This looks like it will be useful for other integration tests - could you factorize it as a decorator? That would also help with the level of indentation.

instance/tests/integration/integration_instance.py
```diff
@@ -56,3 +58,52 @@ def test_provision_instance(self):
...
+
+                uid = str(uuid.uuid4())[:8]
+                instance = OpenEdXInstance.objects.create(
```
antoviaque: Since this duplicates code from the other integration test, it could be worth factorizing it into a instance factory for integration tests?

instance/tests/integration/integration_instance.py
```diff
@@ -56,3 +58,52 @@ def test_provision_instance(self):
...
+                    forum_version='named-release/cypress',
+                    notifier_version='named-release/cypress',
+                    xqueue_version='named-release/cypress',
+                    certs_version='named-release/cypress',
+                )
+                provision_instance(instance.pk)
+                self.assertEqual(instance.status, OpenEdXInstance.PROVISIONING_FAILURE)
+
+                uid = str(uuid.uuid4())[:8]
+                instance = OpenEdXInstance.objects.create(
```
antoviaque: Same remark - and worth splitting this into two separate tests maybe?

instance/tests/models/test_instance.py
```diff
@@ -407,6 +408,25 @@ def test_provision(self, os_server_manager, mock_run_playbook, mock_set_dns_reco
...
     @patch_os_server
+    @patch('instance.models.server.openstack.create_server')
+    @patch('instance.models.server.OpenStackServer.sleep_until_status')
+    @patch('instance.models.server.OpenStackServer.reboot')
+    @patch('instance.models.instance.gandi.set_dns_record')
+    @patch('instance.models.instance.OpenEdXInstance.run_playbook')
+    def test_provision_failed(self, os_server_manager, mock_run_playbook, mock_set_dns_record, mock_server_reboot,
+                              mock_sleep_until_status, mock_openstack_create_server):
+        """
+        Run provisioning sequence
```
antoviaque: Needs a more descriptive docstring - what are we testing?

instance/tests/models/test_instance.py
```diff
@@ -407,6 +408,25 @@ def test_provision(self, os_server_manager, mock_run_playbook, mock_set_dns_reco
...
+        """
+        mock_run_playbook.return_value = (['log'], 1)
+        mock_openstack_create_server.return_value.id = 'test-run-provisioning-server'
+        os_server_manager.add_fixture('test-run-provisioning-server', 'openstack/api_server_2_active.json')
+
+        instance = OpenEdXInstanceFactory(sub_domain='run.provisioning', status=Server.BOOTED)
+        result = instance.provision()
+        self.assertEqual(result[1], ['log'])
```
antoviaque: How does that test that the provisioning failed?

antoviaque:
For the font-awesome dependency, rather than including the files in the repository, could you follow the existing pattern for js depencies, ie downloading them via hte Makefile rules?

* https://github.com/open-craft/opencraft/blob/master/Makefile#L104-L105
* http://cdnjs.com/libraries/font-awesome

instance/tests/models/test_instance.py
chromano: True, I think that was not supposed to be committed. I went ahead and finished implementing this test anyways. Thanks for reporting.

instance/api/instance.py
chromano: Cool idea! We can create a ERROR status and a new field to store the error details.

instance/static/html/instance/index.html
chromano: @antoviaque what do you mean by refactoring the test within the client?

instance/static/js/src/instance.js
chromano: Interesing, I will make it a list of dicts then, so we could easily access meta data about the logging entry, eg: log.level, log.message, etc.  What do you think?

instance/static/js/src/instance.js
chromano: Thanks for catching that, I think it was a poorly resolved conflict.

instance/tests/integration/integration_instance.py
chromano: Yeap, factoryboy for the rescue. Looks much better now, thanks.

instance/static/js/src/instance.js
antoviaque: @chromano That works for me - you could probably actually expose the log entries objects, via the API; since we already have full fledged model objects, that's likely the cleanest and most maintainable.

antoviaque:
@chromano Thanks for the prompt update! I'll have another look tomorrow too.

Note that you have a build failure again : ) https://circleci.com/gh/open-craft/opencraft/173

chromano:
Thanks @antoviaque, I'm still waiting answer on one more question from you and another from @smarnach, but I'm sure these can be addressed tomorrow.

Makefile
```diff
@@ -103,3 +104,6 @@ test_one: clean
 
 js_external:
 	$(MAKE) -C static/js/external
+
+static_external:
+	$(MAKE) -C static/external
```
antoviaque: JS files are static files, so it overlaps to have both `js_external` and `static_external`. Maybe only call `static_external` from here, and then call `js_external` from `static_external`?

instance/models/instance.py
```diff
@@ -94,6 +96,8 @@ class Instance(ValidateModelMixin, TimeStampedModel):
     STOPPED = 'stopped'
     TERMINATING = 'terminating'
 
+    ERR_PROVISIONING_FAILED = 'provisioning_failed'
```
antoviaque: So, having a single error state is better imho, but the way you implemented isn't what I was suggesting. Happy to discuss the merits of each approach (and @smarnach might chime in on this too), but with this, you still have to define a different error for each status, even if you moved it to a different field. What I was suggesting was:

* When an error happens, leave `server.status` with its current status (for what you're implementing, `PROVISIONING`, to preserve that information
* Add a new field, that could have values like `RUNNING`, `SUCCESS`, `FAILED` - maybe called `status_progress` that would further refine the current status. When an error happen, it would go from `RUNNING` to `FAILED`, and we would know that it's a failed provisioning, without the need for adding new values for each of the current states
* Use it in places like https://github.com/open-craft/opencraft/pull/13/files#diff-772e1d190b3c56d74ec801abe130ba8cR50 where we need to know if the user can start a reprovisioning, by simply testing if it isn't `RUNNING`
* Share this field with the client, and use it to display status icons, and whether we show the reprovisioning icon or not.

instance/models/instance.py
```diff
@@ -139,17 +143,37 @@ def active_server_set(self):
...
+    def current_server(self):
         """
-        Instance status
+        Current active server. Raises InconsistentInstanceState if more than
+        one exists.
         """
         active_server_set = self.active_server_set
         if not active_server_set:
-            return self.EMPTY
+            return
```
antoviaque: Why this change?

instance/models/instance.py
```diff
@@ -139,17 +143,37 @@ def active_server_set(self):
...
         elif active_server_set.count() > 1:
             raise InconsistentInstanceState('Multiple servers are active, which is unsupported')
-        else:
-            return active_server_set[0].status
+        return active_server_set[0]
```
antoviaque: Why this change?

instance/models/server.py
```diff
@@ -145,7 +150,8 @@ def _set_status(self, status):
 
         self.status = status
         self.logger.info('Changed status: %s', self.status)
-        self.save()
+        if commit:
+            self.save()
```
antoviaque: Why not keeping saving the status every time it changes? Is it to save writes in the DB? Having it always saved allows to reliably trigger events on status change.

instance/static/js/src/instance.js
```diff
@@ -122,12 +122,33 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
...
                 if($scope.selected.instance && $scope.selected.instance.id === message.data.instance_id) {
                     $scope.$apply(function(){
-                        $scope.selected.instance.log_text += message.data.log_entry + '\n';
+                        $scope.selected.instance.log_entries.push(message.data.log_entry);
                     });
                 }
             }
         };
 
+        $scope.error_log = function() {
```
antoviaque: Could you add a test for this?

instance/static/js/src/instance.js
```diff
@@ -122,12 +122,33 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
...
+            }
+            var log_entries = $scope.selected.instance.log_entries;
+            var error = [];
+            var counting = false;
+            for (var i = log_entries.length; i > 0; i--) {
+                var entry = log_entries[i - 1];
+                if (entry.level != 'ERROR' && entry.level != 'CRITICAL') {
+                    if (counting) {
+                        return error.reverse();
+                    }
```
antoviaque: That assumes that the error would always be the last log_entry, which might not always be the case.

I wonder if it wouldn't be cleaner to isolate the error on the server side, when it actually happens, which would be more reliable - a reference to a set of log_entries could be kept when the error is caught from ansible, to ensure there is always a reference to one, and only one, set of error messages when the instance gets into an error state.

instance/tests/decorators.py
```diff
@@ -0,0 +1,40 @@
...
+"""
+from unittest import mock
+
+
+def patch_git_checkout(func):
+    """
+    Patch git checkout process in order to allow mocking up the checkout
+    process and the working directory
+    """
+    def _wrap(*args, **kwargs):
```
antoviaque: Could you use https://docs.python.org/3.4/library/functools.html#functools.wraps ?

instance/tests/js/instance_app_spec.js
```diff
@@ -105,13 +105,13 @@ describe('Instance app', function () {
 
             it('instance_log', function() {
                 $scope.select('instance', $scope.instanceList[0]);
-                expect($scope.selected.instance.log_text).not.toContain('### Added via websocket ###');
+                expect($scope.selected.instance.log_entries).not.toContain('### Added via websocket ###');
```
antoviaque: Is this still testing the right thing? `log_entries` changed to a list.

static/external/Makefile
```diff
@@ -0,0 +1,33 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Variables ###################################################################
+
+GET = wget -qO
+CDNFONTS = https://fortawesome.github.io/Font-Awesome/assets/
```
antoviaque: Was there files missing from https://cdnjs.com/libraries/font-awesome ? The downloads of assets is currently standardized around font-awesome for JS, and the files are already unpacked there.

instance/static/html/instance/index.html
antoviaque: @chromano I meant that the choice of status that get chosen for each (`provisioning_failure`, `ready`, etc.) is repeated in several places, which would require to update them all every time we add a status, and could thus be factorized in a controller method. But if we go with https://github.com/open-craft/opencraft/pull/13/files#r41096377 this would already be done on the server side, and could make it unecessary - it would guarantee that there would only be one value for each to test.

antoviaque:
@chromano Did a second pass - thanks for all the changes, most of them look good. Posted additional/follow-up comments. If I missed any question let me know : )

instance/models/server.py
chromano: Actually, it was not to save db writes. In the update_status method, there's a place where we are changing the status to ERROR but we can't save the instance right away, because we still need to set the error field accordingly. This change was to support that.

instance/models/instance.py
chromano: @antoviaque  This property (current_server) is the result of the refactory of status and error properties and is meant to be used internally only by these two properties, i.e, it is not going to be returned to outside the models scope. I thought self.EMPTY is meant to be used by callsites outside models scope, but if that's not the case, I will revert this part.

instance/models/instance.py
chromano: It has the same effect, but I think you are right, that's not the point of this PR. I will revert.

instance/tests/js/instance_app_spec.js
chromano: Good point, will make sure it is working as it should.

instance/models/instance.py
chromano: Definitely I think the way you are describing it is better. FTR I thought that my approach would be good because we could retain the error after a provisioning even if we reboot the server; but rethinking this I agree that after a reboot, and having the server at a BOOTED status, it only matters if there was an error during the booting and not previous provisioning failures.

I'm going ahead and make the changes. Thanks!

instance/models/instance.py
smarnach: @chromano This is a very generic implementation for reading from multiple file descriptors, and it would certainly work for this use case.  If we just yield `line`, we won't be able to distinguish between the different streams, though, which is the only point of this approach (see my comment on using `stderr=subprocess.STDOUT` above).  So we need to include the information which stream the line came from to set the log level accordingly.

I just noticed there is a [`selectors`](https://docs.python.org/3/library/selectors.html) module in Python 3.4 I wasn't aware of before.  You can use `selectors.DefaultSelector()` instead of `select.poll()` for better portability – the interface is quite similar.

There definitely are truly asynchronous approaches, by the way.  Using threads is the most obvious one, but since Python 3.4 there is also the event-loop based `asyncio` module – see https://docs.python.org/3.4/library/asyncio-subprocess.html for detail on how to use this with subprocesses.  (I'm not suggesting to use it here – `poll()` should work fine.)

instance/models/instance.py
smarnach: I'm not quite sure `SUCCESS` is a useful value for `status_progress` – wouldn't successful conclusion of a status usually mean we are moving on to the next state?  I think we only need the states `RUNNING` and `FAILED`, so maybe we could make do with a Boolean flag.

chromano: For this case I think being explicit is better. For example, after provisioning we will move to a READY state, but READY progress can't be either RUNNING or FAILED, so we will have to ignore the flag when the status is READY?

instance/static/js/src/instance.js
chromano: Yep

instance/static/html/instance/index.html
chromano: @antoviaque  OK, just confirming before writing the code -- Do you mean that now we can rely on status_progress to display either the success or error icon, correct? But the code above would still be repeated in these two places, is that a problem?

Also, should the "Provisioning in progress" picture display all the time or only when the provisioning is in fact running? And what about the other statuses, like BOOTED, ACTIVE, etc.?

static/external/Makefile
chromano: Thanks

instance/static/js/src/instance.js
chromano: This is a bit tricky actually, because the error can be made of many stderr entries, so I need to iterate over the reversed log entries and capture the maximum subset that are errors and stop once I find something that is not.

Anyways, I still think this logic is necessary, but I agree when you say we should move it server side, I will do it.

chromano: Actually, I changed my mind here; I think it is more coherent and maybe more helpful if we display all the errors we got, since one can have contributed to the other.

instance/models/instance.py
chromano: Thanks for sharing your thoughts, I think I will keep the <code>poll()</code> approach here since we are only dealing with two file descriptors, it doesn't make sense to get more complex than that.

instance/static/html/instance/index.html
antoviaque: Yup, that's what I meant.

instance/tests/models/test_instance.py
```diff
@@ -335,6 +340,29 @@ def test_run_playbook(self, mock_open_repo, mock_run_playbook, mock_inventory, m
...
+
+                stdout.write(b"HELLO\n")
+                stdout.flush()
+                stderr.write(b"HI\n")
+                stderr.flush()
+
+                instance = OpenEdXInstanceFactory()
+                log, returncode = instance._run_playbook("requirements", "playbook")
+
+                self.assertEqual(log, ['HELLO', 'HI'])
```
smarnach: There isn't any guarantee for the order of these two entries, even when flushing after writing to each line.  The OS is actually free to deliver them out of order, since the writes went to different streams.  You can simply use `self.assertItemsEqual()` instead of `self.assertEqual()` here.

instance/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Test cases for functions in the utils module
+    """
+    def test_read_files(self):
+        """
+        Ensure that the lines read are in the order they were written
+        """
+        with NamedTemporaryFile() as wfile1:
+            with NamedTemporaryFile() as wfile2:
+
```
smarnach: You can combine these two lines into one to save one level of indentation:

    with NamedTemporaryFile() as wfile1, NamedTemporaryFile() as wfile2:

(Doesn't really matter in this case, but might come in handy in the future.)

instance/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+                wfile2.write(b"FILE2,LINE3\n")
+
+                wfile1.flush()
+                wfile2.flush()
+
+                self.assertEqual(next(lines), (rfile1, b"FILE1,LINE1\n"))
+                self.assertEqual(next(lines), (rfile2, b"FILE2,LINE1\n"))
+                self.assertEqual(next(lines), (rfile1, b"FILE1,LINE2\n"))
+                self.assertEqual(next(lines), (rfile2, b"FILE2,LINE2\n"))
+                self.assertEqual(next(lines), (rfile2, b"FILE2,LINE3\n"))
```
smarnach: Does this test pass?  You finish reading before you have written anything to the files, so I'd be surprised if `lines` actually contained anything.  (Also note my comment above – there is not guaranteed ordering for writes to differnt streams.)

chromano: Yes, it does pass. Please note that there are different file descriptors for reading and writing and also that they are not closed until test is finished.

instance/tests/models/test_instance.py
chromano: Thanks

instance/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    def test_read_files(self):
+        """
+        Ensure that the lines read are in the order they were written
+        """
+        with NamedTemporaryFile() as wfile1:
+            with NamedTemporaryFile() as wfile2:
+
+                rfile1 = open(wfile1.name, "rb")
+                rfile2 = open(wfile2.name, "rb")
```
smarnach: Ordinary files probably aren't the best candidates to test the `read_files()` function.  They are considered to be _always_ ready reading, so `poll()` won't do anything particularly useful for them.  Probably better to test with pipes, or even easier, a proper subprocess.  You could invoke a shell executing something like

    echo line1; echo line1 >&2; echo line2; echo line2 >&2; echo line3

to simulate a subprocess writing to both stdout and stderr.

instance/tests/test_utils.py
smarnach: @chromano Oh, right, `read_files()` is a generator function.  I completely got this wrong.

instance/tests/test_utils.py
chromano: Excellent, thanks for the suggestion!

instance/static/html/instance/index.html
chromano: OK, done

chromano:
Just pushed new changes to address the remaining suggestions.

antoviaque:
@chromano Thanks! I'll let @smarnach re-review, and once it's good with him I'll also do a last pass - again likely over the week-end (although you're really not expected to fix anything over the week-end, it's just the only moment I really have long stretches of uninterrupted time to review calmly : ) ).

instance/models/instance.py
smarnach: @chromano I think this makes sense.  You could consider prepending `current_server` with an underscore to mark it for internal use only.

instance/tests/integration/ansible/playbooks/failignore.yml
```diff
@@ -0,0 +1,5 @@
+- name: Fail and ignore on purpose
```
smarnach: The first line should be `---`.

instance/tests/models/test_instance.py
```diff
@@ -335,6 +340,29 @@ def test_run_playbook(self, mock_open_repo, mock_run_playbook, mock_inventory, m
...
+
+                stdout.write(b"HELLO\n")
+                stdout.flush()
+                stderr.write(b"HI\n")
+                stderr.flush()
+
+                instance = OpenEdXInstanceFactory()
+                log, returncode = instance._run_playbook("requirements", "playbook")
+
+                self.assertListEqual(sorted(log), sorted(['HELLO', 'HI']))
```
smarnach: `assertItemsEqual()` would be a good fit here – it verifies that two iterables contain the same items, regardless of their order.

chromano: This method isn't available. It belongs to TestCase from python's unittest, but we are relying on django's DjangoTestCase.

instance/tests/models/test_log_entry.py
```diff
@@ -67,14 +67,35 @@ def test_log_text(self):
...
-2015-08-05 18:07:03 |     INFO | instance.models.instance  | instance=my.instance | Line #4, on instance
-2015-08-05 18:07:04 |  WARNING | instance.models.instance  | instance=my.instance | Line #5, on instance (warn)
-2015-08-05 18:07:05 |     INFO | instance.models.server    | instance=my.instance,server=vm1_id | Line #6, on server
-2015-08-05 18:07:06 | CRITICAL | instance.models.server    | instance=my.instance,server=vm1_id | Line #7, exception
-'''))
+        entries = instance.log_entries
+        self.assertEqual(entries[0].level, "INFO")
+        self.assertEqual(entries[0].created.strftime("%Y-%m-%d %H:%M:%S"), "2015-08-05 18:07:00")
+        self.assertEqual(entries[0].text,
+                         "instance.models.instance  | instance=my.instance | Line #1, on instance")
```
smarnach: I'd suggest to compare the whole log entry using something like

    self.assertEqual(entries[0], {

instance/tests/models/test_log_entry.py
```diff
@@ -67,14 +67,35 @@ def test_log_text(self):
...
+
+        self.assertEqual(entries[4].level, "INFO")
+        self.assertEqual(entries[4].created.strftime("%Y-%m-%d %H:%M:%S"), "2015-08-05 18:07:05")
+        self.assertEqual(entries[4].text,
+                         "instance.models.server    | instance=my.instance,server=vm1_id | Line #6, on server")
+
+        self.assertEqual(entries[5].level, "CRITICAL")
+        self.assertEqual(entries[5].created.strftime("%Y-%m-%d %H:%M:%S"), "2015-08-05 18:07:06")
+        self.assertEqual(entries[5].text,
+                         "instance.models.server    | instance=my.instance,server=vm1_id | Line #7, exception")
```
smarnach: This is a bit unwieldy and hard to follow.  Maybe you could factor out a helper function to check a single log entry?

instance/tests/test_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+        expected = [
+            (process.stdout, b"line1\n"),
+            (process.stderr, b"line1\n"),
+            (process.stdout, b"line2\n"),
+            (process.stderr, b"line2\n"),
+            (process.stdout, b"line3\n"),
+        ]
+        self.assertEqual(sorted(lines, key=lambda l: l[1]),
+                         sorted(expected, key=lambda l: l[1]))
```
smarnach: This isn't guaranteed to work, since this sorts only by the lines themselves, and two pairs of them are equal.  What we actually want is to sort by file descriptor, since the lines within each stream are guaranteed to be delivered in order.

    def key(entry):
        return entry[0].fileno()
    self.assertEqual(sorted(lines, key=key), sorted(expected, key=key))

static/external/Makefile
smarnach: You can remove these lines now (line 20 and 21).

templates/base.html
```diff
@@ -18,12 +18,17 @@
     {% endcompress %}
     {% endblock foundation_css %}
 
+    {# Fontawesome #}
+    {% compress css %}
+    <link rel="stylesheet" href='{% static "external/css/font-awesome.min.css" %}'/>
+    {% endcompress %}
```
smarnach: Not sure whether it's worth it to compress font-awesome – the version you include is already minified, and it's big enough to be worth caching on its own.

instance/tests/models/test_instance.py
smarnach: @chromano Djangos `TestCase` class derives from `unittest.TestCase`.  The problem is that the method has been renamed in Python 3 to `self.assertCountEqual()` – I'm still mostly using Python 2, so I forgot about this.

smarnach:
@chromano I did another pass through the code and added a few comments.  Nothing substantial, its' looking good overall.  @antoviaque will do a final pass in the weekend.

chromano:
@smarnach thanks for your review, I made the needed adjustments.

instance/tests/models/test_log_entry.py
chromano: Sure, I didn't create a helper function, but I did improve the readability of the test.

smarnach:
@chromano The changes look good to me, so :+1: from my side. :)  @antoviaque will give this another pass in the weekend.

instance/api/instance.py
```diff
@@ -47,7 +47,9 @@ def provision(self, request, pk=None):
         Start the (re-)provisioning of an instance
         """
         instance = self.get_object()
-        if instance.status not in (instance.EMPTY, instance.READY):
+        if instance.status not in (instance.EMPTY,
+                                   instance.READY,
+                                   instance.ERROR):
```
antoviaque: @chromano Wouldn't it be better to check on `instance.progress` instead here? Ie:

```
if instance.progress is not instance.PROGRESS_RUNNING:
```

instance/models/instance.py
```diff
@@ -86,7 +87,8 @@ class Instance(ValidateModelMixin, TimeStampedModel):
     STARTED = 'started'
     ACTIVE = 'active'
     BOOTED = 'booted'
-    PROVISIONED = 'provisioned'
+    PROVISIONING = 'provisioning'
```
antoviaque: @chromano Will the field value be properly migrated for existing instances, if one happen to be in the `provisioned` state when this is deployed?

instance/models/instance.py
```diff
@@ -86,7 +87,8 @@ class Instance(ValidateModelMixin, TimeStampedModel):
...
+    ERROR = 'error'
```
antoviaque: @chromano You don't need this state now that you have `PROGRESS_FAILED`? I don't think your code is using this state anymore.

instance/models/instance.py
antoviaque: @chromano I see, ok - that makes sense to me too yep.

instance/models/instance.py
```diff
@@ -170,44 +197,62 @@ def on_pre_save(sender, instance, **kwargs):
...
+        """
+        # TODO: Filter out log entries for which the user doesn't have view rights
+        server_log_entry_set = self._current_server.log_entry_set.order_by('pk')\
+                                                                 .iterator()
+        instance_log_entry_set = self.log_entry_set.order_by('pk')\
+                                                   .iterator()
+        return Instance._sort_log_entries(server_log_entry_set, instance_log_entry_set)
+
+    @property
+    def errors(self):
```
antoviaque: @chromano To be consistent with the method name above, can this be `log_error_entries()`?

Actually they both have very similar code - could they be factored into a single method:

```
def get_log_entries(self, level_list=None):
```

which the properties would call?

instance/models/server.py
```diff
@@ -87,7 +87,8 @@ class Server(ValidateModelMixin, TimeStampedModel):
     STARTED = 'started'
     ACTIVE = 'active'
     BOOTED = 'booted'
-    PROVISIONED = 'provisioned'
+    PROVISIONING = 'provisioning'
+    ERROR = 'error'
```
antoviaque: @chromano Same remark about the `ERROR` state - the information is already conveyed by `PROGRESS_FAILED`, and I don't think this is still being used by your code.

instance/serializers/logentry.py
```diff
@@ -0,0 +1,36 @@
...
+
+# Serializers #################################################################
+
+class LogEntrySerializer(serializers.Serializer): #pylint: disable=abstract-method
+    """
+    Log entries API serializer
+    """
+    level = serializers.CharField(read_only=True)
+    text = serializers.CharField(read_only=True)
+    created = serializers.DateTimeField(read_only=True)
```
antoviaque: @chromano Can you add an API test for this? See https://github.com/open-craft/opencraft/blob/master/instance/tests/api/test_instance.py

instance/static/html/instance/index.html
```diff
@@ -24,6 +24,13 @@
       <li ng-repeat="instance in instanceList"
           ng-class="{active: instance == selected.instance}">
         <a ng-click="select('instance', instance)">{{ instance.name }}</a>
+        <div class="status-icon" ng-switch="instance.status">
+          <i class="fa fa-ellipsis-h" tooltip-placement="top" tooltip="Provisioning in progress" ng-switch-default></i>
+        </div>
```
antoviaque: @chromano Doesn't that make it a switch with a single default element, ie always showing? I would have expected to only use the `instance.progress` attribute, and show "Provisioning in progress" when `instance.progress == instance.PROGRESS_RUNNING`? I feel like I'm missing something here.

instance/static/html/instance/index.html
```diff
@@ -58,12 +65,22 @@
...
+          <div class="large-2 columns">
+            <div class="status-icon" ng-switch="selected.instance.status">
+              <button class="button reprovision" ng-switch-when="ready"
+                                                 ng-click="provision(selected.instance)">
+                Reprovision
+              </button>
+              <button class="button reprovision" ng-switch-when="error"
+                                                 ng-click="provision(selected.instance)">
+                Reprovision
+              </button>
```
antoviaque: @chromano Similarly to my remark on the backend side - why not testing for `instance.progress != instance.PROGRESS_RUNNING` here?

instance/static/js/src/instance.js
```diff
@@ -122,7 +122,11 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
             } else if(message.data.type === 'instance_log') {
                 if($scope.selected.instance && $scope.selected.instance.id === message.data.instance_id) {
                     $scope.$apply(function(){
-                        $scope.selected.instance.log_text += message.data.log_entry + '\n';
+                        if (message.data.log_entry.level == 'ERROR' || message.data.log_entry.level == 'CRITICAL') {
+                            $scope.selected.instance.errors.push(message.data.log_entry);
+                        } else {
+                            $scope.selected.instance.log_entries.push(message.data.log_entry);
```
antoviaque: @chromano Don't we need to add errors to both `instance.errors` and `instance.log_entries`? Otherwise the errors will only show up in https://github.com/open-craft/opencraft/pull/13/files#diff-a8e64e20aeb5487bab76b72ae22e9aefR91 but not in https://github.com/open-craft/opencraft/pull/13/files#diff-a8e64e20aeb5487bab76b72ae22e9aefR92 ?

antoviaque:
@chromano I had a few additional minor comments, but now it's looking good to me. I'll still have to test locally once the comments about UI & model fields have been addressed, but if nothing show up there I expect to be able to give you a final +1 then. 

To prepare for the final merge, it would be worth starting to rebase against the current master (to clear the merge commits from your branch history) and squash your commits all together and/or group them into individual logical commits.

instance/api/instance.py
chromano: Actually, no. If the instance is in the <code>NEW</code> state (the default state), it will not pass the test and will proceed with the provisioning. However, we can (and should) remove the <code>instance.ERROR</code> attribute.

instance/models/instance.py
chromano: Excellent catch. I will provide a data migration. Thanks.

instance/models/instance.py
chromano: Sure.

instance/serializers/logentry.py
chromano: Yes

instance/static/js/src/instance.js
chromano: You are right, thanks.

instance/static/html/instance/index.html
chromano: Check my answer above. I think we should just replace the <code>error</code> status here with a <code>empty</code> status checking.

instance/static/html/instance/index.html
chromano: I think it is not clear to me whether to display the "Provisioning in progress" text/icon. It seems like the way you saying we will always display it, but what if there's no provisioning happening? For example, when you just created an instance and didn't start the provisioning process?

instance/models/instance.py
chromano: True.

instance/models/server.py
chromano: OK.

instance/api/instance.py
antoviaque: @chromano Ah, you're right about the NEW state - that's actually indicative of a missing value for the progress field - currently we have SUCCESS and FAILED for when it has finished processing, RUNNING for when it's still in progress, but nothing for when the current state hasn't been started yet. Maybe add a NEW value to the progress field, too? 

This way we would be able to just check for what we really want to check for here, ie that no process is already RUNNING, instead of maintaining a list of states (plus, since we remove the ERROR state from this list, we'd be back to not being able to reprovision instances in error otherwise).

instance/static/html/instance/index.html
antoviaque: @chromano Yup, you're right about that case - see https://github.com/open-craft/opencraft/pull/13#discussion_r41695774 though for a way to still approach it the way I'm suggesting above.

instance/static/html/instance/index.html
antoviaque: @chromano See above too - and here I'm fine just using "reprovision" (or "provision" if you prefer) for the button in all cases for now.

instance/api/instance.py
chromano: IMHO it will provide undesired side effects, for example: if we do it and then proceed with the changes below to the UI, for a short period of time the user will see the provision button and then it will disappear, because the state NEW will only last for a few ms. Plus, it is one more state we will have to maintain.

I suggest to just set the default progress value to SUCCESS on the Server model, since the default status is NEW.

instance/static/html/instance/index.html
chromano: Sure, there's a problem with that approach as described there. If we want to avoid that problem here as well, we would need to explicitly check whether the status is provisioning and the progress is running, then display this icon.

instance/static/html/instance/index.html
chromano: I think this ng-swtich block should all be replaced with a ng-if, like this:

    ng-if="selected.instance.status == 'ready' || selected.instance.status == 'new' || selected.instance.status == 'empty'

Or even better write a controller method:

    $scope.can_be_provisioned = function() {
        return $scope.selected.instance.status != ...
    }

Thoughts?

instance/api/instance.py
antoviaque: @chromano Sounds good, good idea - and it makes sense that the newly created instance status is at PROGRESS_SUCCESS initially, since it was successfully newly created at that point.

And you're right about the UI issue it would cause to test for progress=running here. This issue runs deeper actually - this would happen when the instance is EMPTY too, which also happens upon creating the instance initially, or when the instance is being reprovisioned, when the previous server is deleted but the new one hasn't been created yet. It feels like it will need a larger refactoring, using purely the server status to give the instance status doesn't work well in these cases.

But I don't want to add this to the scope of your task - we have another task about states coming up, to turn the current messy set_status() method into a proper state machine. This could be addressed as part of that scope. For now, even if there are outstanding issues with it, it feels cleaner to try to organize the code around the notion of progress=running for testing whether we're currently already performing actions on the instance, and thus can't allow the user to launch an action like provisioning.

instance/static/html/instance/index.html
antoviaque: It's not just with state=provisioning that this icone would make sense - there is also active, started, etc where it would make sense to show it to the user, otherwise there will be something running, but nothing to show to the user. Imho it makes more sense to try to get to a 1-to-1 matching between the instance progress and the icons we show here for the instance, ie:

* instance.progress=running => fa-ellipsis
* instance.progress=failed => fa-times
* instance.progress=success => fa-check

And then align the instance progress field value to always be at RUNNING when operations are pending or in progress - it's true that it's not consistently the case currently, but I'd rather expose these issues. Checking on individual states would hide them further.

instance/static/html/instance/index.html
antoviaque: Similar remarks as above - I'd rather ensure that the progress gives an accurate representation of whther there are still pending actions in progress, and use that as the basis to decide if the user is allowed to start new actions.

instance/api/instance.py
chromano: OK, I will do it then.

chromano:
OK, so I applied your suggestions and I had to implement some extra for setting the <code>progress</code> field to <code>RUNNING</code> for some original statuses. I also had to implement extra management of the <code>PROVISIONING</code> status so we can get it working together with <code>progress=RUNNING</code>.

I hope you like it.

(The commits were squashed as per request).

antoviaque:
@chromano Thank you - great changes! I'm all good now, and :+1: for the code. Thanks for sticking through this review. We'll still need to test this PR, and I've just started a build to check that everything passes, but assuming we don't find issues there we should be good to merge.

smarnach:
@chromano I tested this patch manually, and it seems to be working basically fine.  At some point, the log on the web page stopped updating, while the console kept showing the log lines.  Reloading the page in the browser fixed the issue, and I couldn't reliably reproduce it, so I can't really tell whether it's a bug in this PR.

One thing I noticed is that it is not quite easy to notice from the web interface that the playbook already failed once or twice.  It only gets obvious that something is wrong once all attempts failed.  Often people will be able to tell what's wrong after the first failure by seeing the error message of the first attempt.  However, I don't know an easy way of achieving this, and I think the PR is already a substantial improvement as it is, so I'd be in favour of merging the PR as it is and improving on the result later.

antoviaque:
@smarnach Thanks for the testing! Sound good for me, let's merge! Congratulations on getting this done @chromano ! This is a great feature.

PR #15, opened 2015-10-01
[WIP] Implement strategy pattern

chromano:
Hello,

This is a first step in the direction to abstract server deployments. The idea is to make it easier to let the user vary the strategy through the admin.

Since this would be much harder to let the user change the strategy via admin, I just wrote the basics to vary the strategy through AppConfig, but it also include some important parts to abstract the references to OpenStackServer.

I'd like your opinion on this PR, I'd be more than happy to implement strategy selection in the admin if you agree with it, just want to confirm before I dig into it.

/cc @antoviaque


antoviaque:
@chromano Thank you for the PR! I can definitely see being really useful. I'll have a closer look at it over the week-end, but on an initial cursory look, my main comment is that we might want to be able to chose the server type on a per-instance basis. So even without implementing the ability to change the strategy through the UI, rather than varying the strategy through AppConfig, I'd rather have a per-instance attribute, with a default defined in the configuration.

Also note that using AppConfig for settings might make sense, but then we would need to have a consistent approach, and likely switch a lot (all?) the application settings to it, which might be a big task for this PR.

chromano:
@antoviaque Yeah, makes sense. Also glad to know it is useful, so I will take it a whirl and come back with issues or just new commits.

antoviaque:
@chromano Since this PR is quite old and would still need work and a rebase, so closing it for now. IF you would like to finish the work let me know, and we'll look at it together.

PR #16, opened 2015-10-02
Add information on how to create a dev user

eeue56:
At the moment the README doesn't have info on how to get set up with a user for an instance.

This PR adds brief instructions to the README on how to create a user and then log in with that account.


antoviaque:
@eeue56 Thank you! LGTM :+1:

PR #17, opened 2015-10-07
Vagrant - Increase VM memory to 1GB to avoid OOM process kills

antoviaque:
@Kelketek To fix an out of memory error I had while reviewing https://github.com/open-craft/opencraft/pull/8 locally - it's a one-liner. Good for you?


Kelketek:
@antoviaque Seems fine, yeah. :+1:

PR #18, opened 2015-10-07
Raise an exception when ansible runs fail

omarkhan:
`OpenEdXInstance.provision` carries on as if nothing had happened if the ansible-playbook run fails. This pull request modifies the `run_playbook` method to raise an exception in this case.

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -69,6 +69,13 @@ class InconsistentInstanceState(Exception):
     pass
 
 
+class AnsibleRunFailed(Exception):
+    """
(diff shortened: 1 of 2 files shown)


antoviaque:
@omarkhan Thanks for this! However that's actually part of the scope of https://github.com/open-craft/opencraft/pull/13 - you might actually want to have a look at that PR, since it's getting closer to a mergeable status, and if it is merged before your you would have to rebase on top of it. You can also ignore this for now if you want : )

antoviaque:
Closing as duplicate of #13

PR #19, opened 2015-10-07
Add the ability to theme sandboxes

eeue56:
## The problem

As defined [here](http://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9541254-theme-for-the-sandboxes), this PR theme variables to an instance. 
## The solution

In order to make this work, I made the data passed down compatible with existing playbooks such as the one found [here](https://github.com/edx/configuration/blob/5c1d71a933769ca55016ba272001099d312a1fa9/playbooks/roles/edxapp/tasks/deploy.yml#L51-L61).
## The changes

The new instance variables added are:
- theme_name (defaults to 'default')
- theme_source_repo (defaults to 'https://github.com/Stanford-Online/edx-theme.git')
- theme_version (defaults to 'master')

To change the default options, you can set the following variables in your `.env` file appropriately:
- THEME_NAME
- THEME_SOURCE_REPO
- THEME_VERSION

These settings are then sent to ansible as the following:

``` yml
edxapp_theme_name: '{{ instance.theme_name }}'
edxapp_theme_source_repo: '{{ instance.theme_source_repo }}'
edxapp_theme_version: '{{ instance.theme_version }}'
```

I've added some tests to make sure ansible gets populated correctly, and that the defaults are as expected.
### Possible future work

It might be sensible to provide some more documentation on the properties in .env. AFAIK there's nothing other than what's provided in the default `.env` examples.

@antoviaque, what do you think? Is there anything I haven't covered?


instance/models/instance.py
```diff
@@ -490,6 +490,10 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
     notifier_version = models.CharField(max_length=50, default='master')
     xqueue_version = models.CharField(max_length=50, default='master')
     certs_version = models.CharField(max_length=50, default='master')
+    
+    theme_name = models.CharField(max_length=50, default=settings.THEME_NAME)
+    theme_source_repo = models.CharField(max_length=256, default=settings.THEME_SOURCE_REPO)
+    theme_version = models.CharField(max_length=50, default=settings.THEME_VERSION)
```
antoviaque: @eeue56 This will result in migrations being generated automatically every time the env changes - see https://github.com/open-craft/opencraft/pull/7#discussion_r40174892

instance/templates/instance/ansible/vars.yml
```diff
@@ -30,6 +30,7 @@ EDXAPP_BULK_EMAIL_DEFAULT_FROM_EMAIL: '{{ instance.email }}'
 
 # Misc
 EDXAPP_TIME_ZONE: 'UTC'
+edxapp_theme_name: '{{ instance.theme_name }}'
```
antoviaque: That means a theme will always be activated. What if we don't want any theme applied?

antoviaque:
@eeue56 Thanks for the PR! It's a good first step, but it's still missing several parts of the requirements - http://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9541254-theme-for-the-sandboxes :

> The default (*on/off*, and which theme repo & branch) would be specified in the application settings, for all sandboxes.

Ie it should be possible to deactivate themes on individual instances, and globally

> Make the default setting be to load a "OpenCraft" theme, *to produce* based on the Stanford theme ( https://github.com/Stanford-Online/edx-theme ), customized to the logo, & colors of OpenCraft.

Currently it only loads the default Stanford theme - not one with the colors & logo of OpenCraft.

eeue56:
So "off" doesn't just mean use the default, but to not pass the theme-based variables down at all, keeping things as they currently are? 

I'll fix the migrations issue 

I'll change the default and make a fork of the Stanford theme.

antoviaque:
@eeue56 Yep, that's correct - not setting the theme at all produces a different look than selecting the default stanford theme, which is already a customization of the platform (ie not having any theme set gives a more "vanilla" look).

I'll start a new build btw (they fail by default for new commits from another fork, as CircleCI only uses the auth for commits on the main fork for security reasons - I'll push your branch there to all a full build to process).

antoviaque:
@eeue56 FYI you have test failures, see https://circleci.com/gh/open-craft/opencraft/219

eeue56:
Great. I've got the solution for disabling/enabling themes. Just need to add a couple of tests for when they're disabled

Is there already an opencraft edx theme? If not, where can I find the colors of opencraft? I've taken the main color from the opencraft website at the moment and that's it.

antoviaque:
@eeue56 Great!

For the OpenCraft theme, currently there are none no, that's why the task scope includes creating one. There are no precise definition of the OpenCraft colors either - only what you already saw on the main website, which is the reference you can use in terms of branding, to give the OpenCraft theme colors & images the same look & feel.

eeue56:
@antoviaque to update:

* I've moved the theme variables into their own .yml file, like the s3 variables.
* I've added test cases for my ThemeableInstanceMixin
* I've currently set the default theme to https://github.com/eeue56/opencraft - the only changes at the moment is renaming stuff to talk about opencraft and the main color


All the tests for my code pass. There are a couple of errors from the API instance testing to do with base_domain not being set correctly - but as far as I can see that's nothing to do with my changes, as I haven't touched the API stuff nor anything to do with the domains. So I'm not sure what's happening there.

antoviaque:
@eeue56 Thanks for the changes! For the tests, they currently pass on master, cf https://circleci.com/gh/open-craft/opencraft/tree/master, so it's likely something related to the code in this branch. I think this might come from the fact that you're overriding the `on_pre_save` method from the instance model, so the setting become missing: https://github.com/open-craft/opencraft/pull/19/files#diff-848aab5153020ca3817d6db9d4d3a144R496 - imho you don't need to use a new class for the themable options, this is already very specific to OpenEdXInstance.

Let me know once it's passing and ready for a review, I'll give it another pass.

PR #20, opened 2015-10-09
Force formatting of text for log entry preemptively.

Kelketek:
When trying to diagnose the cause of the Database connection issues, I discovered that they would not happen if I printed the log entry to standard out right before sending it to the DB. Unfortunately I ran into trouble with the instance manager losing track of the instance it was trying to provision (master) so I didn't get to fully test this idea. However, I suspect the issue is that some part of the log entry is lazily evaluated, and when it tries to evaluate it in the middle of the DB operation, there's a problem. Forcing it to transcribe the log text ahead of time may fix the issue.

@antoviaque

### DIFF
--- a/instance/tests/test_logging.py
+++ b/instance/tests/test_logging.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Logging - Tests
+"""
+from mock import patch
+from instance.logging import DBHandler
+from instance.tests.base import TestCase
+
+
+class DummyObject:
+    """
+    Dummy object for testing forced lazy evaluation.
+    """
+    levelname = 'DEBUG'
+
+    # pylint: disable=no-self-use
+    def __unicode__(self):
+        return u'Rendered'
+
+
+class TestDBHandler(TestCase):
+    """
+    Test the DB Handler
+    """
+    @patch('instance.models.log_entry')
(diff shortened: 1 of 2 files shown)


antoviaque:
@Kelketek Interesting lead, but before considering merging such a fix we would need to know *why* that would fix the DB connection issues. Ie know why this would cause multiple DB connections to be open/not closed. Otherwise we'll just be adding a bit of black magic in this method, and the issue might pop back up elsewhere.

On a related note, this is missing a test.

antoviaque:
@Kelketek See https://github.com/open-craft/opencraft/pull/22 for another approach for addressing the issue of DB connections buildup/leaks. This would hopefully add another layer to prevent this class of issues. Could you review it?

Note that it would still be useful to get to the bottom of the issue you saw in this PR - I'm still curious to know why this would fix the issue in this case, there might still be another bug behind it.

instance/tests/test_logging.py
```diff
@@ -0,0 +1,51 @@
...
+    @patch('instance.models.log_entry')
+    def test_db_handler_text_eval(self, mock_create):
+        """
+        Make sure the text of the record is evaluated early before being sent to the logger.
+        """
+        with patch('instance.logging.DBHandler.format') as mock_format:
+            handler = DBHandler()
+            handler.emit(DummyObject())
+            self.assertTrue(mock_format.called)
+            self.assertTrue(mock_create.called_with('debug', u'Rendered'))
```
antoviaque: @Kelketek Here since this is addressing a specific bug, the test should be testing for the error that it causes when it's not pre-rendered, not just test the change it introduce to fix it. Ie the test should reproduce the bug when run without the change, and pass with the change.

antoviaque:
@Kelketek Were you able to dig the root cause of why it was causing an issue without pre-rendering?

Kelketek:
@antoviaque I was not. I can try looking into it tomorrow.

antoviaque:
@Kelketek That could be good - if only to understand why this was fixing the db connection close. Note that OC-1005 would still have priority.

Kelketek:
@antoviaque I've been checking out the code and have managed to reprovision a box without the error appearing. I'm not seeing any evidence of the issue still occuring-- the other PR likely fixed it.

This PR was, of course, a stab in the dark since this action appeared to fix things. However, since the instance it was reprovisioning hung, and I could not restart it at that time, trying to compare against the original code to see if restarting the instance manager was fixing it rather than my change wasn't possible. So it was not verifiable.

I was able to fully reprovision the master instance, and I haven't seen any messages like the ones we were seeing before. The manager has been running for a few days now-- I imagine if it were going to lock up, it likely would have by now.

I'd therefore like to retract this PR, if it's all the same to you.

Kelketek:
I should also note I haven't seen anything in the code that suggests this fix should have any effect. I did search around but I'm finding myself quite doubtful that it's doing much in reality.

antoviaque:
@Kelketek Ok - I would have preferred getting to the bottom of it, as unexplained mysteries have their way of creeping back at the worst moment, but since you couldn't find it, let's close this.

PR #23, opened 2015-10-12
Mark PR repos on the upstream repo, not the fork.

Kelketek:
@antoviaque This should fix the outstanding issue with the fork name.

### DIFF
--- a/instance/github.py
+++ b/instance/github.py
@@ -86,12 +86,12 @@ def get_settings_from_pr_body(pr_body):
         return ''
 
 
-def get_pr_by_number(fork_name, pr_number):
+def get_pr_by_number(pr_target_fork_name, pr_number):
     """
     Returns a PR object based on the reponse
     """
-    r_pr = get_object_from_url('https://api.github.com/repos/{fork_name}/pulls/{pr_number}'.format(
-        fork_name=fork_name,
+    r_pr = get_object_from_url('https://api.github.com/repos/{pr_target_fork_name}/pulls/{pr_number}'.format(
+        pr_target_fork_name=pr_target_fork_name,
         pr_number=pr_number,
     ))
     pr_fork_name = r_pr['head']['repo']['full_name']
@@ -99,6 +99,7 @@ def get_pr_by_number(fork_name, pr_number):
     pr = PR(
         pr_number,
         pr_fork_name,
+        pr_target_fork_name,
         pr_branch_name,
         r_pr['title'],
         r_pr['user']['login'],
@@ -148,9 +149,11 @@ class PR:
     """
     Representation of a GitHub Pull Request
     """
-    def __init__(self, number, fork_name, branch_name, title, username, body=''):
+    # pylint: disable=too-many-arguments
+    def __init__(self, number, source_fork_name, target_fork_name, branch_name, title, username, body=''):
         self.number = number
-        self.fork_name = fork_name
+        self.fork_name = source_fork_name
+        self.repo_name = target_fork_name
         self.branch_name = branch_name
         self.title = title
         self.username = username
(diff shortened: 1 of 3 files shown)


instance/tests/test_github.py
```diff
@@ -93,7 +93,7 @@ def test_get_pr_by_number(self):
             '**Description**\r\n\r\nHello!\nDesc with unicode «ταБЬℓσ»\r\n'
             '- - -\r\n**Settings**\r\n```yaml\r\nEDXAPP_FEATURES:\r\n  ALLOW: true\r\n```')
         self.assertEqual(pr.number, 8474)
-        self.assertEqual(pr.fork_name, 'open-craft/edx-platform')
+        self.assertEqual(pr.fork_name, 'edx/edx-platform')
```
antoviaque: That's not correct - the PR code is on the origin repository (ie for a PR from open-craft/edx-platform to edx/edx-platform, the code resides on open-craft/edx-platform), the task relies on this to point to the right code to deploy. It's only the PR URL that is on the destination repository (edx/edx-platform).

Kelketek: @antoviaque I believe this change will, in fact, fix the URL, as it will result in a URL like 'https://github.com/edx/edx-platform/pull/4' rather than 'https://github.com/open-craft/edx-platform/pull/4' . The code here uses 'fork-name' to populate the URL.

I suspect what you actually want is for me to rename that attribute, since, after all, it's no longer the fork's name. Does that make sense?

antoviaque: @Kelketek Yes, it will fix the PR URL, but this will break the repository URL, no?

https://github.com/open-craft/opencraft/blob/master/instance/templates/instance/ansible/vars.yml#L42
https://github.com/open-craft/opencraft/blob/master/instance/models/instance.py#L329
https://github.com/open-craft/opencraft/blob/master/instance/models/instance.py#L310

Actually, given that this can be pretty confusing to keep in mind, it might be worth adding a test specifically about that.

Kelketek:
@antoviaque How about this, then?

instance/github.py
```diff
@@ -86,12 +86,12 @@ def get_settings_from_pr_body(pr_body):
...
 
-def get_pr_by_number(fork_name, pr_number):
+def get_pr_by_number(pr_target_repo, pr_number):
     """
     Returns a PR object based on the reponse
     """
-    r_pr = get_object_from_url('https://api.github.com/repos/{fork_name}/pulls/{pr_number}'.format(
-        fork_name=fork_name,
+    r_pr = get_object_from_url('https://api.github.com/repos/{pr_target_repo}/pulls/{pr_number}'.format(
+        pr_target_repo=pr_target_repo,
```
antoviaque: @Kelketek Good idea to make the names more explicit here. To keep variable names consistent, maybe `pr_target_fork_name`? It might not be the best of names, but currently `fork_name` is used consistently across the code base to mean a combination of github org name and repo name - ie `org/repo`.

instance/github.py
```diff
@@ -148,9 +149,11 @@ class PR:
     """
     Representation of a GitHub Pull Request
     """
-    def __init__(self, number, fork_name, branch_name, title, username, body=''):
+    # pylint: disable=too-many-arguments
+    def __init__(self, number, fork_name, repo_name, branch_name, title, username, body=''):
```
antoviaque: `source_fork_name` & `target_fork_name`?

antoviaque:
@Kelketek Looks much better, thanks!

I'll retest one last time after the naming nits above are addressed, to be fully sure we're good this time : ) but LGTM.

Kelketek:
@antoviaque Tests are running now, but if you'd like to run any manual checks you intended to ahead of the sprint meeting, it could be merged as soon as they're done. Just need your thumbs up :)

antoviaque:
@Kelketek Thanks for the fix! Testing looks good, hopefully we got it right this time : ) :+1: 

Merging now to get you the completion before the sprint meeting : )

PR #24, opened 2015-10-14
Regisb/hackathon

regisb:
Re-start instance provisioning whenever a branch is updated.

### DIFF
--- a/instance/github.py
+++ b/instance/github.py
@@ -133,6 +133,18 @@ def get_team_from_organization(organization_name, team_name='Owners'):
     raise KeyError(team_name)
 
 
+def get_watched_usernames():
+    """
+    Retrieve the usernames to watch from both the watched organization and
+    username settings.
+    """
+    usernames = []
+    if settings.WATCH_ORGANIZATION:
+        usernames += get_username_list_from_team(settings.WATCH_ORGANIZATION)
+    if settings.WATCH_USER:
+        usernames.append(settings.WATCH_USER)
+    return sorted(set(usernames))
+
 def get_username_list_from_team(organization_name, team_name='Owners'):
     """
     Retrieve the usernames of a given team's members
--- a/instance/github.py
+++ b/instance/github.py
@@ -181,3 +181,7 @@ class PR:
         Construct the URL for the pull request
         """
         return 'https://github.com/{fork_name}/pull/{number}'.format(fork_name=self.fork_name, number=self.number)
+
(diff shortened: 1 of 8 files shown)


instance/github.py
```diff
@@ -133,6 +133,18 @@ def get_team_from_organization(organization_name, team_name='Owners'):
...
 
 
+def get_watched_usernames():
+    """
+    Retrieve the usernames to watch from both the watched organization and
+    username settings.
+    """
+    usernames = []
+    if settings.WATCH_ORGANIZATION:
+        usernames = get_username_list_from_team(settings.WATCH_ORGANIZATION)
```
antoviaque: Can you append the existing list too here? Just to be consistent, so that someone doesn't miss that the list is overridden if something is added before this step.

instance/tasks.py
```diff
@@ -78,3 +77,16 @@ def watch_pr():
...
+    """
+    Re-provision an existing instance whenever a branch has been updated.
+    TODO: what we really want to do is to create a new instance and delete the
+    previous one once the new one is ready.
+    """
+    for instance in OpenEdXInstance.objects.all():
+        commit_id = instance.commit_id
+        new_commit_id = instance.get_branch_tip_commit_id()
+        if new_commit_id != commit_id:
+            instance.provision()
```
antoviaque: Yup, good approach. I would however add a flag on a per-instance basis to activate this, with a default to false - to ensure that existing sandboxes for which people have done changes manually don't get automatically reprovisioned unless it's explicitly activated. And for sandboxes, the field could be set from the PR description - for an example see:

* https://github.com/open-craft/opencraft/pull/11/files#diff-a3f79e2fa4816b6091c587473626caf2R90
* https://github.com/open-craft/opencraft/pull/11/files#diff-04c6e90faac2675aa89e2176d2eec7d8R235

antoviaque:
@regisb Looks good overall, good approach : ) This will need some tests too, but I imagine you're already aware.

antoviaque:
@regisb Some pep8/pylint errors on your build : ) https://circleci.com/gh/open-craft/opencraft/264

antoviaque:
Fixed at https://github.com/regisb/opencraft/pull/1

Also created a PR rebased against the current master and containing both PRs, to run the tests: #25

regisb:
Replaced by https://github.com/open-craft/opencraft/pull/25

PR #26, opened 2015-10-22
Save Ansible settings at the time of provisioning in the database.

smarnach:
This makes sure the Ansible settings that are actually used for provisioning are stored in the database, instead of generating a settings string with the current values every time the settings are requested.  This is more efficient and more consistent than the old version, and will hopefully resolve the issue with repeated requests to the Github API from the web process.

The data migration will fill in the current version of the Ansible settings for each existing instance.  It's kind of a hack, but I don't think it's possible to do anything fundamentally better.


instance/tests/fixtures/api/instances_list.json
```diff
@@ -42,6 +42,7 @@
...
...'contact@example.com'\nEDXAPP_SERVER_EMAIL: 'contact@example.com'\nEDXAPP_BULK_EMAIL_DEFAULT_FROM_EMAIL: 'contact@example.com'\n\n# Misc\nEDXAPP_TIME_ZONE: 'UTC'\n\n# Security updates\nCOMMON_SECURITY_UPDATES: true\nSECURITY_UNATTENDED_UPGRADES: true\nSECURITY_UPDATE_ALL_PACKAGES: false\nSECURITY_UPGRADE_ON_ANSIBLE: true\n\n# Repositories URLs\nedx_ansible_source_repo: 'https://github.com/edx/configuration.git'\nedx_platform_repo: 'https://github.com/antoviaque/edx-platform.git'\n\n# Pin down dependencies to specific (known to be compatible) commits.\nedx_platform_version: '40952060bc7573c5035620b985b211a24360f311'\nconfiguration_version: 'master'\nforum_version: 'master'\nnotifier_version: 'master'\nxqueue_version: 'master'\ncerts_version: 'master'\n\n# Features\nEDXAPP_FEATURES:\n USE_MICROSITES: false\n AUTH_USE_OPENID: false\n ENABLE_DISCUSSION_SERVICE: true\n ENABLE_INSTRUCTOR_ANALYTICS: true\n ENABLE_INSTRUCTOR_EMAIL: true\n REQUIRE_COURSE_EMAIL_AUTH: false\n ENABLE_PEARSON_HACK_TEST: false\n SUBDOMAIN_BRANDING: false\n SUBDOMAIN_COURSE_LISTINGS: false\n PREVIEW_LMS_BASE: 'pr2.sandbox.opencraft.com'\n ENABLE_DJANGO_ADMIN_SITE: true\n ALLOW_ALL_ADVANCED_COMPONENTS: true\n\n\n"
```
mtyaka: Missing comma at the end.

instance/tests/fixtures/api/instances_list.json
```diff
@@ -64,6 +65,5 @@
     "sub_domain": "pr2.sandbox",
     "url": "http://pr2.sandbox.opencraft.com/",
     "updates_feed": "https://github.com/antoviaque/edx-platform/commits/OC-791-course-timeline.atom",
```
mtyaka: Remove trailing comma.

mtyaka:
@smarnach This looks good, nice improvement!

I have two comments:
* The JS tests are failing, I believe that's because of the faulty commas in `instance/tests/fixtures/api/instances_list.json`
* While you're doing this, can you also change the code in https://github.com/open-craft/opencraft/blob/f99cf233b369ae054e2bf1f2f2b2a3ffd3fb8781/instance/templates/instance/ansible/vars.yml#L67-L71 so that `instance.github_admin_username_list` doesn't get invoked twice while rendering the template?

instance/tests/models/test_instance.py
```diff
@@ -137,9 +137,10 @@ def test_github_admin_username_list_default(self):
         By default, no admin should be configured
         """
         instance = OpenEdXInstanceFactory()
+        instance.reset_ansible_settings()
```
antoviaque: @smarnach Rather than adding this to the tests (and having to remember to do this every time an instance is created), is there a way to ensure the settings are correctly set upon the object creation? Maybe through the manager create method and/or the on_pre_save hook, before the initial object creation?

antoviaque:
Agreed, definitely a good change - thanks!

instance/tests/models/test_instance.py
smarnach: @antoviaque I actually think freezing the configuration at the time the instance is provisioned is the right thing to do.

The manager create method is on the `GitHubInstanceQuerySet` class, which is the wrong place to add anything Ansible-specific.  I would have to create a mixin for the query set that is combined with `GitHubInstanceQuerySet` to take care of this.

The `on_pre_save()` hook is also the wrong place, since we don't want to reset the configuration every time the instance is saved.  The configuration can only possibly change at the time the instance is provisioned, and before it's provisioned, it doesn't really have an Ansible config.

The only place where you have to think of calling the method is in the tests.  I could add a hook to the factory class to avoid that problem.

smarnach:
@mtyaka Thanks for the review and the comments.

The reasion I didn't fix the duplicate invocation of `instance.github_admin_username_list` is that it can't be fixed in template code, since Django template are dumb, and I was too lazy to implement a proper fix, which would be to add some caching infrastructure and cache certain Github API results for a short time.

I can just add another context variable in this case, so I'll do that for now.

smarnach:
@mtyaka Can you take another look, please?

instance/tests/fixtures/api/instances_list.json
```diff
@@ -42,6 +42,7 @@
...
...contact@example.com'\nEDXAPP_SERVER_EMAIL: 'contact@example.com'\nEDXAPP_BULK_EMAIL_DEFAULT_FROM_EMAIL: 'contact@example.com'\n\n# Misc\nEDXAPP_TIME_ZONE: 'UTC'\n\n# Security updates\nCOMMON_SECURITY_UPDATES: true\nSECURITY_UNATTENDED_UPGRADES: true\nSECURITY_UPDATE_ALL_PACKAGES: false\nSECURITY_UPGRADE_ON_ANSIBLE: true\n\n# Repositories URLs\nedx_ansible_source_repo: 'https://github.com/edx/configuration.git'\nedx_platform_repo: 'https://github.com/antoviaque/edx-platform.git'\n\n# Pin down dependencies to specific (known to be compatible) commits.\nedx_platform_version: '40952060bc7573c5035620b985b211a24360f311'\nconfiguration_version: 'master'\nforum_version: 'master'\nnotifier_version: 'master'\nxqueue_version: 'master'\ncerts_version: 'master'\n\n# Features\nEDXAPP_FEATURES:\n USE_MICROSITES: false\n AUTH_USE_OPENID: false\n ENABLE_DISCUSSION_SERVICE: true\n ENABLE_INSTRUCTOR_ANALYTICS: true\n ENABLE_INSTRUCTOR_EMAIL: true\n REQUIRE_COURSE_EMAIL_AUTH: false\n ENABLE_PEARSON_HACK_TEST: false\n SUBDOMAIN_BRANDING: false\n SUBDOMAIN_COURSE_LISTINGS: false\n PREVIEW_LMS_BASE: 'pr2.sandbox.opencraft.com'\n ENABLE_DJANGO_ADMIN_SITE: true\n ALLOW_ALL_ADVANCED_COMPONENTS: true\n\n\n",
```
mtyaka: nit: indentation

mtyaka:
@smarnach One small nit, but otherwise looks great :+1: 

It looks like we are over the OpenStack instance quota limit, which is making integration tests fail. Can we increase the quota or something to make tests green? /cc @antoviaque

smarnach:
@mtyaka Oh, that was an unexpectedly fast review! :)  I'll wait for the integration tests to pass before merging.

instance/tests/models/test_instance.py
antoviaque: @smarnach The problem if you only update it at provisioning is that https://github.com/open-craft/opencraft/pull/26/files#diff-a8e64e20aeb5487bab76b72ae22e9aefR117 is wrong before the provisioning has started, ie you can't preview the configuration on the Web UI that would be applied to the next provisioning. 

That said, imho the right approach for configurations (beyond the scope of this task) would be to eventually have revisions, each provisioning pointing to a specific configuration revision. That would allow to keep track of the current active configuration, as well as the current, up to date, result of the other fields (ie the preview of the configuration that would be applied to the next provisioning). Since my comment could be better taken care of in that context, that's fine to keep your current approach for now.

antoviaque:
@smarnach @mtyaka Fixed for the quotas - seems like the stage environment went a little crazy on the reprovisionings of the continuous integration. Stopped & cleared the VMs, and restarted a build.

PR #27, opened 2015-11-10
Move vagrant provisioning script into separate file

omarkhan:
This allows it to be used for setting up development environments without vagrant, on a vanilla ubuntu install or on an EC2 instance for example.

This PR is a result of my attempts to set up an EC2 instance to run integration tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ so you can access the development server using your web browser.
 
 ### Local install
 
-If you prefer not to install Vagrant, you can install OpenCraft manually.
+If you prefer not to install Vagrant, you can install OpenCraft manually. Refer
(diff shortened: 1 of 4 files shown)


antoviaque:
@omarkhan Thanks for this - definitely a good idea. Could you just move the script to `bin/` directory, so we don't progressively clutter the root directory?

I'm also running the integration tests, but since this isn't affecting the code, it should hopefully pass without issue. I'll test once locally once the script has been moved and merge.

omarkhan:
@antoviaque done

antoviaque:
@omarkhan Thanks! I started the integration tests & a `vagrant up` - currently running.

antoviaque:
@omarkhan Looking all good on both fronts. :+1:

PR #28, opened 2015-11-11
Demo of Resource Abstraction and Finite State Machines in OpenCraft IM

antoviaque:
Reposting https://gist.github.com/bradenmacdonald/f05a269bb8660cd76930 as a PR to make it easier to comment

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -1,0 +1,146 @@
+class Resource(models.Model):
+    client = ForeignKey(Client)
+    owned_by = ForgeignKey(Resource, null=True)
+    name = CharField(blank=False, max_length=128)
+    configuration = JSONField(null=False)
+    types = ResourceTypeRegistry()
+    owned_resources = ResourceManger()
+
+    @property
+    def configuration_schema(self):
+        """ Define the options used to configure this resource. """
+        return {}  # No configurable options by default
+
+    @property
+    def state(self):
+        """
+        The current state of this resource.
+        The return value is always a subclass of ResourceState, and typically will be a subclass of:
+          NeedsRefreshState, ApplyingConfigurationState, NeedsInputState, ReadyState, ErrorState, or DestroyingState
+        Each state supports common methods like get_status_message(), get_status_color(), allow_configuration_changes()
+        """
+        return self._state
+
+    def build_form():
+        """ Convert self.configuration_schema into a Django form """
+        pass
+
+
+class VmProvider(Resource):
+    """ A provider that can provision virtual machines """
+    class Meta:
+        proxy = True
+
+    def get_flavors(self):
+        """ Get available VM sizes """
+        raise NotImplementedError
+    def list_instances(self):
+        """ List instances currently provisioned on this account. """
+        raise NotImplementedError
+    def provision_vm(self, flavor, image):
+        """ Start the process of provisioning a new VM. Returns a VmInstance """
+        raise NotImplementedError
+
+
+class OpenStackVmProvider(VmProvider):
+    # Implementation here.
+
+
+class VmInstance(Resource):
+    """
+    A virtual machine.
+    configuration options: flavor, image, is_powered_on
+    """
+    class Meta:
+        proxy = True
+
+
+class OpenStackVmInstance(VmInstance):
+    # Implementation here
+
+
+class OpenEdxSandboxInstance(Resource):
+    """
+    A single-server installation of Open edX.
+    Useful for development, QA, demos, etc.
+    """
+    class Meta:
+        proxy = True
+
+    @property
+    def configuration_schema(self):
+        return ResourceConfiguration(
+            platform_repo=Option(
+                type=str,
+                required=True,
+                default="https://github.com/edx/edx-platform.git",
+            ),
+            platform_commit=Option(
+                type=str,
+                required=True,
+                default="master",
+            ),
+            vm_provider=Option(
+                type=Resource.types.VmProvider,
+                required=True,
+            ),
+            vm_flavor=Option(
+                type=str,
+                required=True,
+            ),
+            mysql_server=Option(
+                type=Resource.types.MysqlServer,
+                null=True,
+                default=None,
+                description="Leave blank to use an ephemeral MySQl server on the sandbox VM.",
+            ),
+            mongo_server=Option(
+                type=Resource.types.MongodbServer,
+                null=True,
+                default=None,
+                description="Leave blank to use an ephemeral MongoDB server on the sandbox VM.",
+            ),
+            ansible_vars=Option(
+                type=str,
+                default="",
+            ),
+        )
+
+    class ApplyingConfigurationState(ApplyingConfigurationState):
+        # self.conf accesses the Resource's current 'configuration' data
+        # self.owned_resources accesses the Resource's list of owned resources
+        def enter_state(self):
+            if self.conf.vm_provider.changed:
+                if self.owned_resources["vm"]:
+                    # Free the old VM (code will set its owner to none and queue deprovisioning)
+                    del self.owned_resources["vm"]
+                # First things first: Provision a virtual machine for use.
+                self.owned_resources["vm"] = self.conf.vm_provider.provision_vm(flavor=self.conf.vm_flavor, image="ubuntu12.04.5")
+            elif self.conf.vm_flavor.changed:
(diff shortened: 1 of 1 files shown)


demo.py
```diff
@@ -0,0 +1,146 @@
+class Resource(models.Model):
+    client = ForeignKey(Client)
```
antoviaque: Project?

demo.py
```diff
@@ -0,0 +1,146 @@
...
+    owned_by = ForgeignKey(Resource, null=True)
+    name = CharField(blank=False, max_length=128)
+    configuration = JSONField(null=False)
```
antoviaque: How does this configuration field matches/interacts with the existing configuration in OpenCraft IM? Ie:
* The django config set from the env
* The configuration set in model fields on OpenEdXInstance 
* The configuration set in ansible/yaml (from the PR for example)

I assume that Resources are meant to be subclassed, right? Why not keeping it as individual model attributes?

demo.py
```diff
@@ -0,0 +1,146 @@
...
+    types = ResourceTypeRegistry()
```
antoviaque: What would be the role a resource type registry?

demo.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def build_form():
+        """ Convert self.configuration_schema into a Django form """
+        pass
+
+
+class VmProvider(Resource):
+    """ A provider that can provision virtual machines """
+    class Meta:
+        proxy = True
```
antoviaque: Why not using a full-fledged multi-table inheritance, rather than a proxy model? That would allow to have custom fields for each resource type - trying to fit every resource type into a single data model might be very constrained, very fast. And the logic would still have to differ.

demo.py
```diff
@@ -0,0 +1,146 @@
...
+    def configuration_schema(self):
+        """ Define the options used to configure this resource. """
+        return {}  # No configurable options by default
+
+    @property
+    def state(self):
+        """
+        The current state of this resource.
+        The return value is always a subclass of ResourceState, and typically will be a subclass of:
+          NeedsRefreshState, ApplyingConfigurationState, NeedsInputState, ReadyState, ErrorState, or DestroyingState
```
antoviaque: States are already more fine-grained than this - cf https://github.com/open-craft/opencraft/blob/545e0a2d43358f9ca0815e0aa9c19c62a5a144ce/instance/models/server.py#L99-L122

Also, there are two different notions in terms of status - the step it is in ("provisioning", "booting", etc.), and its progression/result within that step (running, completed, error...). Using classes could be a good way to refine or combine these - but with the list above it's not clear how we would express things like "Error while applying configuration" (which would be a combination of ApplyingConfigurationState and ErrorState, and is different from "Error while destroying" or "Error while rebooting" for example). Or "Ready but needs refresh" vs "Error but needs refresh". It's good to refactor, but it would be good to start from the existing implementation, to focus on changing the parts for which we have identified a specific improvement need.

demo.py
```diff
@@ -0,0 +1,146 @@
...
+        Each state supports common methods like get_status_message(), get_status_color(), allow_configuration_changes()
+        """
+        return self._state
+
+    def build_form():
+        """ Convert self.configuration_schema into a Django form """
+        pass
```
antoviaque: Wouldn't it be better if we could use ModelForms the usual way, rather than having to go a bit around Django to generate them automatically? What do we gain from it?

demo.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    class Meta:
+        proxy = True
+
+
+class OpenStackVmInstance(VmInstance):
+    # Implementation here
+
+
+class OpenEdxSandboxInstance(Resource):
```
antoviaque: How would this relate to an OpenEdX production instance resource?

demo.py
bradenmacdonald: Yeah, +1 for "Project" instead of client.

demo.py
bradenmacdonald: My idea here was that subclasses would store all their fields in the `configuration` JSONField inherited from the parent class. The advantage of that is that it allows us to define many different Resource subclasses, and change them as needed, without having to deal with database migrations every time we add a new Resource. It also allows the parent class to "see" the configuration of children and know when it's changed, because we only need to know when that one JSONField changes. If we use multi-table inheritance, then detecting changes to child configuration/fields becomes a little bit harder to maintain (need to use save signals or override the save method consistently, etc. and even then you can't easily tell if any values were actually changed or not).

Also, since we're on PostgreSQL, a `JSONField` can be very efficient and even allow database-level filtering on the field data.

demo.py
bradenmacdonald: @antoviaque This `configuration` is meant to be used by subclasses to store arbitrary data: see earlier comment at https://github.com/open-craft/opencraft/pull/28/files#r44956742 . Admittedly that's not clear here :)

demo.py
bradenmacdonald: See https://github.com/open-craft/opencraft/pull/28/files#r44956742 - if subclasses don't have "real" fields and instead store their fields in `configuration`, that's why. Plus this is more flexible without being *that* complicated - we are still using Django forms, just not Django ModelForms.

But keep in mind, this is just an implementation detail and not a key part of the overall concept here.

demo.py
bradenmacdonald: It would probably be an independent class, with a mixin used to implement common functionality. There would be quite a number of differences I suspect.

demo.py
bradenmacdonald: The idea of finite state machines is that in the rest of your code you generally don't have a list of all the valid states. Admittedly I haven't been particularly strict about that here. But instead of treating states as a fancy enumeration:

```python
if obj.state in (ReadyState, WaitingState):
   # do something
```

you instead just call methods on the state, and you don't care what the state actually is:

```python
if obj.state.is_healthy():
    # do something
```

Also, in terms of expressing things like "Error while applying configuration", you could do that by simply asking each state for a `state.status_message()`, `state.status_message_color()`, `state.full_status_message()`.

antoviaque: I see, interesting - I've never used state machine before, so sorry for the newbie comments : )

Here, does that mean that, for example, NeedRefresh, ReadyState & ErrorState would actually need to be turned into methods on other states? I would be curious to see how the current states in the app match against a state machine.

demo.py
antoviaque: @bradenmacdonald Would adding new resource types be such a frequent occurrence that it needs to be optimised to that extent? Plus, that removes the django migrations, but the need to maintain and migrate that data isn't going to disappear - it means when the "model" of a configuration field is going to evolve, we'll have to do the work of django migrations ourselves.

For detecting configuration changes from children, that seem looks like something the dirty field pattern would solve well, no? There is even a lib for it: https://github.com/romgar/django-dirtyfields

For "seeing" the configuration of children, are you referring to the fact that queries on the Resource manager would return Resource objects, not children objects? Could https://django-model-utils.readthedocs.org/en/latest/managers.html#inheritancemanager help with this?

In a nutshell, I'm wondering if it wouldn't be better to reuse existing well-supported patterns, rather than reimplementing something entirely new of our own baking. Things like Django migrations, Model fields, ModelForms & co work well already, and putting ourselves outside of that pattern will necessarily generate work to replace it, taking the place of other work, which could be more directly linked to our current goals?

demo.py
bradenmacdonald: The resource type registry was to be a class (static) property that can enumerate all of the subclasses of `Resource`. But looking at it now, it's probably not necessary.

There is an example of its use as I was thinking below in this PR:

```
   mysql_server=Option(
                type=Resource.types.MysqlServer,
                null=True,
                default=None,
                description="Leave blank to use an ephemeral MySQl server on the sandbox VM.",
            ),
```

Although if we refactor to Django multi-table inheritance models instead of my idiosyncratic `configuration` scheme, it could probably look instead like:

```
class OpenEdxSandboxInstance(Resource):
    mysql_server = ResourceField(
        type=MysqlServer,
        null=True,
        default=None,
        help_text=_("Leave blank to use an ephemeral MySQl server on the sandbox VM."),
    )    
```

Where `ResourceField` is a subclass of `ForeignKey` that implements an appropriate `limit_choices_to` filter set so that each resource can only link to other resources in the same project, etc.

antoviaque:
Closing now that #51 is open.

PR #29, opened 2015-11-12
Refactor settings for the fork and branch of the Ansible configuration repo.

smarnach:
This commit sets the Ansible configuration repo and fork from the extra settings given in the PR body, and falls back to a default configured in the Django settings if none is given.

### DIFF
--- a/instance/github.py
+++ b/instance/github.py
@@ -22,8 +22,11 @@ GitHub Service API - Helper functions
 
 # Imports #####################################################################
 
+import functools
+import operator
 import re
 import requests
+import yaml
 
 from django.conf import settings
 
@@ -165,6 +168,18 @@ class PR:
         Extra settings contained in the PR body
         """
         return get_settings_from_pr_body(self.body)
+
+    def get_extra_setting(self, name, default=None):
+        """
+        Return the setting given by "name" from extra_settings.
+
+        The name may be a dot-separated path to retrieve nested settings.
+        """
+        extra_settings_dict = yaml.load(self.extra_settings) or {}
+        try:
+            return functools.reduce(operator.getitem, name.split('.'), extra_settings_dict)
+        except KeyError:
+            return default
 
     @property
     def github_pr_url(self):
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -30,7 +30,6 @@ from functools import partial
 from django.conf import settings
 from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models
-from django.db.models.signals import pre_save
 from django.template import loader
 from django.utils import timezone
 from django_extensions.db.models import TimeStampedModel
@@ -184,17 +183,15 @@ class Instance(ValidateModelMixin, TimeStampedModel):
         """
         return {'instance_id': self.pk}
 
-    @staticmethod
-    def on_pre_save(sender, instance, **kwargs):
+    def save(self, **kwargs):
         """
-        Triggered by the pre_save event
+        Set default values before saving the instance.
         """
-        self = instance
-
         # Set default field values from settings - using the `default` field attribute confuses
         # automatically generated migrations, generating a new one when settings don't match
         if not self.base_domain:
             self.base_domain = settings.INSTANCES_BASE_DOMAIN
+        super().save(**kwargs)
 
     @staticmethod
     def _sort_log_entries(server_logs, instance_logs):
@@ -455,10 +452,8 @@ class AnsibleInstanceMixin(models.Model):
     """
     An instance that relies on Ansible to deploy its services
     """
-    ansible_source_repo_url = models.URLField(max_length=256,
-                                              default='https://github.com/edx/configuration.git')
-
-    configuration_version = models.CharField(max_length=50, default='master')
+    ansible_source_repo_url = models.URLField(max_length=256, blank=True)
+    configuration_version = models.CharField(max_length=50, blank=True)
     ansible_playbook_name = models.CharField(max_length=50, default='edx_sandbox')
     ansible_extra_settings = models.TextField(blank=True)
     ansible_settings = models.TextField(blank=True)
@@ -472,6 +467,18 @@ class AnsibleInstanceMixin(models.Model):
 
     class Meta:
         abstract = True
+
+    def save(self, **kwargs):
+        """
+        Set default values before saving the instance.
+        """
+        # Set default field values from settings - using the `default` field attribute confuses
+        # automatically generated migrations, generating a new one when settings don't match
+        if not self.ansible_source_repo_url:
+            self.ansible_source_repo_url = settings.DEFAULT_CONFIGURATION_REPO_URL
+        if not self.configuration_version:
+            self.configuration_version = settings.DEFAULT_CONFIGURATION_VERSION
+        super().save(**kwargs)
 
     @property
     def ansible_playbook_filename(self):
@@ -676,5 +683,3 @@ class OpenEdXInstance(AnsibleInstanceMixin, GitHubInstanceMixin, Instance):
         self.logger.info('Provisioning completed')
 
         return (server, log)
-
-pre_save.connect(OpenEdXInstance.on_pre_save, sender=OpenEdXInstance)
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -73,6 +73,8 @@ def watch_pr():
                             .format(pr=pr, i=instance, truncated_title=truncated_title)
             instance.github_pr_url = pr.github_pr_url
             instance.ansible_extra_settings = pr.extra_settings
+            instance.ansible_source_repo_url = pr.get_extra_setting('edx_ansible_source_repo')
+            instance.configuration_version = pr.get_extra_setting('configuration_version')
             instance.save()
 
             if created:
--- a/pylintrc
+++ b/pylintrc
@@ -94,13 +94,13 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
-    no-init,
     fixme,
     locally-disabled,
+    no-init,
     too-few-public-methods,
     too-many-ancestors,
-    unused-argument,
-    protected-access
+    protected-access,
+    unused-argument
 
 [SPELLING]
 
--- a/instance/migrations/0036_auto_20151112_1122.py
+++ b/instance/migrations/0036_auto_20151112_1122.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0035_reset_ansible_settings'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='ansible_source_repo_url',
+            field=models.URLField(blank=True, max_length=256),
+        ),
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='configuration_version',
+            field=models.CharField(blank=True, max_length=50),
+        ),
+    ]
--- a/instance/tests/test_tasks.py
+++ b/instance/tests/test_tasks.py
@@ -22,6 +22,8 @@ Worker tasks - Tests
 
 # Imports #####################################################################
 
+import textwrap
+
 from mock import patch
 
 from instance import github, tasks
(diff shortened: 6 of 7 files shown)


instance/github.py
```diff
@@ -158,13 +159,8 @@ def __init__(self, number, source_fork_name, target_fork_name, branch_name, titl
...
         self.body = body
-
-    @property
-    def extra_settings(self):
-        """
-        Extra settings contained in the PR body
-        """
-        return get_settings_from_pr_body(self.body)
+        self.extra_settings = get_settings_from_pr_body(body)
+        self.extra_settings_dict = yaml.load(self.extra_settings) or {}
```
antoviaque: Can we avoid duplicating the data? That requires to keep `extra_settings_dict` manually in sync with `extra_settings`, which itself has to be kept in sync with `pr_body`. Duplicating data has a tendency to add lots of subtle bugs on the long term. We should make sure we have our data structure right, if we feel the need to duplicate it it might be indicative of an underlying problem - I'd rather have the data set once, and then computed from that single place by the rest of the code.

instance/tasks.py
```diff
@@ -73,6 +73,14 @@ def watch_pr():
...
             instance.github_pr_url = pr.github_pr_url
             instance.ansible_extra_settings = pr.extra_settings
+            instance.ansible_source_repo_url = pr.extra_settings_dict.get('edx_ansible_source_repo')
+            if not instance.ansible_source_repo_url:
+                instance.ansible_source_repo_url = 'https://github.com/{}.git'.format(
+                    settings.DEFAULT_CONFIGURATION_FORK
+                )
+            instance.configuration_version = pr.extra_settings_dict.get('configuration_version')
+            if not instance.configuration_version:
+                instance.configuration_version = settings.DEFAULT_CONFIGURATION_VERSION
```
antoviaque: Can we keep the tasks small? Defaults here feel out of place - we would have to duplicate this code every time we need to set an instance with the default settings, for example when doing it from the shell.

instance/models/instance.py
```diff
@@ -455,10 +455,8 @@ class AnsibleInstanceMixin(models.Model):
     """
     An instance that relies on Ansible to deploy its services
     """
-    ansible_source_repo_url = models.URLField(max_length=256,
-                                              default='https://github.com/edx/configuration.git')
-
-    configuration_version = models.CharField(max_length=50, default='master')
+    ansible_source_repo_url = models.URLField(max_length=256, blank=True)
+    configuration_version = models.CharField(max_length=50, blank=True)
```
antoviaque: Sorry, I realize that I'm pushing back on your whole PR, but I'd rather always have a value set for these fields; otherwise changing the default in the settings will change the value for all instances, which is what you meant to do here, but imho has unintended consequences - for example, it would also affects old instances, long after they have been deprovisioned, revising the history. Or if someone sets the configuration_version but not the repo_url, changing the default will break the instance provisioning if the commit isn't also on the fork. If we need to change the default, I'd rather explicitly update the fields, to ensure we only change values where it needs to.

Can we use the existing pattern for defaults set from the config, ie https://github.com/open-craft/opencraft/blob/545e0a2d43358f9ca0815e0aa9c19c62a5a144ce/instance/models/instance.py#L194-L197 ?

instance/github.py
smarnach: @antoviaque My viewpoint here was that `PR` instances

(a) are ephemeral.  They are created, used and dropped, but not kept around for a longer time or ever changed.

(b) are duplicating data anyway, since the golden version of the data is on Github, so even if we do keep them around, we have to keep them in sync with the version on Github.  It would be easy to make sure all fields are updated when resyncing with Github.

The motivation of this change is that it becomes quite computationally expensive if we need to parse the whole PR body for the settings, extract them and run the YAML parser everytime we want to access a setting.  I'm happy to turn them into properties again, but I'd like to remind you that the last bug I fixed was caused by too many things being properties and recomputed everytime when actually they shouldn't be, so you have to look at each individual case to figure out the right approach.

instance/models/instance.py
smarnach: @antoviaque Unfortunately, at the moment it's not defined what these fields are supposed to mean.  When I added code to freeze the Ansible settings at provisioning time, you had concerns that now the field with the Ansible settings doesn't reflect the configuration of the _next_ provisioning anymore. This time, you have concerns that changing a field after an instance has been deployed revises history.  In some way, this seems to be the opposite of what you said last time. :)

I think we have to redefine what an `OpenEdxInstance` is, philosophically.  Currently, it's a template that can be used to deploy a set of Open edX instances to an OpenStack cluster, but it's representing the deployed instances as well.  This mixup is also showing in the UI, e.g. if you redeploy an instance, the log for the new provisioning run is simply appended to all existing log entries, making it a bit hard to find where the current provisioning log starts.  We probably need to separate the two mentioned conepts a little better to get clearer notions here.  If we want to keep the history of provisionings correctly, we have to freeze the whole configuration everytime a provisioning run is started, and keep that.  On the other hand, the template for new provisioning runs is allowed to dynamically follow the current settings on the PR – and all global default values, if nothing was set explicitly for the instance.

To me, the default values felt out of place here, since they were the only things that were specific to Open edX on this whole mixin class.

mtyaka: That's a very good observation. If we want to keep a provisioning history, we will need a separate model that will represent a single provisioning attempt/deployment of an `OpenEdxInstance`. It would store frozen configuration variables as they were at the time that particular provisioning was run. I would go further than simply storing `master`, if we wanted accurate history and repeatable builds, we will have to store the commit hash rather than the branch name. We would also have to freeze the version of each service that gets deployed to sandboxes (`cs_comments_service`, `notes_api`, `xqueue`, ...).

However since something like that is clearly out of scope of this PR, I suggest we do this for now:
- Load the default in a pre-save hook, similar to https://github.com/open-craft/opencraft/blob/545e0a2d43358f9ca0815e0aa9c19c62a5a144ce/instance/models/instance.py#L194-L197
- When reprovisioning an instance that was created from a github PR, check whether `edx_ansible_source_repo` and/or `configuration_version` are present, and use that, otherwise set `instance.ansible_source_repo` and `instance.configuration_version` to the defaults (even though they were not blank before)

This would make it possible to switch to a configuration fork when upstream is broken while not breaking existing instances that point to specific configuration repo/version.

Instances not created from PRs (although I'm not really sure we have any like that currently) would not be affected if defaults were changed. They would keep using stored `ansible_source_repo` and `configuration_version`. Existing instances created from PRs that had specific settings for these variables set in the PR body, would also not be affected. They would keep using the version/repo from PR. Existing PR instances without these settings would pick up the new defaults on each provisioning.

instance/tasks.py
```diff
@@ -73,6 +73,14 @@ def watch_pr():
                             .format(pr=pr, i=instance, truncated_title=truncated_title)
             instance.github_pr_url = pr.github_pr_url
             instance.ansible_extra_settings = pr.extra_settings
+            instance.ansible_source_repo_url = pr.extra_settings_dict.get('edx_ansible_source_repo')
+            if not instance.ansible_source_repo_url:
+                instance.ansible_source_repo_url = 'https://github.com/{}.git'.format(
```
mtyaka: Nit/personal preference: I would rather have the default setting hold the entire repo url, rather than interpolating the fork/repo name.

instance/tasks.py
mtyaka: I believe this code needs to be here to force fields to the new defaults when creating instances from PRs (unless values are explicitly provided in the PR body).

However, as mentioned above, I would prefer if the default was also set in a pre-save hook so that the value in the db would never be blank for these fields.

instance/models/instance.py
mtyaka: Actually, looking at the `watch_pr` method, this is already how it works, except that it doesn't set the defaults in a pre-save hook yet. And I am not sure what happens if you click the 'reprovision' button on an instance that was created from a PR - is the PR body re-read when reprovisioning? I believe it should be if it currently isn't.

smarnach:
@mtyaka Please take another look.

smarnach:
@mtyaka Argh, wrong PR.  I'll ping you on the right one as well.

instance/tasks.py
smarnach: My reasoning for the settings to be here was that I considered them specific to the application of creating an instance for a PR.  Since I added the defaults to the `save()` method now, they can be removed here.

instance/models/instance.py
smarnach: @mtyaka My reasoning for what I did was more or less identical to what you wrote above.  I've adopted the approach slightly to set defaults automatically before saving, since this is the approach @antoviaque and you prefer.

instance/tasks.py
smarnach: I actually prefer that as well.  I used the fork to be consistent with the DEFAULT_FORK setting for the platform, changes now.

smarnach:
@mtyaka This is ready for another look now as well.  (This time for real.)

instance/models/instance.py
mtyaka: @smarnach Oh right... I only now realize that your previous code resulted in pretty much same behaviour. I think the idea of ending up with empty fields in the database distracted me from seeing that it doesn't actually matter much in this case. Anyway, thanks for the updates :)

mtyaka:
@smarnach :+1: assuming the build is successful

instance/github.py
antoviaque: @smarnach 

(a) Yup, they are currently ephemeral; I'm not fully clear on how we'll handle production instances yet, but it seem likely that PR instances will remain this way. We'll likely also manage releases of production instances with PRs, but the production instances themeslves would likely more be linked to a branch - although the rc PR might be the equivalent of stage?

They can however be kept for a certain amount of time (some PRs take months to be merged - the longest almost a year!).

But most importantly - they *can* change. Especially the ansible configuration - the initial PR might not have all the settings, some names might change during reviews. And feature branch PRs might end up being quite complex, and evolving a lot.

(b) Yup, there are always cases where the drawbacks of duplicating data are outweight by other criterias. Having the data on another service is usually a good line - the service might become unavailable, we might want to access the data more often than the service allow us to fetch it, the delays might become too big, etc. But from past experience, the threshold can be fairly high - the cost of keeping duplicated data updated expands exponentially as the application increases in size and complexity. I've found it useful to prioritize having getting the right data structure, over things like having efficient code for example - human brain cycles are much more expensive than computer ones : )

> I'd like to remind you that the last bug I fixed was caused by too many things being properties and recomputed everytime when actually they shouldn't be, so you have to look at each individual case to figure out the right approach.

Yup, definitely not an absolute, just something I'd like to be mindful of.

instance/models/instance.py
antoviaque: @smarnach @mtyaka Good point for the concept of deployment - it would definitely deserve its own class. I've added an entry in the uservoice about this: https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/10727073-provisioning-history

instance/models/instance.py
```diff
@@ -473,6 +468,18 @@ class AnsibleInstanceMixin(models.Model):
     class Meta:
         abstract = True
 
+    def save(self, **kwargs):
```
antoviaque: @smarnach That's much better than calling the pre_save hook - silly me for not thinking of this!

Thanks for this change, and for adding the defaults here.

pylintrc
```diff
@@ -94,13 +94,14 @@ confidence=
...
 disable=
-    no-init,
     fixme,
     locally-disabled,
+    no-init,
     too-few-public-methods,
     too-many-ancestors,
-    unused-argument,
-    protected-access
+    too-many-instance-attributes,
```
antoviaque: @smarnach I thought I had commented on this the first time, but I must have forgotten to submit it - can we address pylint's complaint here, or disable it locally if really it doesn't make sense here, rather than disabling it globally? Although that might tie into my comment on https://github.com/open-craft/opencraft/pull/29#discussion_r44755583 and not be necessary once switched back to attributes.

antoviaque:
:+1: from me too, once https://github.com/open-craft/opencraft/pull/29#discussion_r44755583 and https://github.com/open-craft/opencraft/pull/29#discussion_r44954732 have been addressed.

pylintrc
smarnach: @antoviaque I've never found this particular warning useful, so I wasn't shy disabling it globally.

instance/github.py
antoviaque: Although here a middle way could be to set `extra_settings` as a new attribute, and remove `pr_body` altogether, or turn it into a method or setter. You're right that it probably makes more sense to standardize the data structure around `extra_settings`, which would be used by both PRs instances and other types of instances, rather than `pr_body` which would only be used by PR instances.

smarnach: @antoviaque I had already changed this before I saw your last comment.  We should get this PR out, so we can use swap by default.  I'll prepare a new small PR with a suggestion how to make this nicer, if you are fine with committing the latest version of this PR.

antoviaque: @smarnach That works for me! Thanks for the changes.

PR #30, opened 2015-11-12
Allow to set timeouts for playbook runs.

smarnach:
This PR introduces two new settings:

* `ANSIBLE_LINE_TIMEOUT`: The time in seconds to wait for the next log line when running an Ansible playbook.
* `ANSIBLE_GLOBAL_TIMEOUT`: The timeout in seconds for an entire Ansible playbook.

These settings make sure playbook runs don't hang indefintely.

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -41,7 +41,7 @@ from instance.github import fork_name2tuple, get_username_list_from_team
 from instance.logging import log_exception
 from instance.logger_adapter import InstanceLoggerAdapter
 from instance.repo import open_repository
-from instance.utils import read_files
+from instance.utils import poll_streams
 
 from instance.models.utils import ValidateModelMixin
 
@@ -528,7 +528,7 @@ class AnsibleInstanceMixin(models.Model):
             username=settings.OPENSTACK_SANDBOX_SSH_USERNAME,
         ) as process:
             try:
-                log_line_generator = read_files(
+                log_line_generator = poll_streams(
                     process.stdout,
                     process.stderr,
                     line_timeout=settings.ANSIBLE_LINE_TIMEOUT,
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -526,16 +526,26 @@ class AnsibleInstanceMixin(models.Model):
             playbook_path,
             self.ansible_playbook_filename,
             username=settings.OPENSTACK_SANDBOX_SSH_USERNAME,
-        ) as processus:
-            for fd, line in read_files(processus.stdout, processus.stderr):
-                line = line.decode('utf-8').rstrip()
-                if fd == processus.stdout:
-                    self.logger.info(line)
-                elif fd == processus.stderr:
-                    self.logger.error(line)
-                log_lines.append(line)
-            processus.wait()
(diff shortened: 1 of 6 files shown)


instance/models/instance.py
```diff
@@ -526,16 +526,26 @@ def _run_playbook(self, requirements_path, playbook_path):
...
-                log_lines.append(line)
-            processus.wait()
-            return (log_lines, processus.returncode)
+        ) as process:
+            try:
+                log_line_generator = read_files(
+                    process.stdout,
+                    process.stderr,
+                    line_timeout=settings.ANSIBLE_LINE_TIMEOUT,
+                    global_timeout=settings.ANSIBLE_GLOBAL_TIMEOUT,
```
antoviaque: Nice to see that you were able to have both! This will definitely be useful in other cases in the future imho - thanks for looking into it.

instance/utils.py
```diff
@@ -73,26 +75,50 @@ def get_requests_retry(total=10, connect=10, read=10, redirect=10, backoff_facto
...
+            global_timeout = deadline - time.time()
+            if line_timeout is not None:
+                yield min(line_timeout, global_timeout)
+            else:
+                yield global_timeout
+    else:
+        yield from itertools.repeat(line_timeout)
+
+
+def read_files(*files, line_timeout=None, global_timeout=None):
```
smarnach: Maybe we should rename this function.  `read_files()` doesn't seem a good name, given that the main thing it does is multiplexing and not reading, and it doesn't even work for files. :)

mtyaka: Maybe `poll_streams`? :)

mtyaka:
Nice :+1:

instance/tests/models/test_instance.py
```diff
@@ -354,23 +354,20 @@ def test_run_playbook_logging(self, mock_run_playbook):
...
+            mock_run_playbook.return_value.__enter__.return_value.stderr = stderr
+            mock_run_playbook.return_value.__enter__.return_value.returncode = 0
+            os.write(stdout_w, b'Hello\n')
+            os.close(stdout_w)
+            os.write(stderr_w, b'Hi\n')
+            os.close(stderr_w)
+            instance = OpenEdXInstanceFactory()
+            log, returncode = instance._run_playbook("requirements", "playbook")
+            self.assertCountEqual(log, ['Hello', 'Hi'])
+            self.assertEqual(returncode, 0)
```
mtyaka: Is there (a reasonably simple) way to test the new timeouts?

instance/utils.py
smarnach: @mtyaka Thanks, I will go with that name.

instance/tests/models/test_instance.py
smarnach: @mtyaka To test the timeouts as used in `poll_streams()`, I would have to mock out the selector, which is cumbersome and doesn't provide enough benefit in my opinion.  As a basic level of testing, I added a unit test for the helper function that yields the timeouts instead.

smarnach:
@mtyaka Please take another look.

mtyaka:
@smarnach Looking good :+1:

PR #31, opened 2015-11-17
Refactor the PR class.

smarnach:
An (incomplete) suggestion on how to refactor the PR class.

The new version keeps all data retrieved from Github around, and implements most attributes as properties that simply do a lookup in this data dictionary.  The `extra_settings` which are more expensive to retrieve are cached in read-only attributes.  A method `update_data()` allows to update all fields from Github, and makes sure everything is kept in sync.

The repetitiveness of the code can easily be reduced by storing a mapping of property names to paths in the data dict.  This structure would allow to easily add new properties, like the base branch (the branch targetted by the PR) or the timestamp of the last update.

(I'm also not sure we really need to hide `extra_settings` and `extra_settings_dict` behind read-only attributes.  I usually wouldn't do this, since I dislike code doing nothing.)

### DIFF
--- a/instance/github.py
+++ b/instance/github.py
@@ -91,24 +91,12 @@ def get_settings_from_pr_body(pr_body):
 
 def get_pr_by_number(pr_target_fork_name, pr_number):
     """
-    Returns a PR object based on the reponse
+    Retrieves the JSON description of a PR
     """
-    r_pr = get_object_from_url('https://api.github.com/repos/{pr_target_fork_name}/pulls/{pr_number}'.format(
+    return get_object_from_url('https://api.github.com/repos/{pr_target_fork_name}/pulls/{pr_number}'.format(
         pr_target_fork_name=pr_target_fork_name,
         pr_number=pr_number,
     ))
-    pr_fork_name = r_pr['head']['repo']['full_name']
-    pr_branch_name = r_pr['head']['ref']
-    pr = PR(
-        pr_number,
-        pr_fork_name,
-        pr_target_fork_name,
-        pr_branch_name,
-        r_pr['title'],
-        r_pr['user']['login'],
-        body=r_pr['body'],
-    )
-    return pr
 
 
 def get_pr_list_from_username(user_name, fork_name):
@@ -121,7 +109,7 @@ def get_pr_list_from_username(user_name, fork_name):
 
     pr_list = []
     for pr_dict in r_pr_list['items']:
-        pr = get_pr_by_number(fork_name, pr_dict['number'])
+        pr = PR(fork_name, pr_dict['number'])
         pr_list.append(pr)
     return pr_list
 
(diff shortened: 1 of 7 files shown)


instance/github.py
```diff
@@ -118,7 +109,7 @@ def get_pr_list_from_username(user_name, fork_name):
 
     pr_list = []
     for pr_dict in r_pr_list['items']:
-        pr = get_pr_by_number(fork_name, pr_dict['number'])
+        pr = PR(fork_name, pr_dict['number'])
```
antoviaque: Much nicer, +1

instance/github.py
```diff
@@ -149,26 +140,48 @@ class PR:
...
-        self.branch_name = branch_name
-        self.title = title
-        self.username = username
-        self.body = body
+        self.update_data()
+
+    def update_data(self):
+        self._data = get_pr_by_number(target_fork_name, number)
+        self._extra_settings = get_settings_from_pr_body(self.body)
+        self._extra_settings_dict = yaml.load(self.extra_settings) or {}
```
antoviaque: Here, it makes sense to cache `self._data`, since this is retrieved from an external service.

For `self._extra_settings` and `self._extra_settings_dict`, the extra cache layers seem like premature optimization, if this is just for performance concerns. To keep with our previous conversation about the data structure here, I'd rather duplicate data as little as possible, ie only when it proves to be necessary; I don't think we are getting any performance constraints from generating these as properties, no?

instance/models/instance.py
```diff
@@ -455,10 +452,8 @@ class AnsibleInstanceMixin(models.Model):
     """
     An instance that relies on Ansible to deploy its services
     """
-    ansible_source_repo_url = models.URLField(max_length=256,
-                                              default='https://github.com/edx/configuration.git')
-
-    configuration_version = models.CharField(max_length=50, default='master')
+    ansible_source_repo_url = models.URLField(max_length=256, blank=True)
+    configuration_version = models.CharField(max_length=50, blank=True)
```
antoviaque: Since they would always be set in the save method, can we omit the `blank=True`, to keep validating the values set in configuration against empty values?

instance/tasks.py
```diff
@@ -73,6 +73,8 @@ def watch_pr():
                             .format(pr=pr, i=instance, truncated_title=truncated_title)
             instance.github_pr_url = pr.github_pr_url
             instance.ansible_extra_settings = pr.extra_settings
+            instance.ansible_source_repo_url = pr.get_extra_setting('edx_ansible_source_repo')
+            instance.configuration_version = pr.get_extra_setting('configuration_version')
```
antoviaque: With your current PR, that's `pr.extra_settings_dict.get('edx_ansible_source_repo', '')` -- I would however argue here that keeping the current method, `pr.get_extra_setting('edx_ansible_source_repo')`, is more readable.

instance/tests/test_tasks.py
```diff
@@ -59,6 +61,11 @@ def test_watch_pr_new(self, mock_get_username_list, mock_get_pr_list_from_userna
         """
         New PR created on the watched repo
         """
+        ansible_extra_settings = textwrap.dedent("""\
+            WATCH: true
+            edx_ansible_source_repo: https://github.com/open-craft/configuration
+            configuration_version: named-release/elder
+        """)
```
antoviaque: Why do you need to add these variables in this test? Do we still have a test that still uses all the default values, if we set `edx_ansible_source_repo` and `configuration_version` in this one?

antoviaque:
@smarnach Do we have a ticket in Jira tracking this at the moment?

smarnach:
@antoviaque Thanks for your comments.  This PR was just meant to show the idea in the second commit.  I initially targetted the branch of the PR to refactor the configuration settings, and only that second PR was meant to be included.  Since the other branch was rebased, and old version of the refactoring commit now shows up as first commit in this PR, which is entirely unintentional.

We don't have a JIRA ticket for this.  I spent like ten minutes on the PR (which is why it is incomplete) and posted it to get an agreement before doing any further work.

instance/github.py
smarnach: This is a performance optimization, yes.  Once we start accessing more properties parsed from the PR body, we would otherwise have to reparse the whole body, extract the YAML, parse it and extract whatever we need every single time we want to access one of these properties, which seems excessive.

The point of the refactoring is that this way there is practically no chance of cached data getting out of sync, since the attributes are only accessed in `update_data()`, where all of them are updated at once.

Since we don't seem to get an agreement on this, I'll create a new PR against the master branch with a stripped down version on the second commit in this PR.

antoviaque: > This is a performance optimization, yes. Once we start accessing more properties parsed from the PR body, we would otherwise have to reparse the whole body, extract the YAML, parse it and extract whatever we need every single time we want to access one of these properties, which seems excessive.

I guess that's where we diverge - unless this actually causes performance issues, I don't find this excessive, given that it allows to avoid additional cache layers. On the other side, even isolated in a `update_data()` method, this only gives "*practically* no chance of cached data getting out of sync" - instead of no chance *at all* of getting data out of sync. And over time and increased code complexity, that small difference does tend to get bigger. Why do you want to optimise the speed of this and add that cache layer, before we actually need the speed improvement?

smarnach: @antoviaque It's just a difference in assessing risks, which might be related to the different backgrounds we are coming from.  I've seen many problems arising from performance issues like this in the past, so I tend to fix them right away if it's easy to do and I deem it safe, since doing it later might be more work than getting it right now.  I'm also aware of problems with data getting out of sync, but I think this is just one class of problem to keep in mind, not the only one and maybe not even the most important one.

My background is in high-performance computing and at Google; performance was paramount for both of these.  You are probably right that performance simply doesn't matter here, and I don't think we need to discuss this any further.  We should focus on solving actual problems instead. :)

I'll prepare a stripped-down version of this PR anyway with a few clean-ups, but I won't include caching the settings.

antoviaque: Ah, I see - it's true that Google seem to have a strong culture of high performance/speed, at least from what shows from the outside, so I can see how that would become an habit : )

Yup, that works for me to focus on other things - that said, I think it's useful that we took the time to explain our view points, and understand each other's approach and preferences. It will likely show up again in other places in reviews, so that's likely not lost time, even if the changes here are definitely minor.

And sounds good for the next steps - thanks!

antoviaque:
@smarnach This PR is quite old - do you still plan on pushing a stripped down version without the caching (maybe the next time you take on an OpenCraft IM task?)? Or should we close this?

smarnach:
@antoviaque I still plan to push an updated version at one point.  However, we can close the PR for now and I open a new one.

PR #32, opened 2015-11-26
Fix misnamed env variable.

smarnach:
This typo was introduced when addressing review comments in #29.

cc @antoviaque


antoviaque:
:+1: Thanks for fixing it : )

PR #34, opened 2015-12-06
Add libmysqlclient-dev to the system dependencies.

smarnach:
This package is needed since recently to install the Ansible prerequisites, see
https://github.com/edx/configuration/pull/2537.

Also remove a redundant command from the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ Instructions based on Ubuntu 14.04.
 Install the system package dependencies & virtualenv:
 
 ```
-$ sudo apt-get update
 $ make install_system_dependencies
 $ pip3 install --user virtualenv && pip3 install --user virtualenvwrapper
 ```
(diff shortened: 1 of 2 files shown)


smarnach:
@antoviaque Could you take a look, please?  This is not urgent at all, and I mostly filed it to avoid forgetting about it.

antoviaque:
:+1: For the changes, look good.

There is a test failure in the build though - `msg: '/usr/bin/aptitude safe-upgrade' failed: E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)`. Not sure what caused this on the integration VM - started a rebuild in case this was a transitory issue from the cloud provider.

smarnach:
@antoviaque The restarted build succeeded, so I'd assume it was a transient error.  We should watch out if we see this particular error again, and if we do, we need to add some retry behaviour to make the tests more resilient.  They are already starting to become somewhat flaky, which can't be completely avoided since we depend on many external resources.

Merging now.

PR #37, opened 2016-02-16
Don't instantiate novaclient V2 directly

omarkhan:
It doesn't like it and prints a warning:

    UserWarning: 'novaclient.v2.client.Client' is not designed to be initialized directly. It is inner class of novaclient. Please, use 'novaclient.client.Client' instead. Related lp bug-report: 1493576

### DIFF
--- a/instance/openstack.py
+++ b/instance/openstack.py
@@ -24,7 +24,7 @@ OpenStack - Helper functions
 
 import requests
 
-from novaclient.v2.client import Client as NovaClient
+from novaclient.client import Client as NovaClient
 
 from django.conf import settings
 
@@ -39,11 +39,12 @@ logger = logging.getLogger(__name__)
 
 # Functions ###################################################################
 
-def get_nova_client():
+def get_nova_client(api_version=2):
     """
     Instanciate a python novaclient.Client() object with proper credentials
     """
     nova = NovaClient(
+        api_version,
         settings.OPENSTACK_USER,
         settings.OPENSTACK_PASSWORD,
         settings.OPENSTACK_TENANT,


antoviaque:
@omarkhan You need to get a reviewer on this - maybe the reviewer of the task you were doing when you originally pushed this, or the current sprint fireman if there is nothing urgent pending? Otherwise a task for next sprint.

CC @bradenmacdonald

omarkhan:
@pomegranited if you have time could you have a quick look at this? Thanks

pomegranited:
@omarkhan I haven't run the code, but did look at the [novaclient documentation](http://docs.openstack.org/developer/python-novaclient/api.html) which agrees with your change, and even [warns against the old code](https://bugs.launchpad.net/horizon/+bug/1493576).

:+1: once the checks pass.

omarkhan:
Thanks @pomegranited

antoviaque:
@pomegranited Thanks for the prompt review! But please don't give your +1 to a PR without testing it - otherwise here it means the code would enter production with only Omar ever testing it; the one case where we have sometimes been more lax in terms of giving a +1 without testing is when another reviewer has already tested it - but even there it should only be the case when the time constraints make it necessary. Testing and ensuring proper behavior is fully part of the review process, and an important part of it.

omarkhan:
This is very quick to test if you have a development environment set up: just run the unit tests and check that the warning is not printed to the console. @pomegranited if you don't have a development environment set up for this it might be better for someone else to review it. @jbzdak maybe if you have time? Shouldn't take more than 5 minutes.

antoviaque:
Also, if needed this can be put in a formal task next week - I don't want to hijack any current work that might be more urgent.

CC @bradenmacdonald

jbzdak:
@antoviaque @pomegranited I can volunteer to test it, I will have working IM shortly, so I might as well provision one instance on @omarkhan branch.

omarkhan:
Thanks Jacek. You don't have to provision anything, just checkout this
branch and run the unit tests.

antoviaque:
@omarkhan Sorry to contradict again on this - but running the tests isn't enough to consider a PR tested; the CI already runs the tests; it's the manual testing which is important - not just verifying that the warning message disappeared, but also that it still works correctly. @jbzdak had the right approach proposing to test a provisioning, which will allow to do both.

omarkhan:
@antoviaque got it, sorry - I thought the integration tests were enough for that, but I see your point.

antoviaque:
@omarkhan No worries - we never had a chance to explicitly discuss this I think, and that might not be explained explicitly in the on-boarding docs; so it's actually a good occasion to clarify it.

pomegranited:
@antoviaque Gotcha, thanks for the reminder.

jbzdak:
@omarkhan Just a quick note I'll start provisioning an instance using this settings tomorrow :)

jbzdak:
@omarkhan I have successfully provisioned a sandbox using your code, and also checked that instance was successfully deleted during re provisioning. Works as advertised :+1: (sorry it took so long, due to the certs error it provisioned successfully at third attempt).

omarkhan:
Great, thanks @jbzdak! Merging

antoviaque:
@jbzdak @omarkhan @pomegranited Thank you for getting this merged! : )

PR #38, opened 2016-02-16
Limit the number of log lines displayed

omarkhan:
See [OC-1035](https://tasks.opencraft.com/browse/OC-1035).

#### Problem

The interface takes a very long time to load, as it tries to fetch all log entries for all instances and there can be a lot of data sent over the wire.

#### Solution

This pull request adds a configurable limit on the number of log entries returned in the initial request. You can set the limit using the `LOG_LIMIT` environment variable (default: 10000).

In addition, loading log entries from the server is deferred until the user actually selects an instance.

#### Testing

To run the js tests in your development environment, see https://github.com/open-craft/opencraft/pull/39.

To test manually, create a server and instance with some log entries. We can use the test factories for this:

```python
from instance.tests.models.factories.server import OpenStackServerFactory

server = OpenStackServerFactory.create()
server.log_entry_set.create(level='INFO', text='server log message')
server.instance.log_entry_set.create(level='INFO', text='instance log message')
```

Then, start the dev server with

```
make rundev
```

With your web console open, go to http://localhost:5000/ and select the instance we just created. You will see a second request to fetch the log entries.

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -113,8 +113,6 @@ class Instance(ValidateModelMixin, TimeStampedModel):
 
     last_provisioning_started = models.DateTimeField(blank=True, null=True)
 
-    logger = InstanceLoggerAdapter(logger, {})
-
     class Meta:
         abstract = True
         unique_together = ('base_domain', 'sub_domain')
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -227,30 +227,42 @@ class Instance(ValidateModelMixin, TimeStampedModel):
 
         return log
 
-    def _get_log_entries(self, level_list=None):
+    def _get_log_entries(self, level_list=None, limit=None):
         """
         Return the list of log entry instances for the instance and its current active server,
-        optionally filtering by logging level.
+        optionally filtering by logging level. If a limit is given, only the latest records are
+        returned.
         """
         # TODO: Filter out log entries for which the user doesn't have view rights
         server_log_entry_set = self._current_server.log_entry_set
         if level_list:
             server_log_entry_set = server_log_entry_set.filter(level__in=level_list)
-        server_log_entry_set = server_log_entry_set.order_by('pk').iterator()
+        server_log_entry_set = server_log_entry_set.order_by('pk')
+        if limit:
+            server_log_entry_set = reversed(server_log_entry_set.reverse()[:limit])
+        else:
+            server_log_entry_set = server_log_entry_set.iterator()
 
         instance_log_entry_set = self.log_entry_set
         if level_list:
             instance_log_entry_set = instance_log_entry_set.filter(level__in=level_list)
-        instance_log_entry_set = instance_log_entry_set.order_by('pk').iterator()
+        instance_log_entry_set = instance_log_entry_set.order_by('pk')
+        if limit:
+            instance_log_entry_set = reversed(instance_log_entry_set.reverse()[:limit])
+        else:
+            instance_log_entry_set = instance_log_entry_set.iterator()
 
-        return Instance._sort_log_entries(server_log_entry_set, instance_log_entry_set)
+        entries = self._sort_log_entries(server_log_entry_set, instance_log_entry_set)
+        if limit:
+            return entries[-limit:]
+        return entries
 
     @property
     def log_entries(self):
         """
         Return the list of log entry instances for the instance and its current active server
         """
-        return self._get_log_entries()
+        return self._get_log_entries(limit=settings.LOG_LIMIT)
 
     @property
     def log_error_entries(self):
--- a/instance/static/js/src/instance.js
+++ b/instance/static/js/src/instance.js
@@ -75,9 +75,18 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
             });
         };
 
-        $scope.select = function(selection_type, value) {
-            $scope.selected[selection_type] = value;
-            console.log('Selected ' + selection_type + ':', value);
+        $scope.select = function(instance) {
+            $scope.loading = true; // Display loading message
+            console.log('Selected instance', instance.id);
+
+            return OpenCraftAPI.one('openedxinstance', instance.id).get().then(function(instance) {
+                console.log('Fetched instance', instance.id);
+                $scope.selected.instance = instance;
+            }, function(response) {
+                console.log('Error from server: ', response);
+            }).finally(function () {
+                $scope.loading = false;
+            });
         };
 
         $scope.provision = function(instance) {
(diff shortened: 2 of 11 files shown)


instance/models/instance.py
```diff
@@ -207,52 +205,61 @@ def _sort_log_entries(server_logs, instance_logs):
...
 
         while instance_log_entry is not None:
-            log.append(instance_log_entry)
+            yield instance_log_entry
             instance_log_entry = next_instance_log_entry()
 
         while server_log_entry is not None:
-            log.append(server_log_entry)
+            yield server_log_entry
             server_log_entry = next_server_log_entry()
```
smarnach: Out of scope: Are we using Python 3.5?  If we do, the whole function above should be deleted, and we should use `heapa.merge()` instead.  (In Python 3.4, this function doesn't have a `key` parameter, but it might be worthwhile to rewrite the function anyway.)

instance/models/instance.py
```diff
@@ -207,52 +205,61 @@ def _sort_log_entries(server_logs, instance_logs):
...
+        returned.
         """
         # TODO: Filter out log entries for which the user doesn't have view rights
         server_log_entry_set = self._current_server.log_entry_set
         if level_list:
             server_log_entry_set = server_log_entry_set.filter(level__in=level_list)
-        server_log_entry_set = server_log_entry_set.order_by('pk').iterator()
+        server_log_entry_set = server_log_entry_set.order_by('pk')
+        if limit:
+            server_log_entry_set = reversed(server_log_entry_set.reverse()[:limit])
```
smarnach: What about `server_log_entry_set[-limit:]`?  Would that fetch all log entries from the database?

instance/models/instance.py
```diff
@@ -207,52 +205,61 @@ def _sort_log_entries(server_logs, instance_logs):
...
             instance_log_entry_set = instance_log_entry_set.filter(level__in=level_list)
-        instance_log_entry_set = instance_log_entry_set.order_by('pk').iterator()
+        instance_log_entry_set = instance_log_entry_set.order_by('pk')
+        if limit:
+            instance_log_entry_set = reversed(instance_log_entry_set.reverse()[:limit])
+        else:
+            instance_log_entry_set = instance_log_entry_set.iterator()
 
-        return Instance._sort_log_entries(server_log_entry_set, instance_log_entry_set)
+        entries = list(self._sort_log_entries(server_log_entry_set, instance_log_entry_set))
```
smarnach: If we use `itertools.islice()` here to limit the number of entries retrieved, couldn't we avoid having to limit it separately for the server logs and the instance logs, and then a third time below?

instance/static/js/src/instance.js
```diff
@@ -75,9 +75,18 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
             });
         };
 
-        $scope.select = function(selection_type, value) {
-            $scope.selected[selection_type] = value;
-            console.log('Selected ' + selection_type + ':', value);
+        $scope.select = function(instance) {
+            $scope.loading = true; // Display loading message
+            console.log('Selected instance', instance.id);
```
smarnach: Nit: I'd move this into the anonymous function below, because that's the function the clean-up is relating to.

instance/static/js/src/instance.js
```diff
@@ -75,9 +75,18 @@ app.controller("Index", ['$scope', 'Restangular', 'OpenCraftAPI', '$q',
             });
         };
 
-        $scope.select = function(selection_type, value) {
-            $scope.selected[selection_type] = value;
-            console.log('Selected ' + selection_type + ':', value);
+        $scope.select = function(instance) {
+            $scope.loading = true; // Display loading message
```
smarnach: Nit: I'd move this into the anonymous function below, because that's the function the clean-up is relating to.

smarnach:
@omarkhan Well done, looks good already!  I've added a few comments, and will continue testing this tomorrow.

instance/models/instance.py
omarkhan: I don't think we're using python 3.5, as pylint doesn't support it yet.
Thanks for the tip about heapq.merge, I didn't know that. Should I revert
this?

instance/models/instance.py
omarkhan: The problem is that we want the *last* X entries. We could do that with `itertools.islice` by iterating backwards and then reversing the order.

instance/models/instance.py
omarkhan: Django querysets do not support negative indexing.

instance/models/instance.py
omarkhan: Actually pylint does support python 3.5 now, so we could make the switch. See https://github.com/open-craft/opencraft/pull/6

omarkhan:
@smarnach I have added some manual test instructions. Let me know if you have any problems.

instance/models/instance.py
smarnach: For the time being, let's just leave the code the way it is.  While it's easy to write a helper function implements the `key` parameter around the old `heapq.merge()`, it's rather cumbersome (and slow) to write something that supports the `reversed` parameter.  For reversing, you really need to implement custom comparisons, and can't simply annotate the objects with their keys.

We can clean this up once we're using Python 3.5.

instance/models/instance.py
smarnach: Right, I've run into this before, but I didn't remember.

instance/models/instance.py
smarnach: @omarkhan I think this is good enough for now.  Once we use `heapq.merge()`, the whole code can be simplified, and we will clean it up then.

smarnach:
@omarkhan Tested manually, looks all good now.  :+1:

instance/models/instance.py
omarkhan: I have reverted this change for the time being.

omarkhan:
Thanks @smarnach, rebasing

instance/models/instance.py
smarnach: @omarkhan Argh, that's not really what I meant (though it's really what I wrote).  I actually meant: just leave the code the way you have it.

Anyway, doesn't really matter, either way is fine, and we'll make sure to simplify this once we are on Python 3.5.

omarkhan:
@antoviaque will this deploy automatically?

antoviaque:
@omarkhan Sadly no, at least not yet - definitely on the list of future improvements: https://opencraft.uservoice.com/forums/280754-opencraft/suggestions/9937191-self-deployment . You will have to do it yourself for now (which you can since updates aren't handled by ansible currently):

* `ssh ubuntu@console.opencraft.com`
* then access `screen -rd`
* pull the latest code from master (in a `sudo su www-data` shell, opencraft lives in `/var/www/opencraft`)
* then switch to the screen that runs `make run`, stop it, run the migrations and restart it.
* also restart the `make shell` to ensure it uses the latest code

Let me know if you have any question.

smarnach:
@omarkhan I have deployed this to production.  (It happened automatically when I deployed my changes.)  I recommend trying out http://console.opencraft.com/ with this PR installed; it's great!  Thanks for getting this done!

omarkhan:
Thanks Sven! Yeah it was almost unusable for me before

antoviaque:
@omarkhan Yes, thank you so much for that fix! Really great :D

PR #40, opened 2016-02-18
Set default log level to valid 'INFO' value

omarkhan:
Lowercase 'info' is not a valid choice, and will trigger a ValidationError:

```
ValidationError: {'level': ["Value 'info' is not a valid choice."]}
```

### DIFF
--- a/instance/tests/models/test_log_entry.py
+++ b/instance/tests/models/test_log_entry.py
@@ -25,6 +25,7 @@ Logger models & mixins - Tests
 from freezegun import freeze_time
 from mock import patch
 
+from instance.models.log_entry import GeneralLogEntry
 from instance.tests.base import TestCase
 from instance.tests.models.factories.instance import OpenEdXInstanceFactory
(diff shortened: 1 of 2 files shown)


antoviaque:
@omarkhan Thanks! Good catch. Could you add a unit test for this?

omarkhan:
@antoviaque unit test added.

instance/tests/models/test_log_entry.py
```diff
@@ -39,6 +40,13 @@ class LogEntryTestCase(TestCase):
     """
     Test cases for LoggerInstanceMixin
     """
+    def test_default_log_level(self):
+        """
+        Check that the default log level is INFO
+        """
+        log_entry = GeneralLogEntry(text='OHAI')
+        self.assertEqual(log_entry.level, 'INFO')
```
smarnach: You could also relax this a bit to just check that the default value is a valid choice (but I'm fine with it either way).

smarnach:
:+1: Looks good to me.

PR #42, opened 2016-03-01
Use utf8mb4 charset for mysql databases

omarkhan:
Supports all utf-8 code points, including 4-byte chars

### DIFF
--- a/instance/tests/models/test_instance.py
+++ b/instance/tests/models/test_instance.py
@@ -385,7 +385,8 @@ class MySQLInstanceTestCase(TestCase):
     """
     def check_mysql(self, instance):
         """
-        Check that the mysql databases and user have been created, then remove them
+        Check that the mysql databases and user have been created, and that we can write utf-8 data
+        to them. Then remove the databases and user.
         """
         self.assertIs(instance.mysql_provisioned, True)
         self.assertTrue(instance.mysql_user)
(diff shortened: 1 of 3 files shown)


instance/tests/models/test_instance.py
```diff
@@ -394,13 +395,24 @@ def check_mysql(self, instance):
...
                     password=instance.mysql_pass,
                     db_name=database,
                 )
-                tables = subprocess.call(mysql_cmd, shell=True)
-                self.assertEqual(tables, 0)
+                process = subprocess.Popen(mysql_cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
+                test_sql = ("CREATE TABLE test (value TEXT);"
+                            "INSERT INTO test (value) VALUES ('OHAI 😜');"
+                            "SELECT value FROM test;")
+                output, _ = process.communicate(input=test_sql.encode('utf-8'))
```
smarnach: Might be good to capture stderr as well, and add include it in the error message when the assert in the next line fails (the error message can be given in the third parameter of `assertEqual()`).

smarnach:
Code changes look good to me, but the tests are failing.  One of the integration tests failed as well, and the error might very well be related.  (Basically, the MySQL engine we use has a key size limit, and using a different character set might implicitly increase the key of some celery table.)

omarkhan:
@smarnach yes you are right. I have opened https://github.com/edx/configuration/pull/2823 which I am currently deploying to a sandbox. Let's see if it works.

omarkhan:
@smarnach looks like that sandbox deploy failed too. I can't seem to ssh in to debug it (permission denied). Shouldn't it add my github public key to authorized_keys?

omarkhan:
@smarnach never mind I sorted out my ssh issues.

it looks like the only way we are going to get this working is with mysql 5.7, which increases the default limit on prefix and index sizes. It is theoretically possible to do it with mysql 5.6, but it involves modifying a bunch of settings and specifying a row storage type of `DYNAMIC` or `COMPRESSED` with every `CREATE TABLE` command.

This was supposed to be a 1-line change - do we care enough about this to go to all this trouble? The `edxlocal` role uses `utf8` and not `utf8mb4` for databases, if that's good enough for upstream maybe it's good enough for us?

smarnach:
@omarkhan Thanks for the additional information.  We heard from other people running Open edX instances that the utf8 character encoding caused issues for the because someone entered a non-BMP character in some field that ended up in the database, resulting in errors etc.  We haven't experienced this issue ourselves so far, but thought it didn't hurt to be pre-emptive.  I also read an [article on the subject](https://mathiasbynens.be/notes/mysql-utf8mb4) which concluded that you should always use utf8mb4.

Anyway, given that this is more difficult than expected, I don't think it's justified to spend more time on this unless we have more than a bit of anecdotal evidence about how much of a problem this issue actually is.

@mtyaka @bradenmacdonald Opinions?

bradenmacdonald:
@smarnach If you don't believe there's any reasonably low-effort way to resolve the issues: Given that this hasn't caused any issues so far, and upstream isn't using this (yet), I'd be fine with holding off on this change until MySQL 5.7 is used in the edX configuration roles or the upstream code is modified to work around the key length problems.

bradenmacdonald:
@omarkhan @smarnach Do we need to consider reverting https://github.com/open-craft/documentation/pull/30/files ?

CC @mtyaka

mtyaka:
That's unfortunate :(

PR #41, opened 2016-02-26
Improve documentation

omarkhan:
This pull request updates the instance manager documentation:

- Explain what the instance manager does
- List the configuration options and explain what they mean
- Add example configuration
- Explain how to set up an ssh key pair for sandbox provisioning
- Explain how to configure openstack images and flavours
- Show how to manage sandboxes from the shell

PR #43, opened 2016-03-02
Fix shell instructions

omarkhan:
You have to use `OpenEdxInstance.objects.create`, you can't instatiate the instance first and save it later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,12 +411,17 @@ provision_instance(instance.pk)
 To reprovision an instance from the shell, simply run the `provision_instance`
 task again:
 
-    provision_instance(instance.pk)
+```python
+instance = OpenEdXInstance.objects.get(name__contains='...')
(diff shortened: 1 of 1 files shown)


omarkhan:
@pomegranited this is just a quick follow up to https://github.com/open-craft/opencraft/pull/41. Does it look ok?

pomegranited:
@omarkhan Looks good, just one question: When re-provisioning or deleting, how do you get the ``instance`` instance?

omarkhan:
@pomegranited you can use django's usual queryset methods, like this:

```python
instance = OpenEdXInstance.objects.get(name__contains='...')
```

or

```python
instance = OpenEdXInstance.objects.get(github_pr_url='...')
```

I will add this to the README

pomegranited:
Thank you, @omarkhan ! :+1:

PR #44, opened 2016-03-09
The nova client should retry in case of server errors

omarkhan:
The nova client currently only retries failed DNS lookups, socket connections and connection timeouts, as per the [requests defaults](http://docs.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter). This pull request extends the retry policy to include server errors (5XX responses).


Kelketek:
@omarkhan You have a quality failure, which is breaking the build. Approach makes sense to me, though. :+1: once the build is green.

PR #45, opened 2016-03-11
Terminate servers on unhandled exception during provisioning

omarkhan:
This ensures that we don't end up with zombie VMs when provisioning fails because of a bug in the Instance Manager.


omarkhan:
@Kelketek this is ready for review

Kelketek:
:+1:

PR #46, opened 2016-03-16
Display error when provisioning instance for deleted branch

omarkhan:
### Problem

If a PR's branch has been deleted from GitHub and the user clicks "Provision", the provisioning fails silently, without any relevant error appearing in the web console error log (also the status says "provisioning" even though it is not).

### Solution

- Return a client error response with a useful error message when the user attempts to provision an instance for a branch that does not exist
- Display that error message as a notification to the user

### DIFF
--- a/instance/api/instance.py
+++ b/instance/api/instance.py
@@ -27,6 +27,7 @@ from rest_framework.decorators import detail_route
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
+from instance import github
 from instance.models.instance import OpenEdXInstance
 from instance.serializers.instance import (OpenEdXInstanceListSerializer,
                                            OpenEdXInstanceDetailSerializer)
@@ -49,11 +50,17 @@ class OpenEdXInstanceViewSet(viewsets.ModelViewSet):
         instance = self.get_object()
         if instance.progress == instance.PROGRESS_RUNNING:
             return Response({'status': 'Instance is not ready for reprovisioning'},
-                            status=status.HTTP_403_FORBIDDEN)
+                            status=status.HTTP_400_BAD_REQUEST)
 
-        instance.set_to_branch_tip()
+        try:
+            instance.set_to_branch_tip()
+        except github.ObjectDoesNotExist:
+            return Response({
+                'status': ("Branch '{0}' not found."
+                           'Has it been deleted on GitHub?'.format(instance.branch_name))
+            }, status=status.HTTP_400_BAD_REQUEST)
+
         provision_instance(pk)
-
         return Response({'status': 'Instance provisioning started'})
 
     def get_serializer_class(self):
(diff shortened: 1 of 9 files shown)


instance/api/instance.py
```diff
@@ -51,9 +52,13 @@ def provision(self, request, pk=None):
             return Response({'status': 'Instance is not ready for reprovisioning'},
                             status=status.HTTP_403_FORBIDDEN)
 
-        instance.set_to_branch_tip()
-        provision_instance(pk)
+        try:
+            instance.set_to_branch_tip()
+        except github.ObjectDoesNotExist:
+            return Response({'status': "Ref '{0}' not found".format(instance.branch_name)},
+                            status=status.HTTP_403_FORBIDDEN)
```
omarkhan: Not sure that 403 is the best status code to return here. I picked it because it's the one used above.

omarkhan:
@smarnach this is ready for review

instance/api/instance.py
smarnach: The status code doesn't really matter too much, but a 403 seems weird.  I'd just go with a 400 for both this one and the one above.  Up to you whether you want to change it.

instance/api/instance.py
```diff
@@ -51,9 +52,13 @@ def provision(self, request, pk=None):
             return Response({'status': 'Instance is not ready for reprovisioning'},
                             status=status.HTTP_403_FORBIDDEN)
 
-        instance.set_to_branch_tip()
-        provision_instance(pk)
+        try:
+            instance.set_to_branch_tip()
+        except github.ObjectDoesNotExist:
+            return Response({'status': "Ref '{0}' not found".format(instance.branch_name)},
```
smarnach: I think we can be a bit more specific here.  While probably any git ref would work, for auto-created sandboxes this will always be a branch name.  I'd also include the hint that the branch has probably been deleted.

smarnach:
@omarkhan Code changes look good, and everything worked fine when testing locally.

The notifications you added are a bit ad-hoc.  We already have the "loading" indicator that pops up at the top of the screen, which looks completely different.  However, I don't think consistency in the UI is important as this point.  We will have to see how the instance manager evolves, and improve the UI at some point in the future.

I'm also not too fond of notifications that disappear after a few seconds, because I'll often miss them (e.g. because I switch tabs immediately after clicking "Provision").  I'd prefer a notification with an "x" button in the corner, but that's mostly personal taste.

:+1: after addressing the comments.

omarkhan:
@smarnach thanks for the review.

I agree that the visual styling is not very consistent, and that it doesn't matter very much at this point. It's mostly using the same css so it shouldn't be hard to improve in the future.

See edbdc5a - an error will be logged in the instance log if the branch has been deleted, so you can see what happened even if you miss the popup notification.

smarnach:
@omarkhan :+1:

PR #48, opened 2016-03-23
Clean temporary directories left from previous provisioning runs.

jbzdak:
# Reproduce steps

1. Install IM and setup environment settings
2. Go to `/localhost/admin` and create an instance. 
3. Wait until provisioning is started
4. Log in to shell account to you vagrant instance ``vagrant ssh`` and go to ``/tmp`` dir. Note that quite a lot of folders was created. 
5. Run ``ps -aef | grep python``, and kill the ansible process (killing step is not-neccessary, as venv's are not deleted at all, but it makes it faster)
6. Due to the fact provisioning is restarted you might need to kill it three times. 
7. Note that some of the folders in ``/tmp`` are not deleted. There should be as many folders left as many propositioning runs you started, these folders contain virtualenvs created during provisioning runs. 
8. (For bonus points) If you kill ansible in a "proper" moment it will left a tmp file with a role contents.

# Verification steps

1. Repeat the procedure and note that virtualenves are removes
2. (For bonus points) Note that ansible role files are not being left over. 
3. (For bonus points) Check also that it works for full provisionings (that is if you provision succesfully the sandbox all temporary files/folders are deleted). 
4. Start provisioning, note which tmp directory contains virtualenv, remove ``u-rwx`` permissions from this directory, and kill ansible process. Note that errors about impossibility of cleaning up are logged (to the GeneralLog in /admin)

### DIFF
--- a/instance/ansible.py
+++ b/instance/ansible.py
@@ -23,11 +23,14 @@ Ansible - Helper functions
 # Imports #####################################################################
 
 import os
+import shutil
 import subprocess
+import sys
+import traceback
 import yaml
 
 from contextlib import contextmanager
-from tempfile import mkdtemp, mkstemp
+from tempfile import mkdtemp, NamedTemporaryFile
 
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


jbzdak:
@omarkhan This is ready for review.

instance/tests/test_ansible.py
```diff
@@ -96,41 +98,105 @@ class AnsibleTestCase(TestCase):
...
+        is deleted after we exit the context manager.
+        """
+        saved_temp_dir = None
+        try:
+            with ansible.create_temp_dir() as temp_dir:
+                saved_temp_dir = temp_dir
+                self.assertTrue(os.path.exists(temp_dir))
+                self.assertTrue(os.path.isdir(temp_dir))
+                raise KeyboardInterrupt()
+        except KeyboardInterrupt:
```
omarkhan: Would it be worth checking that the exception propagates outside the `with` block here? If someone makes a change in the future that catches the exception and doesn't re-raise it, the assertion here will be skipped and the test will pass even if the temp dir still exists.

instance/tests/test_ansible.py
```diff
@@ -96,41 +98,105 @@ class AnsibleTestCase(TestCase):
...
+            self.assertFalse(os.path.exists(saved_temp_dir))
+
+    def test_string_to_file_path_ok(self):
+        """
+        Check if string_to_file_path behaves correctly when no exception is
+        raised from inside it.
+
+        By behave correctly we mean: creates a file with proper contents,
+        and this file is deleted when we exit the context manager.
```
omarkhan: This test doesn't check that the file is deleted. Also `string_to_file_path` isn't a context manager anymore, right?

instance/tests/test_ansible.py
```diff
@@ -96,41 +98,105 @@ class AnsibleTestCase(TestCase):
...
+                playbook_name='playbook_name',
+                remote_username='root',
+                venv_path='/tmp/tempdir/venv'
             )
-            self.assertEqual(
-                mock_popen.mock_calls,
-                [call(run_playbook_cmd, bufsize=1, stdout=-1, stderr=-1, cwd='/play/book', shell=True)]
+
+            mock_popen.assert_called_once_with(
+                "ANSIBLE CMD", bufsize=1, stdout=-1, stderr=-1, cwd='/play/book', shell=True, env=mock.ANY
```
omarkhan: Should we check that the `TMPDIR` environment variable is set correctly?

instance/ansible.py
```diff
@@ -67,59 +68,96 @@ def dict_merge(dict1, dict2):
...
-    Runs ansible-playbook in a dedicated venv
+    A context manager that creates a temporary directory, returns it. Directory is deleted upon context manager exit.
+    """
+    try:
+        temp_dir = mkdtemp()
+        yield temp_dir
+    finally:
+        # This is really a "best effort" solution, if something will block the deletion (like file being locked somehow)
+        # this should just move on rather than raise.
+        shutil.rmtree(temp_dir, ignore_errors=True)
```
omarkhan: If we are going to ignore errors here, should we at least log them?

omarkhan:
@jbzdak I had a look through the diff, looks great. See my comments on the diff. I'm going to test this manually now...

instance/tests/test_ansible.py
jbzdak: Right now ``string_to_file_path`` doesn't delete files, there is a previous commit where it did, but then I had `with`` with four objects created. 

Right now a temporary dir is created, and files created by ``test_string_to_file_path_ok`` are created inside it. That directory will be deleted at the end of run.

omarkhan: That's fine - it might just be worth updating the docstring here to match what is being tested.

jbzdak: I totally missed the doctring needs updating, thanks!

instance/tests/test_ansible.py
jbzdak: Done

instance/tests/test_ansible.py
jbzdak: Done

instance/ansible.py
jbzdak: This is now logged, however messages do not appear in instance logs, to make them appear in instance logs, I'd have to somehow pass `{"obj": instance.pk}` as a logging context, and instance is nowhere near this method. This can be done, for example by storing instance pk in a threadlocal varibable, or passing it though couple of levels of call stack, but I think it is not worthwile.

jbzdak:
@omarkhan Addressed all PR comments, thanks, they ware usefull.

omarkhan:
@jbzdak I have tested this manually, seems to work fine. :+1:

instance/ansible.py
omarkhan: You could *not* ignore errors here, and catch and log them in the Instance method instead.

jbzdak: Yes, but errors raised from ``shutil.rmtree`` would stop the cleanup action --- and possibly leave more garbage in the ``/tmp``, which for me would be a step back. ``rmtree`` bails out after first error, but if it ignores the error it could continue. 

 So this really a decision: 'make maters slightly worse --- by leaving more garbage' vs 'make error less obvious'

Possibly we are discussing a very edge case, we could leave it as it is, or just let ``rmtree`` raise the exception, which would remove some error handling code.

omarkhan: I think the most important thing here is for failures to be obvious. So I think the best solution would be to just raise an exception if anything failed, and handle that exception somewhere in the `Instance` model code. But I don't think that this is likely to cause too many problems, so if you want to do it another way I'm fine with that.

jbzdak: In the end I couldn't find a sensible way to eventual handle errors raised by rmtree. For now I don't believe that this error will occur on production, so I decided to just raise the exception, which will log the error, and terminate the server set. 

Ideal solution would be to implement a sort of logging context that would transparently handle passing server context. And just log these errors, but leave instance state unchanged (if it provisioned don't tear it down). 

For now I'd suggest just raising this exception. This also the same way rmtree errors are handled when [deleting github repository.](https://github.com/open-craft/opencraft/blob/d56624843485f588ba76759fb09f8fe50952684d/instance/repo.py#L41)

omarkhan: +1

omarkhan:
:+1:

PR #49, opened 2016-03-29
Logrotate - Set logging to use `RotatingFileHandler`

antoviaque:
Switch the logging file handler in the Django configuration to `RotatingFileHandler`, to avoid letting log files grow too large. 

Two new configuration variables:

* `LOGGING_ROTATE_MAX_KBYTES`: The max size of each log file (in KB)
* `LOGGING_ROTATE_MAX_FILES`: The max number of log files to keep

How to test:

1. Set `LOGGING_ROTATE_MAX_KBYTES=1` in your local OpenCraft IM `.env` file - this is to trigger log rotation without needing to generate huge logs
2. Generate some logs, for example by provisioning a new instance
3. Check `log/main.*` - you should see lots of small log files, a few KB each (it seem to still log a full message even if it exceeds slightly the max size, which is a desired behavior).

Note: No unit test since this is only a configuration change, but if you see a meaningful way to test this I'm happy to add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ flavor:
 * `REDIS_URL` (default: `redis://localhost:6379/`)
 * `HUEY_ALWAYS_EAGER` Set to True to run huey tasks synchronously, in the web
   process. Use in development only (default: False)
+* `LOGGING_ROTATE_MAX_KBYTES`: The max size of each log file (in KB, default: 10MB)
+* `LOGGING_ROTATE_MAX_FILES`: The max number of log files to keep (default: 60)
 
 ### OpenStack credentials
 
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -305,6 +305,8 @@ ADMINS = env.json('ADMINS', default=set())
 
 BASE_HANDLERS = env.json('BASE_HANDLERS', default=["file", "console"])
 HANDLERS = BASE_HANDLERS + ['db']
+LOGGING_ROTATE_MAX_KBYTES = env.json('LOGGING_ROTATE_MAX_KBYTES', default=10 * 1024)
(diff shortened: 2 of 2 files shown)


pomegranited:
@antoviaque I was able to watch the logs rotate, and verified that the `LOGGING_ROTATE_MAX_KBYTES` and `LOGGING_ROTATE_MAX_FILES` settings are respected.  :+1:

Just wondering, did you want to compress them too? Could be done with `/etc/logrotate.d/`, since it sounds like RotatingFileHandler won't do it in python3.

Testing steps taken:

* Set up a vagrant dev instance and superuser account according to [opencraft/README](https://github.com/open-craft/opencraft).
* With default `LOGGING_ROTATE` settings:
    * `make rundev`
    * In another shell: `watch -d ls -l log/`
    * Refresh http://localhost:5000/ several times
    * Watched log file grow.. but didn't refresh enough to see it tip over 10Mb and rotate.
* Added to `.env`: `LOGGING_ROTATE_MAX_KBYTES=1`, and restarted dev server.
    * Watched initial log file rotate immediately.
    * Subsequent log files maintain a max size of ~1Kb, with rotated file count of 60.
* Added to `.env`: `LOGGING_ROTATE_MAX_FILES=5`, and restarted dev server.
    * The previously generated log files main.log.6-60 remained untouched
    * Only main.log.1-5 were updated.
* Changed `LOGGING_ROTATE_MAX_KBYTES=2`, and restarted dev server.
   * Watched main.log.1-10 maintain max size of ~2Kb.

pomegranited:
@antoviaque Sorry!  hit the wrong button :(

antoviaque:
@pomegranited Thanks!

For the compression, that's probably not necessary at this stage - with the default settings, which is what I think we'd start with on production, that gives 600MB, which won't really impact disk usage much.

PR #50, opened 2016-04-06
Alert admins by email if provisioning fails

e-kolpakov:
**Description:** Sends email notifications to server admins if provisioning fails

**Testing instructions:**
1. Provide email settings in .env file:
1.1. `ADMINS='[["you", "you@opencraft.com"]]'`
1.2. `SERVER_EMAIL = 'you@opencraft.com'`  # SMTP Server refuses connections from unknown hosts
1.3. EMAIL_HOST, EMAIL_HOST_USER, EMAIL_HOST_PASSWORD - from opencraft/secrets repository
2. Patch `instance.provision` so that it is guaranteed to fail, or otherwise ensure provisioning will fail.
3. Provision an instance.

You should receive an email with details on what went wrong:
image

Note there are two flavors of "failure":
1. Unhandled exception
2. `deploy` returned non-zero exit code.

Emails are slightly different, example screenshot is for `deploy` returned non-zero exit code failure.

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -23,33 +23,25 @@ Instance app models - Instance
 # Imports #####################################################################
 
 import logging
-import os
 import string
 
-from functools import partial
-
 from django.conf import settings
-from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models
 from django.db.backends.utils import truncate_name
 from django.template import loader
 from django.utils import timezone
 from django.utils.crypto import get_random_string
 from django_extensions.db.models import TimeStampedModel
+from functools import partial
 
-import MySQLdb as mysql
-import pymongo
-
-from instance import ansible, github
 from instance.gandi import GandiAPI
-from instance.github import fork_name2tuple, get_username_list_from_team
-from instance.logging import log_exception
 from instance.logger_adapter import InstanceLoggerAdapter
-from instance.repo import open_repository
-from instance.utils import poll_streams
-
+from instance.logging import log_exception
+from instance.models.mixins.ansible import AnsibleInstanceMixin
+from instance.models.mixins.database import MySQLInstanceMixin, MongoDBInstanceMixin
+from instance.models.mixins.utilities import EmailInstanceMixin
+from instance.models.mixins.version_control import GitHubInstanceMixin
 from instance.models.utils import ValidateModelMixin
-
 
 # Constants ###################################################################
 
@@ -74,13 +66,8 @@ class InconsistentInstanceState(Exception):
     """
     pass
 
-
-# Validators ##################################################################
-
-sha1_validator = RegexValidator(regex='^[0-9a-f]{40}$', message='Full SHA1 hash required')
-
-
 # Models ######################################################################
+
 
 class Instance(ValidateModelMixin, TimeStampedModel):
     """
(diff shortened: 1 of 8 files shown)


instance/models/mixins/utilities.py
```diff
@@ -0,0 +1,82 @@
...
+
+class EmailInstanceMixin(object):
+    """
+    An instance class that can send emails
+    """
+    class EmailSubject(object):
+        """
+        Class holding email subject constants
+        """
+        PROVISION_FAILED = u"Instance {instance_name} ({instance_url}) failed to provision"
```
omarkhan: Why does this need to be in a nested class?

instance/models/instance.py
```diff
@@ -909,6 +465,7 @@ def provision(self):
             log, exit_code = self.deploy()
             if exit_code != 0:
                 server.update_status(provisioning=True, failed=True)
+                self.provision_failed_email(self.ProvisionMessages.PROVISION_ERROR)
```
omarkhan: It would be nice to include the ansible log in the email (at least the last part with the failure), or a link to the log in the instance manager console.

omarkhan:
Tested, works great. Very nice! LGTM apart from my comments above

instance/models/mixins/utilities.py
e-kolpakov: @omarkhan I does not *need* to be in a nested class of course, but it *can* be there :) And the benefit of having a nested class for it is that it acts much like a namespace, so

* All email subject constans are co-located
* They are isolated/separated from other constants
* Don't neded to use some prefix/suffix like `PROVISION_FAILED _EMAIL_SUBJECT`

instance/models/instance.py
e-kolpakov: @omarkhan Good catch - I planned to do so initially, but that completely forgot about it. To save some time, my approach is to add log as an attached file (including it in the body will further bloat the message) - do you have any objections/suggestions/comments to doing so?

instance/models/mixins/utilities.py
omarkhan: Understood, but it adds a little complexity as well:

- Harder to override in subclasses
- Using a class as a namespace feels a bit odd

I feel that simple prefixed constants would make more sense here. To quote the zen of python, "flat is better than nested". But then again the zen of python also says namespaces are great, so what do I know :smile: 

I don't feel too strongly about this, it's your call.

instance/models/instance.py
omarkhan: No objections, that sounds fine.

instance/models/mixins/utilities.py
e-kolpakov: @omarkhan So far we don't have a use case for overriding them (and it would be relatively straightforward to modify the code to ease overriding), so according to YAGNI principle, I'll keep it as is :)

e-kolpakov:
@omarkhan Could you please take another look and approve/provide more feedback?

instance/tests/models/test_instance.py
```diff
@@ -503,12 +504,130 @@ def test_provision_mongo_again(self):
         self.check_mongo(instance)
 
 
+class EmailMixinInstanceTestCases(TestCase):
```
omarkhan: Nit: this is singular for all the other test cases (`TestCases` -> `TestCase`)

instance/tests/models/test_instance.py
```diff
@@ -503,12 +504,130 @@ def test_provision_mongo_again(self):
...
+    @patch("instance.models.mixins.utilities.EmailMultiAlternatives")
+    @override_settings(ADMINS=(("admin1", "admin1@localhost"),))
+    def test_provision_failed_email(self, mock_email_class):
+        """
+        Tests that provision_failed sends email when called from normal program flow
+        """
+        instance = OpenEdXInstanceFactory(name='test', sub_domain='test')
+        reason = "something went wrong"
+        mock_email_instance = Mock()
+        mock_email_class.return_value = mock_email_instance
```
omarkhan: I think these tests could be simpler and easier to read if you use [the dummy outbox that django sets up](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#email-services). Not a big deal, just a thought

omarkhan:
@e-kolpakov I'm not seeing the log attached to the email. I think you forgot to pass it to the `provision_failed_email` call. Apart from that, :+1:

instance/models/instance.py
omarkhan: You still need to pass the log here, right?

e-kolpakov:
@omarkhan one last review, please :)

omarkhan:
Looks good. Tested and working :+1:

PR #51, opened 2016-04-09
State machine (OC-906)

bradenmacdonald:
This replaces the `status` and `progress` attributes of the `Server` class with a finite state machine implementation.

The immediate benefits are:
* Strong guarantee that the `Server` instance is always in one of the specified states.
* Transitions between states are clearly defined. One cannot use public methods to change from "New" to "Provisioning" for example.
* Removed states that weren't being used

The future benefits are much greater, but this is just "a step in the right direction". See below for next steps to help give context.

### Features:

* Any "resource" (such as a server) can have finite state machines attached to it
* Each state machine is always in one of a pre-defined number of states
* The value of the state can optionally be saved to the database if the resource is a django model
* The state can only be changed through the use of a *transition*; the allowed transitions ('allow changing from Loading to Ready') are clearly defined, and an attempt to make an invalid transition always raises an exception
* A resource can annotate its methods with annotations like `@status.only_for(Status.Booted)` - this will automatically raise an exception if the method is called when not appropriate, and also gives access to the handy `if method.is_available()` property which can be used to determine if you're allowed to call a method without actually calling it, in a fully generic fashion (you don't need to know what states support the method and what don't). So you can do things like `while not server.run_ssh_command.is_available(): wait()` which will wait until a server is ready to run SSH commands, without you needing to know what state(s) the server is in.
* When reading some state of a resource, you get an *instance* of the state, which has a pointer to the resource instance. This means that you can simplify resources by implementing methods on the state class rather than the resource. This is completely optional though.
    Example:

    ```python
    # Instead of 
    class Resource:
        def update(self):
            if self.status == New:
                ...
            elif self.status == Ready:
                ...
            else:
                raise Exception()

    # You can just use:
    class Resource:
        def update(self):
            self.status.update()
    ```
    When the behavior of an object varies greatly depending on its state, or you want to do cool things with subclassing, this type of abstraction is very powerful. For example, say you have many types of server: OpenStackServer, AWSServer, LinodeServer, and each provider offers different error states such as "network_offline", "alien_attack", etc. If your code has a `server` instance, you could just check the health of the server with `server.state.is_healthy()` rather than `server.state in (Server.State.Booting, Server.State.Running, OpenStackServer.State.RestoringImage, LinodeServer.State.Cloning)`

### Suggested Next Steps

* Change the server states to Unknown, Pending, Building, Booting, Ready, BuildFailed, and Terminated. (Unknown for when the OpenStack API is not responsive, BuildFailed for when OpenStack fails to create the server, e.g. due to a quota issue).
* Rename OpenEdXInstance to SingleVMOpenEdXInstance. Give it a set of states: New, WaitingForServer, ConfiguringServer, Running, ConfigurationFailed, Error.
* Create a new class OpenEdXSandbox, which manages of a collection of SingleVMOpenEdXInstance. When a new instance successfully provisions, the old ones get deleted. When provisioning fails, any  old sandbox that still runs will be kept.
* Give the Server states these properties: is_steady_state (the status is not expected to change, so don't wait for it to do so - e.g. Ready, BuildFailed, Terminated), and accepts_ssh_commands. Then refactor the code so that outside of the Server class we don't need to know what set of server states there are; we can just display `server.status.name` and `server.status.description` in the UI, and we can do things like `sleep() until server.status.accepts_ssh_commands`. This will allow us to do things in the future, like support more of the OpenStack API by adding more possible server states, without being coupled to other parts of the codebase.

### DIFF
--- a/.prospector/opencraft.yml
+++ b/.prospector/opencraft.yml
@@ -31,7 +31,7 @@ profile-validator:
   options: {}
   run: null
 pyflakes:
-  disable: [F401, F403]
+  disable: [F401, F403, F841]
   enable: []
   options: {}
   run: null
--- a/.prospector/opencraft.yml
+++ b/.prospector/opencraft.yml
@@ -31,7 +31,10 @@ profile-validator:
   options: {}
   run: null
 pyflakes:
-  disable: [F401, F403, F841]
+  disable:
+    - F401
+    - F403
+    - F841  # "local variable is assigned to but never used" - pylint has an equivalent check, and the pylint version correctly ignores names like 'dummy'
   enable: []
   options: {}
   run: null
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -122,8 +122,8 @@ class Progress(ResourceState.Enum):
         is_final = True
 
 
-
 # Models ######################################################################
+
 
 class ServerQuerySet(models.QuerySet):
     """
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -39,7 +39,7 @@ from instance.logger_adapter import ServerLoggerAdapter
 from instance.utils import is_port_open, to_json
 
 from instance.models.instance import OpenEdXInstance
-from instance.models.utils import ValidateModelMixin, ResourceState, ResourceStateDescriptor
+from instance.models.utils import ValidateModelMixin, ResourceState, ModelResourceStateDescriptor
 
 
 # Logging #####################################################################
@@ -151,7 +151,9 @@ class Server(ValidateModelMixin, TimeStampedModel):
     A single server VM
     """
     Status = Status
-    status = ResourceStateDescriptor(state_classes=Status.states, default_state=Status.New, model_field_name='_status')
+    status = ModelResourceStateDescriptor(
+        state_classes=Status.states, default_state=Status.New, model_field_name='_status'
+    )
     _status = models.CharField(
         max_length=20,
         default=status.default_state_class.state_id,
@@ -173,7 +175,7 @@ class Server(ValidateModelMixin, TimeStampedModel):
     _status_to_terminated = status.transition(to_state=Status.Terminated)
 
     Progress = Progress
-    progress = ResourceStateDescriptor(
+    progress = ModelResourceStateDescriptor(
         state_classes=Progress.states,
         default_state=Progress.Running,
         model_field_name='_progress',
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -39,7 +39,7 @@ from instance.logger_adapter import ServerLoggerAdapter
 from instance.utils import is_port_open, to_json
 
 from instance.models.instance import OpenEdXInstance
-from instance.models.utils import ValidateModelMixin
+from instance.models.utils import ValidateModelMixin, ResourceState, ResourceStateDescriptor
 
 
 # Logging #####################################################################
@@ -47,13 +47,80 @@ from instance.models.utils import ValidateModelMixin
 logger = logging.getLogger(__name__)
 
 
-# Exceptions ##################################################################
+# States ######################################################################
 
-class ServerNotReady(Exception):
+
+class ServerState(ResourceState):
     """
-    Raised when an action is attempted in a status that doesn't allow it
+    A [finite state machine] state describing a virtual machine.
     """
-    pass
+
+
+class Status(ResourceState.Enum):
+    """
+    The states that a server can be in.
+
+    TODO:
+    * Reduce these to Unknown, Pending, Building, Booting, Ready, BuildFailed, and Terminated.
+    """
+    class New(ServerState):
+        """ Not yet loaded """
+        state_id = 'new'
+
+    class Started(ServerState):
+        """ Running but not active yet """
+        state_id = 'started'
+
+    class Active(ServerState):
+        """ Running but not booted yet """
+        state_id = 'active'
+
+    class Booted(ServerState):
+        """ Booted but not ready to be added to the application """
+        state_id = 'booted'
+
+    class Provisioning(ServerState):
+        """ Provisioning is in progress """
+        state_id = 'provisioning'
+
+    class Rebooting(ServerState):
+        """ Reboot in progress, to apply changes from provisioning """
+        state_id = 'rebooting'
+
+    class Ready(ServerState):
+        """ Booted and ready to add to the application """
+        state_id = 'ready'
+
+    #class Stopped(ServerState):
+    #    """ Stopped temporarily """
+    #    state_id = 'stopped'
+
+    class Terminated(ServerState):
+        """ Stopped forever """
+        state_id = 'terminated'
+
+
+class Progress(ResourceState.Enum):
+    """
+    The progress states that a server can be in.
+
+    TODO: Refactor and perhaps remove this second state.
+    """
+    class Running(ResourceState):
+        """ Running """
+        state_id = 'running'
+        is_final = False
+
+    class Success(ResourceState):
+        """ Success """
+        state_id = 'success'
+        is_final = True
+
+    class Failed(ResourceState):
+        """ Failed """
+        state_id = 'failed'
+        is_final = True
+
 
 
 # Models ######################################################################
@@ -67,7 +134,7 @@ class ServerQuerySet(models.QuerySet):
         """
         Terminate the servers from the queryset
         """
-        qs = self.filter(~Q(status=Server.TERMINATED), *args, **kwargs)
+        qs = self.filter(~Q(_status=Status.Terminated.state_id), *args, **kwargs)
         for server in qs:
             server.terminate()
         return qs
@@ -76,54 +143,52 @@ class ServerQuerySet(models.QuerySet):
         """
         Filter out terminated servers from the queryset
         """
-        return self.filter(~Q(status=Server.TERMINATED))
+        return self.filter(~Q(_status=Status.Terminated.state_id))
 
 
 class Server(ValidateModelMixin, TimeStampedModel):
     """
     A single server VM
     """
-    NEW = 'new'
-    STARTED = 'started'
-    ACTIVE = 'active'
-    BOOTED = 'booted'
-    PROVISIONING = 'provisioning'
-    REBOOTING = 'rebooting'
-    READY = 'ready'
-    LIVE = 'live'
-    STOPPING = 'stopping'
-    STOPPED = 'stopped'
-    TERMINATING = 'terminating'
-    TERMINATED = 'terminated'
-
-    STATUS_CHOICES = (
-        (NEW, 'New - Not yet loaded'),
-        (STARTED, 'Started - Running but not active yet'),
-        (ACTIVE, 'Active - Running but not booted yet'),
-        (BOOTED, 'Booted - Booted but not ready to be added to the application'),
-        (PROVISIONING, 'Provisioning - Provisioning is in progress'),
-        (REBOOTING, 'Rebooting - Reboot in progress, to apply changes from provisioning'),
-        (READY, 'Ready - Rebooted and ready to add to the application'),
-        (LIVE, 'Live - Is actively used in the application and/or accessed by users'),
-        (STOPPING, 'Stopping - Stopping temporarily'),
-        (STOPPED, 'Stopped - Stopped temporarily'),
-        (TERMINATING, 'Terminating - Stopping forever'),
-        (TERMINATED, 'Terminated - Stopped forever'),
+    Status = Status
+    status = ResourceStateDescriptor(state_classes=Status.states, default_state=Status.New, model_field_name='_status')
+    _status = models.CharField(
+        max_length=20,
+        default=status.default_state_class.state_id,
+        choices=status.model_field_choices,
+        db_index=True,
+        db_column='status',
     )
-
-    PROGRESS_RUNNING = 'running'
-    PROGRESS_SUCCESS = 'success'
-    PROGRESS_FAILED = 'failed'
-
-    PROGRESS_CHOICES = (
-        (PROGRESS_RUNNING, 'Running'),
-        (PROGRESS_SUCCESS, 'Success'),
-        (PROGRESS_FAILED, 'Failed'),
+    # State transitions:
+    _status_to_started = status.transition(from_states=Status.New, to_state=Status.Started)
+    _status_to_active = status.transition(from_states=Status.Started, to_state=Status.Active)
+    _status_to_booted = status.transition(from_states=Status.Active, to_state=Status.Booted)
+    _status_to_provisioning = status.transition(from_states=Status.Booted, to_state=Status.Provisioning)
+    _status_to_rebooting = status.transition(
+        from_states=(Status.Active, Status.Booted, Status.Ready, Status.Provisioning), to_state=Status.Rebooting
     )
+    _status_to_ready = status.transition(
+        from_states=(Status.Active, Status.Booted, Status.Provisioning, Status.Rebooting), to_state=Status.Ready
+    )
+    _status_to_terminated = status.transition(to_state=Status.Terminated)
+
+    Progress = Progress
+    progress = ResourceStateDescriptor(
+        state_classes=Progress.states,
+        default_state=Progress.Running,
+        model_field_name='_progress',
+    )
+    _progress = models.CharField(
+        max_length=7,
+        default=progress.default_state_class.state_id,
+        choices=progress.model_field_choices,
+        db_column='progress',
+    )
+    _progress_success = progress.transition(from_states=(Progress.Running, Progress.Success), to_state=Progress.Success)
+    _progress_failed = progress.transition(from_states=Progress.Running, to_state=Progress.Failed)
+    _progress_reset = progress.transition(to_state=Progress.Running)
 
     instance = models.ForeignKey(OpenEdXInstance, related_name='server_set')
-    status = models.CharField(max_length=20, default=NEW, choices=STATUS_CHOICES, db_index=True)
-    progress = models.CharField(max_length=7, default=PROGRESS_RUNNING, choices=PROGRESS_CHOICES)
 
     objects = ServerQuerySet().as_manager()
 
@@ -147,29 +212,43 @@ class Server(ValidateModelMixin, TimeStampedModel):
             'server_id': self.pk,
         }
 
-    def _set_status(self, status, progress):
+    def _transition(self, state_transition, progress=Progress.Running):
         """
-        Update the current status variable, to be called when a status change is detected
-        """
-        if status not in (s[0] for s in self.STATUS_CHOICES):
-            raise ValueError(status)
+        Helper method to update the state and the progress.
 
-        self.status = status
-        self.progress = progress
+        Mostly exists to ensure the 'progress' is in sync with 'state' and the django DB.
+        """
+        state_transition()
+        self._set_progress(progress, expected_status=state_transition.to_state)
+
+    def _set_progress(self, progress, expected_status=None):
+        """
+        Helper method to update the progress.
+
+        Mostly exists to ensure the 'progress' is in sync with the django DB.
+        """
+        if expected_status:
+            assert isinstance(self.status, expected_status)
+        if progress is Progress.Running:
+            self._progress_reset()
+        elif progress is Progress.Success:
+            self._progress_success()
+        else:
+            assert progress is Progress.Failed
+            self._progress_failed()
         self.logger.info('Changed status: %s (%s)', self.status, self.progress)
+        # The '_progress' field needs to be saved:
         self.save()
-        return self.status
 
-    def sleep_until_status(self, target_status):
+    def sleep_until_status(self, *target_status_list):
         """
-        Sleep in a loop until the server reaches one of the specified status
+        Sleep in a loop until the server reaches one of the specified statuses
         """
-        target_status_list = [target_status] if isinstance(target_status, str) else target_status
         self.logger.info('Waiting to reach status %s...', target_status_list)
 
         while True:
             self.update_status()
-            if self.status in target_status and self.progress != self.PROGRESS_RUNNING:
+            if self.progress.is_final and isinstance(self.status, target_status_list):
                 break
             time.sleep(1)
         return self.status
@@ -185,7 +264,7 @@ class Server(ValidateModelMixin, TimeStampedModel):
             'server_pk': self.pk,
         })
 
-    def update_status(self, provisioning=False, rebooting=False, failed=None):
+    def update_status(self):
         """
         Check the current status and update it if it has changed
         """
@@ -214,7 +293,7 @@ class OpenStackServer(Server):
         OpenStack nova server API endpoint
         """
         if not self.openstack_id:
-            assert self.status == self.NEW
+            assert self.status == Status.New
             self.start()
         return self.nova.servers.get(self.openstack_id)
 
@@ -232,7 +311,7 @@ class OpenStackServer(Server):
 
         return public_addr['addr']
 
-    def update_status(self, provisioning=False, rebooting=False, failed=None):
+    def update_status(self):
         """
         Refresh the status by querying the openstack server via nova
         """
@@ -240,36 +319,43 @@ class OpenStackServer(Server):
         os_server = self.os_server
         self.logger.debug('Updating status from nova (currently %s):\n%s', self.status, to_json(os_server))
 
-        if self.status == self.STARTED:
+        if self.status == Status.Started:
             self.logger.debug('OpenStack: loaded="%s" status="%s"', os_server._loaded, os_server.status)
             if os_server._loaded and os_server.status == 'ACTIVE':
-                self._set_status(self.ACTIVE, self.PROGRESS_SUCCESS)
+                self._transition(self._status_to_active, Progress.Success)
             else:
-                self._set_status(self.ACTIVE, self.PROGRESS_RUNNING)
+                self._transition(self._status_to_active, Progress.Running)
 
-        elif self.status == self.ACTIVE and self.public_ip and is_port_open(self.public_ip, 22):
-            self._set_status(self.BOOTED, self.PROGRESS_RUNNING)
+        elif self.status == Status.Active and self.public_ip and is_port_open(self.public_ip, 22):
+            self._transition(self._status_to_booted, Progress.Success)
 
-        elif self.status == self.BOOTED:
-            if provisioning:
-                self._set_status(self.PROVISIONING, self.PROGRESS_RUNNING)
-            else:
-                self._set_status(self.BOOTED, self.PROGRESS_SUCCESS)
-
-        elif self.status == self.PROVISIONING and failed is not None:
-            if failed:
-                self._set_status(self.PROVISIONING, self.PROGRESS_FAILED)
-            else:
-                self._set_status(self.PROVISIONING, self.PROGRESS_SUCCESS)
-
-        elif self.status in (self.PROVISIONING, self.READY) and rebooting:
-            self._set_status(self.REBOOTING, self.PROGRESS_RUNNING)
-
-        elif self.status == self.REBOOTING and not rebooting and is_port_open(self.public_ip, 22):
-            self._set_status(self.READY, self.PROGRESS_SUCCESS)
+        elif self.status == Status.Rebooting and is_port_open(self.public_ip, 22):
+            self._transition(self._status_to_ready, Progress.Success)
 
         return self.status
 
+    @Server.status.only_for(Status.Booted)
+    def mark_as_provisioning(self):
+        """
+        Indicate that this server is being provisioned.
+
+        TODO: Remove this. A 'server' resource shouldn't know or care is ansible is running or not.
+        """
+        self._transition(self._status_to_provisioning, Progress.Running)
+
+    @Server.status.only_for(Status.Provisioning)
+    def mark_provisioning_finished(self, success):
+        """
+        Indicate that this server is done provisioning, either due to success or failure.
+        """
+        if success:
+            # Status does not switch to Ready at this point because the instance reboots the
+            # server before declaring it ready.
+            self._set_progress(Progress.Success, expected_status=Status.Provisioning)
+        else:
+            self._set_progress(Progress.Failed, expected_status=Status.Provisioning)
+
+    @Server.status.only_for(Status.New)
     def start(self):
         """
         Get a server instance started and an openstack_id assigned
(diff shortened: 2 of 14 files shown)


.prospector/opencraft.yml
```diff
@@ -31,7 +31,7 @@ profile-validator:
   options: {}
   run: null
 pyflakes:
-  disable: [F401, F403]
+  disable: [F401, F403, F841]
```
bradenmacdonald: This disables "pyflakes: F841 / local variable 'dummy' is assigned to but never used" since pylint has an equivalent check, and the pylint version supports inline disabling.

smarnach: Maybe add this as a comment to the YAML file?

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+    @ClassProperty
+    @classmethod
+    def description(cls):
+        """
+        description: Human-readable explanation of this state (1-2 sentences)
+
+        Defaults to the state class's docstring, but you can override it.
+        Doesn't need to be a property; override in the subclass using just 'description = "..."'
+        """
+        return cls.__doc__.strip()
```
smarnach: I suggest using [`textwrap.dedent()`](https://docs.python.org/3.4/library/textwrap.html#textwrap.dedent) instead of `str.strip()` here, since it results in a nicer formatted output.

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+
+    # Constants:
+    PROGRESS_UNKNOWN = -1
+
+    def __str__(self):
+        return "{} [{}]".format(self.name, self.state_id)
+
+    def __eq__(self, obj):
```
smarnach: It's generally recommended to implement `__hash__()` as well when implementing `__eq__()` (since having objects that compare equal return different hashes breaks assumptions).  Simply return the hash of `type(self)` should do the trick.

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+        """
+        Syntactic sugar to make comparing states easier.
+
+        obj should be a ResourceState class or instance.
+        """
+        try:
+            return issubclass(self.__class__, obj)
+        except TypeError:  # The provided argument is not a class
+            return type(obj) is type(self)
```
smarnach: This has slightly different semantics than the `issubclass()` check above.  Why not use `isinstance()` here, which would give the same semantics?

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+
+class ResourceState:
+    """
+    A [finite state machine] state class representing the overall state of a resource,
+    such as a server, a database, a DNS entry, etc.
+    """
+    # state_id: A string uniquely identifying this state
+    state_id = None
+
+    @ClassProperty
```
smarnach: People generally use meta-classes to support properties on classes, but I actually prefer this decorator for this use case.

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+
+        Then you can use MyStates.states to get a list of the state classes.
+        """
+        @ClassProperty
+        @classmethod
+        def states(cls):
+            """
+            Get an iterable listing all the classes defined within this class
+            """
+            return tuple(state for name, state in cls.__dict__.items() if name[:1] != '_' and name != 'states')
```
smarnach: This approach won't work together with inheritance, since `cls.__dict__` (or equivalently, `vars(cls)`) does not contain inherited members.  I suggest using `inspect.getmembers()` instead.  (Filtering out `states` isn't necessary with the current approach, since it won't be in the list, but will be necessary when using `inspect.getmembers()`.)

instance/models/server.py
```diff
@@ -76,54 +143,52 @@ def exclude_terminated(self):
...
-        (BOOTED, 'Booted - Booted but not ready to be added to the application'),
-        (PROVISIONING, 'Provisioning - Provisioning is in progress'),
-        (REBOOTING, 'Rebooting - Reboot in progress, to apply changes from provisioning'),
-        (READY, 'Ready - Rebooted and ready to add to the application'),
-        (LIVE, 'Live - Is actively used in the application and/or accessed by users'),
-        (STOPPING, 'Stopping - Stopping temporarily'),
-        (STOPPED, 'Stopped - Stopped temporarily'),
-        (TERMINATING, 'Terminating - Stopping forever'),
-        (TERMINATED, 'Terminated - Stopped forever'),
+    Status = Status
```
e-kolpakov: @bradenmacdonald nit: might be a bit confusing - maybe move `Status` and `Progress` classes into a dedicated module and import the module, so `Status = Status` becomes `Status=state_machine.Status`?

instance/models/server.py
```diff
@@ -76,54 +143,52 @@ def exclude_terminated(self):
...
+    _status_to_active = status.transition(from_states=Status.Started, to_state=Status.Active)
+    _status_to_booted = status.transition(from_states=Status.Active, to_state=Status.Booted)
+    _status_to_provisioning = status.transition(from_states=Status.Booted, to_state=Status.Provisioning)
+    _status_to_rebooting = status.transition(
+        from_states=(Status.Active, Status.Booted, Status.Ready, Status.Provisioning), to_state=Status.Rebooting
+    )
+    _status_to_ready = status.transition(
+        from_states=(Status.Active, Status.Booted, Status.Provisioning, Status.Rebooting), to_state=Status.Ready
+    )
+    _status_to_terminated = status.transition(to_state=Status.Terminated)
```
e-kolpakov: @bradenmacdonald nit: a bit cluttered. Maybe these transitions should become attributes of (e.g.) `ServerStatusStateMachine` class, and used as `server.state.machine.status_to_ready`?

instance/models/server.py
```diff
@@ -76,54 +143,52 @@ def exclude_terminated(self):
...
+        db_index=True,
+        db_column='status',
     )
+    # State transitions:
+    _status_to_started = status.transition(from_states=Status.New, to_state=Status.Started)
+    _status_to_active = status.transition(from_states=Status.Started, to_state=Status.Active)
+    _status_to_booted = status.transition(from_states=Status.Active, to_state=Status.Booted)
+    _status_to_provisioning = status.transition(from_states=Status.Booted, to_state=Status.Provisioning)
+    _status_to_rebooting = status.transition(
+        from_states=(Status.Active, Status.Booted, Status.Ready, Status.Provisioning), to_state=Status.Rebooting
```
e-kolpakov: @bradenmacdonald nit: specifying `from_states` as a list might become problematic in future - I'll elaborate more in comment to entire PR (when I'm done with line-by-line nitpicks :)).

instance/models/server.py
```diff
@@ -76,54 +143,52 @@ def exclude_terminated(self):
...
 
-    PROGRESS_RUNNING = 'running'
-    PROGRESS_SUCCESS = 'success'
-    PROGRESS_FAILED = 'failed'
-
-    PROGRESS_CHOICES = (
-        (PROGRESS_RUNNING, 'Running'),
-        (PROGRESS_SUCCESS, 'Success'),
-        (PROGRESS_FAILED, 'Failed'),
+    Progress = Progress
```
e-kolpakov: @bradenmacdonald I might be wrong, but looks like you've tried to normalize statuses like `provision_running`, `provision_failed`, `rebooting_running`, `rebooting_failed`, etc, and as a result there are some helper methods and orthogonal state machine for this. Have you considered "denormalized" approach, i.e. single `Status` state machine with `*_started`, `*_running`, `*_failed`, `*_complete` statuses?

Also, I believe some transitions are invalid, e.g. you can't go to Provisioning.Failed from any status except Provisioning.Running; if I do understand right, current "orthogonal" implementation does not enforce it in any way, and rely on "callers" to keep that invariant.

instance/models/server.py
```diff
@@ -232,44 +311,51 @@ def public_ip(self):
 
         return public_addr['addr']
 
-    def update_status(self, provisioning=False, rebooting=False, failed=None):
+    def update_status(self):
         """
         Refresh the status by querying the openstack server via nova
```
e-kolpakov: @bradenmacdonald nit: is this docstring still correct? I don't think this method queries anything.

instance/models/server.py
```diff
@@ -147,29 +212,43 @@ def event_context(self):
...
-        Sleep in a loop until the server reaches one of the specified status
+        Sleep in a loop until the server reaches one of the specified statuses
         """
-        target_status_list = [target_status] if isinstance(target_status, str) else target_status
         self.logger.info('Waiting to reach status %s...', target_status_list)
 
         while True:
             self.update_status()
-            if self.status in target_status and self.progress != self.PROGRESS_RUNNING:
+            if self.progress.is_final and isinstance(self.status, target_status_list):
```
e-kolpakov: @bradenmacdonald nit: can't `self.status.one_of` be used here?

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+
+    @classmethod
+    def one_of(cls, *state_classes):
+        """ Syntactic sugar to make comparing an instance to a set of types easier """
+        return issubclass(cls, state_classes)
+
+    def __init__(self, resource, state_manager):
+        """ Instantiate this state, saving a reference to the parent resource """
+        self._resource = resource
+        self._state_manager = state_manager
```
e-kolpakov: @bradenmacdonald nit: are these two attributes used anywhere? Can't find any references in the diff.

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+        try:
+            state = self.cache[resource]
+        except KeyError:
+            # 'resource' refers to a brand new object, whose state property hasn't been accessed
+            # yet. Load the state from the django field, or use the default state.
+            state = self._set_state(resource, self._get_initial_state(resource))
+        assert isinstance(state, self.state_classes)
+        if self.model_field_name:
+            assert getattr(resource, self.model_field_name) == state.state_id, \
+                "The reource's {} field has been tampered with".format(self.model_field_name)
```
e-kolpakov: @bradenmacdonald nit: typo: *reource's*

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+                if state_class:
+                    return state_class
+        # Otherwise, just use the default:
+        return self.default_state_class
+
+    def _get_state_class_from_id(self, state_id):
+        """
+        Given a state_id, get the ResourceState subclass, or None
+        """
+        for state_class in self.state_classes:
```
e-kolpakov: @bradenmacdonald very minor nit: not very impactful, but this method has O(N) performance. Consider building a dictionary of `state_id` -> `StateClass` in `init`.

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+        """
+        Get a tuple of (state_id, name) pairs.
+
+        Suitable for passing to a django CharField choices parameter.
+        """
+        return tuple((state.state_id, state.__class__.__name__) for state in self.state_classes)
+
+    # Internal helper methods:
+
+    def _get_initial_state(self, resource):
```
e-kolpakov: @bradenmacdonald nit: a bit confusing name - it clashes with "initial_state" of the actual state-machine-backed property (that becomes `default_state_class` here). Actually, it tries fetching *current* value from DB, or falls back to default if it's not in DB yet. `_load_state(self, resource, default=self.default_state)` maybe?

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+        default_state: If no state has been set, assume the state is this state class.
+        model_field_name: The name of a django CharField to keep updated with the name of the
+            current state, or None. (caching the state in a field is optional)
+        """
+        self.state_classes = state_classes
+        assert len(set(state_class.state_id for state_class in state_classes)) == len(state_classes), \
+            "A resource's states must each have a unique state_id"
+        assert default_state in state_classes
+        self.default_state_class = default_state
+        self.model_field_name = model_field_name
```
e-kolpakov: @bradenmacdonald nit: extract `model_field_name` related code into a subclass?

instance/models/utils.py
```diff
@@ -44,3 +55,263 @@ def save(self, *args, **kwargs):
...
+
+        def do_transition(resource):
+            """ The method that performs a specific transition """
+            current_state = self.__get__(resource)
+            if from_states and not isinstance(current_state, from_states):
+                raise WrongStateException("This transition cannot be used to move from {} to {}".format(
+                    current_state.name, to_state.name  # pylint: disable=no-member
+                ))
+            self._set_state(resource, to_state)
+        do_transition.from_states = from_states  # Convenient way for other code to inspect this transition
```
e-kolpakov: @bradenmacdonald nit: if `from_states` is None, this will also be `None`; I believe in such case `self.state_classes` would make a bit more sense.

instance/tests/models/test_server.py
```diff
@@ -129,21 +130,28 @@ def test_sleep_until_status(self, mock_sleep, mock_update_status):
...
         mock_update_status.side_effect = update_status
 
-        self.assertEqual(server.sleep_until_status(server.BOOTED), server.BOOTED)
-        self.assertEqual(server.progress, server.PROGRESS_SUCCESS)
-        self.assertEqual(server.status, server.BOOTED)
-        self.assertEqual(server.progress, server.PROGRESS_SUCCESS)
+        self.assertEqual(server.sleep_until_status(ServerStatus.Booted), ServerStatus.Booted)
+        self.assertEqual(server.progress, ServerProgress.Success)
+        self.assertEqual(server.status, ServerStatus.Booted)
+        self.assertEqual(server.progress, ServerProgress.Success)
```
e-kolpakov: @bradenmacdonald nit: double assertion on server.progress - here and two lines above

e-kolpakov:
@bradenmacdonald ok, now the promised overall-PR comment. I see a couple of problems with the current approach:

1. Orthogonal "State" and "Progress" statuses - see inline comment. In two words, it result in two state machines for what essentially is a single state - as a result, (1) the code allows for statuses not existing in reality (like "Active-Failed") and (2) can't control transitions to "substatuses" (see example in inline comment)
2. Not 100% sure but looks like sometimes raising "WrongStatusException" might happen after "event"/"action" causing transition already performed - `server.update_status`.
3. I might be intermixing the concept of state machine with event sourcing, so this might not be immediately applicable to this PR, but still it would be nice to have. So, as I saw it, instead of "external" things telling state machine to transition from one state to the other, state machine should respond to *events*, causing state transitions and each transition somehow declare a (potentially empty) list of actions to be performed. Say, event: "server_startup_complete" causes SM to transition from `BOOTING` to `PROVISIONING` and execute `instance.deploy`. This is a big rewrite, but it does not feel like we could transition to that model without more or less completely rewriting this PR.

Final note: needs rebasing :)

.prospector/opencraft.yml
bradenmacdonald: Good idea.

instance/models/utils.py
bradenmacdonald: Good point. But I'll use `inspect.getdoc().strip()` - getdoc is more relevant, and the `.strip()` is necessary with either `dedent` or `getdoc` to avoid a string like `""" Docstring """` getting returned with a trailing space.

instance/models/utils.py
bradenmacdonald: Thanks; good point. "the hash of type(self)" sounds good to me.

instance/models/utils.py
bradenmacdonald: I rewrote this a couple times and changed my mind each time about the semantics I wanted. I think the most sensible thing for this semantic sugar is strict type checking, rather than `issubclass` or `isinstance` (just use those explicitly if that's what you want, rather than using `==`), so I've changed the method to that and expanded the tests accordingly.

instance/models/utils.py
bradenmacdonald: I agree :)

instance/models/utils.py
bradenmacdonald: `states` was actually in the list with `__dict__`, FYI. I changed this code to use `dir()`. I couldn't use `inspect.getmembers()` since it loads the value of each attribute before calling the predicate method, and that caused infinite recursion on the `states` method. I think one would have to change to using a metaclass to avoid that.

instance/models/server.py
bradenmacdonald: These states are only for `Server`, and we will be adding more state machines and states for other models, so I'd rather keep these states in this file. If you feel strongly about this I could rename `Status` to `ServerStatus` and remove the `Server.Status` scope-helper-attribute.

instance/models/server.py
bradenmacdonald: @e-kolpakov I agree with your general sentiment, but what I'd like to do next is actually to remove the "Progress" state machine altogether. Every status should be considered a "Running" or "Success" status depending on its type, and if something fails, then the status of the whole server should change to an "OfflineStatus" or similar. That's what I put in the code as a TODO item. I jsut didn't want to make that change during this PR, in order to keep the diff smaller and make it easier to review. (more smaller PRs and changes rather than one giant one)

instance/models/server.py
bradenmacdonald: Yes, the docstring is correct - the `os_server` property queries the OpenStack API, and then this code checks the resulting `status`.

instance/models/server.py
bradenmacdonald: Good call. Changed.

instance/models/utils.py
bradenmacdonald: @e-kolpakov They are not currently used, but I think they're important, to allow putting code into the state classes as I described in the PR description. I wrote this part first, and I was a bit more ambitious when starting this, until I realized that adding this implementation plus refactoring all the states of Instance and Server and OpenStack server was too big of change to do in one PR.

instance/models/utils.py
bradenmacdonald: My thought was that if you have so many states that this matters, you should refactor your code. But on second thought, it's easy to change to a `frozenset` and that has nicer semantics anyways, so that's what I did.

bradenmacdonald:
Rebased from 3d170526357a76c56b856f545d60dafe4813e791 onto master.

I have pushed 9096be5bf8216b2aeee3596401af447a289e39fd to address most of the review comments so far - thanks @smarnach and @e-kolpakov :)

@e-kolpakov:

> Orthogonal "State" and "Progress" statuses

See my comment earlier. I agree, but I want to combine them as the first of the "Suggested Next Steps"

> Not 100% sure but looks like sometimes raising "WrongStatusException" might happen after "event"/"action" causing transition already performed

Since the check wraps the method and checks before calling the inner method, I don't think this is possible. But please provide an example, because that sounds like a bug if it is.

> state machine should respond to events

I like that approach and I was somewhat keeping it in mind as I coded this. I tried to keep this approach as flexible as possible. I think we could achieve what you are describing using this PR by:
* Give each `ResourceState` subclass an `enter_state()` method, which is called automatically when the state machine changes to that state.
* Give each `ResourceState` subclass an `update()` method, which can do things like pop events from an event queue and poll external services for changes (like OpenStack)
* Allow each state's `update()` method to cause a transition to another state - this is already possible.

I was actually hoping to write something like the above during this PR (remove States from Server, make them OpenStack-specific, give them methods and events), but I realized it would be too much work, so I put that into "Next Steps".

instance/models/server.py
bradenmacdonald: @e-kolpakov I agree. The problem is that as written, `server.state._state_manager` doesn't have access to the `resource` instance, so couldn't hold transition methods. But since I wanted to implement `ResourceStateDescriptor` as a descriptor, I didn't yet see a simple way to organize the transitions nicely other than what I already put in this PR. Open to suggestions here.

instance/models/server.py
bradenmacdonald: @e-kolpakov What did you mean by this? `from_states` is optional, and can be a single class/mixin, or a tuple of classes/mixins, and the verification uses `isinstance`, so I think it's a pretty flexible approach.

instance/models/utils.py
bradenmacdonald: Fixed.

instance/models/utils.py
bradenmacdonald: I like that idea, though currently every usage we have uses model_field_name. I didn't get time to do this refactoring today, but I do like it in principle.

instance/models/utils.py
bradenmacdonald: @e-kolpakov I don't see any variables/methods called `initial_state`. In this case, I meant `initial` as in "the state when this python class instance is first initialized", not as in "the state when this type of resource is first created". I do see how that's a bit confusing but I prefer the current name to `_load_state`; there's no fetching from DB, though it does read a django field if that field is being used and already loaded from the DB.

instance/models/server.py
e-kolpakov: @bradenmacdonald imho, it is a bit "wrong" from mathematical point of view. FSM transitions are given as function of (current_state, event) -> new state. This transitions differ in two ways:

1. They only take into account current state (i.e. they are (current_state, ANY_EVENT) -> new_state functions)
2. They specify a list of current states instead of single state.

It might not be practical to pursue this right now, but without events taken into account current implementation is not a FSM, but a collection of rules to allow/deny transitions between states.

instance/models/utils.py
e-kolpakov: @bradenmacdonald than I suggest removing it - we can add these two later, when we need them.

instance/models/server.py
e-kolpakov: @bradenmacdonald I'm not in the loop for making a sensible suggestions, so let's leave it as is for now.

instance/models/server.py
e-kolpakov: @bradenmacdonald not worth it right now, but pylint will complain of "too many lines in module" sooner or later :)

instance/models/server.py
e-kolpakov: @bradenmacdonald ok, do we have tasks for those TODOs? Also, this delayed features might become a technical debt.

instance/models/utils.py
e-kolpakov: @bradenmacdonald ok.

e-kolpakov:
@bradenmacdonald :+1: provided "Next Steps" are logged into JIRA

bradenmacdonald:
@e-kolpakov I found a small bug in manual testing, and the best way to fix it was the django model refactor you suggested. Please review 15cba1946ef53af33b563e5a571bea71a52acda3 and 8f9b711f59217dbb9388bc3058c2dd6ba86e3614.

If they're good, I'll merge then create new JIRA stories.

e-kolpakov:
@bradenmacdonald reviewed, :+1:

instance/models/server.py
bradenmacdonald: @e-kolpakov Just to respond to your comments about the formal definition of an FSM:

You are correct that this does not match a formal description as you describe. I actually started implementing something more like that at first, but I found this approach to be more practical (for now) for a few reasons:
* There is no event loop in the app right now. There is a sort of "local event loop" on the server class with its `update_status` and `sleep_until_status` methods, and together they kind of approximate what you are describing. For example, you can think of the "mark_as_provisioning()" method as an event dispatcher which dispatches the PROVISIONING_STARTED event, and Booted state will catch that event and then transition to the "Provisioning" status. There is no other way to use a public method and get the `Server` instance into the "Provisioning" state besides sending that specific event when it is already in the Booted status. I'll admit this is a bit of a stretch though as it's not formally implemented using "events".
* We have to be careful with event handling since this app runs as multiple processes. For example, with a naive event implementation, if we dispatched a "server is ready for provisioning" event to a global event queue, then it's possible that either (A) multiple processes would catch that event and handle it by starting competing ansible runs, or (B) only one processes would catch that event (by design) and then the other processes' `Server.status` info would be out of date. These problems can be worked around, but I just want to point them out. Currently we work around them by having one worker "own" the process of provisioning and explicitly tell the server to update its status.
* The "list of states" vs. single state only is a pragmatic shortcut, to represent the fact that the server could fail to an error state or a terminated state at any time (due to a problem on the provider side), and saving us from having to define a separate transition on each state to the error/terminated states.

smarnach:
@bradenmacdonald You've asked me on the discovery document to create the stories for the future steps, but here you said you will create them yourself.  We have to make sure only one of us actually does this. :)  It probably makes sense if you do it, and you've already created the story for the last bullet point.

bradenmacdonald:
@smarnach Sure, I will create them today and then you and Eugeny and Xavier can review them and adjust as you see fit.

bradenmacdonald:
@e-kolpakov @itsjeyd I made a tiny change (comments only) and now integration tests are failing due to some sort of issue with OVH's apt mirrors :-/

e.g.

```
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration | TASK: [edxlocal | install apt repository] *************************************
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration | failed: [149.202.185.48] => {"failed": true, "parsed": false}
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration | BECOME-SUCCESS-pbvjggtzjrpotpchfwpodfxbbhrbotml
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration | Traceback (most recent call last):
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |   File "/home/ubuntu/.ansible/tmp/ansible-tmp-1460574641.18-189577319084662/apt_repository", line 2524, in <module>
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |     main()
13/Apr/2016 19:11:14 |    ERROR | process=10645 | instance.models.instance  | instance=35d64f50.integration | You should consider upgrading via the 'pip install --upgrade pip' command.
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |   File "/home/ubuntu/.ansible/tmp/ansible-tmp-1460574641.18-189577319084662/apt_repository", line 436, in main
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |     cache.update()
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |   File "/usr/lib/python2.7/dist-packages/apt/deprecation.py", line 98, in deprecated_function
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |     return func(*args, **kwds)
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |   File "/usr/lib/python2.7/dist-packages/apt/cache.py", line 441, in update
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration |     raise FetchFailedException(e)
13/Apr/2016 19:11:14 |     INFO | process=10645 | instance.models.instance  | instance=35d64f50.integration | apt.cache.FetchFailedException: W:Failed to fetch http://nova.clouds.archive.ubuntu.com/ubuntu/dists/precise/main/source/Sources  404  Not Found [IP: 91.189.91.23 80]
```

```
13/Apr/2016 19:10:40 |     INFO | process=10639 | instance.models.instance  | instance=aed9deef.integration | TASK: [common | Install role-independent useful system packages] **************
13/Apr/2016 19:10:40 |     INFO | process=10639 | instance.models.instance  | instance=aed9deef.integration | failed: [149.202.185.49] => {"failed": true}
13/Apr/2016 19:10:40 |    ERROR | process=10639 | instance.models.instance  | instance=aed9deef.integration | You should consider upgrading via the 'pip install --upgrade pip' command.
13/Apr/2016 19:10:40 |     INFO | process=10639 | instance.models.instance  | instance=aed9deef.integration | msg: Could not fetch updated apt files
13/Apr/2016 19:10:40 |     INFO | process=10639 | instance.models.instance  | instance=aed9deef.integration | 
13/Apr/2016 19:10:40 |     INFO | process=10639 | instance.models.instance  | instance=aed9deef.integration | FATAL: all hosts have already failed -- aborting
```

I assume this is a temporary issue. Could one of you please try rebuilding this in the morning and then merge it once the build is green?

smarnach:
@bradenmacdonald That's a weird error.  The apt client should try to fetch `Sources.gz` (which exists), not the uncompressed `Sources` (which isn't usually available on an apt mirror).  If the error message is correct, this is a client-side problem, and not a problem with the OVH Ubuntu mirror.

bradenmacdonald:
@smarnach I've been seeing other errors too, including these 403 errors which now seem to be resolved:

```
13/Apr/2016 17:42:08 |     INFO | process=10655 | instance.models.instance  | instance=52067d63.integration | Failed to fetch http://nova.clouds.archive.ubuntu.com/ubuntu/pool/main/n/ntp/ntp_4.2.6.p3+dfsg-1ubuntu3.9_amd64.deb  403  Forbidden [IP: 91.189.91.23 80]
```

bradenmacdonald:
Update: I re-ran the build and I see now that it's passing :)  However, I'm out of time and am wary of doing a deployment just before I sign off for a long weekend. Could one of you please deploy this for me? (perhaps @itsjeyd as part of his PR which is based on this branch anyways). Thanks team!

antoviaque:
FYI, `nova.clouds.archive.ubuntu.com` should still be the canonical ubuntu repositories/mirrors, not a mirror specific to OVH.

antoviaque:
For the deployment of this, it could also be @e-kolpakov since he has to deploy     OC-1517 still - I'll let you two decide : )

e-kolpakov:
@antoviaque you've got outdated information - I've deployed OC-1517 on Tuesday morning :)

antoviaque:
@e-kolpakov Alright, good! Where could I have seen this information? I don't see an update on OC-1517 and it's still in "merged" status.

@itsjeyd Do you want to try your hand at deploying this, then? If so I'll add you to the authorized_keys on console.opencraft.com

e-kolpakov:
@antoviaque I thought I have posted a comment on OC-1517, but apparently I didn't (maybe I've hit "Submit" but it didn't arrive to JIRA - I had some connectivity issues back then). So, you're right - you couldn't obtain it from anywhere.

antoviaque:
@e-kolpakov Ok - no worries; I'll move that ticket to delivered then. Thanks!

itsjeyd:
@antoviaque Yes, I can take on the deployment :) I'll ping you about it on IRC soon.

PR #52, opened 2016-04-13
Make code agnostic to specific server states

itsjeyd:
... and use new `name` and `description` properties when displaying information about current state of an instance to a user.

Follows up on https://github.com/open-craft/opencraft/pull/51.

cf. [OC-1535](https://tasks.opencraft.com/browse/OC-1535)

**Update**: As of https://github.com/open-craft/opencraft/pull/52/commits/0378c167170e18c55fe87a18a6201d6e7a0898d6, also addresses [this request](https://github.com/open-craft/opencraft/pull/52#discussion_r59595130) from @bradenmacdonald.

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -426,10 +426,14 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
         server = self.server_set.create()
         server.start()
 
+        def can_proceed():
+            """ Is server in a state from which we can proceed? """
+            return server.status.accepts_ssh_commands
+
         try:
             # DNS
             self.logger.info('Waiting for IP assignment on server %s...', server)
-            server.sleep_until_status(server.Status.Active, server.Status.Booted)
+            server.sleep_until(can_proceed)
             self.logger.info('Updating DNS: LMS at %s...', self.domain)
             gandi.set_dns_record(type='A', name=self.sub_domain, value=server.public_ip)
             self.logger.info('Updating DNS: Studio at %s...', self.studio_domain)
@@ -442,7 +446,6 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
                 self.provision_mongo()
 
             # Provisioning (ansible)
-            server.sleep_until_status(server.Status.Booted)
             server.mark_as_provisioning()
             self.reset_ansible_settings(commit=True)
             log, exit_code = self.deploy()
@@ -456,7 +459,7 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
             # Reboot
             self.logger.info('Rebooting server %s...', server)
             server.reboot()
-            server.sleep_until_status(server.Status.Ready)
+            server.sleep_until(can_proceed)
             self.logger.info('Provisioning completed')
 
             return (server, log)
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -426,14 +426,14 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
         server = self.server_set.create()
         server.start()
 
-        def can_proceed():
-            """ Is server in a state from which we can proceed? """
+        def accepts_ssh_commands():
+            """ Does server accept SSH commands? """
             return server.status.accepts_ssh_commands
 
         try:
             # DNS
             self.logger.info('Waiting for IP assignment on server %s...', server)
-            server.sleep_until(can_proceed)
+            server.sleep_until(accepts_ssh_commands)
             self.logger.info('Updating DNS: LMS at %s...', self.domain)
             gandi.set_dns_record(type='A', name=self.sub_domain, value=server.public_ip)
             self.logger.info('Updating DNS: Studio at %s...', self.studio_domain)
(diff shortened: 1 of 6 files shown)


instance/models/server.py
```diff
@@ -54,6 +56,8 @@ class ServerState(ResourceState):
     """
     A [finite state machine] state describing a virtual machine.
     """
+    is_steady_state = False
+    accepts_ssh_commands = False
```
bradenmacdonald: These should have comments explaining them here.

instance/models/server.py
```diff
@@ -242,18 +252,36 @@ def _set_progress(self, progress, expected_status=None):
...
         """
-        self.logger.info('Waiting to reach status %s...', target_status_list)
+        if condition():
+            return
+        else:
+            if self.status.is_steady_state:
+                raise SteadyStateException(
+                    "The current state ({name}) does not fulfill the desired condition "
+                    "and is steady, i.e., it is not expected to change.".format(name=self.status.name)
+                )
```
bradenmacdonald: This check should be inside the `while` loop. When `self.update_status()` is called, the server could transition into a steady state, and then this would loop forever.

instance/models/server.py
```diff
@@ -242,18 +252,36 @@ def _set_progress(self, progress, expected_status=None):
...
+
+        self.logger.info('Waiting to reach status from which we can proceed...')
 
         while True:
             self.update_status()
-            if self.progress.is_final and self.status.one_of(*target_status_list):
+            if self.progress.is_final and condition():
+                self.logger.info('Reached appropriate status. Proceeding.')
```
bradenmacdonald: BTW, this isn't part of this task, but if you have time, I'd love to see us add a timeout parameter to this method as well, so that if the loop sleeps for more than, say, a default of 60 minutes, it will raise an exception.

instance/models/instance.py
```diff
@@ -426,10 +426,14 @@ def provision(self):
         server = self.server_set.create()
         server.start()
 
+        def can_proceed():
```
e-kolpakov: @itsjeyd nit: `can_proceed` is a bit too generic: maybe `server_accepts_ssh_commands`?

instance/models/server.py
```diff
@@ -242,18 +262,49 @@ def _set_progress(self, progress, expected_status=None):
...
-        while True:
+        self.logger.info('Waiting to reach status from which we can proceed...')
+
+        while timeout > 0:
             self.update_status()
-            if self.progress.is_final and self.status.one_of(*target_status_list):
-                break
+            if condition():
+                if self.progress.is_final:
+                    self.logger.info('Reached appropriate status. Proceeding.')
```
e-kolpakov: @itsjeyd nit: include status into log message

instance/models/server.py
```diff
@@ -242,18 +262,49 @@ def _set_progress(self, progress, expected_status=None):
...
+                    return
+            else:
+                if self.status.is_steady_state:
+                    raise SteadyStateException(
+                        "The current state ({name}) does not fulfill the desired condition "
+                        "and is steady, i.e., it is not expected to change.".format(name=self.status.name)
```
e-kolpakov: @itsjeyd nit: "and is not expected to change" would be enough :)

instance/tests/models/test_server.py
```diff
@@ -125,51 +125,117 @@ def authenticate(client):
 
     @patch('instance.models.server.OpenStackServer.update_status')
     @patch('instance.models.server.time.sleep')
-    def test_sleep_until_status(self, mock_sleep, mock_update_status):
+    def test_sleep_until_condition_already_fulfilled(self, mock_sleep, mock_update_status):
```
e-kolpakov: @itsjeyd nit: `test_sleep_until_condition_fulfilled` feels like slightly "more correct" :)

instance/tests/models/test_server.py
```diff
@@ -125,51 +125,117 @@ def authenticate(client):
...
+                server._status_to_active,
+                server._status_to_booted,
+            ]
+            status_queue.reverse() # To be able to use pop()
+
+            mock_update_status.side_effect = scoped_update_status(server, status_queue)
+            mock_sleep.call_count = 0
+
+            # Sleep until condition is fulfilled
+            server.sleep_until(condition)
```
e-kolpakov: @itsjeyd nit: Imagine that someone break something so that status is stuck in transient (non-steady) state, and wait condition is not satisfied. Do I understand right that this test will wait for status change until default timeout have passed? If so, it might be good to have `timeout` set to 1 or other significantly small value.

instance/tests/models/test_server.py
```diff
@@ -125,51 +125,117 @@ def authenticate(client):
...
+            server._transition(server._status_to_started, progress=ServerProgress.Success)
         mock_update_status.side_effect = update_status
 
-        self.assertEqual(server.sleep_until_status(ServerStatus.Booted), ServerStatus.Booted)
-        self.assertEqual(server.progress, ServerProgress.Success)
-        self.assertEqual(server.status, ServerStatus.Booted)
-        self.assertEqual(mock_sleep.call_count, 3)
-        self.assertEqual(status_queue, [server._status_to_terminated])
+        # Pretend that Status.Started (which doesn't accept SSH commands) is a steady state
+        ServerStatus.Started.is_steady_state = True
```
e-kolpakov: @itsjeyd Shouldn't this line affect all the test running after the test? COnsider wrapping everything in catfch-all exception handler and restoring old value, or just using `mock.patch`

e-kolpakov:
@itsjeyd :+1: conditional tests pass

instance/models/instance.py
itsjeyd: @e-kolpakov I went for just `accepts_ssh_commands`, since it reads a bit better when it's being used below: `server.sleep_until(accepts_ssh_commands)` vs. `server.sleep_until(server_accepts_ssh_commands)`.

instance/tests/models/test_server.py
itsjeyd: @e-kolpakov Yes, you're right, it would wait until the default timeout has passed. I changed it to use a much smaller timeout (here and in a couple more places).

itsjeyd:
@e-kolpakov Thanks! Comments addressed. I'll squash my commits once I get a green build.

PR #53, opened 2016-04-16
Add pending migrations.

smarnach:
These are migrations currently missing on the master branch.  I also needed to normalise the order of choices for the progress and status fields, since with the randomised order we had before we would get a new migration everytime you run `makemigrations`.

cc @bradenmacdonald

(Shouldn't the tests be able to catch missing migrations?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ collectstatic: clean static_external
 migrate: clean
 	$(HONCHO_MANAGE) migrate
 
+# Check for unapplied migrations
 migration_check: clean
 	!(($(HONCHO_MANAGE) showmigrations | grep '\[ \]') && printf "\n\033[0;31mERROR: Pending migrations found\033[0m\n\n")
 
@@ -92,6 +93,10 @@ test_unit: clean
 	@echo -e "\nCoverage HTML report at file://`pwd`/build/coverage/index.html\n"
 	@coverage report --fail-under 94 || (echo "\nERROR: Coverage is below 95%\n" && exit 2)
 
+# Check whether migrations need to be generated
+test_migrations_missing: clean
+	! honcho -e .env.test run ./manage.py makemigrations --dry-run --exit
+
 test_integration: clean
 	@if [ -a .env.integration ] ; then \
 		echo -e "\nRunning integration tests..." ; \
@@ -106,7 +111,7 @@ test_js: clean static_external
 test_js_web: clean static_external
 	cd instance/tests/js && jasmine --host 0.0.0.0
 
-test: clean test_prospector test_unit test_js test_integration
+test: clean test_prospector test_unit test_migrations_missing test_js test_integration
 	@echo -e "\nAll tests OK!\n"
 
 test_one: clean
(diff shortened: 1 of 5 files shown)


bradenmacdonald:
@smarnach This migration is wrong. In this migration you can see some things like "RemoveField status... AddField _status", which will delete all data in the `status` column that exists in the database. That's not what we want. If you look in models.py you can see that the new `_status` column says `db_column='status'` so that the "real" column name is unchanged. (Which is why the tests continue to pass just fine; the existing migrations create the correct tables and columns.)

FYI, these migrations primarily reflect changes to the `choices` setting of each field. Since Django does not enforce `choices` at the database level, I find it annoying that `choices` are reflected in migrations at all, but it's apparently helpful for checking the correctness of data migrations.

smarnach:
@bradenmacdonald You are right.  It's not really possible to model "rename a field, but keep the DB column name the same" in Django migrations, so I had to do it in two steps, and the resulting SQL is a bit daft:

```sql
BEGIN;
ALTER TABLE "instance_openstackserver" RENAME COLUMN "progress" TO "_progress";
ALTER TABLE "instance_openstackserver" RENAME COLUMN "status" TO "_status";
ALTER TABLE "instance_openstackserver" RENAME COLUMN "_progress" TO "progress";
ALTER TABLE "instance_openstackserver" RENAME COLUMN "_status" TO "status";
ALTER TABLE "instance_generallogentry" ALTER COLUMN "level" SET DEFAULT 'INFO';
ALTER TABLE "instance_generallogentry" ALTER COLUMN "level" DROP DEFAULT;
ALTER TABLE "instance_instancelogentry" ALTER COLUMN "level" SET DEFAULT 'INFO';
ALTER TABLE "instance_instancelogentry" ALTER COLUMN "level" DROP DEFAULT;
ALTER TABLE "instance_serverlogentry" ALTER COLUMN "level" SET DEFAULT 'INFO';
ALTER TABLE "instance_serverlogentry" ALTER COLUMN "level" DROP DEFAULT;
COMMIT;
```

Should be good enough though (and I don't have a better idea anyway).

smarnach:
@bradenmacdonald Should the tests be modified to catch missing migrations?  I'm not sure we would want the tests to fail locally when a migration is missing, but I think we want it on CircleCI to prevent PRs that don't include all migrations from being merged.

bradenmacdonald:
@smarnach That's pretty ridiculous SQL, but I think that should be fine. Technically one can create custom `Operation` classes within django's migration framework to cover edge cases like this, but I don't think that's worthwhile right now. It's too bad the `RenameField` `Operation` doesn't check if `from_model._meta.get_field(self.old_name) == to_model._meta.get_field(self.new_name)`.

And yes, it would be awesome if we can add a check for missing migrations. EdX recently did that on edx-platform.

smarnach:
@bradenmacdonald Could you please take another look?  I'd like to get this merged as soon as possible, since I've got two more PRs based on this one, and one of them is ready for merge.

bradenmacdonald:
Migration seems to work well, and I confirmed that the `test_migrations_missing` test works correctly.

👍

PR #54, opened 2016-04-16
Support provisioning Swift containers.

smarnach:
Add support to provision Swift containers.  This is implemented analogously to MySQL and MongoDB provisioning (see #11).  Swift containers are only provisioned for instances with persistent databases, but I also added a separate flag to disable Swift provisioning if it's unwanted for some reason.

Unlike MySQL and MongoDB provisioning, Swift provisioning can't be tested locally with reasonable effort (see the [Swift All-In-One documentation](http://docs.openstack.org/developer/swift/development_saio.html) to get an idea of the complexity of setting up a local Swift cluster.), so the unit tests only test whether the API call to the Swift API is made.  The integration test will provision Swift containers though.  They aren't cleaned up when the instance is terminated (which is kind of the point of persistent databases).  In the long run we need some clean-up process to remove unused Swift containers. (The same holds for unused MongoDB and MySQL databases, which aren't cleaned up either.)

I had to modify the hack I implemented in one of the migrations, since it didn't work anymore for the latest version of `OpenEdXInstance`.  The new version of the hack should be more resilient, though it will fail to migrate instances correctly for people with a very old version of the instance manager installed, which shouldn't be a problem in practice.

As discussed on Jira, the full OpenStack credentials are passed on to the instance.  We should get a separate OpenStack project that doesn't have the permission to provision VMs soon, so we at least don't have the OpenStack credentials used to provision the VMs themselves on the servers.

**Testing instructions**

On the development Vagrant instance, run `pip install -r requirements.txt` to make sure the Swift client is installed.  Make sure OpenStack credentials are configured in `.env`, and provision a new instance from the development installation, e.g. by running `make shell` and entering

    In [1]: from instance.tasks import provision_instance
    In [2]: instance = OpenEdXInstance(sub_domain='swift-test', name='Swift Test')
    In [3]: provision_instance(instance.pk)

Once the provisioning finishes, verify the value of `instance.swift_provisioned` and copy the value of `instance.swift_container_name`.  In the shell, set the environment variables `OS_USERNAME`, `OS_TENANT_NAME`, `OS_PASSWORD`, `OS_AUTH_URL` and `OS_REGION_NAME` to the corresponding values from the `.env` file (note that the variable names in that file are different), and run

    swift stat <container_name>

with the container name noted above.  Verify that the container has the `.r:*` read ACL.

An even better test would be to provision an instance using the branch from https://github.com/edx/edx-platform/pull/11286, and then test the instance actually works with the container (which I didn't test yet).

### DIFF
--- a/instance/migrations/0039_auto_20160416_1729.py
+++ b/instance/migrations/0039_auto_20160416_1729.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0038_merge'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='openstackserver',
+            old_name='progress',
+            new_name='_progress',
+        ),
+        migrations.RenameField(
+            model_name='openstackserver',
+            old_name='status',
+            new_name='_status',
+        ),
+        migrations.AlterField(
+            model_name='openstackserver',
+            name='_progress',
+            field=models.CharField(choices=[('failed', 'type'), ('running', 'type'), ('success', 'type')], default='running', db_column='progress', max_length=7),
+        ),
+        migrations.AlterField(
+            model_name='openstackserver',
+            name='_status',
+            field=models.CharField(choices=[('active', 'type'), ('booted', 'type'), ('new', 'type'), ('provisioning', 'type'), ('ready', 'type'), ('rebooting', 'type'), ('started', 'type'), ('terminated', 'type')], default='new', db_column='status', max_length=20, db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='generallogentry',
+            name='level',
+            field=models.CharField(choices=[('DEBUG', 'Debug'), ('INFO', 'Info'), ('WARNING', 'Warning'), ('ERROR', 'Error'), ('CRITICAL', 'Critical')], default='INFO', max_length=9, db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='instancelogentry',
+            name='level',
+            field=models.CharField(choices=[('DEBUG', 'Debug'), ('INFO', 'Info'), ('WARNING', 'Warning'), ('ERROR', 'Error'), ('CRITICAL', 'Critical')], default='INFO', max_length=9, db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='serverlogentry',
+            name='level',
+            field=models.CharField(choices=[('DEBUG', 'Debug'), ('INFO', 'Info'), ('WARNING', 'Warning'), ('ERROR', 'Error'), ('CRITICAL', 'Critical')], default='INFO', max_length=9, db_index=True),
+        ),
+    ]
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -38,7 +38,7 @@ from instance.gandi import GandiAPI
 from instance.logger_adapter import InstanceLoggerAdapter
 from instance.logging import log_exception
 from instance.models.mixins.ansible import AnsibleInstanceMixin
-from instance.models.mixins.database import MySQLInstanceMixin, MongoDBInstanceMixin
+from instance.models.mixins.database import MongoDBInstanceMixin, MySQLInstanceMixin, SwiftContainerInstanceMixin
 from instance.models.mixins.utilities import EmailInstanceMixin
 from instance.models.mixins.version_control import GitHubInstanceMixin
 from instance.models.utils import ValidateModelMixin
@@ -243,8 +243,8 @@ class Instance(ValidateModelMixin, TimeStampedModel):
 
 
 # pylint: disable=too-many-instance-attributes
-class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceMixin,
-                      GitHubInstanceMixin, EmailInstanceMixin, Instance):
+class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftContainerInstanceMixin,
+                      AnsibleInstanceMixin, GitHubInstanceMixin, EmailInstanceMixin, Instance):
     """
     A single instance running a set of Open edX services
     """
@@ -263,6 +263,7 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
         'ansible_s3_settings',
         'ansible_mysql_settings',
         'ansible_mongo_settings',
+        'ansible_swift_settings',
     ]
 
     class ProvisionMessages(object):
@@ -324,6 +325,21 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
                                 'port': settings.INSTANCE_MONGO_URL_OBJ.port or 27017,
                                 'database': self.mongo_database_name,
                                 'forum_database': self.forum_database_name})
+
+    @property
+    def ansible_swift_settings(self):
+        """
+        Ansible settings for Swift access.
+        """
+        if self.use_ephemeral_databases or not settings.SWIFT_ENABLE:
+            return ''
+
+        template = loader.get_template('instance/ansible/swift.yml')
+        return template.render({'user': self.swift_openstack_user,
+                                'password': self.swift_openstack_password,
+                                'tenant': self.swift_openstack_tenant,
+                                'auth_url': self.swift_openstack_auth_url,
+                                'region': self.swift_openstack_region})
 
     @property
     def studio_sub_domain(self):
@@ -392,6 +408,20 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
         """
         return [self.mongo_database_name, self.forum_database_name]
 
+    @property
+    def swift_container_name(self):
+        """
+        The name of the Swift container used by the instance.
+        """
+        return self.database_name
+
+    @property
+    def swift_container_names(self):
+        """
+        The list of Swift container names to be created.
+        """
+        return [self.swift_container_name]
+
     def save(self, **kwargs):
         """
         Set this instance's default field values
@@ -441,9 +471,12 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
 
             # Provisioning (external databases)
             if not self.use_ephemeral_databases:
-                self.logger.info('Provisioning external databases...')
+                self.logger.info('Provisioning MySQL database...')
                 self.provision_mysql()
+                self.logger.info('Provisioning MongoDB databases...')
                 self.provision_mongo()
+                self.logger.info('Provisioning Swift container...')
+                self.provision_swift()
 
             # Provisioning (ansible)
             server.mark_as_provisioning()
@@ -486,3 +519,18 @@ class OpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, AnsibleInstanceM
             self.mongo_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
             self.mongo_pass = get_random_string(length=32)
         return super().provision_mongo()
+
+    def provision_swift(self):
+        """
+        Set Swfit credentials and create the Swift container.
+        """
+        if settings.SWIFT_ENABLE and not self.swift_provisioned:
+            # TODO: Figure out a way to use separate credentials for each instance.  Access control
+            # on Swift containers is granted to users, and there doesn't seem to be a way to create
+            # Keystone users in OpenStack public clouds.
+            self.swift_openstack_user = settings.SWIFT_OPENSTACK_USER
+            self.swift_openstack_password = settings.SWIFT_OPENSTACK_PASSWORD
+            self.swift_openstack_tenant = settings.SWIFT_OPENSTACK_TENANT
+            self.swift_openstack_auth_url = settings.SWIFT_OPENSTACK_AUTH_URL
+            self.swift_openstack_region = settings.SWIFT_OPENSTACK_REGION
+        return super().provision_swift()
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -19,10 +19,13 @@
 """
 Instance app model mixins - Database
 """
-import MySQLdb as mysql
-import pymongo
+
 from django.conf import settings
 from django.db import models
+import MySQLdb as mysql
+import pymongo
+
+from instance.openstack import create_swift_container
 
 
 class MySQLInstanceMixin(models.Model):
@@ -96,4 +99,43 @@ class MongoDBInstanceMixin(models.Model):
                 mongo[database].add_user(self.mongo_user, self.mongo_pass)
 
             self.mongo_provisioned = True
+            self.save()
+
+
+class SwiftContainerInstanceMixin(models.Model):
+    """
+    Mixin to provision Swift containers for an instance.
+    """
+    swift_openstack_user = models.CharField(max_length=32, blank=True)
+    swift_openstack_password = models.CharField(max_length=64, blank=True)
+    swift_openstack_tenant = models.CharField(max_length=32, blank=True)
+    swift_openstack_auth_url = models.URLField(blank=True)
+    swift_openstack_region = models.CharField(max_length=16, blank=True)
+    swift_provisioned = models.BooleanField(default=False)
+
+    class Meta:
+        abstract = True
+
+    @property
+    def swift_container_names(self):  # pylint: disable=no-self-use
+        """
+        An iterable of Swift container names.
+        """
+        return NotImplementedError
+
+    def provision_swift(self):
+        """
+        Create the Swift containers if necessary.
+        """
+        if settings.SWIFT_ENABLE and not self.swift_provisioned:
+            for container_name in self.swift_container_names:
+                create_swift_container(
+                    container_name,
+                    user=self.swift_openstack_user,
+                    password=self.swift_openstack_password,
+                    tenant=self.swift_openstack_tenant,
+                    auth_url=self.swift_openstack_auth_url,
+                    region=self.swift_openstack_region,
+                )
+            self.swift_provisioned = True
             self.save()
--- a/instance/templates/instance/ansible/swift.yml
+++ b/instance/templates/instance/ansible/swift.yml
@@ -1,0 +1,8 @@
+# The following line is commented out until the openstack role has been merged upstream.
+# EDXAPP_DEFAULT_FILE_STORAGE: 'swift.storage.SwiftStorage'
+EDXAPP_SWIFT_AUTH_VERSION: '2'
+EDXAPP_SWIFT_USERNAME: '{{ user }}'
+EDXAPP_SWIFT_KEY: '{{ password }}'
+EDXAPP_SWIFT_TENANT_NAME: '{{ tenant }}'
+EDXAPP_SWIFT_AUTH_URL: '{{ auth_url }}'
+EDXAPP_SWIFT_REGION_NAME: '{{ region }}'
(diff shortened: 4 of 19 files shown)


antoviaque:
@smarnach For cleaning up containers - it is by design yes, but did you mean that integration tests also leave the containers behind? The tests themselves should make sure to remove them in that case, they should cleanup after themselves. It is an issue for MySQL & MongoDB too (there is actually a bug about this in the backlog, OC-1543 ), but the fact that there is existing debt elsewhere doeesn't make it ok to add to that debt on this PR for SWIFT containers...

Also, I don't see any change in the integration tests in the diff - I didn't look at it closely so I might have missed something, but are you checking that the integration tests correctly setup the swift containers?

For the additional project, I'll create it - I have (again) to see this with the support, their quotas checks are broken again. For integration we can just use a different zone maybe for now? Or would it be worth creating a dedicated project there too?

instance/migrations/0039_auto_20160416_1729.py
```diff
@@ -0,0 +1,47 @@
...
+            model_name='instancelogentry',
+            name='level',
+            field=models.CharField(choices=[('DEBUG', 'Debug'), ('INFO', 'Info'), ('WARNING', 'Warning'), ('ERROR', 'Error'), ('CRITICAL', 'Critical')], default='INFO', max_length=9, db_index=True),
+        ),
+        migrations.AlterField(
+            model_name='serverlogentry',
+            name='level',
+            field=models.CharField(choices=[('DEBUG', 'Debug'), ('INFO', 'Info'), ('WARNING', 'Warning'), ('ERROR', 'Error'), ('CRITICAL', 'Critical')], default='INFO', max_length=9, db_index=True),
+        ),
+    ]
```
mtyaka: I find django migrations quite confusing... I guess this file is only here to be able to get the next migration to run and will disappear from this diff after https://github.com/open-craft/opencraft/pull/53 is merged and this PR rebased.

instance/models/instance.py
```diff
@@ -440,6 +470,7 @@ def provision(self):
                 self.logger.info('Provisioning external databases...')
                 self.provision_mysql()
                 self.provision_mongo()
+                self.provision_swift()
```
mtyaka: nit: it would be nice to put a log statement before each step: "Provisioning MySQL...", "Provisioning Mongo...", "Provisioning Swift...".

instance/models/mixins/database.py
```diff
@@ -97,3 +100,42 @@ def provision_mongo(self):
...
+                create_swift_container(
+                    container_name,
+                    user=self.swift_openstack_user,
+                    password=self.swift_openstack_password,
+                    tenant=self.swift_openstack_tenant,
+                    auth_url=self.swift_openstack_auth_url,
+                    region=self.swift_openstack_region,
+                )
+                self.swift_provisioned = True
+                self.save()
```
mtyaka: I think the above two lines should be outside the `for` loop.

instance/tests/models/test_instance.py
```diff
@@ -555,6 +558,50 @@ def test_provision_mongo_again(self):
...
+    def test_swift_disabled(self, mock_swift_connection):
+        """
+        Verify disabling Swift provisioning works.
+        """
+        instance = OpenEdXInstanceFactory(use_ephemeral_databases=False)
+        instance.provision_swift()
+        self.assertIs(instance.swift_provisioned, False)
+        self.assertFalse(mock_swift_connection.called)
+
+
```
mtyaka: We should also add some tests to `OpenEdXInstanceTestCase`, to test that `provision_swift` is/is not invoked, and that swift ansible vars are/are not set with different combinations of values for `SWIFT_ENABLE` and `use_ephemeral_databases`.

instance/templates/instance/ansible/swift.yml
```diff
@@ -0,0 +1,6 @@
+EDXAPP_SWIFT_AUTH_VERSION: '2'
+EDXAPP_SWIFT_USERNAME: '{{ user }}'
+EDXAPP_SWIFT_KEY: '{{ password }}'
+EDXAPP_SWIFT_TENANT_NAME: '{{ tenant }}'
+EDXAPP_SWIFT_AUTH_URL: '{{ auth_url }}'
+EDXAPP_SWIFT_REGION_NAME: '{{ region }}'
```
mtyaka: To be able to use Swift containers with https://github.com/edx/edx-platform/pull/11286, we will have to set `EDXAPP_DEFAULT_FILE_STORAGE` to `'swift.storage.SwiftStorage'`. I suppose this would be the right place to do that.
We will also have to set `EDXAPP_SETTINGS` to `opencraft` somewhere - but I think we can do that in [ansible/vars.yml](https://github.com/open-craft/opencraft/blob/smarnach/swift/instance/templates/instance/ansible/vars.yml).

instance/migrations/0039_auto_20160416_1729.py
smarnach: I split #53 out to get that PR merged as quickly as possible, and based this PR on it.  I needed to do so to be able to generate the correct migrations in this PR.  And yes, the commit will disappear from this PR once we merged the other one.

instance/templates/instance/ansible/swift.yml
smarnach: @mtyaka Right, we need to add `EDXAPP_SETTINGS: openstack` to vars.yml, but can't do this right now, since it would break provisioning any branch that doesn't include edx/edx-platform#11286.  I'm not sure how to handle this transition.  For now, I'd suggest to manually add this to the PR settings if we need it, since I can't think of a good way to automatically decide which set of settings to use.

We won't need to set `EDXAPP_DEFAULT_FILE_STORAGE` explicitly, since the is implied by `EDXAPP_SETTINGS: openstack` provided you pass in Swift credentials, but we will need these settings as well:

    EDXAPP_SETTINGS: 'openstack'
    XQUEUE_SETTINGS: 'openstack_settings'
    COMMON_VHOST_ROLE_NAME: 'openstack'

mtyaka: @smarnach I don't think `EDXAPP_SETTINGS` is used anywhere before Omar introduces it in https://github.com/edx/configuration/pull/2723, so in theory I think we could set it now, and it will just be ignored until that PR gets merged. It probably makes more sense to wait until Omar's PRs get merged, though.

Even when using `EDXAPP_SETTINGS: openstack`, I believe `EDXAPP_DEFAULT_FILE_STORAGE` still defaults to "django.core.files.storage.FileSystemStorage"`, so we'll have to set it explicitly: https://github.com/edx/edx-platform/pull/11286/files#diff-8126d0664ba3411882f0ceac7d41e2e5R23

smarnach:
@antoviaque Thanks for your comments!

> For cleaning up containers - it is by design yes, but did you mean that integration tests also leave the containers behind? The tests themselves should make sure to remove them in that case, they should cleanup after themselves. It is an issue for MySQL & MongoDB too (there is actually a bug about this in the backlog, OC-1543 ), but the fact that there is existing debt elsewhere doeesn't make it ok to add to that debt on this PR for SWIFT containers...

Yes, the integration tests do leave (empty) Swift containers behind at the moment.  However, I don't think there is a robust way to fix this within from the integration tests (since they can die at any time;  we regularly leak VMs as well).  We need some external clean-up job that takes care of removing "old" MySQL databases, MongoDB databases, Swift containers and VMs, both for the production environment as well as for the integration test environment.  I felt that this is out of scope for the current task, and that adding a half-baked solution to the integration tests isn't worth the effort.

I suggest running the clean-up task on Google App Engine.  It supports scheduled jobs, it's secure and maintenance-free, and probably wouldn't cost any money either.

> Also, I don't see any change in the integration tests in the diff - I didn't look at it closely so I might have missed something, but are you checking that the integration tests correctly setup the swift containers?

So far the philosophy was to bring up the instance and see whether provisioning succeeds, and take that as an implicit confirmation that all the steps succeeded.  If the API call to provision the Swift container fails, the integration tests will fail.  If it succeeds but the container isn't created correctly anyway, chances are it wasn't our fault.

Well, since we can't test this locally (in contrast to MySQL and MongoDB provisioning), it might make sense to make an exception for the Swift containers and add a check to test they have been successfully provisioned.

> For the additional project, I'll create it - I have (again) to see this with the support, their quotas checks are broken again. For integration we can just use a different zone maybe for now? Or would it be worth creating a dedicated project there too?

I would suggest a dedicated project for the integration tests as well.  The integration tests often provision older versions of edx-platform, so they might not have the latest security patches, so we don't want credentials that can create VMs on these servers either.

instance/templates/instance/ansible/swift.yml
smarnach: @mtyaka Oh, good, that makes things a whole lot easier.  In that case, I would add `EDXAPPS_SETTINGS` now; it saves us the trobule of thinking of it later, and the only downside I can think of is that it might confuse someone, which can be addressed by a simple comment.

Regarding the file storage class, I was wrongly assuming that the code in https://github.com/open-craft/edx-platform/blob/omar/remove-s3/lms/envs/openstack.py#L22-L27 handles this for us, but since we explicitly set this in Ansible, we have to override it there as well.

antoviaque: Also, currently I don't think we use persistent DBs on any of our sandboxes, no? Since you mentioned SWIF is only activated when persitent databases are, the impact should be minimal in any case?

antoviaque:
@smarnach 

> Yes, the integration tests do leave (empty) Swift containers behind at the moment. However, I don't think there is a robust way to fix this within from the integration tests (since they can die at any time; we regularly leak VMs as well). We need some external clean-up job that takes care of removing "old" MySQL databases, MongoDB databases, Swift containers and VMs, both for the production environment as well as for the integration test environment. I felt that this is out of scope for the current task, and that adding a half-baked solution to the integration tests isn't worth the effort.

Yes, if the process just dies it's one thing, it is beyond the scope of this story. However, taking care of not leaving behind SWIFT containers after a normal integration test run isn't -- cleaning up state in `tearDown()` is part of the scope of writing a test?

> So far the philosophy was to bring up the instance and see whether provisioning succeeds, and take that as an implicit confirmation that all the steps succeeded. 

Can you point me to areas of the code where we do this? That seem very wrong - tests need to be explicit or they aren't testing much. An exception or an error would only happen in a small proportion of the cases.

> If the API call to provision the Swift container fails, the integration tests will fail. If it succeeds but the container isn't created correctly anyway, chances are it wasn't our fault.

But how do you know that the API call to provision the swift container was done in the first place, if the result isn't tested?

> I would suggest a dedicated project for the integration tests as well. The integration tests often provision older versions of edx-platform, so they might not have the latest security patches, so we don't want credentials that can create VMs on these servers either.

Yup, here that's already the case : ) The integration VMs are created on a dedicated project.

antoviaque:
@smarnach FYI, new project for production SWIFT containers created, and added to the secrets repository.

smarnach:
@antoviaque We seem to have some crossed wires here.  I really don't think we disagree that much. :) (Well, on the first point I was wrong, so there has been a disagreement there.)

> Yes, if the process just dies it's one thing, it is beyond the scope of this story. However, taking care of not leaving behind SWIFT containers after a normal integration test run isn't -- cleaning up state in `tearDown()` is part of the scope of writing a test?

I was trying to make a pragmatic decision here to save a bit of time.  The reasoning wasn't that tests don't need to clean up after themselves, but rather that we need a different mechanism to clean up anyway, so we could just as well completely rely on that mechanism.  However, I can see that this is wrong, for various reasons, including that our other mechanisms probably won't apply to our development Vagrant instances.  I've already started implementing this (and I've got a ticket for the clean-ups now anyway).

> > So far the philosophy was to bring up the instance and see whether provisioning succeeds, and take that as an implicit confirmation that all the steps succeeded. 

> Can you point me to areas of the code where we do this? That seem very wrong - tests need to be explicit or they aren't testing much. An exception or an error would only happen in a small proportion of the cases.

Just to make sure, I'm talking about the integration tests here.  The full code for the main integration tests is [these few lines](https://github.com/open-craft/opencraft/blob/14f1eeb46da78303bc836a431e289c920f0f305c/instance/tests/integration/integration_instance.py#L64-L86), so it's literally just: bring up the instance, verify it's reachable, and nothing more.  Everything more specific is in the unit tests.  I think this approach is reasonable, but as I said above I''ve implemented a check that the Swift container has been provisioned as well, as we can't really do this in the unit tests, in contrast to almost everything else.

> But how do you know that the API call to provision the swift container was done in the first place, if the result isn't tested?

The unit tests verify that this is done during provisioning, without actually performing the provisioning (since everything is mocked out).

> Yup, here that's already the case : ) The integration VMs are created on a dedicated project.

I know the integration tests are on a separate project.  What I meant is that I would prefer to push credentials to VMs that don't allow to create further VMs.  Otherwise, a break-in on one of the integration test VMs would allow an attacker to create ten more VMs and send spam from all of them at the same time.  (The risk is rather low, though, and we could simply destroy the whole project and create a new one if this happens.)

antoviaque:
> I was trying to make a pragmatic decision here to save a bit of time. The reasoning wasn't that tests don't need to clean up after themselves, but rather that we need a different mechanism to clean up anyway, so we could just as well completely rely on that mechanism. However, I can see that this is wrong, for various reasons, including that our other mechanisms probably won't apply to our development Vagrant instances. I've already started implementing this (and I've got a ticket for the clean-ups now anyway).

Sounds good!

> Just to make sure, I'm talking about the integration tests here. The full code for the main integration tests is these few lines, so it's literally just: bring up the instance, verify it's reachable, and nothing more. Everything more specific is in the unit tests. I think this approach is reasonable, but as I said above I''ve implemented a check that the Swift container has been provisioned as well, as we can't really do this in the unit tests, in contrast to almost everything else.

Yup, I was talking about the integration tests.

If you look at the first integration test, it does test the result, which is that an instance was setup and is now responding a 200 on the HTTP port: https://github.com/open-craft/opencraft/blob/14f1eeb46da78303bc836a431e289c920f0f305c/instance/tests/integration/integration_instance.py#L64-L72 . You could (rightly) argue that it could test it more precisely, but that's still a test that tests its result.

For the second integration test however, you are right, it currently doesn't test anything, beyond setting up the settings and hoping for the best. If the name of the setting was changed, it wouldn't even notice it. This is an issue imho, and I've added a bug to the backlog to fix this.

Another point that you could rightly make here is that the integration test coverage is very poor - this is due to the fact that integration tests were added late, and was only focused on the most important thing to test at the time, and which was already covering a lot of ground - the fact that a provisioning of an edxapp_sandbox playbook would complete and give a working Open edX instance. But we should try to improve this over time - not every little feature or fix requires an integration test, but a lot of them do, especially as it will help improving the integration coverage over time. It could be good to make this a deliberate/explicit decision on each PR.

>> But how do you know that the API call to provision the swift container was done in the first place, if the result isn't tested?
> The unit tests verify that this is done during provisioning, without actually performing the provisioning (since everything is mocked out).

That's the issue - with the unit tests alone, we actually don't know if the result is actually as expected. The way I see the difference between unit and integration is that unit tests are a good way to ensure interface stability, as well as reminding of the different scenarios code can be used in, ie an help/reminder during development. while integration tests aims to replace QA testers, checking that at the end, all the smaller pieces unit tests checks individually do actually provide the right result when put together. It doesn't need to be as thorough and detailed as unit tests, but it needs to ensure it tests all the features from the user perspective, and that the results are as expected. Does that make sense?

> I know the integration tests are on a separate project. What I meant is that I would prefer to push credentials to VMs that don't allow to create further VMs. Otherwise, a break-in on one of the integration test VMs would allow an attacker to create ten more VMs and send spam from all of them at the same time. (The risk is rather low, though, and we could simply destroy the whole project and create a new one if this happens.)

I might be missing something, but I don't think we create credentials that don't allow to create VMs? OVH only allows to create one user for each project, and doesn't give any control over the rights. The user from the project dedicated to host SWIFT containers can create VMs there, for example. Or did I understand it wrong?

instance/templates/instance/ansible/swift.yml
smarnach: If I do include `EDXAPP_DEFAULT_FILE_STORAGE` in the Swift config snippet, this would result in persistent databases not working until the Swift PR has been merged in edx_platform.  We could just disable Swift using the SWIFT_ENABLE switch on the instance manager for now, and see how long it takes until the Swift PR gets merged.

@antoviaque Two of the sandboxes are using persistent databases at the moment.

antoviaque: @smarnach What are the two sandboxes? Do we need them?

smarnach: They are both belonging to @omarkhan:

    PR#11871: Custom icons for SSO ... (omarkhan) - open-craft/omar/sso-icon-images (2fc7721)
    PR#11854: Allow for tooltips with ... (omarkhan) - open-craft/omar/rich-tooltips (d673de3)

The first PR is merged, the second isn't.  Anyway, even if we got rid of those, we don't want to break the persistent databases feature, do we?

smarnach: @omarkan Would it be possible to implement logic similar to https://github.com/edx/edx-platform/pull/11286/files#diff-8126d0664ba3411882f0ceac7d41e2e5R22 in edx/configuration#2723?  This would allow us not to set explicitly set `EDXAPP_DEFAULT_FILE_STORAGE`, so the code would work for edx-platform versions that do include Swift support as well as those that don't.

smarnach:
@mtyaka This is ready for another look.  I think I've addressed all review notes, but I have some doubts the integration tests will pass in this attempt, since I simply used Omar's Swift development branch for the integration tests, which probably isn't going to work.

This includes part of a fix for OC-1543 now as well, which Omar is reviewer for, but it doesn't make sense to split that out, and it's small anyway.

smarnach:
@mtyaka I wrote more code for OC-1543, and then it started to make sense to put that in a separate PR, so I've removed the code for the storage backend cleanup from this PR.

I also commented out `EDXAPP_DEFAULT_FILE_STORAGE` for now, to be able to use the default branches and have the integration tests pass.  We will have to figure out a better solution for this, maybe by implementing some logic in the configuration repo to set this automatically.

instance/templates/instance/ansible/swift.yml
antoviaque: > even if we got rid of those, we don't want to break the persistent databases feature, do we?

Nope, of course - my understanding was that we were discussing breaking it only until the SWIFT support work was merged upstream, right?

mtyaka:
Thanks for the updates :+1:

smarnach:
@antoviaque

> For the second integration test however, you are right, it currently doesn't test anything, beyond setting up the settings and hoping for the best […] If the name of the setting was changed, it wouldn't even notice it.

When considered in isolation, the test doesn't test much.  In combination with the unit tests, it does test something.  The unit tests make sure that the correct provisioning methods are called when persistent databases are enabled, and that provisioning databases works correctly, so we assume here that these parts work.  Based on this assumption, the test will verify that everything works together.

The problem with this approach is that future changes to the tests can break the implicit assumptions they are based on, since these assumptions are not explicitly stated in the code, so I agree it would be better to change it.

> I might be missing something, but I don't think we create credentials that don't allow to create VMs? OVH only allows to create one user for each project, and doesn't give any control over the rights. The user from the project dedicated to host SWIFT containers can create VMs there, for example. Or did I understand it wrong?

I don't have access to the OVH control panel, so I've got no idea what settings you can make.  I was hoping you were able to set project quotas in a way that you don't allow any VMs at all for a project, and then we use that project for Swift only.  If that's not possible anyway, there is no point in creating a separate project just for Swift containers in the integration tests.

instance/templates/instance/ansible/swift.yml
omarkhan: @smarnach I replied to your question on the JIRA ticket but I'll include it here as well:

It *should* be possible to set the default value of `DEFAULT_FILE_STORAGE` in the configuration repo to null. This should cause both the `aws.py` and the `openstack.py` settings files to use the correct storage backend as a fallback.

The downside is that for deployments that use the default settings and include AWS credentials, this will change the storage backend from the file system to S3. This will break backwards compatibility, and upstream may not be happy with it. Since upstream have already taken ages to review the PR I'm reluctant to add any more breaking changes to it.

What we can do for the instance manager is explicitly set `DEFAULT_FILE_STORAGE` to null. Would that work?

smarnach: @omarkhan Thanks for the suggestion.  This sounds like it will work.  I will test it tomorrow.

OmarKan: Dear Marnach,

I have been mentioned in this forum due to a spelling mistake, please stop
mentioning me :)

Thank you.

PR #57, opened 2016-04-21
Separate server and instance states

itsjeyd:
Follows up on #51.

cf. [OC-1540](https://tasks.opencraft.com/browse/OC-1540)

### DIFF
--- a/instance/migrations/0040_auto_20160420_0754.py
+++ b/instance/migrations/0040_auto_20160420_0754.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0039_auto_20160419_0921'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='singlevmopenedxinstance',
+            name='status',
+            field=models.CharField(choices=[('terminated', 'Terminated'), ('configuring', 'ConfiguringServer'), ('failed', 'ConfigurationFailed'), ('waiting', 'WaitingForServer'), ('new', 'New'), ('running', 'Running'), ('error', 'Error')], default='new', max_length=20, db_index=True)
+        )
+    ]
--- a/instance/migrations/0040_auto_20160420_0754.py
+++ b/instance/migrations/0040_auto_20160420_0754.py
@@ -4,6 +4,27 @@ from __future__ import unicode_literals
 from django.db import migrations, models
 
 
+def set_instance_states(apps, schema_editor):
+    SingleVMOpenEdXInstance = apps.get_model("instance", "SingleVMOpenEdXInstance")
+    for instance in SingleVMOpenEdXInstance.objects.iterator():
+        server_status = instance.server_status
+        if server_status:
+            if server_status.state_id in ("new", "started", "active", "rebooting"):
+                instance._status_to_waiting_for_server()
+            elif server_status.state_id in ("booted", "ready"):
+                instance._status_to_waiting_for_server()
+                instance._status_to_configuring_server()
+                instance._status_to_running()
+            elif server_status.state_id == "provisioning":
+                instance._status_to_waiting_for_server()
+                instance._status_to_configuring_server()
+            elif server_status.state_id == "terminated":
+                instance._status_to_waiting_for_server()
+                instance._status_to_configuring_server()
+                instance._status_to_running()
+                instance._status_to_terminated()
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -15,5 +36,6 @@ class Migration(migrations.Migration):
             model_name='singlevmopenedxinstance',
             name='status',
             field=models.CharField(choices=[('terminated', 'Terminated'), ('configuring', 'ConfiguringServer'), ('failed', 'ConfigurationFailed'), ('waiting', 'WaitingForServer'), ('new', 'New'), ('running', 'Running'), ('error', 'Error')], default='new', max_length=20, db_index=True)
-        )
+        ),
+        migrations.RunPython(set_instance_states, migrations.RunPython.noop),
     ]
--- a/instance/migrations/0040_auto_20160420_0754.py
+++ b/instance/migrations/0040_auto_20160420_0754.py
@@ -3,26 +3,27 @@ from __future__ import unicode_literals
 
 from django.db import migrations, models
 
+from instance.models.instance import Status as InstanceStatus
+
 
 def set_instance_states(apps, schema_editor):
     SingleVMOpenEdXInstance = apps.get_model("instance", "SingleVMOpenEdXInstance")
-    for instance in SingleVMOpenEdXInstance.objects.iterator():
-        server_status = instance.server_status
-        if server_status:
-            if server_status.state_id in ("new", "started", "active", "rebooting"):
-                instance._status_to_waiting_for_server()
-            elif server_status.state_id in ("booted", "ready"):
-                instance._status_to_waiting_for_server()
-                instance._status_to_configuring_server()
-                instance._status_to_running()
-            elif server_status.state_id == "provisioning":
-                instance._status_to_waiting_for_server()
-                instance._status_to_configuring_server()
-            elif server_status.state_id == "terminated":
-                instance._status_to_waiting_for_server()
-                instance._status_to_configuring_server()
-                instance._status_to_running()
-                instance._status_to_terminated()
+    SingleVMOpenEdXInstance.objects.filter(
+        server_set__status__in=("new", "started", "active", "rebooting")).update(
+            _status=InstanceStatus.WaitingForServer.state_id
+        )
+    SingleVMOpenEdXInstance.objects.filter(
+        server_set__status__in=("booted", "ready")).update(
+            _status=InstanceStatus.Running.state_id
+        )
+    SingleVMOpenEdXInstance.objects.filter(
+        server_set__status="provisioning").update(
+            _status=InstanceStatus.ConfiguringServer.state_id
+        )
+    SingleVMOpenEdXInstance.objects.filter(
+        server_set__status="terminated").update(
+            _status=InstanceStatus.Terminated.state_id
+        )
 
 
 class Migration(migrations.Migration):
@@ -34,8 +35,8 @@ class Migration(migrations.Migration):
     operations = [
         migrations.AddField(
             model_name='singlevmopenedxinstance',
-            name='status',
-            field=models.CharField(choices=[('terminated', 'Terminated'), ('configuring', 'ConfiguringServer'), ('failed', 'ConfigurationFailed'), ('waiting', 'WaitingForServer'), ('new', 'New'), ('running', 'Running'), ('error', 'Error')], default='new', max_length=20, db_index=True)
+            name='_status',
+            field=models.CharField(choices=[('terminated', 'Terminated'), ('configuring', 'ConfiguringServer'), ('failed', 'ConfigurationFailed'), ('waiting', 'WaitingForServer'), ('new', 'New'), ('running', 'Running'), ('error', 'Error')], db_column='status', default='new', max_length=20, db_index=True)
         ),
         migrations.RunPython(set_instance_states, migrations.RunPython.noop),
     ]
--- a/instance/migrations/0043_change_server_statuses.py
+++ b/instance/migrations/0043_change_server_statuses.py
@@ -7,14 +7,14 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('instance', '0040_auto_20160420_0754'),
+        ('instance', '0042_add_instance_status'),
     ]
 
     operations = [
         migrations.AlterField(
             model_name='openstackserver',
-            name='status',
-            field=models.CharField(choices=[('pending', 'Pending'), ('building', 'Building'), ('booting', 'Booting'), ('ready', 'Ready'), ('terminated', 'Terminated'), ('unknown', 'Unknown'), ('failed', 'BuildFailed')], max_length=20, db_index=True, default='pending'),
+            name='_status',
+            field=models.CharField(choices=[('booting', 'Booting'), ('building', 'Building'), ('failed', 'BuildFailed'), ('pending', 'Pending'), ('ready', 'Ready'), ('terminated', 'Terminated'), ('unknown', 'Unknown')], default='pending', db_column='status', max_length=20, db_index=True),
         ),
         migrations.RunSQL(
             [
--- a/instance/migrations/0043_change_server_statuses.py
+++ b/instance/migrations/0043_change_server_statuses.py
@@ -19,6 +19,7 @@ class Migration(migrations.Migration):
         migrations.RunSQL(
             [
                 "UPDATE instance_openstackserver SET status = 'pending' WHERE status = 'new'",
+                "UPDATE instance_openstackserver SET status = 'failed' WHERE status = 'started' AND progress = 'failed'",
                 "UPDATE instance_openstackserver SET status = 'building' WHERE status = 'started'",
                 "UPDATE instance_openstackserver SET status = 'booting' WHERE status = 'active' OR status = 'rebooting'",
                 "UPDATE instance_openstackserver SET status = 'ready' WHERE status = 'booted' OR status = 'provisioning'",
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -214,16 +214,6 @@ class Instance(ValidateModelMixin, TimeStampedModel):
         return None
 
     @property
-    def progress(self):
-        """
-        Instance's current status progress
-        """
-        server = self._current_server
-        if server:
-            return server.progress
-        return None
-
-    @property
     def event_context(self):
         """
         Context dictionary to include in events
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -89,8 +89,7 @@ class InstanceState(ResourceState):
     is_steady_state = True
 
     # An instance is healthy if it is part of a normal (expected) workflow.
-    # This information can be used to delay execution of an operation
-    # until the target server has reached a healthy state.
+    # This information can be used to detect problems and highlight them in the UI or notify users.
     # Healthy states include:
     # - Status.New
     # - Status.WaitingForServer
@@ -112,11 +111,13 @@ class Status(ResourceState.Enum):
     class WaitingForServer(InstanceState):
         """ Server not yet accessible """
         state_id = 'waiting'
+        name = 'Waiting for server'
         is_steady_state = False
 
     class ConfiguringServer(InstanceState):
         """ Running Ansible playbooks on server """
         state_id = 'configuring'
+        name = 'Configuring server'
         is_steady_state = False
 
     class Running(InstanceState):
@@ -126,6 +127,7 @@ class Status(ResourceState.Enum):
     class ConfigurationFailed(InstanceState):
         """ Instance was not configured successfully (but may be partially online) """
         state_id = 'failed'
+        name = 'Configuration failed'
         is_healthy_state = False
 
     class Error(InstanceState):
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -558,7 +558,8 @@ class SingleVMOpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftCon
                 'Provision attempt {attempt} of {attempts}'.format(attempt=attempt_num, attempts=self.attempts)
             )
             server, deploy_log, provisioned = self._provision_attempt()
-            logs.extend(deploy_log)
+            if deploy_log is not None:  # If server fails to build, no deployment logs will be available
+                logs.extend(deploy_log)
             attempt_num += 1
 
         return (server, logs)
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -75,6 +75,18 @@ class InstanceState(ResourceState):
     """
     A [finite state machine] state describing an instance.
     """
+    # A state is "steady" if we don't expect it to change.
+    # This information can be used to:
+    # - delay execution of an operation until the target server reaches a steady state
+    # - raise an exception when trying to schedule an operation that depends on a state change
+    #   while the target server is in a steady state
+    # Steady states include: Status.Running, Status.ConfigurationFailed, Status.Error, Status.Terminated
+    is_steady_state = False
+
+    # An instance is healthy if it has Status.Running.
+    # This information can be used to delay execution of an operation
+    # until the target instance has reached this status.
+    is_healthy_state = False
 
 
 class Status(ResourceState.Enum):
@@ -97,18 +109,23 @@ class Status(ResourceState.Enum):
     class Running(InstanceState):
         """ Instance is up and running """
         state_id = 'running'
+        is_steady_state = True
+        is_healthy_state = True
 
     class ConfigurationFailed(InstanceState):
         """ Instance was not configured successfully (but may be partially online) """
         state_id = 'failed'
+        is_steady_state = True
 
     class Error(InstanceState):
         """ Instance never got up and running (something went wrong when trying to build new VM) """
         state_id = 'error'
+        is_steady_state = True
 
     class Terminated(InstanceState):
         """ Instance was running successfully and has been shut down """
         state_id = 'terminated'
+        is_steady_state = True
 
 
 # Models ######################################################################
@@ -221,15 +238,6 @@ class Instance(ValidateModelMixin, TimeStampedModel):
         Context dictionary to include in events
         """
         return {'instance_id': self.pk}
-
-    def _transition(self, state_transition):
-        """
-        Helper method to update the state of this instance.
-
-        Ensures that state is in sync with the Django DB.
-        """
-        state_transition()
-        self.save()
 
     def save(self, **kwargs):
         """
@@ -546,7 +554,7 @@ class SingleVMOpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftCon
 
         # Instance
         self.logger.info('Prepare for (re-)provisioning')
-        self._transition(self._status_to_waiting_for_server)
+        self._status_to_waiting_for_server()
 
         # Server
         self.logger.info('Terminate servers')
@@ -558,7 +566,7 @@ class SingleVMOpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftCon
         try:
             server.sleep_until(lambda: server.status.vm_available)
         except SteadyStateException:
-            self._transition(self._status_to_error)
+            self._status_to_error()
             return (server, None, False)  # No deploy logs available yet; instance has not been provisioned
 
         def accepts_ssh_commands():
@@ -584,22 +592,24 @@ class SingleVMOpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftCon
                 self.provision_swift()
 
             # Provisioning (ansible)
-            self.mark_as_provisioning()
+            self.logger.info('Provisioning server...')
+            self._status_to_configuring_server()
             self.reset_ansible_settings(commit=True)
             deploy_log, exit_code = self.deploy()
             if exit_code != 0:
-                self.mark_provisioning_finished(success=False)
+                self.logger.info('Provisioning failed')
+                self._status_to_configuration_failed()
                 self.provision_failed_email(self.ProvisionMessages.PROVISION_ERROR, deploy_log)
                 return (server, deploy_log, False)
 
             # Reboot
+            self.logger.info('Provisioning completed')
             self.logger.info('Rebooting server %s...', server)
             server.reboot()
             server.sleep_until(accepts_ssh_commands)
-            self.logger.info('Provisioning completed')
 
             # Declare instance up and running
-            self.mark_provisioning_finished(success=True)
+            self._status_to_running()
 
             return (server, deploy_log, True)
 
@@ -607,23 +617,6 @@ class SingleVMOpenEdXInstance(MySQLInstanceMixin, MongoDBInstanceMixin, SwiftCon
             self.server_set.terminate()
             self.provision_failed_email(self.ProvisionMessages.PROVISION_EXCEPTION)
             raise
-
-    @Instance.status.only_for(Status.WaitingForServer)
-    def mark_as_provisioning(self):
-        """
-        Indicate that this instance is being provisioned.
-        """
-        self._transition(self._status_to_configuring_server)
-
-    @Instance.status.only_for(Status.ConfiguringServer)
-    def mark_provisioning_finished(self, success):
-        """
-        Indicate that this instance is done provisioning, either due to success or failure.
-        """
-        if success:
-            self._transition(self._status_to_running)
-        else:
-            self._transition(self._status_to_configuration_failed)
 
     def provision_mysql(self):
         """
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -160,19 +160,26 @@ class Instance(ValidateModelMixin, TimeStampedModel):
     )
     # State transitions:
     _status_to_waiting_for_server = status.transition(
-        from_states=Status.New, to_state=Status.WaitingForServer
-    )
-    _status_to_configuring_server = status.transition(
-        from_states=Status.WaitingForServer, to_state=Status.ConfiguringServer
+        from_states=(
+            Status.New,
+            Status.Error,
+            Status.ConfigurationFailed,
+            Status.Running,
+            Status.Terminated
+        ),
+        to_state=Status.WaitingForServer
     )
     _status_to_error = status.transition(
         from_states=Status.WaitingForServer, to_state=Status.Error
     )
-    _status_to_running = status.transition(
-        from_states=Status.ConfiguringServer, to_state=Status.Running
+    _status_to_configuring_server = status.transition(
+        from_states=Status.WaitingForServer, to_state=Status.ConfiguringServer
     )
     _status_to_configuration_failed = status.transition(
         from_states=Status.ConfiguringServer, to_state=Status.ConfigurationFailed
+    )
+    _status_to_running = status.transition(
+        from_states=Status.ConfiguringServer, to_state=Status.Running
     )
     _status_to_terminated = status.transition(
         from_states=Status.Running, to_state=Status.Terminated
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -41,7 +41,7 @@ from instance.models.mixins.ansible import AnsibleInstanceMixin
 from instance.models.mixins.database import MongoDBInstanceMixin, MySQLInstanceMixin, SwiftContainerInstanceMixin
 from instance.models.mixins.utilities import EmailInstanceMixin
 from instance.models.mixins.version_control import GitHubInstanceMixin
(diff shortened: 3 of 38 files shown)


instance/migrations/0040_auto_20160420_0754.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('instance', '0039_auto_20160419_0921'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='singlevmopenedxinstance',
+            name='status',
+            field=models.CharField(choices=[('terminated', 'Terminated'), ('configuring', 'ConfiguringServer'), ('failed', 'ConfigurationFailed'), ('waiting', 'WaitingForServer'), ('new', 'New'), ('running', 'Running'), ('error', 'Error')], default='new', max_length=20, db_index=True)
```
bradenmacdonald: Do we also need a data migration to set the status to something other than `new` for existing instances?

instance/models/instance.py
```diff
@@ -140,22 +216,21 @@ def status(self):
...
         return {'instance_id': self.pk}
 
+    def _transition(self, state_transition):
+        """
+        Helper method to update the state of this instance.
+
+        Ensures that state is in sync with the Django DB.
+        """
+        state_transition()
+        self.save()
```
bradenmacdonald: Hmm, maybe we should just move this logic to `ModelResourceStateDescriptor`.

You could put it into `_set_state` as follows:

```python
def _set_state(self, resource, new_state_class):
    ...
    # Save changes to this one field only:
    resource.save(update_fields=[self.model_field_name]) 
    return new_state
```

Then you can remove the `_transition` method from Instance and Server, and we can just call the transition methods (like self._status_to_waiting_for_server()) directly.

instance/models/instance.py
```diff
@@ -446,29 +531,47 @@ def accepts_ssh_commands():
...
+
+    @Instance.status.only_for(Status.ConfiguringServer)
+    def mark_provisioning_finished(self, success):
+        """
+        Indicate that this instance is done provisioning, either due to success or failure.
+        """
+        if success:
+            self._transition(self._status_to_running)
+        else:
+            self._transition(self._status_to_configuration_failed)
```
bradenmacdonald: I think you can get rid of these methods (`mark_as_provisioning` and `mark_provisioning_finished`). The transitions already check that the current state is correct, so the `only_for` is redundant.

Instead of `self.mark_provisioning_finished(success=True)` you can just call `self._status_to_running()` directly (if you also remove `_transition` as I suggested).

instance/models/server.py
```diff
@@ -61,85 +61,55 @@ class ServerState(ResourceState):
...
     """
-    class New(ServerState):
-        """ Not yet loaded """
-        state_id = 'new'
-
-    class Started(ServerState):
-        """ Running but not active yet """
-        state_id = 'started'
+    class Pending(ServerState):
+        """ Not yet loaded (need to request new VM from OpenStack API) """
```
bradenmacdonald: Nit: technically this class is not supposed to be specific to OpenStack.

instance/models/server.py
```diff
@@ -419,30 +324,35 @@ def start(self):
...
+        self._transition(self._status_to_building)
+        try:
+            os_server = openstack.create_server(
+                self.nova,
+                self.instance.sub_domain,
+                settings.OPENSTACK_SANDBOX_FLAVOR,
+                settings.OPENSTACK_SANDBOX_BASE_IMAGE,
+                key_name=settings.OPENSTACK_SANDBOX_SSH_KEYNAME,
+            )
+        except novaclient.exceptions.ClientException:
```
bradenmacdonald: Would this cover "quota limitations & waiting list" as mentioned in the docstring above? If so, it's a partial implementation of that TODO item :)

instance/models/utils.py
```diff
@@ -358,7 +355,7 @@ def model_field_choices(self):
 
         Suitable for passing to a django CharField choices parameter.
         """
-        return tuple((state.state_id, state.__class__.__name__) for state in self.state_classes)
+        return tuple((state.state_id, state.__name__) for state in self.state_classes)
```
bradenmacdonald: Looks like this was missing test coverage - my bad :-/

instance/serializers/instance.py
```diff
@@ -71,23 +71,23 @@ class Meta:
     def to_representation(self, obj):
         output = super().to_representation(obj)
         # Convert the state values from objects to strings:
+        if output['server_status'] is None:
+            output['server_status'] = 'empty'  # 'empty' for backwards compatibility
+        else:
+            output['server_status'] = obj.server_status.state_id
         if output['status'] is None:
```
bradenmacdonald: When would this `status` be `None`? I don't think it should ever be.

instance/serializers/instance.py
```diff
@@ -71,23 +71,23 @@ class Meta:
     def to_representation(self, obj):
         output = super().to_representation(obj)
         # Convert the state values from objects to strings:
+        if output['server_status'] is None:
+            output['server_status'] = 'empty'  # 'empty' for backwards compatibility
```
bradenmacdonald: If we're changing the key name, we could also change the value here - unless `empty` is more convenient than `null` for the client-side code.

instance/static/html/instance/index.html
```diff
@@ -27,10 +27,12 @@
...
-        <div class="status-icon" ng-switch="instance.progress">
-          <i class="fa fa-ellipsis-h" tooltip-placement="top" tooltip="In progress" ng-switch-when="running"></i>
-          <i class="fa fa-times" tooltip-placement="top" tooltip="Failed" ng-switch-when="failed"></i>
-          <i class="fa fa-check" tooltip-placement="top" tooltip="Success" ng-switch-when="success"></i>
+        <div class="status-icon" ng-switch="instance.status">
+          <i class="fa fa-ellipsis-h" tooltip-placement="top" tooltip="Waiting for server" ng-switch-when="waiting"></i>
+          <i class="fa fa-ellipsis-h" tooltip-placement="top" tooltip="Configuring server" ng-switch-when="configuring"></i>
+          <i class="fa fa-times" tooltip-placement="top" tooltip="Error" ng-switch-when="error"></i>
+          <i class="fa fa-times" tooltip-placement="top" tooltip="Configuration Failed" ng-switch-when="failed"></i>
+          <i class="fa fa-check" tooltip-placement="top" tooltip="Running" ng-switch-when="running"></i>
```
bradenmacdonald: Could we instead do this generically and re-use the existing `status.description` attribute (which defaults to the docstring):

```
<div class="status-icon">
    <i ng-attr-class="{{ status.is_steady && (status.is_healthy && 'fa fa-check' || 'fa fa-times') || 'fa fa-ellipsis-h' }}"
        tooltip-placement="top"
        ng-attr-tooltip="{{status.description}}">
    </i>
</div>
```

instance/static/html/instance/index.html
```diff
@@ -49,9 +51,10 @@
                 title="{{ server.status_description }}">
             {{ server.status_name }}
           </span>
-          <span ng-switch="selected.instance.progress">
-            <i class="fa fa-times" tooltip-placement="top" tooltip="Failed" ng-switch-when="failed"></i>
-            <i class="fa fa-check" tooltip-placement="top" tooltip="Success" ng-switch-when="success"></i>
+          <span ng-switch="selected.instance.status">
+            <i class="fa fa-times" tooltip-placement="top" tooltip="Error" ng-switch-when="error"></i>
+            <i class="fa fa-times" tooltip-placement="top" tooltip="Configuration Failed" ng-switch-when="failed"></i>
+            <i class="fa fa-check" tooltip-placement="top" tooltip="Running" ng-switch-when="running"></i>
```
bradenmacdonald: Same comment about making this generic.

instance/static/html/instance/index.html
```diff
@@ -73,12 +76,20 @@
...
-            <div class="status-icon" ng-if="selected.instance.progress == 'running'">
+            <div class="status-icon" ng-if="selected.instance.status == 'running'">
+              <button class="button reprovision" ng-click="provision(selected.instance)">
+                Provision
+              </button>
+            </div>
+            <div class="status-icon" ng-if="selected.instance.status == 'waiting'">
+              <img src="/static/img/running.png" tooltip-placement="bottom" tooltip="Waiting for server..." alt="Waiting for server..."/>
+            </div>
+            <div class="status-icon" ng-if="selected.instance.status == 'configuring'">
```
bradenmacdonald: Would be nicer to do:
```
<div ng-if="selected.instance.status.supports_reprovisioning" ng-attr-class="status-icon status-{{selected.instance.status.state_id}}">
    <button>...</button>
</div>
<div ng-if=" ! selected.instance.status.supports_reprovisioning" ng-attr-class="status-icon status-{{selected.instance.status.state_id}}">
    <span ng-attr-title="{{selected.instance.status.description}}">{{ selected.instance.status.name }}</span>
</div>
```

instance/serializers/instance.py
itsjeyd: @bradenmacdonald I checked the client-side code and it turns out `server_status` wasn't being used in a meaningful way, so I completely removed it from the instance serializer.

instance/models/server.py
itsjeyd: @bradenmacdonald I'm not sure :) Based on [this](http://docs.openstack.org/developer/python-novaclient/ref/exceptions.html), there doesn't seem to be a dedicated exception for quota limits.

instance/migrations/0040_auto_20160420_0754.py
```diff
@@ -4,6 +4,27 @@
 from django.db import migrations, models
 
 
+def set_instance_states(apps, schema_editor):
+    SingleVMOpenEdXInstance = apps.get_model("instance", "SingleVMOpenEdXInstance")
+    for instance in SingleVMOpenEdXInstance.objects.iterator():
+        server_status = instance.server_status
```
bradenmacdonald: This is now giving `AttributeError: 'SingleVMOpenEdXInstance' object has no attribute 'server_status'`

instance/migrations/0040_auto_20160420_0754.py
```diff
@@ -4,6 +4,27 @@
...
+                instance._status_to_configuring_server()
+                instance._status_to_running()
+            elif server_status.state_id == "provisioning":
+                instance._status_to_waiting_for_server()
+                instance._status_to_configuring_server()
+            elif server_status.state_id == "terminated":
+                instance._status_to_waiting_for_server()
+                instance._status_to_configuring_server()
+                instance._status_to_running()
+                instance._status_to_terminated()
```
bradenmacdonald: A much more robust way to write this sort of migration would be:

```python
SingleVMOpenEdXInstance.objects.filter(server_set___status__in=("new", "started", "active", "rebooting")).update(_status=SingleVMOpenEdXInstance.Status.WaitingForServer.state_id)
SingleVMOpenEdXInstance.objects.filter(server_set___status__in=("booted", "ready")).update(_status=SingleVMOpenEdXInstance.Status.Running.state_id)
...
```

I believe that in the code now, `server_set` should only ever return one server. However, In the event that `server_set` contains multiple servers, the instance will be returned separately for each server, and this would potentially use the "wrong" server for the final status. Shouldn't happen though, so I'm not worried about that.

instance/models/instance.py
```diff
@@ -75,6 +75,18 @@ class InstanceState(ResourceState):
...
+    # - delay execution of an operation until the target server reaches a steady state
+    # - raise an exception when trying to schedule an operation that depends on a state change
+    #   while the target server is in a steady state
+    # Steady states include: Status.Running, Status.ConfigurationFailed, Status.Error, Status.Terminated
+    is_steady_state = False
+
+    # An instance is healthy if it has Status.Running.
+    # This information can be used to delay execution of an operation
+    # until the target instance has reached this status.
+    is_healthy_state = False
```
bradenmacdonald: I think that New, WaitingForServer, ConfiguringServer, and Running should all be considered healthy states, since they're all part of a normal/expected workflow. "Unhealthy" should mean a state that is not expected and indicates that there is a problem.

In other words, we should be able to use this in the UI to highlight error states when is_healthy_state is false, and we don't want to highlight a "ConfiguringServer" instance in red as though there's an error.

instance/tests/models/test_utils.py
```diff
@@ -486,3 +486,35 @@ class DjangoResourceTest(SimpleResourceTestCase):
     Run the same tests as in SimpleResourceTestCase, but using DjangoResource.
     """
     make_resource = DjangoResource
+
+    def setUp(self):
+        self.make_resource.save = Mock()
```
bradenmacdonald: This will permanently alter the `save` method of `DjangoResource` and could interfere with other tests. The correct way to mock this in the `setUp` method is to use `patch()`, since it will undo the change when the test ends:

```python
def setUp(self):
    super().setUp()
    patcher = patch.object(self.make_resource, 'save')
    patcher.start()
    self.addCleanup(patcher.stop)
```

instance/tests/models/test_utils.py
```diff
@@ -486,3 +486,35 @@ class DjangoResourceTest(SimpleResourceTestCase):
...
+        """
+        Test that ModelResourceStateDescriptor produces a sensible set of field choices.
+        """
+        model_field_choices = set(self.make_resource.state.model_field_choices)
+        expected_model_field_choices = set((
+            ('state1', 'State1'),
+            ('state2', 'State2'),
+            ('state3', 'State3'),
+        ))
+        self.assertEqual(model_field_choices, expected_model_field_choices)
```
bradenmacdonald: When you rebase the code you might want to tweak this test, since I think Sven has made the result ordered in #53, and we should test for that.

bradenmacdonald:
One **optional** comment: Personally I don't think it's very helpful to have a ton of migrations with names like "0040_auto_20160420_0754". If you get a chance, it would be nice if you could give the migrations friendly names like "0040_instance_states".

instance/migrations/0040_auto_20160420_0754.py
```diff
@@ -0,0 +1,41 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0039_auto_20160419_0921'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='singlevmopenedxinstance',
+            name='status',
```
bradenmacdonald: Shouldn't this be `name='_status'` and shouldn't there be `db_column='status'` on the line below?

instance/models/instance.py
```diff
@@ -66,13 +68,104 @@ class InconsistentInstanceState(Exception):
...
+class InstanceState(ResourceState):
+    """
+    A [finite state machine] state describing an instance.
+    """
+    # A state is "steady" if we don't expect it to change.
+    # This information can be used to:
+    # - delay execution of an operation until the target server reaches a steady state
+    # - raise an exception when trying to schedule an operation that depends on a state change
+    #   while the target server is in a steady state
+    # Steady states include: Status.Running, Status.ConfigurationFailed, Status.Error, Status.Terminated
```
bradenmacdonald: Shouldn't `New` also be a steady state? Because I can instantiate an instance, and it will have status `New`, but it will stay in that status forever until we call `provision_instance()`.

This would also fix an issue in the UI, where if you create an instance but don't call `provision_instance`, there is no option to provision it seen in the UI, and it looks like something is in progress, when in reality nothing is happening.

instance/static/html/instance/index.html
```diff
@@ -49,9 +50,11 @@
                 title="{{ server.status_description }}">
             {{ server.status_name }}
```
bradenmacdonald: I don't think we need to display the server status and the instance status separately here. And I like the way the server status appears now more than the icon used for the instance status.

So maybe we could combine these?

```
      <h3>
        {{ selected.instance.name }}
        <span class="status radius label"
          ng-class="{warning: !selected.instance.status.is_healthy, success: selected.instance.status.is_healthy}"
          tooltip="{{ selected.instance.status_description }}">
            <i ng-attr-class="{{ selected.instance.is_steady && (selected.instance.is_healthy && 'fa fa-check' || 'fa fa-times') || 'fa fa-ellipsis-h' }}"></i>
            {{selected.instance.status_name}}
        </span>
      </h3>
```

And then below in the `UL.instance-description`, we could add a new `<li>` that lists each server and states its IP and status.

```
              <li class="instance-server-list">
                Servers: 
                <span ng-repeat="server in selected.instance.active_server_set">
                  vm{{server.id}} ({{server.status_name}})
                </span>
              </li>
```

End result:
screen shot 2016-04-24 at 3 08 59 pm

Which is much nicer than:
screen shot 2016-04-24 at 3 10 26 pm

instance/tests/models/test_utils.py
itsjeyd: @bradenmacdonald Hm, using `patch.object` doesn't work because `DjangoResource` doesn't define a `save` method itself. Now that `ModelResourceStateDescriptor._set_state` is responsible for persisting the current status of a resource, most of the tests that `DjangoResourceTest` inherits from `SimpleResourceTestCase` depend on the `save` method being available, but they are agnostic to what it actually does. So we could add a no-op `save` method to `DjangoResource` and patch it in `DjangoResourceTest` as you suggest, but this would be kind of redundant. Alternatively we could keep the existing implementation -- that shouldn't do any harm since `DjangoResource` is only being used by `DjangoResourceTest`.

bradenmacdonald: @itsjeyd  Oh, I forgot that DjangoResource was just defined in this file. Right. In that case, I think you could just add `save = Mock()` right in the definition of `DjangoResource`, rather than patching it or modifying it here.

itsjeyd: @bradenmacdonald 

> I think you could just add save = Mock() right in the definition of DjangoResource, rather than patching it or modifying it here.

This makes testing call counts for `save` in `test_mutator` a bit inconvenient: That test currently relies on having a fresh mock available, and it would be kind of awkward to use a "global" mock for the remaining tests and a separate one for `test_mutator` (or mess with the global mock directly in `test_mutator`). I kept the current approach for now.

bradenmacdonald:
@itsjeyd There is an integration test failure:

```
F
FAIL: test_external_databases (instance.tests.integration.integration_instance.InstanceIntegrationTestCase)
  File "/home/ubuntu/opencraft/instance/tests/integration/integration_instance.py", line 103, in test_external_databases
  File "/home/ubuntu/opencraft/instance/tests/integration/integration_instance.py", line 52, in assert_instance_up
    self.assertEqual(instance.status, InstanceStatus.Running)
AssertionError: <instance.models.instance.Status.ConfigurationFailed object at 0x2ac15f743390> != <class 'instance.models.instance.Status.Running'>

Ran 1 test in 4700.507s
```

instance/models/instance.py
```diff
@@ -80,13 +80,24 @@ class InstanceState(ResourceState):
...
+    # - Status.ConfigurationFailed
+    # - Status.Error
+    # - Status.Terminated
+    is_steady_state = True
+
+    # An instance is healthy if it is part of a normal (expected) workflow.
     # This information can be used to delay execution of an operation
-    # until the target instance has reached this status.
-    is_healthy_state = False
+    # until the target server has reached a healthy state.
```
bradenmacdonald: Nit: I would say that is_healthy_state is more useful as a way "to detect problems and highlight them in the UI or notify users"

instance/tests/api/test_instance.py
```diff
@@ -28,9 +28,12 @@
 
 from instance import github
 from instance.models.instance import SingleVMOpenEdXInstance
-from instance.models.server import OpenStackServer
 from instance.tests.api.base import APITestCase
-from instance.tests.models.factories.instance import SingleVMOpenEdXInstanceFactory
+from instance.tests.models.factories.instance import (
+    ConfiguringSingleVMOpenEdXInstanceFactory, ConfigFailedSingleVMOpenEdXInstanceFactory,
+    ErrorSingleVMOpenEdXInstanceFactory, RunningSingleVMOpenEdXInstanceFactory,
+    SingleVMOpenEdXInstanceFactory, TerminatedSingleVMOpenEdXInstanceFactory, WaitingSingleVMOpenEdXInstanceFactory
```
bradenmacdonald: Nit: with a big list of long ugly imported names like this, it's nicer to put one name per line.

bradenmacdonald: Optional: Also, it would arguably be nicer to just import `SingleVMOpenEdXInstanceFactory` and `Status` and then use things like `SingleVMOpenEdXInstanceFactory(status=Status.Configuring)` instead of defining a separate class for each status.

instance/tests/api/test_instance.py
```diff
@@ -82,22 +85,20 @@ def test_get_domain(self):
...
         self.assertIn(('studio_url', 'http://studio.domain.api.example.com/'), instance_data)
 
-    def test_provision_waiting_for_server(self):
+    def test_provision_not_allowed(self):
         """
-        POST /:id/provision - Instance is waiting for server
+        POST /:id/provision - Instance is waiting for server or configuring server
         """
+        instance_factories = (
+            WaitingSingleVMOpenEdXInstanceFactory, ConfiguringSingleVMOpenEdXInstanceFactory
```
bradenmacdonald: Nit: better to use `ddt`. Though you'd need to add it. I'm surprised, but despite the fact that OpenCraft IM has 110 requirements listed in requirements.txt it is not currently using `ddt`.

instance/tests/api/test_instance.py
```diff
@@ -114,22 +115,27 @@ def test_provision_github_branch_deleted(self, mock_get_commit_id_from_ref):
     @patch('instance.api.instance.provision_instance')
     def test_provision(self, mock_provision_instance, mock_get_commit_id_from_ref):
         """
-        POST /:id/provision
+        POST /:id/provision - Instance is in a state that allows provisioning
         """
+        instance_factories = (
+            SingleVMOpenEdXInstanceFactory, RunningSingleVMOpenEdXInstanceFactory,
+            ErrorSingleVMOpenEdXInstanceFactory, ConfigFailedSingleVMOpenEdXInstanceFactory,
+            TerminatedSingleVMOpenEdXInstanceFactory
```
bradenmacdonald: Nit: ddt would be better, and this could be a list of status values passed to `SingleVMOpenEdXInstanceFactory` if you wanted.

instance/tests/models/test_instance.py
```diff
@@ -961,6 +996,25 @@ def test_provision(self, mocks):
...
+        self.assertIsNone(instance.server_status)
+
+        mocks.mock_create_server.side_effect = novaclient.exceptions.ClientException(400)
+        instance.provision()
+
+        self.assertEqual(instance.status, Instance.Status.Error)
+        self.assertEqual(instance.server_status, Server.Status.BuildFailed)
+        # If server fails to build, provisioning sequence never gets to the point
+        # where it starts interacting with the server:
+        mocks.mock_set_dns_record.assert_not_called()
```
bradenmacdonald: Nice - happy to see this functionality and this test :)

instance/tests/models/test_server.py
```diff
@@ -33,7 +33,8 @@
 from instance.tests.base import AnyStringMatching, TestCase
 from instance.tests.models.factories.instance import SingleVMOpenEdXInstanceFactory
 from instance.tests.models.factories.server import (
-    OpenStackServerFactory, BuildingOpenStackServerFactory, ReadyOpenStackServerFactory
+    OpenStackServerFactory, BootingOpenStackServerFactory, BuildingOpenStackServerFactory,
+    BuildFailedOpenStackServerFactory, ReadyOpenStackServerFactory, TerminatedOpenStackServerFactory
```
bradenmacdonald: nit: better to have one import per line at this point.

instance/tests/models/test_server.py
```diff
@@ -158,10 +176,23 @@ def test_sleep_until_state_changes(self, mock_sleep, mock_update_status):
...
+                'check_function': lambda: server.status.accepts_ssh_commands,
+                'required_transitions': 3,
+                'expected_status': ServerStatus.Ready,
+            },
+            'vm_available': {
+                'check_function': lambda: server.status.vm_available,
+                'required_transitions': 2,
+                'expected_status': ServerStatus.Booting,
+            },
+        }
```
bradenmacdonald: Nit: ddt

instance/migrations/0043_change_server_statuses.py
```diff
@@ -0,0 +1,27 @@
...
+            model_name='openstackserver',
+            name='_status',
+            field=models.CharField(choices=[('booting', 'Booting'), ('building', 'Building'), ('failed', 'BuildFailed'), ('pending', 'Pending'), ('ready', 'Ready'), ('terminated', 'Terminated'), ('unknown', 'Unknown')], default='pending', db_column='status', max_length=20, db_index=True),
+        ),
+        migrations.RunSQL(
+            [
+                "UPDATE instance_openstackserver SET status = 'pending' WHERE status = 'new'",
+                "UPDATE instance_openstackserver SET status = 'building' WHERE status = 'started'",
+                "UPDATE instance_openstackserver SET status = 'booting' WHERE status = 'active' OR status = 'rebooting'",
+                "UPDATE instance_openstackserver SET status = 'ready' WHERE status = 'booted' OR status = 'provisioning'",
```
bradenmacdonald: I think we need to check the old `progress` field here too... Because I had an instance in state "Provisioning Failed", then I upgraded to this branch, and now it just shows as "Ready" in the UI.

So I think you need a line before this one that sets status to failed if progress = failed.

instance/models/instance.py
```diff
@@ -66,13 +68,114 @@ class InconsistentInstanceState(Exception):
...
+    """
+    The states that an instance can be in.
+    """
+
+    class New(InstanceState):
+        """ Newly created """
+        state_id = 'new'
+
+    class WaitingForServer(InstanceState):
+        """ Server not yet accessible """
```
bradenmacdonald: This status class should also define `name = "Waiting for Server"`. Same with `ConfiguringServer` and `ConfigurationFailed`.

Otherwise, we see class names in the UI, instead of names with a space in them:
screen shot 2016-04-26 at 9 42 50 pm

instance/static/html/instance/index.html
```diff
@@ -70,16 +66,27 @@
...
+                </span>
+              </li>
             </ul>
           </div>
           <div class="large-2 columns">
-            <div class="status-icon" ng-if="selected.instance.progress != 'running'">
-              <button class="button reprovision" ng-click="provision(selected.instance)">
+            <div ng-if="selected.instance.is_steady"
+                 ng-attr-class="status-icon status-{{ selected.instance.status }}">
+              <button>
```
bradenmacdonald: This is supposed to be `<button class="button reprovision" ng-click="provision(selected.instance)">`

Currently, the Provision button is broken (nothing happens when I click on it).

instance/migrations/0043_change_server_statuses.py
itsjeyd: @bradenmacdonald Thanks for pointing this out. I think we mainly need to attack the issue of incorporating `progress` info in the migration that precedes this one (i.e., `0042_add_instance_status.py`): Servers don't track the provisioning process anymore, so it is fine for a server to have status `Ready` even if the provisioning process failed. And `failed` (`BuildFailed`) is supposed to indicate that OpenStack failed to create the server, so it wouldn't be correct to set "Provisioning Failed" to `failed`. So we need to use appropriate instance states (`Error` and `ConfigurationFailed`) to indicate build failures and provisioning failures. This migration should only take into account `progress` info for servers with a status of `started`; if "Started Failed", status needs to be set to `failed`. (I couldn't find any transitions from "Started Running" to "Started Failed" on master, so modifying this migration to that extent would mostly be a safety net.)

bradenmacdonald: @itsjeyd Yeah, I guess the old "Server status: provisioning, progress: failed" should show up now as "Instance: ConfigurationFailed, Server: Ready". So that sounds good.

And I don't think "Started Failed" happened on the old version, as you say, so it's not really important to spend any time on that.

instance/tests/api/test_instance.py
```diff
@@ -85,20 +86,19 @@ def test_get_domain(self):
...
-
-        for instance_factory in instance_factories:
-            instance = instance_factory()
-            OpenStackServerFactory(instance=instance)
-            response = self.api_client.post('/api/v1/openedxinstance/{pk}/provision/'.format(pk=instance.pk))
-            self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        instance = SingleVMOpenEdXInstanceFactory(status=instance_status)
+        OpenStackServerFactory(instance=instance)
+        response = self.api_client.post('/api/v1/openedxinstance/{pk}/provision/'.format(pk=instance.pk))
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
bradenmacdonald: I think this looks much nicer and more readable :)

bradenmacdonald:
@itsjeyd Sven just merged something to master so this has a merge conflict now that needs to be fixed. I  think the conflict is only in test code though.

bradenmacdonald:
@itsjeyd  I had a server in `ConfigurationFailed` and when I click the provision button I get this error:

```
12:10:57 worker.1    | Traceback (most recent call last):
12:10:57 worker.1    |   File "/home/vagrant/.virtualenvs/opencraft/lib/python3.4/site-packages/huey/consumer.py", line 157, in process_task
12:10:57 worker.1    |     self.huey.execute(task)
12:10:57 worker.1    |   File "/home/vagrant/.virtualenvs/opencraft/lib/python3.4/site-packages/huey/api.py", line 244, in execute
12:10:57 worker.1    |     result = task.execute()
12:10:57 worker.1    |   File "/home/vagrant/.virtualenvs/opencraft/lib/python3.4/site-packages/huey/api.py", line 427, in execute
12:10:57 worker.1    |     return func(*args, **kwargs)
12:10:57 worker.1    |   File "/home/vagrant/.virtualenvs/opencraft/lib/python3.4/site-packages/huey/djhuey/__init__.py", line 106, in inner
12:10:57 worker.1    |     return fn(*args, **kwargs)
12:10:57 worker.1    |   File "/vagrant/instance/tasks.py", line 50, in provision_instance
12:10:57 worker.1    |     instance.provision()
12:10:57 worker.1    |   File "/vagrant/instance/logging.py", line 52, in wrapper
12:10:57 worker.1    |     return method(self, *args, **kwds)
12:10:57 worker.1    |   File "/vagrant/instance/models/instance.py", line 549, in provision
12:10:57 worker.1    |     self._status_to_waiting_for_server()
12:10:57 worker.1    |   File "/vagrant/instance/models/utils.py", line 285, in do_transition
12:10:57 worker.1    |     current_state.name, to_state.name  # pylint: disable=no-member
12:10:57 worker.1    | instance.models.utils.WrongStateException: This transition cannot be used to move from Configuration failed to Waiting for server
```

Seems like that transition should be allowed, right? Hopefully this is an easy fix.

itsjeyd:
@bradenmacdonald Yup, that transition should be allowed. Fixed!

instance/tests/models/test_instance.py
```diff
@@ -1027,3 +1084,112 @@ def deploy():
...
+        self.assertEqual(instance_bad_server.status, Instance.Status.Error)
+        self._assert_status_conditions(instance_bad_server, is_healthy_state=False)
+
+        # Provisioning fails
+        instance_provisioning_failed = ConfiguringSingleVMOpenEdXInstanceFactory()
+        instance_provisioning_failed._status_to_configuration_failed()
+        self.assertEqual(instance_provisioning_failed.status, Instance.Status.ConfigurationFailed)
+        self._assert_status_conditions(instance_provisioning_failed, is_healthy_state=False)
+
+        # Reprovisioning from steady states
```
itsjeyd: @bradenmacdonald FYI, I'm not using `ddt` here because I wanted to be able to cover all happy paths (workflows) in a single test, without having portions of the test that don't depend on changing data run more than once.

itsjeyd:
@bradenmacdonald This is up-to-date with master now (had to rebase again after #62 was merged earlier). Could you have another look please?

instance/tests/models/test_server.py
```diff
@@ -306,38 +306,46 @@ def test_reboot_server_wrong_status(self, server_status):
...
+    def test_terminate_server_vm_unavailable(self, openstack_id, server_status):
         """
-        Terminate a server
+        Terminate a server that is unavailable
         """
         server = OpenStackServerFactory(openstack_id=openstack_id, status=server_status)
         server.terminate()
         self.assertEqual(server.status, ServerStatus.Terminated)
-        server.os_server.delete.assert_called_once_with()
+        server.os_server.delete.assert_not_called()
```
bradenmacdonald: This seems mostly right, but if we terminate a server that is `Building`, shouldn't `os_server.delete` be called in that case? Otherwise, the server will finish building and then we'll have an orphaned VM in OpenStack that never gets deleted. Or am I missing something?

instance/tests/models/test_instance.py
```diff
@@ -552,3 +555,61 @@ def test_status_transitions(self):
...
+    def test_invalid_status_transitions(self, transition):
+        """
+        Test that invalid status transitions raise exception
+        """
+        invalid_from_states = (state for state in Instance.Status.states if state not in transition['from_states'])
+        for invalid_from_state in invalid_from_states:
+            instance = SingleVMOpenEdXInstanceFactory(status=invalid_from_state)
+            self.assertEqual(instance.status, invalid_from_state)
+            with self.assertRaises(WrongStateException):
+                getattr(instance, transition['name'])()
```
bradenmacdonald: Comment: This test is fine, but this behavior should already be guaranteed and unit tested by the `ResourceState` code, and we shouldn't have to have separate tests that "invalid status transitions raise exception" every time we use `ResourceState`.

instance/tests/models/test_server.py
```diff
@@ -417,6 +418,50 @@ def test_status_transitions(self):
...
+    def test_invalid_status_transitions(self, transition):
+        """
+        Test that invalid status transitions raise exception
+        """
+        invalid_from_states = (state for state in ServerStatus.states if state not in transition['from_states'])
+        for invalid_from_state in invalid_from_states:
+            instance = OpenStackServerFactory(status=invalid_from_state)
+            self.assertEqual(instance.status, invalid_from_state)
+            with self.assertRaises(WrongStateException):
+                getattr(instance, transition['name'])()
```
bradenmacdonald: Same here... you can leave this test in but I think it's overkill.

bradenmacdonald:
Other than my one question ("This seems mostly right, but if we terminate a server that is Building, shouldn't os_server.delete be called in that case?"), this seems good and is working for me!

👍  if you reply to or address that question.

I do think that my local testing can't cover all the possibilities that we'll find on the live instance though. I think when we deploy this we should take a snapshot of the RDS PostgreSQL database first, so that if the migrations don't work properly, we won't lose any important data. Also, when deploying, be sure to wait until nobody else is actively running any provisioning tasks.

instance/tests/models/test_instance.py
itsjeyd: @bradenmacdonald I'm aware that we have unit tests for the `ResourceState` code but they only test the behavior in general. The reason for adding this test (and the one you commented on below) was to make sure that specific transitions associated with instance (and server) states are not allowed. It's a sad path test that is meant to compliment the previous happy path test. It would probably be better if the docstring said something like "Test that status transitions that do not make sense are not allowed" :)

instance/tests/models/test_server.py
itsjeyd: @bradenmacdonald That's right - if the server has status `Building` and an OpenStack ID, we should delete it. Fixed in https://github.com/open-craft/opencraft/pull/57/commits/21244a1e382d1207574cac6ba888c0e507df7cb8.

PR #58, opened 2016-04-21
Improve storage backend clean up after running the unit and integration tests.

smarnach:
This change introduces `deprovision_*()` methods for deprovisioning storage backends, and calls these functions in the `tearDown()` methods for applicable unit and integration tests.

The unit tests already did a clean-up, but not in `tearDown()`, so they would leak the databases if they errored out earlier.

This is implemented on top of #54.  Only the last two commits need to be reviewed here.

### DIFF
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -24,8 +24,22 @@ from django.conf import settings
 from django.db import models
 import MySQLdb as mysql
 import pymongo
+from swiftclient.exceptions import ClientException as SwiftClientException
 
-from instance.openstack import create_swift_container
+from instance import openstack
+
+
+def _get_mysql_cursor():
+    """
+    Get a database cursor.
+    """
+    connection = mysql.connect(
+        host=settings.INSTANCE_MYSQL_URL_OBJ.hostname,
+        user=settings.INSTANCE_MYSQL_URL_OBJ.username,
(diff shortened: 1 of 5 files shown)


instance/models/mixins/database.py
```diff
@@ -139,3 +171,26 @@ def provision_swift(self):
...
+                        user=self.swift_openstack_user,
+                        password=self.swift_openstack_password,
+                        tenant=self.swift_openstack_tenant,
+                        auth_url=self.swift_openstack_auth_url,
+                        region=self.swift_openstack_region,
+                    )
+                except SwiftClientException:
+                    # If deleting a Swift container fails, we still want to continue.
+                    self.logger.exception(
+                        'Could not delete Swift container "%s".', container_name, exc_info=True
```
omarkhan: Nit: `exc_info` is not strictly necessary with `.exception()`

omarkhan:
@smarnach looks good. I tested this by forcing the unit tests to error out early, and checking that all databases were cleaned up. Works perfectly. 👍 once circleci is happy and the merge conflict is resolved

smarnach:
@omarkhan Thanks for the review!  I've addressed your nit about `exc_info`.  CircleCI is now passing (RDS instance ran out of space, not an issue with this PR), and there hasn't been any merge conflict.

PR #59, opened 2016-04-22
Beta test registration

omarkhan:
This pull request adds a registration form for the OpenCraft beta. The form is available at `/beta/register/`.

#### Testing

- Checkout this branch
- Run the migrations
- In a separate terminal window inside the development VM, start a dummy smtp server with `sudo python -m smtpd -n -c DebuggingServer localhost:25`
- Start the development server
- Navigate to http://localhost:5000/beta/register/
- Play around with the input fields. Error messages should appear for invalid input
- Submit the form. You should see 2 verification emails in the smtp server output
- Follow the links in the verification emails
- In the django admin, check that a registration, user, and profile have been created, and that the email addresses have been verified

Note that I have not done any styling. This will be done by a web design agency next month.

### DIFF
--- a/betatest/admin.py
+++ b/betatest/admin.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # OpenCraft -- tools to aid developing and hosting free software projects
-# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
@@ -17,27 +17,20 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-Global URL Patterns
+Admin for the betatest app
 """
 
 # Imports #####################################################################
 
-from django.conf.urls import include, url
 from django.contrib import admin
-from django.views.generic.base import RedirectView
 
-import instance
+from betatest.models import BetaTestApplication
 
 
-# URL Patterns ################################################################
+# ModelAdmins #################################################################
 
-urlpatterns = [
-    url(r'^grappelli/', include('grappelli.urls')),
-    url(r'^admin/', include(admin.site.urls)),
-    url(r'^api/', include('api.urls', namespace="api")),
-    url(r'^instance/', include('instance.urls', namespace="instance")),
-    url(r'^beta/', include('betatest.urls', namespace='beta')),
-    url(r'^email-verification/', include('email_verification.urls', namespace='email-verification')),
-    url(r'^favicon\.ico$', RedirectView.as_view(url='/static/img/favicon/favicon.ico', permanent=False)),
-    url(r'^$', instance.views.index),
-]
+class BetaTestApplicationAdmin(admin.ModelAdmin): #pylint: disable=missing-docstring
+    list_display = ('user', 'domain')
+
+
+admin.site.register(BetaTestApplication, BetaTestApplicationAdmin)
--- a/betatest/forms.py
+++ b/betatest/forms.py
@@ -130,7 +130,7 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
         Check that the username is unique.
         """
         username = self.cleaned_data.get('username')
-        if username and User.objects.filter(username=username).exists():
+        if username and User.objects.filter(username=username).exists(): #pylint: disable=no-member
             raise forms.ValidationError(
                 'This username is already taken.',
                 code='unique',
@@ -142,7 +142,7 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
         Check that the email address is unique.
         """
         email = self.cleaned_data.get('email')
-        if email and User.objects.filter(email=email).exists():
+        if email and User.objects.filter(email=email).exists(): #pylint: disable=no-member
             raise forms.ValidationError(
                 'This email address is already registered.',
                 code='unique',
--- a/betatest/forms.py
+++ b/betatest/forms.py
@@ -1,0 +1,199 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Forms for the Instance Manager beta test
+"""
+
+# Imports #####################################################################
+
+from django import forms
+from django.contrib.auth.models import User
+from django.db import transaction
+from django.utils.text import capfirst
+from djng.forms import NgFormValidationMixin, NgModelForm, NgModelFormMixin
+from zxcvbn import password_strength
+
+from betatest.models import BetaTestApplication
+from userprofile.models import UserProfile
+
+
+# Forms #######################################################################
+
+class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelForm):
+    """
+    Application form for beta testers. Creates instances of User, UserProfile,
+    and BetaTestApplication models on submit.
+    """
+    class Meta: #pylint: disable=missing-docstring
+        model = BetaTestApplication
+        exclude = ('user', 'status')
+        widgets = {
+            'public_contact_email': forms.widgets.EmailInput(attrs={
+                'validate-email': True,
+            }),
+        }
+
+    full_name = forms.CharField(
+        max_length=255,
+        help_text='Example: Albus Dumbledore',
+    )
+    username = forms.RegexField(
+        regex=r'^[\w.+-]+$',
+        max_length=30,
+        help_text=('This would also be the username of the administrator '
+                   'account on the Open edX instance.'),
+        error_messages={
+            'invalid': ('Usernames may contain only letters, numbers, and '
+                        './+/-/_ characters.'),
+            'unique': 'This username is already taken.',
+        },
+    )
+    email = forms.EmailField(
+        help_text=('This is also your account name, and where we will send '
+                   'important notices.'),
+        widget=forms.widgets.EmailInput(attrs={'validate-email': True}),
+    )
+    password = forms.CharField(
+        widget=forms.PasswordInput,
+        help_text=('Pick a password for your OpenCraft account. You will be '
+                   'able to use it to login and access your account.'),
+    )
+    password_confirmation = forms.CharField(
+        widget=forms.PasswordInput,
+        help_text=('Please use a strong password: avoid common patterns and '
+                   'make it long enough to be difficult to crack.'),
+    )
+    accept_terms = forms.BooleanField(
+        required=True,
+        label='',
+        help_text=('I understand that this is a beta test, that bugs and '
+                   'crashes are expected, and that the instance is provided '
+                   'for free for the duration of the beta-test, without any '
+                   'guarantee.'),
+        error_messages={
+            'required': 'You must accept these terms to register.',
+        },
+    )
+
+    # This field is created automatically from the model field, but the regex
+    # validator is not copied over. We need to define the field manually so
+    # that validation will work client-side. We do this by copying from the
+    # model field (DRY).
+    _subdomain_field = Meta.model._meta.get_field('subdomain')
+    _subdomain_validator = next(v for v in _subdomain_field.validators
+                                if hasattr(v, 'regex'))
+    subdomain = forms.RegexField(
+        regex=_subdomain_validator.regex,
+        max_length=_subdomain_field.max_length,
+        label=capfirst(_subdomain_field.verbose_name),
+        help_text=_subdomain_field.help_text,
+        error_messages={
+            'invalid': _subdomain_validator.message,
+        },
+    )
+
+    # Form values will be available in the angular scope under this namespace
+    scope_prefix = 'registration'
+
+    # This form itself will be available in the angular scope under this name
+    form_name = 'form'
+
+    def clean_subdomain(self):
+        """
+        Strip the base domain from the end of the subdomain, if given.
+        """
+        subdomain = self.cleaned_data.get('subdomain')
+        if subdomain:
+            match = subdomain.rfind('.' + BetaTestApplication.BASE_DOMAIN)
+            if match != -1:
+                return subdomain[:match]
+        return subdomain
+
+    def clean_username(self):
+        """
+        Check that the username is unique.
+        """
+        username = self.cleaned_data.get('username')
+        if username and User.objects.filter(username=username).exists():
+            raise forms.ValidationError(
+                'This username is already taken.',
+                code='unique',
+            )
+        return username
+
+    def clean_email(self):
+        """
+        Check that the email address is unique.
+        """
+        email = self.cleaned_data.get('email')
+        if email and User.objects.filter(email=email).exists():
+            raise forms.ValidationError(
+                'This email address is already registered.',
+                code='unique',
+            )
+        return email
+
+    def clean_password(self):
+        """
+        Check password strength.
+        """
+        password = self.cleaned_data.get('password')
+        if password and password_strength(password)['score'] < 2:
+            raise forms.ValidationError(
+                ('Please use a stronger password: avoid common patterns and '
+                 'make it long enough to be difficult to crack.'),
+                code='invalid',
+            )
+        return password
+
+    def clean_password_confirmation(self):
+        """
+        Check that the password confirmation field matches the password field.
+        """
+        password = self.cleaned_data.get('password')
+        password_confirmation = self.cleaned_data.get('password_confirmation')
+        if password and password_confirmation and password != password_confirmation:
+            raise forms.ValidationError(
+                "The two password fields didn't match.",
+                code='password_mismatch',
+            )
+        return password
+
+    def save(self, commit=True):
+        """
+        Create User, UserProfile, and BetaTestApplication instances.
+        """
+        application = super().save(commit=False)
+        application.user = User(
+            username=self.cleaned_data['username'],
+            email=self.cleaned_data['email'],
+            is_active=False,
+        )
+        application.user.set_password(self.cleaned_data['password'])
+        application.user.profile = UserProfile(
+            full_name=self.cleaned_data['full_name'],
+        )
+        if commit:
+            with transaction.atomic():
+                application.user.save()
+                application.user.profile.user_id = application.user.pk
+                application.user_id = application.user.pk
+                application.user.profile.save()
+                application.save()
+        return application
--- a/betatest/migrations/0001_initial.py
+++ b/betatest/migrations/0001_initial.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.core.validators
+import betatest.models
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='BetaTestApplication',
+            fields=[
+                ('id', models.AutoField(serialize=False, primary_key=True, verbose_name='ID', auto_created=True)),
+                ('subdomain', models.CharField(validators=[django.core.validators.RegexValidator('^[\\w.-]+$', "Please include only letters, numbers, '_', '-' and '.'"), betatest.models.validate_available_subdomain], help_text='The URL students will visit. In the future, you will also have the possibility to use your own domain name.\n\nExample: hogwarts.opencraft.hosting', error_messages={'unique': 'This domain is already taken.'}, verbose_name='domain name', unique=True, max_length=255)),
+                ('instance_name', models.CharField(help_text='The name of your institution, company or project.\n\nExample: Hogwarts Online Learning', max_length=255)),
+                ('public_contact_email', models.EmailField(help_text='The email your instance of Open edX will be using to send emails, and where your users should send their support requests.\n\nThis needs to be a valid email.', max_length=254)),
+                ('project_description', models.TextField(verbose_name='your project', help_text='What are you going to use the instance for? What are your expectations?')),
+                ('subscribe_to_updates', models.BooleanField(verbose_name='', default=False, help_text='I want OpenCraft to keep me updated about the progress of the beta test, and occasionally send me an email about it.')),
+                ('status', models.CharField(choices=[('pending', 'Pending'), ('accepted', 'Accepted'), ('rejected', 'Rejected')], default='pending', max_length=255)),
+                ('user', models.OneToOneField(to=settings.AUTH_USER_MODEL)),
+            ],
+        ),
+    ]
--- a/betatest/models.py
+++ b/betatest/models.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Models for the Instance Manager beta test
+"""
+
+# Imports #####################################################################
+
+from django.conf import settings
+from django.core import validators
+from django.core.exceptions import ValidationError
+from django.db import models
+
+from instance.models.instance import SingleVMOpenEdXInstance
+
+
+# Models ######################################################################
+
+def validate_available_subdomain(subdomain):
+    """
+    Check that the given subdomain is not already used by an existing
+    instance.
+    """
+    if SingleVMOpenEdXInstance.objects.filter(sub_domain=subdomain).exists():
+        raise ValidationError(
+            message='This domain is already taken.',
+            code='unique',
+        )
+
+
+class BetaTestApplication(models.Model):
+    """
+    An application to beta test the Instance Manager.
+    """
+    BASE_DOMAIN = 'opencraft.hosting'
+
+    PENDING = 'pending'
+    ACCEPTED = 'accepted'
+    REJECTED = 'rejected'
+    STATUS_CHOICES = (
+        (PENDING, 'Pending'),
+        (ACCEPTED, 'Accepted'),
+        (REJECTED, 'Rejected'),
+    )
+
+    user = models.OneToOneField(settings.AUTH_USER_MODEL)
+    subdomain = models.CharField(
+        max_length=255,
+        unique=True,
+        verbose_name='domain name',
+        help_text=('The URL students will visit. In the future, you will also '
+                   'have the possibility to use your own domain name.'
+                   '\n\nExample: hogwarts.{0}').format(BASE_DOMAIN),
+        validators=[
+            validators.RegexValidator(
+                r'^[\w.-]+$',
+                "Please include only letters, numbers, '_', '-' and '.'",
+            ),
+            validate_available_subdomain,
+        ],
+        error_messages={
+            'unique': 'This domain is already taken.',
+        },
+    )
+    instance_name = models.CharField(
+        max_length=255,
+        help_text=('The name of your institution, company or project.'
+                   '\n\nExample: Hogwarts Online Learning'),
+    )
+    public_contact_email = models.EmailField(
+        help_text=('The email your instance of Open edX will be using to '
+                   'send emails, and where your users should send their '
+                   'support requests.'
+                   '\n\nThis needs to be a valid email.'),
+    )
+    project_description = models.TextField(
+        verbose_name='your project',
+        help_text=('What are you going to use the instance for? What are '
+                   'your expectations?'),
+    )
+    subscribe_to_updates = models.BooleanField(
+        default=False,
+        verbose_name='',  # Hide the label
+        help_text=('I want OpenCraft to keep me updated about the progress '
+                   'of the beta test, and occasionally send me an email '
+                   'about it.'),
+    )
+    status = models.CharField(
+        max_length=255,
+        choices=STATUS_CHOICES,
+        default=PENDING,
+    )
+
+    def __str__(self):
+        return self.domain
+
+    @property
+    def domain(self):
+        """
+        The full domain requested for this application.
+        """
+        return '{0}.{1}'.format(self.subdomain, self.BASE_DOMAIN)
--- a/betatest/tests/browser/browser_registration.py
+++ b/betatest/tests/browser/browser_registration.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Beta registration browser tests
+"""
+
+# Imports #####################################################################
+
+from collections import defaultdict
+import re
+import time
+
+from django.core.urlresolvers import reverse
+from django.contrib.staticfiles.testing import StaticLiveServerTestCase
+from selenium import webdriver
+from selenium.webdriver.support import expected_conditions
+from selenium.webdriver.support.ui import WebDriverWait
+
+from betatest.tests.test_views import BetaTestApplicationViewTestMixin
+
+
+# Tests #######################################################################
+
+class BetaTestBrowserTestCase(BetaTestApplicationViewTestMixin,
+                              StaticLiveServerTestCase):
+    """
+    Tests the beta test registration flow with a real browser.
+    """
+    def setUp(self):
+        super().setUp()
+        self.client = webdriver.Firefox()
+
+    def tearDown(self):
+        self.client.quit()
+        super().tearDown()
+
+    def register(self, form_data):
+        """
+        Fill in the registration form and click the submit button, if the form
+        is valid.
+        """
+        self.client.get('{host}{path}'.format(
+            host=self.live_server_url,
+            path=reverse('beta:register'),
+        ))
+
+        for field, value in form_data.items():
+            element = self.form.find_element_by_name(field)
+            if (element.get_attribute('type') == 'checkbox' and
+                    bool(value) != element.is_selected()):
+                element.click()
+            else:
+                element.send_keys(value)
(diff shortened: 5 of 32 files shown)


omarkhan:
@smarnach this is ready for a look now.

omarkhan:
Oops, @itsjeyd this one is for you

betatest/forms.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        Make sure fields are displayed in the correct order.
+        """
+        super().__init__(*args, **kwargs)
+        fields = self.fields.copy()
+        self.fields = OrderedDict()
+        for field in self.field_order:
+            self.fields[field] = fields[field]
```
smarnach: Why is this necessary?  I think fields are rendered in the order in which you define them in the form class, so you could just define them in the correct order.  (I vaguely remember they keep a global counter when a Field class is instantiated, and use that global counter for ordering.)
In any case, you don't need to copy `self.fields`, since you don't even modify it, and even if you did that would be OK, since the fields shared between all instances of the class are in `self.base_fields`.

smarnach: I see, some of the fields come from the model, and some of the fields are defined here, that makes sense.

(I just had a quick look at the code because I'm curious, by the way.  I'm not actually doing a review.)

betatest/templates/betatest/application.html
```diff
@@ -0,0 +1,38 @@
...
+{% endblock head_js %}
+
+{% block content %}
+<div class="container content" ng-app="BetaTestApp">
+    <div class="row">
+        <div class="large-12 columns">
+            <h1>Register for Beta</h1>
+            <p>This is the application form for the beta of the self-service
+            hosting offered by OpenCraft. For enterprise or institutional
+            offerings, contact us.</p>
```
antoviaque: Can you make "contact us" a link to contact@opencraft.com ?

(Just lurking quickly too : ) The PR looks neat! Glad to see the integration of forms with angular : ) )

betatest/models.py
```diff
@@ -0,0 +1,102 @@
...
+        max_length=255,
+        unique=True,
+        verbose_name='domain name',
+        help_text=('The URL students will visit. In the future, you will also '
+                   'have the possibility to use your own domain name.'
+                   '\n\nExample: hogwarts.{0}').format(BASE_DOMAIN),
+        validators=[
+            validators.RegexValidator(
+                r'^[\w.]+$',
+                'Please include only letters, numbers, and the . character.',
```
smarnach: The regex also allows `_`.  Shouldn't we allow hyphens as well?  they are quite common in domain names.

betatest/forms.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+
+    # This field is created automatically from the model field, but the regex
+    # validator is not copied over. We need to define the field manually so
+    # that validation will work client-side. We do this by copying from the
+    # model field (DRY).
+    _subdomain_field = Meta.model._meta.get_field('subdomain')
+    _subdomain_validator = next(v for v in _subdomain_field.validators
+                                if hasattr(v, 'regex'))
+    subdomain = forms.RegexField(
```
smarnach: Do we validate whether the subdomain is already taken?  We should look at both the instances that are currently running and the other beta test applications.  It might also make sense to blacklist certain values (we don't really want anyone to register `www.openedxhosting.com`), but we could do that manually for now.

omarkhan:
I'm having some trouble writing js tests. The way they are currently set up, they are run completely in isolation, without any of the html from the django templates. This doesn't really make sense here, as most of the functionality I want to test relates to the form, and what django-angular does with it. @itsjeyd @smarnach @antoviaque what should I do here? Just skip these and focus on selenium tests? Extend the js tests to include a fake DOM with something like [jsdom](https://github.com/tmpvar/jsdom)?

betatest/admin.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
```
itsjeyd: @omarkhan This should probably be updated to say `2015-2016` (also for the other files you added).

betatest/forms.py
```diff
@@ -0,0 +1,205 @@
...
+from betatest.models import BetaTestApplication
+from userprofile.models import UserProfile
+
+
+# Forms #######################################################################
+
+class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelForm):
+    """
+    Application form for beta testers. Creates User, UserProfile, and
+    BetaTestApplication models.
```
itsjeyd: @omarkhan Nit: "Creates instances of User, UserProfile, and BetaTestApplication models on submit."

betatest/models.py
```diff
@@ -0,0 +1,102 @@
...
+        verbose_name='your project',
+        help_text=('What are you going to use the instance for? What are '
+                   'your expectations?'),
+    )
+    subscribe_to_updates = models.BooleanField(
+        default=False,
+        verbose_name='',
+        help_text=('I want OpenCraft to keep me updated about the progress '
+                   'of the beta test, and send me an email occasionally '
+                   'about it.'),
```
itsjeyd: @omarkhan Nit: "and occasionally send me an email about it" maybe?

betatest/models.py
```diff
@@ -0,0 +1,102 @@
...
+                   '\n\nExample: Hogwarts Online Learning'),
+    )
+    public_contact_email = models.EmailField(
+        help_text=('The email your instance of Open edX will be using to '
+                   'send emails, and where your users should send their '
+                   'support requests.'
+                   '\n\nThis needs to be a valid email.'),
+    )
+    project_description = models.TextField(
+        verbose_name='your project',
```
itsjeyd: @omarkhan Capitalize first letter? "Your project"

betatest/models.py
```diff
@@ -0,0 +1,102 @@
...
+        help_text=('What are you going to use the instance for? What are '
+                   'your expectations?'),
+    )
+    subscribe_to_updates = models.BooleanField(
+        default=False,
+        verbose_name='',
```
itsjeyd: @omarkhan Any specific reason for setting this to `''`?

betatest/tests.py
```diff
@@ -0,0 +1,158 @@
...
+        self.application = {
+            'subdomain': 'hogwarts',
+            'instance_name': 'Hogwarts',
+            'full_name': 'Albus Dumbledore',
+            'username': 'albus',
+            'email': 'albus.dumbledore@hogwarts.edu',
+            'public_contact_email': 'support@hogwarts.edu',
+            'password': 'gryffindor',
+            'password_confirmation': 'gryffindor',
+            'project_description': 'Online courses in Witchcraft and Wizardry',
```
itsjeyd: @omarkhan Where can I sign up? :D

betatest/tests.py
```diff
@@ -0,0 +1,158 @@
...
+        for user_field in ('username', 'email'):
+            self.assertEqual(getattr(user, user_field),
+                             self.application[user_field])
+        self.assertTrue(user.check_password(self.application['password']))
+
+        # Check the profile fields
+        self.assertEqual(profile.full_name, self.application['full_name'])
+
+        # Test email verification flow
+        self.assertEqual(len(mail.outbox), 2)
```
itsjeyd: @omarkhan Should we check if email is *unverified* before clicking the links below?

betatest/admin.py
smarnach: Do we really need copyright notes in every file?  Google stopped doing this since their lawyers said that they don't have any legal relevance whatsoever, and that having a single license file in the project achieves exactly the same.  (I can't judge whether that's true, but I find having these redundant comments everywhere rather annoying.)

email_verification/models.py
```diff
@@ -0,0 +1,145 @@
...
+        try:
+            instance = cls.objects.get(email=email)
+        except cls.DoesNotExist:
+            instance = cls(email=email)
+        if force_send_email or not instance.verification_email_sent:
+            instance.send_verification_email(base_url)
+        return instance
+
+    @classmethod
+    def check_verification_code(cls, code):
```
itsjeyd: @omarkhan I think it would make sense to rename this method to `verify_mail`, to better reflect its main purpose. The fact that it checks the verification code is sort of incidental (it might be doing other checks as well before marking an email address as valid).

email_verification/models.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Check that the given verification code is valid, and mark the email
+        address as verified if it is. Returns the corresponding
+        `EmailVerification` instance.
+        """
+        instance = cls.objects.get(verification_code=code)
```
itsjeyd: @omarkhan `instance` is a bit generic (and slightly confusing since we also have an `instance` model). Rename to `email_verification`, perhaps?

email_verification/templates/email_verification/verify.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+
+{% block title %}Verify email address{% endblock %}
+
+{% block content %}
+<div class="container content">
+    {% if verification.verified %}
+    Thank you for verifying your email address ({{ verification.email }}).
+    {% elif verification.verification_code_expired %}
+    {# TODO: Resend verification email #}
```
itsjeyd: @omarkhan Leftover TODO, just pointing out to make sure we don't forget about it.

email_verification/templates/email_verification/verify.html
```diff
@@ -0,0 +1,16 @@
...
+    Your email verification link has expired.
+    {% else %}
+    Your email address could not be verified - please contact us.
```
itsjeyd: @omarkhan Make "contact us" a link here to when you address @antoviaque's comment :)

email_verification/tests/test_models.py
```diff
@@ -0,0 +1,117 @@
...
+        verification = EmailVerification.verify(self.verification.email,
+                                                base_url='http://example.com/',
+                                                force_send_email=True)
+        self.assertEqual(len(mail.outbox), 2)
+        self.assertNotEqual(verification.verification_code, original_code)
+
+        email = mail.outbox[-1]
+        self.assertIn(verification.verification_code, email.body)
+
+    def test_check_verification_code(self):
```
itsjeyd: @omarkhan This will need to be updated as well when you rename `check_verification_code` to `verify_email`.

email_verification/tests/test_views.py
```diff
@@ -0,0 +1,73 @@
...
+            email='raspberryberet@example.com',
+            base_url='http://example.com/',
+        )
+        self.verification_url = self.verification.get_verification_url(
+            base_url='http://example.com/',
+        )
+
+    def test_verify_email(self):
+        """
+        Test that we can check a verification code.
```
itsjeyd: @omarkhan "... that we can verify an email address."

email_verification/tests/test_views.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        response = self.client.get(self.verification_url)
+        self.assertContains(response, 'Thank you for verifying your email address')
+        self.verification.refresh_from_db()
+        self.assertEqual(self.verification.verified, True)
+
+    @patch.object(EmailVerification, 'verification_code_expiry', timedelta(0))
+    def test_verify_email_code_expired(self):
+        """
+        Test attempting to verify an expired verification code.
```
itsjeyd: @omarkhan Nit: "... to verify an email address with an expired verification code."

betatest/urls.py
```diff
@@ -0,0 +1,37 @@
...
+from betatest import views
+
+
+# URL Patterns ################################################################
+
+app_name = 'beta'
+urlpatterns = [
+    url(r'^register/$', views.BetaTestApplicationView.as_view(), name='register'),
+    url(r'^success/$', views.success, name='success'),
+    url(r'^api/register/validate/$', views.validate_registration, name='validate'),
```
antoviaque: Would you be able to use the standard API URL pattern & listing registration here? Cf https://github.com/open-craft/opencraft/tree/master/api

betatest/admin.py
bradenmacdonald: I know Xavier likes these but I feel the same way :)

email_verification/templates/email_verification/verify.html
antoviaque: +1 :)

email_verification/models.py
```diff
@@ -0,0 +1,145 @@
...
+from django.conf import settings
+from django.core.mail import send_mail
+from django.core.urlresolvers import reverse
+from django.db import models
+from django.utils.crypto import get_random_string
+
+
+# Models ######################################################################
+
+class EmailVerification(models.Model):
```
antoviaque: Was there an advantage to reimplement this over reusing an existing one, like https://github.com/mfogel/django-simple-email-confirmation ?

smarnach:
@omarkhan Sorry, I can't help regarding the question how to implement JS tests for this, that's a topic I should learn more about.

betatest/admin.py
antoviaque: I don't know either, but in doubt on licensing issues, I would pick the opinion of the FSF over the one from Google. Here in any case it seem minor, and it doesn't have much inconvenience to it. I don't personally see that section negatively - it's a place we talk to other devs or companies looking to work on it, and can even be part of the visual esthetic : )

However we should change the contact email - to contact@opencraft.com : )

antoviaque:
@omarkhan 

>I'm having some trouble writing js tests. The way they are currently set up, they are run completely in isolation, without any of the html from the django templates. This doesn't really make sense here, as most of the functionality I want to test relates to the form, and what django-angular does with it. @itsjeyd @smarnach @antoviaque what should I do here? Just skip these and focus on selenium tests? Extend the js tests to include a fake DOM with something like jsdom?

I'm not completely sure I understand what is blocking you - you have the dom of the browser? But yes you need to test the result of the operations based on changes on a DOM fixture: https://github.com/open-craft/opencraft/tree/master/instance/tests/fixtures . Here at first sight you would probably want to mock both the responses the Django server sends to the client.

In any case please don't add any node/npm dependency, I beg you : )

betatest/admin.py
smarnach: All the section on how to apply the licence says is

> It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty.

It doesn't say this is necessary for the licence to be valid or anything.

(And I wouldn't trust legal advice from the FSF too much.  It seems to be generally agreed that substantial parts of the AGPL are unenforcable under US law; at least that's what I heard from an FSF member.)

betatest/urls.py
omarkhan: @antoviaque you want me to move this into the api app? I don't think that makes sense, as the functionality is completely specific to this app, used only for ajax validation of the registration form. If we move it into the api app we create an unnecessary coupling between apps.

email_verification/models.py
omarkhan: @antoviaque I evaluated [django-simple-email-confirmation](https://github.com/mfogel/django-simple-email-confirmation), [django-registration](https://github.com/macropin/django-registration), and [django-userena](https://github.com/bread-and-pepper/django-userena) for this. All of them implement email verification but tie email addresses to the `User` model. In this case we also want to verify an email address that belongs to the `BetaTestApplication` model. I considered using django-simple-email-confirmation and adding some mappings between its `EmailAddress` model and out `BetaTestApplication` model, but in the end I decided it would be cleaner and easier to just implement the basic email verification functionality as a separate app.

betatest/forms.py
omarkhan: @smarnach we do validate whether the subdomain is already taken, but only by other beta registrations, not by instances that are currently running. I will add a validation step for that.

betatest/models.py
omarkhan: Django does this automatically when displaying a form. The [example in the docs](https://docs.djangoproject.com/en/1.8/ref/models/options/#verbose-name) is lowercase, so I followed that convention.

betatest/models.py
omarkhan: There is no label on the mockup. I tried setting this to `None`, but that just causes django to fall back to its default behaviour of using the field name as a label. The empty string seems to work. I will add a comment to clarify

email_verification/models.py
omarkhan: I'm ok to rename this, but then we should also rename the `verify` method above. What would you call that?

betatest/admin.py
antoviaque: @smarnach Are you sure trolling about the enforceability of the GPL over the format of headers is a good use of our time here? I'd be happy to chat about the FSF and the GPL around a beer though, and I will definitely be interested in knowing which parts of the GPL are unenforceable - once you sprint is clean can you send me the details?

Also, besides the technical legal skills, there are other elements to consider when considering legal advice, such as the motivations of the organization giving them, and how this might fit their larger strategy. I don't always agree with everything the FSF does or say, but at least their motivations are clear and strongly held to. They align in many ways with ours. On the other side, on licensing Google is quite opposed to many of the principles of free software - cf the fact that the vast majority of the code they produce is proprietary, and their position on software patents (which is behind a lot of their opposition to the GPLv3, btw).

betatest/urls.py
antoviaque: @omarkhan If you look at the link I gave, you will see that this module doesn't contain code (which is always in the app itself, cf how it's done for the instance app), it's just a place where we register the api router - the idea being to have a registry that can list all the API calls supported, as well as a standard URL pattern: http://console.opencraft.com/api/v1/

email_verification/models.py
antoviaque: @omarkhan What about contributing the ability to confirm multiple addresses to one of these upstream projects? Email verification is a very standard feature, the ecosystem would likely benefit more from improving the projects that already exist rather than adding our own.

betatest/urls.py
omarkhan: I understand, but this is an internal api that is used for something very specific. If you think we should expose this endpoint anyway, I can add it to the `api` urlconf, but I would prefer to keep it here as well to keep the `betatest` app self-contained.

email_verification/models.py
itsjeyd: @omarkhan Good point, how about `send_verification_email`?

email_verification/models.py
omarkhan: @antoviaque OK I will work on a PR for django-simple-email-confirmation once I have finished the selenium tests for the registration form.

betatest/admin.py
smarnach: @antoviaque My main intention in the last comment was to point out that even the FSF recommendation to include the header everywhere isn't that strongly worded.  I didn't mean to imply any critique of the goals of the FSF in general, which I definitely agree with.

betatest/urls.py
antoviaque: @omarkhan I'd like to follow the same pattern for all APIs - they are currently all internal, only meant to be used by the browser client.

email_verification/models.py
omarkhan: That works. Done

betatest/urls.py
omarkhan: @antoviaque done, see https://github.com/open-craft/opencraft/pull/59/commits/f2a2931ac6232e574a36767d886c91de04a336b1. It feels like a bit of a hack but it works.

omarkhan:
@itsjeyd this is ready for another look now. I have addressed your comments, as well as @smarnach's and @antoviaque's. @antoviaque wants me to contribute the email verification functionality to an existing lib and use that instead, so that's next on my list.

I have added js tests and selenium tests. The js tests will run when you run `make test_js`. For the selenium tests, run `make test_browser`.

betatest/urls.py
antoviaque: @omarkhan Thanks! Does it make the API call discoverable through the API web interface? 

Also, why does it feel like a hack to you? It does introduce a dependency between the api app and the beta app, but the api app is used solely for managing the API URL scheme, so that seem fine. Or is there a better way to manage a common API scheme, including disoverability & browsable web interface, which I missed?

omarkhan: Yes this makes it discoverable through the browsable api web interface.

I'm not happy with the coupling between apps, as it means that we can no longer move the app into its own django project, for example. But that is not why this feels like a hack to me.

It feels like a hack because DRF routers only support viewsets, which are an abstraction designed to automatically generate a RESTful api for models. In this case we are simply validating a form, models are not involved. So to get this to work with the browsable api I had to repurpose the viewset's `list` method, even though we are not listing any models. That feels like a hack to me because we are not using the abstraction the way it was designed. It just happens to work.

The quick and easy way I have done this in the past is to simply list the available endpoints in the project's documentation. Then you can visit the url in your browser and see DRF's autogenerated docs.

A more complete solution would be to use something like [DRF Docs](https://github.com/ekonstantinidis/django-rest-framework-docs). Other options are listed here: http://www.django-rest-framework.org/topics/documenting-your-api/

antoviaque: @omarkhan Thanks for the details. DRF Docs looks neat, it's a good idea imho - do you want to add it to the uservoice, so we consider it to replace the current approach in the future?

omarkhan: Sounds good, I will do that

omarkhan:
I can't reproduce a bug @itsjeyd reported. Can someone else see if they can reproduce it?

- Type some text into an input box
- Select the text and cut it. 'This field is required' should appear
- Paste the text back into the input box using the keyboard

For @itsjeyd the 'This field is required' message is still displayed. For me it disappears, as expected:
registration

betatest/tests/js/betatest_app_spec.js
```diff
@@ -0,0 +1,82 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(function(){
```
itsjeyd: @omarkhan Nit: Missing space between `()` and `{`

antoviaque:
@omarkhan congratz on getting the Js tests done :) Now you still need to add an integration test, you know that, right? :)

@open-craft/core I meant to mention this in the meeting yesterday, but here is just as well - when you send a PR sharing open craft im, please consider adding an integration test. Sometimes it might not make sense to do one, but then please explain the rationale for it in the PR description. The coverage of integration tests is quite low at the moment, we should be careful to add them, especially for user facing features. Otherwise soon we'll be needing QA testers :)

omarkhan:
@antoviaque I did write [integration tests](https://github.com/open-craft/opencraft/blob/omar/beta-registration/betatest/tests/browser/browser_registration.py). Did you have something else in mind?

betatest/tests/browser/browser_registration.py
```diff
@@ -0,0 +1,83 @@
...
+from selenium.webdriver.support import expected_conditions
+from selenium.webdriver.support.ui import WebDriverWait
+
+from betatest.tests.test_views import BetaTestApplicationViewMixin
+
+
+# Tests #######################################################################
+
+class BetaTestBrowserTestCase(BetaTestApplicationViewMixin,
+                              StaticLiveServerTestCase):
```
antoviaque: Is it the expected practice to be mixing a View mixin with the TestCase? These are two different types of objects?

betatest/tests/browser/browser_registration.py
```diff
@@ -0,0 +1,83 @@
...
+        if submit.is_enabled():
+            # Click the submit button and wait for the next page to load
+            html = self.client.find_element_by_tag_name('html')
+            submit.click()
+            WebDriverWait(self.client, timeout=3) \
+                .until(expected_conditions.staleness_of(html))
+
+        return self.client.page_source
+
+    def assert_registration_fails(self, form_data):
```
antoviaque: What is being run by the test runner? just the `assert_*()` function? Why not use `test_*()` like in other test cases?

betatest/tests/browser/browser_registration.py
```diff
@@ -0,0 +1,83 @@
...
+        for field, value in form_data.items():
+            element = form.find_element_by_name(field)
+            if (element.get_attribute('type') == 'checkbox' and
+                    bool(value) != element.is_selected()):
+                element.click()
+            else:
+                element.send_keys(value)
+
+        submit = form.find_element_by_tag_name('button')
+        if submit.is_enabled():
```
antoviaque: When is this run?

It would be good to test the expected results more - currently I only see testing the activation status of a button, and that the next page loads. What page should load, What messages should be displayed in each step?

antoviaque:
@omarkhan My bad, I had missed them - that will teach me to reply on my phone : )  Cool, thanks. I was looking for an `integration` folder, but it makes sense to handle the browser tests separately from the server integration tests. I had a few questions/comments, which I posted inline in that code.

betatest/tests/browser/browser_registration.py
omarkhan: Oops it's actually a test mixin, I just didn't name it properly. Fixed: 8131b8a

betatest/tests/browser/browser_registration.py
omarkhan: The test methods are defined in the mixin here: https://github.com/open-craft/opencraft/blob/omar/beta-registration/betatest/tests/test_views.py#L39

The code in this class fills in and submits the form using selenium instead of sending the form data directly using the django test client.

antoviaque: I see - nice, thanks for the explanation. I'm good then : ) Maybe just a short comment so that it's more obvious? With the `..TestMixin` class that's already much more clear, but it can't hurt.

omarkhan:
@itsjeyd this is ready for a final review now.

betatest/admin.py
itsjeyd: @antoviaque @smarnach @omarkhan If we are keeping the notices, should the years be updated?

betatest/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
...
+                ('id', models.AutoField(verbose_name='ID', primary_key=True, auto_created=True, serialize=False)),
+                ('subdomain', models.CharField(verbose_name='domain name', unique=True, max_length=255, help_text='The URL students will visit. In the future, you will also have the possibility to use your own domain name.\n\nExample: hogwarts.opencraft.hosting')),
+                ('instance_name', models.CharField(help_text='The name of your institution, company or project.\n\nExample: Hogwarts Online Learning', max_length=255)),
+                ('public_contact_email', models.EmailField(help_text='The email your instance of Open edX will be using to send emails, and where your users should send their support requests.\n\nThis needs to be a valid email.', max_length=254)),
+                ('project_description', models.TextField(verbose_name='your project', help_text='What are you going to use the instance for? What are your expectations?')),
+                ('subscribe_to_updates', models.BooleanField(help_text='I want OpenCraft to keep me updated about the progress of the beta test, and send me an email occasionally about it.', default=False)),
```
itsjeyd: @omarkhan Can you update this to match the new wording? I.e., "... and occasionally send me an email about it." (I think the help texts for the other fields should still match but I didn't double-check.)

betatest/admin.py
omarkhan: Whatever we decide here, can we do it in a separate PR?

betatest/migrations/0001_initial.py
omarkhan: Good catch @itsjeyd, I had forgotten to update the migration. Fixed

betatest/admin.py
itsjeyd: @omarkhan I'd be fine with that, since we'll also have to update the notices in existing files. Let's see what @antoviaque and @smarnach think.

smarnach: My suggestion would be to change the email address and remove the year in a separate PR.  The year is bound to get outdated anyway.

antoviaque: Yup, updating the email (to contact@opencraft.com) in a separate PR works for me.

smarnach:
@omarkhan I cannor reproduce the issue with "This field is required" continuing to be displayed even when there is data in the input field again.  I tried on Firefox 46 and Chrome 49, and tried pasting by Keyboard short cut and context menu as well as Linux middle-click pasting.  All combinations worked fine, and I couldn't reproduce any issue.

omarkhan:
Thanks @smarnach. @itsjeyd OK to merge once I get a green build and squash the commits?

itsjeyd:
@omarkhan 

> OK to merge once I get a green build and squash the commits?

Yup, this looks great :+1:

antoviaque:
@omarkhan I still need to do a product review before the merge given that there is user-facing functionality - I've planned it for today though.

omarkhan:
@antoviaque OK. I have squashed the commits.

antoviaque:
@omarkhan Looks good! Kudos, this is nice both from the technical and user perspective : )

The comments I had while testing were:
* "Please use a stronger password." => do we state anywhere what the definition of a strong password is? It should be added both to the description on the right-hand column, and to the error message.
* With the default dev environment configuration, submitting the form crashes with https://i.imgur.com/c1F4XcD.png - and the only way to go around it is to setup proper SMTP credentials, and changing the from email to match what is accepted by this mailer. Could you add `EMAIL_BACKEND` to the settings, defaulting to `django.core.mail.backends.console.EmailBackend` and to `django.core.mail.backends.smtp.EmailBackend` on console.plebia.org?
* Currently, the approach is that the submit button is disabled until all the required fields are filled or checked. Which is good, except for the fact that it currently doesn't provide the user with any way to know what is missing, if he doesn't fill all the required fields. For example, only one of the two checkboxes at the end is mandatory - but if I don't check any of them, nothing tells me explicitly which field I am missing, and I can't to try to submit it to get an explicit error. I can't think of a good way to provide the user with that feedback at the right time (and only at the right time) if the submit button is disabled - showing it all the time would add a lot of red to the whole form until it is filled; we could also just add a "* Required" label next to all the fields, but the user would still have to look through all of these to search for the one he missed. The only simple and straightforward way that I see is to allow to click on the button when not all the requirements are met, and to show the error message on the missing fields then. That's what was (implicitly) intended in the requirements for "_error message at the top of the page, and one next to the issue explaining the issue and how to fix it_". But maybe there is a better approach?

betatest/admin.py
antoviaque: Done in https://github.com/open-craft/opencraft/pull/6/commits/cb0cf5df710789a5eedbb12eca33bfd1a8d54a38 . Updating the year only takes running a sed command, so I've left it, and changed it to `2015-2016` in all files.

omarkhan:
@antoviaque I have addressed your feedback, can you take another look?

antoviaque:
@omarkhan Thank you! I'll try, but I'm not sure I will have time for another review today, especially because that means reprovisioning my vagrant VM. Although you will be updating it to the current version once https://github.com/open-craft/opencraft/pull/6 is merged - ping me when it's done that would help. And if I don't get time to rereview before the meeting, feel free to merge to avoid the spillover - we would then take on any additional changes in a future task.

antoviaque:
@omarkhan I would have had time now to review, but this still needs a rebase. I'll try to reprovision a VM with your branch in the meantime, but you will need the rebase to merge on time before the meeting - note that it will require rebuilding your VM, so make sure to start it ahead of the meeting, to avoid the spillover.

betatest/forms.py
```diff
@@ -0,0 +1,199 @@
...
+    )
+    password = forms.CharField(
+        widget=forms.PasswordInput,
+        help_text=('Pick a password for your OpenCraft account. You will be '
+                   'able to use it to login and access your account.'),
+    )
+    password_confirmation = forms.CharField(
+        widget=forms.PasswordInput,
+        help_text=('Please use a strong password: avoid common patterns and '
+                   'make it long enough to be difficult to crack.'),
```
antoviaque: By definition, I meant giving the precise constraints - like the minimum number of characters, if there needs to be a combination of special characters, etc.

antoviaque:
Kudos for the upstream PR to django-agular - quickly merged and released, awesome : )

requirements.txt
```diff
@@ -9,7 +10,7 @@ debtcollector==0.8.0
 decorator==4.0.4
 dj-static==0.0.6
 Django==1.8.5
-django-angular==0.7.15
+-e git+https://github.com/open-craft/django-angular.git@add307dbaa5df05f5839361ba899d88e83d658b2#egg=django-angular
```
antoviaque: You can use 0.8.2 from https://github.com/jrief/django-angular/pull/256 now, right?

betatest/forms.py
smarnach: @antoviaque I specifically asked in the task description _not_ to have this kind of restrictions, and instead use a library that estimates password strength based on entropy, since restrictions that require special characters etc. are annoying.

@bradenmacdonald You might want to chime in here. :)

antoviaque: @smarnach The problem is that the user needs to know what are the constraints - I hadn't noticed this requirement, or more precisely didn't realize that it meant that we can't have a clear definition for the user of the constraint here. Can we provide an example of what will be considered strong enough, in a reliable manner? Like "For example, a 15 letters word or series of words"? Would that be reliable, or will that depend on the letters?

omarkhan:
@antoviaque rebased. I don't see anything in https://github.com/open-craft/opencraft/pull/6 that would cause any problems with this. Once I get a green build I will merge.

betatest/forms.py
omarkhan: @antoviaque some background if you're interested: https://tech.dropbox.com/2012/04/zxcvbn-realistic-password-strength-estimation/

antoviaque:
@omarkhan Alright - but https://github.com/open-craft/opencraft/pull/59#discussion_r61754286 still needs to be addressed. Can you have a look at my question there? If there is no time that can be put in a follow-up task, but if it's manageable by a string change it would be simpler to do it now.

betatest/forms.py
smarnach: Fifteen times the letter `a` certainly wouldn't be strong enough.  I think the nicest feedback we could give would be to include a progress bar that indicates how strong the entered password is, with a mark at the required strength.  Otherwise, we can just give the general advice to add more lettters, or a bigger diversity of letters.

omarkhan: @antoviaque it all depends. 'qwertyuiopasdfghjklzxcvbnm' is more than 15 letters but would fail validation, as it should. A series of words would work but again it depends on the words.

How about this: 'try to use at least 10 characters and avoid common patterns such as "password" or "qwerty".'

antoviaque: We need something reliable to give to the user, so he doesn't get stuck and get accurate feedback when something is wrong - if he enters a 10 character password, and it's not enough, what should he do? Add more characters? Change the words he is using in it, because they are too common? He won't know.

If we can't provide complete instructions & clear error messages, unless we find a better idea, we will need to change the validation method. Since it's bigger than what you can do now, you can leave it aside - I'll create a task in the backlog about this.

bradenmacdonald: I strongly prefer entropy-based password validation, using a tool like `zxcvbn`. What most websites do that use it is simply show a client-side "strength bar" while the user enters their password. Then, as they type every single character in their password, they can instantly see if the password is strong enough or not, and they can get an intuitive sense of what makes a strong password.

In such cases, it is typically not necessary to provide any further guidance to the user. People are used to making up passwords and generally have an idea of how to make a password stronger. I don't think it hurts to provide some specific examples though, such as something like "An example of a strong password would be thinking of a sentence with eight or more words and using the first letter of each word in the sentence, e.g. 'I love taking online courses with Open edX!' becomes 'IltocwOe!'"

omarkhan: I agree with Braden. The only problem is that the python port of zxcvbn
does not always give the same results as the js reference implementation.
When Tim was testing this he found a password that passed client-side
validation but failed on the server. It would not be acceptable for a
client-side password strength meter to say that a password is very strong
only for the server to reject it when the form is submitted.

Possible solutions:

- Fix the python port of zxcvbn
- Use pyv8 or similar to call the js version of zxcvbn on the server

smarnach: I also think the strength bar Braden described would be a good solution (but I'm bad at UX).  I have some doubts that we want pyv8 on the server, since I'd expect it to be more trouble than it's worth.  Fixing the Python port would be a great solution though, if it's feasible.

I also just saw the xkcd included in the link Omar gave above.  It's a very good summary why I don't like the usual numeral/uppercase/symbol rules: https://xkcd.com/936/

antoviaque: I still don't see a solution here for the problem I was pointing out:

> We need something reliable to give to the user, so he doesn't get stuck and can get accurate feedback when something is wrong - if he enters a 10 character password, and it's not enough, what should he do? Add more characters? Change the words he is using in it, because they are too common? He won't know.

The progress bars are better than what there is now, but they don't answer this.

I agree that entropy based validation is better technically, and I would like to keep it. But users come first, so we would only keep it if we can avoid putting them in a situation where we tell them they are doing something wrong, without pointing out what, exactly. Especially in a critical step such as registration.

omarkhan: The js implementation of zxcvbn includes nice user-readable messages
explaining how a password can be made stronger. These messages are absent
from the python port. We could port them over.

bradenmacdonald: Alternately, you could only validate the password strength on the client side for now. If the user deliberately hacks around your JS validation to submit a weak password, it'll be their own fault.

The client form could submit the entropy score along with the password, so that the server can at least verify that the client-side checks seem to have run, and the entropy is reasonable.

That way, you get nice suggestion messages, and don't have to worry about inconsistencies between JS and python implementations.

antoviaque: @omarkhan @bradenmacdonald Ah, cool - I've had a look at https://dl.dropboxusercontent.com/u/209/zxcvbn/test/index.html and the messages displayed there would definitely solve the problem on my side. Thanks for this. I let you update OC-1587 once you agree on the exact approach?

omarkhan: I agree with Braden. I have updated OC-1587

antoviaque: @omarkhan Perfect - thanks for the update!

PR #62, opened 2016-04-27
Automatic reprovisioning attempts rebuild servers

e-kolpakov:


instance/models/instance.py
```diff
@@ -447,6 +447,23 @@ def provision(self):
 
         Returns: (server, log)
         """
+        attempt_num = 1
+        provisioned = False
+        server, logs = None, []
+
+        while attempt_num <= self.attempts and not provisioned:
+            server, deploy_log, provisioned = self._provision_attempt()
```
smarnach: It would be nice to continue logging the attempt number before each attempt, as we do now.  I find this information useful (and I search for it in the web interface to find the next attempt).

instance/models/mixins/ansible.py
```diff
@@ -140,29 +140,14 @@ def deploy(self):
...
-                self.logger.info(log)
-                log, returncode = self._run_playbook(requirements_path, playbook_path)
-                if returncode != 0:
-                    self.logger.error(
-                        'Playbook failed for instance {}'.format(self))
-                    continue
-                else:
-                    break
+            log = 'Running playbook "{path}/{name}":'.format(path=playbook_path, name=self.ansible_playbook_name)
+            self.logger.info(log)
```
smarnach: We don't need to store the string in a variable first, in particular given that the same variable name is used for something else in the next line.  (I know you didn't introduce this, but we might just as well fix it while we are here.)

instance/tests/models/test_instance.py
```diff
@@ -705,6 +705,7 @@ def test_provision_failed_exception_email(self):
         self.assertEqual(mime_type, "text/html")
 
 
+# pylint: disable=too-many-public-methods
```
smarnach: Great that the linter tells us that we have too many tests. :)

instance/tests/models/test_instance.py
```diff
@@ -977,6 +978,38 @@ def test_provision_failed(self, mocks):
...
+        ]
+        mocks.os_server_manager.add_fixture('test-run-provisioning-server-1', 'openstack/api_server_2_active.json')
+        mocks.os_server_manager.add_fixture('test-run-provisioning-server-2', 'openstack/api_server_3_active.json')
+        mock_reboot1 = mocks.os_server_manager.get_os_server('test-run-provisioning-server-1').reboot
+        mock_reboot2 = mocks.os_server_manager.get_os_server('test-run-provisioning-server-2').reboot
+
+        instance = OpenEdXInstanceFactory(sub_domain='run.provisioning', use_ephemeral_databases=True, attempts=2)
+
+        instance.provision()
+        print(mocks.mock_set_dns_record.mock_calls)
```
smarnach: This can be deleted.

smarnach:
@e-kolpakov I've added a few comments, and this needs to be rebased.  I'll do the manual testing in the next round.

instance/tests/models/test_instance.py
antoviaque: Can we address that comment instead of disabling it, especially at the global level on that file? pylint warns for a good reason here. During the pylint upgrade I've noticed that we have taken this shortcut a few times throughout the codebase, and created a ticket to address this (OC-1572). It would be good to not add to that debt.

e-kolpakov: @antoviaque ok. I just wonder, why "stop suppressing pylint errors" always (often?) arises on my PRs? :)

smarnach: I don't think having many test methods on a single test class is "technical debt" just because pylint warns about it.  On the contrary, I think it's in itself a good thing.

However, I do think we have an underlying problem here, which is that the instance model does basically everything.  The fact that the functionality is split into several mix-ins helps a bit, but it doesn't fundamentally change the fact that we have one big, monolithic, strongly coupled class in the end.  That big class needs many tests, and it's good that we have many tests.  Splitting up the tests in two classes doesn't do anything to reduce technical debt.  The only thing it achieves is getting rid of a pylint directive.

I hope that this will get improved a bit by the work @bradenmacdonald is doing this week, which will split the responsibilities up into two classes.

antoviaque: @e-kolpakov Thanks! And not sure why it comes up in your PRs : ) I don't think the pylint suppressions I saw in the upgrade were from you :p although I haven't checked.

@smarnach You can always still split in smaller chunks, tests can be group logically. If anything it will make it easier to find the right test, just by categorizing the tests better.

For pylint in general, I know sometimes it can look frivolous or counter productive to address the warnings, but avoiding pylint exclusions has the merit of making it easier to review them, and to have a more unified code style. And they aren't usually that hard to follow.

e-kolpakov:
@smarnach @antoviaque notes addressed - please take another look

antoviaque:
@e-kolpakov No more objections from my side : ) Thanks for making the changes!

smarnach:
:+1: Thanks, looks good to me!

PR #63, opened 2016-04-28
OC-1541: Keep the old instance running while deploying a new one

bradenmacdonald:
Changes:
- Fix occasional pylint warning about undocumented Meta classes
- Combine LogEntry classes
    * ^ These two commits have been copied to #67 so they can be reviewed more easily and merged sooner
- Separated `[OpenEdX]Instance` into `Instance` (long-lived, mutable, creates AppServers and marks one as active), and `AppServer` (immutable, owns one VM)
- Factored Open edX specific code into `OpenEdXInstance` and `OpenEdXAppServer` respectively
- Factored Open edX specific code out of the ansible mixin
- Change the ansible mixin so it can optionally run multiple playbooks
- Removed `instance.last_provisioning_started` since you can compute that from the most recent `AppServer.created` field
- Combined `forum_version`, `notifier_version`, `xqueue_version`, and `certs_version` into a single `openedx_release` setting, since these were always set to the same value, and if one needs to be set differently it can be done using ansible_extra_settings anyways.
- Separated PR-watching code to a separate app (`OpenEdXInstance` should not need to know about pull requests)
- Updated API
- Allows provisioning a new AppServer at any time
- Updated UI accordingly
   * Lovely new home screen ("empty state"):
     screen shot 2016-05-08 at 7 54 01 pm
   * Instance list indicates the state of the active app server for each instance, as well as the state of any newer app server waiting to be provisioned / reviewed / made active.
     screen shot 2016-05-08 at 7 54 57 pm
   * New "Pull Request" tab:
      screen shot 2016-05-08 at 7 56 40 pm
   * When you click "Update Instance Settings from PR", it displays an animation while fetching updates from GitHub, then switches to the "Settings" tab to show you the new settings / commit hash.
   * New Settings tab for instance settings (should be made editable in the future)
       screen shot 2016-05-08 at 7 58 03 pm
   * New App Servers tab to review app servers, launch new ones, or change which one is active:
       screen shot 2016-05-08 at 7 59 37 pm
   * Deep linking to any instance or app server :)
- Behavior: The first time a PR is seen, an instance will be auto-provisioned; on failure, the provisioning will be retried (on a new AppServer) up to 1 additional time; on success, the AppServer will automatically be marked active. On the other hand, whenever the user manually provisions a new instance or AppServer, there is no automatic retry or activation. (To give users time to inspect the errors or the new appserver.)

TODO:
- Update Jasmine tests
- Fix `Instance` and `InstanceReference` model `delete()` methods so they delete each other.

[...]

PR #67, opened 2016-04-29
Combine log entries (Prereq for OC-1541)

bradenmacdonald:
This change combines `GeneralLogEntry`, `InstanceLogEntry`, `ServerLogEntry` into a single `LogEntry` class that uses [the contenttypes framework](https://docs.djangoproject.com/en/1.9/ref/contrib/contenttypes/) to hold a reference to the object that "owns" each log entry.

This change is required in order to accommodate the refactoring in #63. Specifically, there is a new intermediate class, `AppServer`, which would need a new table to hold its logs. And `AppServer` is abstract (since multi-table inheritance in django gets very messy), so there are potentially a number of concrete `AppServer` subclasses that we may have, and we can't expect to create a separate logging table for each one. Luckily, django already has a wonderful and longstanding solution to this problem, which is the contenttypes framework.

Pros:
* Code is way simpler
* Loading logs way faster/simpler
* Less load on app server
* Ignoring the migrations, this change resulting in a net reduction of 77 lines of code :)
* ~~Log entries not deleted when instance/server is deleted.~~
* Includes a fully reversible data migration, so this can be tested, deployed, or even reverted without any loss of data.
* The various existing logging test cases cover this code and pass with (almost) no modifications.

Cons:
* Foreign keys not enforced at DB level.
* ~~(As a result:) Log entries not deleted when instance/server is deleted. (Not sure if this is a pro or a con)~~

Also, while I was at it I fixed an occasional pylint warning about undocumented Meta classes: We don't need docstrings on Meta, and now pylint knows that, without inline disable comments

### DIFF
--- a/instance/migrations/0047_generic_logging3.py
+++ b/instance/migrations/0047_generic_logging3.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django_extensions.db.fields
+import instance.models.utils
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contenttypes', '0002_remove_content_type_name'),
+        ('instance', '0046_generic_logging2'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='GeneralLogEntry',
+        ),
+        migrations.RemoveField(
+            model_name='instancelogentry',
+            name='obj',
+        ),
+        migrations.RemoveField(
+            model_name='serverlogentry',
+            name='obj',
+        ),
+        migrations.DeleteModel(
+            name='InstanceLogEntry',
+        ),
+        migrations.DeleteModel(
+            name='ServerLogEntry',
+        ),
+    ]
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -26,7 +26,9 @@ import logging
 import string
 
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from django.db.models import Q
 from django.db.backends.utils import truncate_name
 from django.template import loader
 from django.utils import timezone
@@ -36,6 +38,7 @@ from django_extensions.db.models import TimeStampedModel
 from instance.gandi import GandiAPI
 from instance.logger_adapter import InstanceLoggerAdapter
 from instance.logging import log_exception
+from instance.models.log_entry import LogEntry
 from instance.models.mixins.ansible import AnsibleInstanceMixin
 from instance.models.mixins.database import MongoDBInstanceMixin, MySQLInstanceMixin, SwiftContainerInstanceMixin
 from instance.models.mixins.utilities import EmailInstanceMixin
@@ -272,6 +275,8 @@ class Instance(ValidateModelMixin, TimeStampedModel):
         Return the list of log entry instances for the instance and its current active server,
         optionally filtering by logging level. If a limit is given, only the latest records are
         returned.
+
+        Returns oldest entries first.
         """
         # TODO: Filter out log entries for which the user doesn't have view rights
         instance_type = ContentType.objects.get_for_model(self)
@@ -282,17 +287,12 @@ class Instance(ValidateModelMixin, TimeStampedModel):
             (Q(content_type=server_type) & Q(object_id__in=server_ids))
         )
         if level_list:
-            instance_log_entry_set = instance_log_entry_set.filter(level__in=level_list)
-        instance_log_entry_set = instance_log_entry_set.order_by('pk')
+            entries = entries.filter(level__in=level_list)
         if limit:
-            instance_log_entry_set = reversed(instance_log_entry_set.reverse()[:limit])
-        else:
-            instance_log_entry_set = instance_log_entry_set.iterator()
-
-        entries = self._sort_log_entries(server_log_entry_set, instance_log_entry_set)
-        if limit:
-            return entries[-limit:]
-        return entries
+            # Apply the limit at the SQL/DB level while sorted by descending date, then reverse.
+            # Otherwise, we'd have to retrieve all rows and then apply the limit using python.
+            return reversed(list(entries[:limit]))
+        return entries.order_by('created')
 
     @property
     def log_entries(self):
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -22,7 +22,6 @@ Instance app models - Instance
 
 # Imports #####################################################################
 
-from functools import partial
 import logging
 import string
 
@@ -268,36 +267,6 @@ class Instance(ValidateModelMixin, TimeStampedModel):
             self.base_domain = settings.INSTANCES_BASE_DOMAIN
         super().save(**kwargs)
 
-    @staticmethod
-    def _sort_log_entries(server_logs, instance_logs):
-        """
-        Helper method to combine the instance and server log outputs in chronological order
-        """
-        next_server_log_entry = partial(next, server_logs, None)
-        next_instance_log_entry = partial(next, instance_logs, None)
-
-        log = []
-        instance_log_entry = next_instance_log_entry()
-        server_log_entry = next_server_log_entry()
-
-        while instance_log_entry is not None and server_log_entry is not None:
-            if server_log_entry.created < instance_log_entry.created:
-                log.append(server_log_entry)
-                server_log_entry = next_server_log_entry()
-            else:
-                log.append(instance_log_entry)
-                instance_log_entry = next_instance_log_entry()
-
-        while instance_log_entry is not None:
-            log.append(instance_log_entry)
-            instance_log_entry = next_instance_log_entry()
-
-        while server_log_entry is not None:
-            log.append(server_log_entry)
-            server_log_entry = next_server_log_entry()
-
-        return log
-
     def _get_log_entries(self, level_list=None, limit=None):
         """
         Return the list of log entry instances for the instance and its current active server,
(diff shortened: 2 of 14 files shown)


antoviaque:
@bradenmacdonald Cool, I didn't knew about content types, this looks like a good solution. Thanks for doing the change!

>Cons:
>
>    Foreign keys not enforced at DB level.
>    (As a result:) Log entries not deleted when instance/server is deleted. (Not sure if this is a pro or a con)

This sounds like things you can handle with signals, no? Ie pre_save & pre_delete? Would be good to not leave objects behind when the instances/servers are deleted, especially. 

Or for foreign keys is it already enforced at the application level?

bradenmacdonald:
Yes, exactly. I can add a post_delete signal to delete related log entries when a server or instance is deleted. I notice that we weren't deleting the general logs ever so I thought we might also want to preserve the server/instance logs forever too, and maybe delete them after a year or so rather than when the server is deleted. But I'm not sure that would actually be useful, so I will go ahead with a post_delete signal.

bradenmacdonald:
I've now added code to delete old log entries automatically.

antoviaque:
@bradenmacdonald For the general log entries, they should have been deleted with the instance itself, no? If not that would have been a bug. Generally speaking we'll probably want to avoid deleting instance/server objects in any case (and flag them as deleted instead), but it's always good to make sure that if we do we don't keep orphaned objects; the database could end up growing to a point where that becomes necessary, especially for the logentries.

Btw my remark about signals was also about your other point, "_Foreign keys not enforced at DB level._" - even if it can't be enforced at the DB level, `pre_save` could help to enforce this at the application level, if it's not already?

bradenmacdonald:
@antoviaque As far as I know, the general log is full of entries like "200 GET /" ... but maybe that's only when run in debug mode. Those general log entries are usually not associated with a particular instance, so would never be deleted.

Foreign keys are not strictly enforced at the application level, but the code as it stands would never create an invalid foreign key. I don't think there's any reason to add a separate validation layer.

antoviaque:
@bradenmacdonald You shouldn't be reading your email on your birthday :p 

For these general log entries, it could be good having a closer look - are they generated by a provision run? Could be worth adding a bug in the backlog if that's the case, and if that's not addressed by the current PR.

> Foreign keys are not strictly enforced at the application level, but the code as it stands would never create an invalid foreign key. I don't think there's any reason to add a separate validation layer.

Today yes (or maybe? How can you be sure there isn't a bug?) - but an explicit validation of this in pre_save would ensure it remains the case, and doesn't take much to put in place.

instance/migrations/0047_generic_logging3.py
```diff
@@ -0,0 +1,45 @@
...
+    def data_forward(apps, schema_editor):
+        """
+        Copy log entries from GeneralLogEntry, InstanceLogEntry, and ServerLogEntry into the new
+        combined LogEntry table.
+        """
+
+    def data_backward(apps, schema_editor):
+        """
+        Reverse the data migration
+        """
```
itsjeyd: @bradenmacdonald I think you can remove `data_forward` and `data_backward` from this migration.

instance/migrations/0047_generic_logging3.py
```diff
@@ -0,0 +1,45 @@
...
+            name='GeneralLogEntry',
+        ),
+        migrations.RemoveField(
+            model_name='instancelogentry',
+            name='obj',
+        ),
+        migrations.RemoveField(
+            model_name='serverlogentry',
+            name='obj',
+        ),
```
itsjeyd: @bradenmacdonald Why do we need to remove these fields before deleting the model?

instance/models/instance.py
```diff
@@ -268,65 +270,27 @@ def save(self, **kwargs):
...
-            instance_log_entry_set = reversed(instance_log_entry_set.reverse()[:limit])
-        else:
-            instance_log_entry_set = instance_log_entry_set.iterator()
-
-        entries = self._sort_log_entries(server_log_entry_set, instance_log_entry_set)
+            entries = entries.filter(level__in=level_list)
         if limit:
-            return entries[-limit:]
-        return entries
+            return reversed(list(entries[:limit]))
```
itsjeyd: @bradenmacdonald Nit: When I first read this I was tempted to suggest that you could do `entries[:limit].order_by('created')` here, to be consistent with the code in the next line. But that doesn't work because "Cannot reorder a query once a slice has been taken." Maybe add a comment to that extent, to avoid potential confusion?

instance/models/log_entry.py
```diff
@@ -52,41 +50,38 @@ class LogEntry(ValidateModelMixin, TimeStampedModel):
...
+        """
+        Whenever an object is deleted, check if there are corresponding log entries to delete.
+        """
+        if isinstance(instance, LogEntry) or instance._meta.app_label == 'migrations':
+            return  # Avoid pointless database queries when deleting log entries themselves or saving migration history
+        if not isinstance(instance.pk, int):
+            return  # Current LogEntry schema requires integer objects IDs, so we know this object isn't in the table
+        content_type = ContentType.objects.get_for_model(instance)
+        LogEntry.objects.filter(content_type=content_type, object_id=instance.pk).delete()
+        # TODO: Use the following instead once we upgrade to Django 1.9:
```
itsjeyd: @bradenmacdonald Leftover TODO, just marking it to ensure we don't forget

instance/tests/models/test_log_entry.py
```diff
@@ -133,6 +133,36 @@ def test_log_publish(self, mock_publish_data):
...
+        server2.logger.info('Line #3, on server 2')
+
+        self.assertEqual(LogEntry.objects.count(), 3)
+        # Delete server 1:
+        server1.delete()
+        # Now its log entry should be deleted:
+        self.assertEqual(LogEntry.objects.count(), 2)
+        entries = LogEntry.objects.order_by('pk').all().values_list('text', flat=True)
+        self.assertIn('Line #1, on instance', entries[0])
+        self.assertIn('Line #3, on server 2', entries[1])
```
itsjeyd: @bradenmacdonald Nit: Maybe add another check asserting that "Line #2, on server 1" is missing from `entries[1]`.

instance/models/log_entry.py
```diff
@@ -21,24 +21,22 @@
 """
 
 # Imports #####################################################################
-
```
itsjeyd: Nit: We should keep this blank line to stay consistent with the formatting in other files.

instance/models/log_entry.py
```diff
@@ -21,24 +21,22 @@
...
-from instance.models.server import OpenStackServer
-from instance.models.utils import ValidateModelMixin
-
+from .utils import ValidateModelMixin
 
 # Logging #####################################################################
 
 logger = logging.getLogger(__name__)
 
-
```
itsjeyd: @bradenmacdonald Nit: We should keep this blank link to stay consistent with the formatting in other files.

itsjeyd:
@bradenmacdonald Happy birthday! :)

I reviewed this and left a few comments. Manual testing went fine. I was able to 

* run forward and backward migrations 
* get rid of log entries by deleting corresponding server and instance objects
* provision a new instance

without issue. Running the migrations produced expected results (in both directions).

instance/migrations/0047_generic_logging3.py
bradenmacdonald: Whoops, good point - thanks.

instance/migrations/0047_generic_logging3.py
bradenmacdonald: That was auto-generated by Django's migration framework. I believe the reason is that these are foreign keys, and some databases require you to delete the foreign key constraints before deleting the table.

instance/models/instance.py
bradenmacdonald: @itsjeyd  In SQL, there's no simple way to say "Sort the rows by increasing date and return the last 200 rows"; all you can say is "sort the rows by *decreasing* date and return up to 200 rows". So since `order_by` directives must be translated to SQL, it makes sense that Django can't do that.


I will add a comment that says "# Apply the limit at the SQL/DB level while sorted by descending date, then reverse. Otherwise, we'd have to retrieve all rows and then apply the limit using python."

instance/models/log_entry.py
bradenmacdonald: @itsjeyd Django 1.9 was merged so I updated this PR with master and uncommented this.

bradenmacdonald:
@antoviaque @itsjeyd All review comments addressed. Is this good to merge now?

instance/tests/models/test_log_entry.py
```diff
@@ -134,6 +136,108 @@ def test_log_publish(self, mock_publish_data):
...
+            self.server.logger.info(msg)
+        log_entry = LogEntry.objects.order_by('-pk')[0]
+        self.assertEqual(
+            str(log_entry),
+            '2015-10-20 20:10:15 |     INFO | instance.models.server    | instance=my.instance,server=vm1_id | ' + msg,
+        )
+
+    def test_invalid_content_type_object_id_combo(self):
+        """
+        Test that content_type cannot be set on its own
```
itsjeyd: @bradenmacdonald Nit: "Test that content_type and object_id cannot be set on their own"

I'm fine with updating this docstring as part of #63; it's definitely not a blocker to merging this PR :)

itsjeyd:
@bradenmacdonald Thanks for the updates, this LGTM :+1:

antoviaque:
@bradenmacdonald Yup, LGTM - I have only had a look at the changes we discussed, but I don't have any remaining objections.

bradenmacdonald:
Thanks guys!

Addressed nit, rebased from 473371f7cfa1d07d8c09d68e0f75aa2c184d0f39 onto master... will merge once green.

Rebased diff is identical, except for the nit I fixed.

PR #68, opened 2016-05-01
Add Ansible playbook to deploy the instance manager.

smarnach:
**Testing instructions**

1. Bring up a PostgreSQL test server, and possibly MongoDB and MySQL test servers as well.
2. Bring up a new OpenStack server using the unmodified xenial image I uploaded to the SBG1 region of the Instance Manager production project.  (Credentials can be found on the production instance.)  A small instance will be enough, no external volume needed for testing.  (The playbooks don't mount or format any volumes.)  Edit the security groups of the database servers to give access to the new instance.
3. Follow the instructions in the `README.md` file added in this PR.  For a test private settings file, see the Jira ticket.  You need to update the IP address of the Postgres server, and possibly add the MySQL and MongoDB servers to the configuration.


deploy/opencraft.yml
```diff
@@ -0,0 +1,11 @@
+- name: Deploy the OpenCraft Instance Manager
+  hosts: all
+  become: true
+  roles:
+    - name: common-server
+      COMMON_SERVER_NO_BACKUPS: true    # The instance manager VM is stateless
+      FORWARD_MAIL_MYDOMAIN: "{{ inventory_hostname }}"
```
jbzdak: You'll need to add {{forward-server-mail}} role explicitly, the only way to fix the 20-30 mails sent during the initial provisioning run was to remove this role from dependencies of {{common-server}}  and add it to playbook after commmon server role. I thought about different solutions, but this one was least bad.

deploy/roles/.gitignore
```diff
@@ -0,0 +1 @@
+*/
```
jbzdak: Just of sheer curiosity: what does it ignore?

smarnach: All subdirectories, i.e. the roles installed here via ansible-galaxy.  (You achieved a similar goal by ignoring `roles/` one level further up.)

deploy/opencraft.yml
smarnach: Thanks.  I didn't update my repos with the lates changes yet to reduce the number of times I have to do that.

deploy/README.md
```diff
@@ -0,0 +1,63 @@
...
+        pip install -r requirements.txt
+
+3. Install all required roles:
+
+        ansible-galaxy install -r requirements.yml
+
+4. Prepare your server with a stock Ubuntu 16.04 image, and make sure you can SSH to it.  Manually
+   install Python 2 by running
+
+        sudo apt-get update && sudo apt-get install -y python
```
jbzdak: I think we should try to minimize amount of manual tasks, maybe this could be automated, [this random blog on the internet](http://blog.swwomm.com/2016/04/xenial-ansible.html), suggested that something like this would work, would you try this approach? 

```
- name: bootstrap python 2
  gather_facts: no
  tasks:
  - raw: sudo apt-get update -qq && sudo apt-get install -qq python2.7
```

deploy/README.md
```diff
@@ -0,0 +1,63 @@
...
+   (The trailing comma must be preserved.)
+
+After deployment, the server doesn't run the instance manager automatically.  You need to log in
+and run it manually inside a `screen` or `tmux` session:
+
+1. SSH into your server.
+
+2. Start `screen`.
+
+3. Become the `www-data` user: `sudo -s -H -u www-data`
```
jbzdak: I also needed to `cd /var/www/opencraft/`.

deploy/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+4. Run the server and workers: `/var/www/.virtualenvs/opencraft/bin/exec make run WORKERS=5`
```
jbzdak: I needed to (manually make change from this PR: https://github.com/open-craft/opencraft/pull/70)

jbzdak:
@smarnach Some documentation to the provisioning should be added about `OPENSTACK_SANDBOX_SSH_KEYNAME`, ideally we should generate key on the instance and then push it to proper openstack project, but given how often we deploy new IM instances this might be an overkill. But just add this as an explicit step to readme.

deploy/README.md
smarnach: @jbzdak Sorry, I meant to ping you about that.  #70 is merged now, so this problem should be fixed.

deploy/README.md
smarnach: This sounds good, thanks for the hint.  I stopped looking at random blogs after reading an FAQ entry in the Ansible documentation that dealt with the specific issue of not having Python 2 installed on the target machine, and they just recommended to install it, so I assumed there is no better way.

smarnach:
@jbzdak You are perfectly right about `OPENSTACK_SANDBOX_SSH_KEYNAME`.  I wonder why I didn't notice this, since I waited until the instance manager at least started provisioning an instance (I didn't wait for it to complete, but it looked like it was working fine so far).

We actually need to include the private ssh key in the configuration, since we don't want to use a new key everytime we provision a new instance.  I'll implement that tomorrow.

jbzdak:
@smarnach There is a problem with non-ephemereal databases, only user configured for ``mysql-im`` in the databases playbook is the ``root`` user, and geeringguy.mysql role [prohibits root user from logging in remotely](https://github.com/geerlingguy/ansible-role-mysql/blob/master/tasks/secure-installation.yml#L3). So you'd have to create another user with privileges for everything.

deploy/README.md
smarnach: Just for the record, this is implemented now and it works nicely.

jbzdak:
:+1: Once the tests pass, either here or on #75. Most probably nothing to do.

jbzdak:
Build is green. Merging.

antoviaque:
@jbzdak You probably didn't knew about this, but on OpenCraft IM for merges/rebases in PRs, we have agreed to only merge during the review to facilitate the job of the reviewer, but to still rebase before merging to keep an ordered history on master - this was missing the rebase before the merge.

CC @open-craft/core FYI, in case others didn't knew about it.

PR #70, opened 2016-05-04
Remove `--periodic` option from Huey (unavailable in the latest version)

antoviaque:
This was preivously a no-op option (since this is the default), but kept in the command-line for clarity. The latest version of Huey removed the command altogether, so we do too.

@bradenmacdonald @smarnach


bradenmacdonald:
👍

smarnach:
Thanks @bradenmacdonald and @antoviaque for fixing this quickly!

PR #71, opened 2016-05-04
Add factory functions for creating sandbox and production instances

itsjeyd:
cf. [OC-1554](https://tasks.opencraft.com/browse/OC-1554).

**Changes**

Adds two factory functions for creating instances from the shell:

- `instance_factory` creates instances with settings that are appropriate for sandboxes.
- `production_instance_factory` creates instances with settings that are appropriate for production instances.

The only mandatory argument for each function is `sub_domain`. Defaults for other fields can be overridden by passing additional keyword arguments to the functions. If necessary it is also possible to modify instances after they have been created by the functions (before starting the provisioning process).

**Test instructions**

1. Start the shell via `make shell`.

2. Import the new functions via:

   ```python
   from instance.factories import instance_factory, production_instance_factory
   ```

3. Create a new sandbox instance via:

   ```python
   sandbox = instance_factory(sub_domain="hello.sandbox")
   ```

4. Check that `sandbox` has 
   * `use_ephemeral_databases` set to `True`
   * `configuration_version` set to `master`
   * `openedx_release` set to `master`
   * `configuration_extra_settings` set to `""`

5. Create a new production instance via:

   ```python
   production_instance = production_instance_factory(sub_domain="hello.production")
   ```

6. Check that `production_instance` has 
   * `use_ephemeral_databases` set to `False`
   * `configuration_version` set to `named-releases/dogwood.rc`
   * `openedx_release` set to `named-releases/dogwood.rc`
   * `configuration_extra_settings` set to the contents of [prod-vars.yml](https://github.com/open-craft/opencraft/blob/production-instance/instance/templates/instance/ansible/prod-vars.yml)

7. Create a few more instances, passing in non-default values for some fields. Check that relevant fields are set to correct values on the resulting instances.

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -70,8 +70,13 @@ def instance_factory(**kwargs):
     # Ensure caller provided required arguments
     assert "sub_domain" in kwargs
 
+    # Ensure instance uses ephemeral databases by default,
+    # irrespective of current value of INSTANCE_EPHEMERAL_DATABASES setting
+    instance_kwargs = dict(use_ephemeral_databases=True)
+    instance_kwargs.update(kwargs)
+
     # Create instance
-    instance = OpenEdXInstance.objects.create(**kwargs)
+    instance = OpenEdXInstance.objects.create(**instance_kwargs)
     return instance
 
 
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -27,6 +27,7 @@ import logging
 from django.conf import settings
 from django.template import loader
 
+from instance import ansible
 from instance.models.openedx_instance import OpenEdXInstance
 
 
@@ -108,16 +109,18 @@ def production_instance_factory(**kwargs):
     # Check environment and report potential problems
     _check_environment()
 
+    # Gather settings
+    production_settings = loader.get_template('instance/ansible/prod-vars.yml').render({})
+    configuration_extra_settings = kwargs.pop("configuration_extra_settings", "")
+    extra_settings = ansible.yaml_merge(production_settings, configuration_extra_settings)
+    instance_kwargs = dict(
+        use_ephemeral_databases=False,
+        configuration_version=settings.LATEST_OPENEDX_RELEASE,
+        openedx_release=settings.LATEST_OPENEDX_RELEASE,
+        configuration_extra_settings=extra_settings,
+    )
+    instance_kwargs.update(kwargs)
+
     # Create instance
-    production_instance = OpenEdXInstance(**kwargs)
-    if "use_ephemeral_databases" not in kwargs:
-        production_instance.use_ephemeral_databases = False
-    if "configuration_version" not in kwargs:
-        production_instance.configuration_version = settings.LATEST_OPENEDX_RELEASE
-    if "openedx_release" not in kwargs:
-        production_instance.openedx_release = settings.LATEST_OPENEDX_RELEASE
-    if "configuration_extra_settings" not in kwargs:
-        template = loader.get_template('instance/ansible/prod-vars.yml')
-        production_instance.configuration_extra_settings = template.render({})
-    production_instance.save()
+    production_instance = OpenEdXInstance.objects.create(**instance_kwargs)
     return production_instance
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -115,8 +115,8 @@ def production_instance_factory(**kwargs):
     extra_settings = ansible.yaml_merge(production_settings, configuration_extra_settings)
     instance_kwargs = dict(
         use_ephemeral_databases=False,
-        configuration_version=settings.LATEST_OPENEDX_RELEASE,
-        openedx_release=settings.LATEST_OPENEDX_RELEASE,
+        configuration_version=settings.OPENEDX_RELEASE_STABLE_REF,
+        openedx_release=settings.OPENEDX_RELEASE_STABLE_REF,
         configuration_extra_settings=extra_settings,
     )
     instance_kwargs.update(kwargs)
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Instance app - Factory functions for creating instances
+"""
+
+# Imports #####################################################################
+
+from django.conf import settings
+from django.template import loader
+
+from instance.models.openedx_instance import OpenEdXInstance
+
+
+# Functions ###################################################################
+
+def instance_factory(**kwargs):
+    """
+    Factory function for creating instances.
+
+    Returns a newly created OpenEdXInstance.
+
+    Callers can use keyword arguments to pass in non-default values
+    for any field that is defined on the OpenEdXInstance model.
+
+    When called without any arguments, the instance that is returned
+    will have its fields set to default values that are appropriate
+    for *sandbox* instances.
+
+    To create an instance with default settings that are suitable for production,
+    use `production_instance_factory`.
+    """
+    return OpenEdXInstance.objects.create(**kwargs)
+
+
+def production_instance_factory(**kwargs):
+    """
+    Factory function for creating production instances.
+
+    Returns a newly created OpenEdXInstance.
+
+    Callers can use keyword arguments to pass in non-default values
+    for any field that is defined on the OpenEdXInstance model.
+
+    When called without any arguments, the instance that is returned
+    will have its fields set to default values that are appropriate
+    for *production* instances.
+
+    To create an instance with default settings that are suitable for sandboxes,
+    use `instance_factory`.
+    """
+    production_instance = OpenEdXInstance.objects.create(**kwargs)
+    if "use_ephemeral_databases" not in kwargs:
+        production_instance.use_ephemeral_databases = False
+    if "configuration_version" not in kwargs:
+        production_instance.configuration_version = settings.LATEST_OPENEDX_RELEASE
+    if "openedx_release" not in kwargs:
+        production_instance.openedx_release = settings.LATEST_OPENEDX_RELEASE
+    if "configuration_extra_settings" not in kwargs:
+        template = loader.get_template('instance/ansible/prod-vars.yml')
+        production_instance.configuration_extra_settings = template.render({})
+    production_instance.save()
+    return production_instance
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -65,6 +65,9 @@ def production_instance_factory(**kwargs):
     To create an instance with default settings that are suitable for sandboxes,
     use `instance_factory`.
     """
+    # NOTE: The long-term goal is to eliminate differences between sandboxes
+    # and production instances, and for this function to disappear.
+    # Please do not add behavior that is specific to production instances here.
     production_instance = OpenEdXInstance.objects.create(**kwargs)
     if "use_ephemeral_databases" not in kwargs:
         production_instance.use_ephemeral_databases = False
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -22,13 +22,32 @@ Instance app - Factory functions for creating instances
 
 # Imports #####################################################################
 
+import logging
+
 from django.conf import settings
 from django.template import loader
 
 from instance.models.openedx_instance import OpenEdXInstance
 
 
+# Logging #####################################################################
+
+logger = logging.getLogger(__name__)
+
+
 # Functions ###################################################################
+
+def _check_environment():
+    """
+    Check environment and report potential problems for production instances
+    """
+    if not settings.SWIFT_ENABLE:
+        logger.warning("Swift support is currently disabled. Adjust SWIFT_ENABLE setting.")
+    if settings.INSTANCE_MYSQL_URL is None:
+        logger.warning("URL for external MySQL database is missing. Adjust INSTANCE_MYSQL_URL setting.")
+    if settings.INSTANCE_MONGO_URL is None:
+        logger.warning("URL for external Mongo database is missing. Adjust INSTANCE_MONGO_URL setting.")
+
 
 def instance_factory(**kwargs):
     """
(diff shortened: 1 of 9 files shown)


instance/models/openedx_instance.py
```diff
@@ -160,10 +167,41 @@ def spawn_appserver(self):
             # Copy the current value of each setting into the AppServer, preserving it permanently:
             configuration_database_settings=self.get_database_settings(),
             configuration_storage_settings=self.get_storage_settings(),
+            configuration_extra_settings=self.get_extra_settings(),
```
bradenmacdonald: `configuration_extra_settings` is already included in `instance_config` so this is not necessary

instance/models/openedx_instance.py
```diff
@@ -160,10 +167,41 @@ def spawn_appserver(self):
...
             # finish and replace the second as the active server. We are not really worried about that for now.
             self.set_appserver_active(app_server.pk)
+
+
+class OpenEdXProductionInstance(OpenEdXInstance):
+    """
+    OpenEdXProductionInstance: An OpenEdXInstance using production settings
+    """
+
+    LATEST_RELEASE = "named-release/dogwood.rc"
```
bradenmacdonald: Should be a django setting

instance/models/openedx_instance.py
```diff
@@ -160,10 +167,41 @@ def spawn_appserver(self):
...
+
+    def save(self, **kwargs):
+        """
+        Set default field values specific to production instances before saving.
+        """
+        # Ansible-specific settings:
+        self.configuration_version = self.LATEST_RELEASE
+        self.openedx_release = self.LATEST_RELEASE
+        # Also set edx_platform_commit in case behavior of "set_field_defaults" changes
+        self.edx_platform_commit = self.LATEST_RELEASE
```
bradenmacdonald: I'd rather here that you just override `set_field_defaults` and set `openedx_release = settings.LATEST_RELEASE`, and set `configuration_version`, then just call `super().set_field_defaults()`. Setting `edx_platform_commit` is unnecessary and you can just add a test for that, rather than writing redundant code to do that "in case behavior...changes"

instance/models/openedx_instance.py
```diff
@@ -160,10 +167,41 @@ def spawn_appserver(self):
...
+        # Misc settings:
+        self.use_ephemeral_databases = False
+
+        super().save(**kwargs)
+
+    def get_extra_settings(self):
+        """
+        Get configuration_extra_settings to pass to a new AppServer
+        """
+        template = loader.get_template('instance/ansible/prod-vars.yml')
```
bradenmacdonald: Instead of putting these into `get_extra_settings`, you could also change the OpenEdXAppServer code to change `CONFIGURATION_VARS_TEMPLATE` from a constant to a field, set to either 'vars' or 'prod-vars', and then just made the 'prod-vars.yml' template `{% include "vars.yml" %}`

instance/models/openedx_instance.py
```diff
@@ -160,10 +167,41 @@ def spawn_appserver(self):
...
+
+
+class OpenEdXProductionInstance(OpenEdXInstance):
+    """
+    OpenEdXProductionInstance: An OpenEdXInstance using production settings
+    """
+
+    LATEST_RELEASE = "named-release/dogwood.rc"
+
+    def save(self, **kwargs):
```
antoviaque: Also - sorry if I sound like a broken record on this, but I thought we agreed to only have different options/attributes on the Production instance, but no difference in logic. Adding methods to this class modifies the logic, even if it's to set default values. @bradenmacdonald 's suggestions already address how to move most of the logic on the parent class, though - just a note to make sure this class doesn't include methods at the end : )

instance/factories.py
```diff
@@ -0,0 +1,79 @@
...
+    production_instance = OpenEdXInstance.objects.create(**kwargs)
+    if "use_ephemeral_databases" not in kwargs:
+        production_instance.use_ephemeral_databases = False
+    if "configuration_version" not in kwargs:
+        production_instance.configuration_version = settings.LATEST_OPENEDX_RELEASE
+    if "openedx_release" not in kwargs:
+        production_instance.openedx_release = settings.LATEST_OPENEDX_RELEASE
+    if "configuration_extra_settings" not in kwargs:
+        template = loader.get_template('instance/ansible/prod-vars.yml')
+        production_instance.configuration_extra_settings = template.render({})
```
antoviaque: Moving the code to a factory method still carries this issue, btw - I can see this function growing over time. I'd rather add a way to make the defaults depend on class attributes that can be overridden in the proxy model, for example.

instance/models/openedx_instance.py
itsjeyd: @antoviaque I should have mentioned on the ticket that I don't intend to keep this class :) It is part of the approach that I implemented yesterday (in https://github.com/open-craft/opencraft/pull/71/commits/cc963ecddd47a313d4c31477099e5acc3c2f2e09). The reason I didn't remove it yet was to make it easier to review the second approach that I implemented today (in https://github.com/open-craft/opencraft/pull/71/commits/5eb1d3227bcb0d544f8df6f4415335d9e5601018).

instance/factories.py
itsjeyd: @antoviaque Why would the function grow over time? If we are going to eliminate the differences between sandboxes and production instances, we won't need this function anymore (or at least be able to reduce the number of settings that it customizes).

> I'd rather add a way to make the defaults depend on class attributes that can be overridden in the proxy model, for example.

What do you mean by "proxy model" here?

---

I have to admit that I am getting confused about what we are after here. Based on my current understanding, it is already possible to create instances with settings that are suitable for production; it currently just requires a few manual steps. So it seems like the goal is to be able to make it easier to create production instances (as I mentioned in my most recent [comment](https://tasks.opencraft.com/browse/OC-1554?focusedCommentId=22242&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22242) on OC-1554); and one way of achieving that is via factory functions (such as the ones I implemented in https://github.com/open-craft/opencraft/pull/71/commits/5eb1d3227bcb0d544f8df6f4415335d9e5601018). Once we unify the settings for sandbox and production instances (which is supposed to happen when implementing continuous deployment), we can just remove the factory functions. This should be easier than having to modify the class hierarchy again at that point (which would happen if we introduced a separate class for production instances now). Does that make sense, or am I off track here?

CC @bradenmacdonald

instance/models/openedx_instance.py
antoviaque: Yup, I saw the commits sequentially so I hadn't seen the factory approach when I posted this : )

instance/factories.py
antoviaque: @itsjeyd I'm saying this function will grow because I imagine someone seeing this while implementing another feature or sorting out a bug specific to a prod instance, and thinking this will be a convenient place to add specific behavior, without realizing that the goal is for it to disappear.

That said, we can watch for this in reviews, and I might be stretching it a bit given that someone could still add logic anywhere anyway. Maybe just add a comment there, to explain the eventual goal?

itsjeyd: @antoviaque Thanks for the clarifications, that makes sense about why the function would grow. Adding a comment to avoid that sounds good :)

instance/models/mixins/openedx_database.py
```diff
@@ -0,0 +1,123 @@
...
+    def set_field_defaults(self):
+        """
+        Set default values for mysql and mongo credentials.
+        """
+        if not self.mysql_provisioned:
+            self.mysql_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
+            self.mysql_pass = get_random_string(length=32)
+        if not self.mongo_provisioned:
+            self.mongo_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
+            self.mongo_pass = get_random_string(length=32)
```
bradenmacdonald: IMHO this should have `super().set_field_defaults()`. Same in the storage one.

instance/models/openedx_instance.py
```diff
@@ -0,0 +1,197 @@
...
+        if not self.edx_platform_repository_url:
+            self.edx_platform_repository_url = DEFAULT_EDX_PLATFORM_REPO_URL
+        if not self.edx_platform_commit:
+            self.edx_platform_commit = self.openedx_release
+
+        # Database settings
+        OpenEdXDatabaseMixin.set_field_defaults(self)
+
+        # Storage settings
+        OpenEdXStorageMixin.set_field_defaults(self)
```
bradenmacdonald: Is it necessary to call these mixins explicitly? I feel like they should be called via `super().set_field_defaults()` and python's MRO.

instance/models/mixins/openedx_database.py
itsjeyd: @bradenmacdonald None of the superclasses of `OpenEdXDatabaseMixin`/`OpenEdXStorageMixin` currently define `set_field_defaults`, so that doesn't seem to be possible.

instance/models/openedx_instance.py
itsjeyd: @bradenmacdonald It's necessary to call them explicitly because the leftmost superclass of `OpenEdXInstance` that defines `set_field_defaults` is `ValidateModelMixin`, and that's what `super().set_field_defaults()` ends up calling.

itsjeyd:
@bradenmacdonald  Just a heads-up that this is up-to-date with your `provision-while-running` branch again.

bradenmacdonald:
@itsjeyd Great, thanks. Hope the rebase was simple.

itsjeyd:
@bradenmacdonald Thanks, I ran into a couple conflicts but they were pretty easy to resolve :)

itsjeyd:
@bradenmacdonald This is up-to-date with your branch again.

instance/factories.py
```diff
@@ -0,0 +1,118 @@
...
+    # Create instance
+    production_instance = OpenEdXInstance(**kwargs)
+    if "use_ephemeral_databases" not in kwargs:
+        production_instance.use_ephemeral_databases = False
+    if "configuration_version" not in kwargs:
+        production_instance.configuration_version = settings.LATEST_OPENEDX_RELEASE
+    if "openedx_release" not in kwargs:
+        production_instance.openedx_release = settings.LATEST_OPENEDX_RELEASE
+    if "configuration_extra_settings" not in kwargs:
+        template = loader.get_template('instance/ansible/prod-vars.yml')
```
smarnach: Instead of only loading the template when no `configuration_extra_settings` are provided, can we merge them using the merge function in `ansible.py`?  It would be convenient to provide more settings without losing the defaults.

I also suggest reversing the logic here to avoid the somewhat repetitive ifs:

    extra_settings = ... merge code using kwargs.pop("configuration_extra_settings") ...
    instance_kwargs = dict(
        use_ephemeral_databases=False,
        configuration_version=settings.LATEST_OPENEDX_RELEASE,
        openedx_release=settings.LATEST_OPENEDX_RELEASE,
        configuration_extra_settings=extra_settings,
    )
    instance_kwargs.update(kwargs)
    production_instance = OpenEdXInstance(**instance_kwargs)

(Though this is probably just a matter of personal taste.)

bradenmacdonald: +1 for merging instead of the either-or approach to defaults.

instance/factories.py
```diff
@@ -0,0 +1,122 @@
...
+    # Please do not add behavior that is specific to production instances here.
+
+    # Ensure caller provided required arguments
+    assert "sub_domain" in kwargs
+
+    # Check environment and report potential problems
+    _check_environment()
+
+    # Create instance
+    production_instance = OpenEdXInstance(**kwargs)
```
bradenmacdonald: Is there a reason you use `.objects.create()` for sandboxes but instantiate-update-save() for the production factory? I think it would be easier to compare the two methods if you took the same approach as above here, i.e. update `instance_kwargs` then just call `.create(**instance_kwargs)`

instance/models/mixins/openedx_database.py
```diff
@@ -75,23 +75,17 @@ def mongo_database_names(self):
...
-    def provision_mongo(self):
-        """
-        Set mongo credentials and provision the database.
-        """
-        if not self.mongo_provisioned:
-            self.mongo_user = get_random_string(length=16, allowed_chars=string.ascii_lowercase)
-            self.mongo_pass = get_random_string(length=32)
-        return super().provision_mongo()
+        if not self.use_ephemeral_databases:
+            if not self.mysql_provisioned:
```
bradenmacdonald: Since this is now a `set_field_defaults` method, I think this `if` should be `if not self.mysql_user:` (instead of checking `if not self.mysql_provisioned`). Otherwise multiple calls to `set_field_defaults()` would change the values of `mysql_user` and `mysql_pass` every time (until the DB has been provisioned).

bradenmacdonald: Actually, I realized there is a problem with making this into `set_field_defaults`: If I create an instance with `use_ephemeral_databases=True` but then edit it so it has `use_ephemeral_databases=False` (which is totally legitimate), then `spawn_appserver()` will fail because `mysql_user` and `mysql_pass` will never be generated.

One fix for this could be to always generate `mysql_user` regardless of whether `set_field_defaults` is set. Since we want to eventually drop support for ephemeral databases entirely, I think that is reasonable - as long as it won't interfere with `edxlocal` creating local DBs in the ephemeral case.

instance/models/mixins/openedx_storage.py
```diff
@@ -54,11 +54,11 @@ def swift_container_names(self):
         """
         return [self.swift_container_name]
 
-    def provision_swift(self):
+    def set_field_defaults(self):
```
bradenmacdonald: Same comments apply here as for provision_mysql/set_field_defaults

bradenmacdonald:
@itsjeyd Code is looking good, besides the issues Sven and I commented on.

itsjeyd:
@bradenmacdonald Thanks! Comments addressed, and I also updated the README.

antoviaque:
@itsjeyd I think you might need to get the last review from someone else, as @bradenmacdonald is away until Thursday. @smarnach since you have also participated to this and use the dependent PR, are you able to give a thumb up to this?

bradenmacdonald:
@antoviaque I was hoping to get to this yesterday, but didn't. I do have a bit of time today. I will do a final review shortly.

smarnach:
@bradenmacdonald Thanks!  I just assigned myself as the second reviewer of this ticket, but if you can finish this, that's even better.

opencraft/settings.py
```diff
@@ -295,9 +295,9 @@
 # for provisioning Open edX app servers.
 DEFAULT_CONFIGURATION_VERSION = env('DEFAULT_CONFIGURATION_VERSION', default=DEFAULT_OPENEDX_RELEASE)
 
-# Git tag for latest Open edX release. Used as a default refspec for
+# Git ref for stable Open edX release. Used as a default refspec for
 # configuration, edx-platform, forum, notifier, xqueue, and certs when creating production instances.
-LATEST_OPENEDX_RELEASE = env('LATEST_OPENEDX_RELEASE', default='named-release/dogwood.rc')
+OPENEDX_RELEASE_STABLE_REF = env('OPENEDX_RELEASE_STABLE_REF', default='named-release/dogwood.rc')
```
bradenmacdonald: I'm curious if you tried `named-release/dogwood.3` ? Because that tag should also be available on all these repos (if using the `edx` fork, but not `open-craft` fork etc). And I think that's nicer.

No need to change this though.

bradenmacdonald:
👍  This looks great!

antoviaque:
@bradenmacdonald Cool, thanks! Hope you get to enjoy your holidays now : )

opencraft/settings.py
itsjeyd: @bradenmacdonald I hadn't tried `named-release/dogwood.3`, so leaving this as is for now.

itsjeyd:
@bradenmacdonald Thanks! Rebased on master and got a green build. Merging now.

PR #72, opened 2016-05-04
Eliminate warnings

mtyaka:
Removes the following warnings:

**djangular_tags deprecation** (`make test_unit` or `make rundev`)
```
/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/djng/templatetags/djangular_tags.py:4: UserWarning: Templatetags `djangular_tags` have been renamed to `djng_tags`.
  "Templatetags `djangular_tags` have been renamed to `djng_tags`."
```

**MySQL auth warnings** (`make test_unit`):

```
/home/vagrant/opencraft/instance/models/mixins/database.py:68: Warning: Using GRANT for creating new user is deprecated and will be removed in future release. Create new user with CREATE USER statement.
  (self.mysql_user, self.mysql_pass))

mysql: [Warning] Using a password on the command line interface can be insecure.
.mysql: [Warning] Using a password on the command line interface can be insecure.
..mysql: [Warning] Using a password on the command line interface can be insecure.
```

**OptionParser -> ArgumentParser warning** (`make rundev`)

```
09:58:17 worker.1    | /home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/django/core/management/__init__.py:345: RemovedInDjango110Warning: OptionParser usage for Django management commands is deprecated, use ArgumentParser instead
09:58:17 worker.1    |   self.fetch_command(subcommand).run_from_argv(self.argv)
```

**NoArgsCommand -> BaseCommand** (`make shell`)

```
/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/django/core/management/base.py:577: RemovedInDjango110Warning: NoArgsCommand class is deprecated and will be removed in Django 1.10. Use BaseCommand instead, which takes no arguments by default.
  RemovedInDjango110Warning
```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,13 @@ machine:
 dependencies:
   pre:
     - pip install --upgrade pip
+    - pip install --upgrade virtualenv
+    # Update python to latest 2.6.* release from the deadsnakes PPA.
+    # Ubuntu 14.04 which we currently use on CircleCI comes with 2.7.6 preinstalled,
+    # which does not support SNI. When CircleCI starts providing Ubuntu 16.04 images,
+    # this won't be needed anymore.
+    - sudo add-apt-repository ppa:fkrull/deadsnakes-python2.7 -y
+    - sudo apt-get update; sudo apt-get install python2.7 python2.7-dev
   override:
     - pip install -r requirements.txt
 test:
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   python:
     version: 3.5.1
+  services:
+    - redis
 dependencies:
   pre:
     - pip install --upgrade pip
--- a/instance/tests/models/test_instance_mixins.py
+++ b/instance/tests/models/test_instance_mixins.py
@@ -381,12 +381,14 @@ class MySQLInstanceTestCase(TestCase):
         databases = subprocess.check_output("mysql -u root -e 'SHOW DATABASES'", shell=True).decode()
         for database in self.instance.mysql_database_names:
             self.assertIn(database, databases)
-            mysql_cmd = "mysql -u {user} --password={password} -e 'SHOW TABLES' {db_name}".format(
+            # Pass password using MYSQ_PWD environment variable rather than the --password
+            # parameter so that mysql command doesn't print a security warning.
+            env = {'MYSQL_PWD': self.instance.mysql_pass}
+            mysql_cmd = "mysql -u {user} -e 'SHOW TABLES' {db_name}".format(
                 user=self.instance.mysql_user,
-                password=self.instance.mysql_pass,
                 db_name=database,
             )
-            tables = subprocess.call(mysql_cmd, shell=True)
+            tables = subprocess.call(mysql_cmd, shell=True, env=env)
             self.assertEqual(tables, 0)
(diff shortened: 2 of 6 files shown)


circle.yml
```diff
@@ -1,9 +1,14 @@
 machine:
   python:
     version: 3.5.1
+  services:
+    - redis
 dependencies:
   pre:
     - pip install --upgrade pip
+    - pip install --upgrade virtualenv
+    - sudo add-apt-repository ppa:fkrull/deadsnakes-python2.7 -y
```
jbzdak: Is this still needed since we want to use Xenial as a base box in CircleCI (I might be lacking context, so this is a genuine question).

instance/tests/models/test_instance_mixins.py
```diff
@@ -381,9 +381,11 @@ def check_mysql(self):
         databases = subprocess.check_output("mysql -u root -e 'SHOW DATABASES'", shell=True).decode()
         for database in self.instance.mysql_database_names:
             self.assertIn(database, databases)
-            mysql_cmd = "mysql -u {user} --password={password} -e 'SHOW TABLES' {db_name}".format(
-                user=self.instance.mysql_user,
+            # Pass password using MYSQ_PWD environment variable rather than the --password
+            # parameter so that mysql command doesn't print a security warning.
+            mysql_cmd = "MYSQL_PWD={password} mysql -u {user} -e 'SHOW TABLES' {db_name}".format(
```
jbzdak: `subprocess.call` allows you to pass `env` dictionary, you could pass variable there. This would increase security as to read environmental variables of another process you need to be an administrative user (AFAIK).

requirements.txt
```diff
@@ -15,7 +15,7 @@ debtcollector==1.3.0
 decorator==4.0.9
 dj-static==0.0.6
 Django==1.9.5
-django-angular==0.8.2
+git+https://github.com/jrief/django-angular.git@c21d7f013fa5108ba7ecc3c7a33de0251aafc604#egg=django-angular==0.8.3
```
jbzdak: Why hashes instead of versions, once again this is a genuine question). If the reason is that we want to use version that is not yet released, please add a comment: "Can be replaced with django-angular==0.8.3 when it is released".

jbzdak:
On the lighter note: I really liked the way that each commit was a separate change, with good message. This is something I should do more often :)

circle.yml
mtyaka: @jbzdak We can't use Xenial on CircleCI because it only supports Ubuntu 12.04 and 14.04 images at this time. Good idea about the comment - I'll mention this won't be needed anymore after upgrade to 16.04.

jbzdak: Ach! I mixed ubuntu versions :)

instance/tests/models/test_instance_mixins.py
mtyaka: @jbzdak Oh, I didn't know about the `env` parameter, nice! I'll do that.

requirements.txt
mtyaka: @jbzdak Good idea, done.

jbzdak:
:+1: Once the build is green.

PR #76, opened 2016-05-09
Run integration tests with credentials from environment

omarkhan:
The integration tests are currently only run if a `.env.integration` file is present. We run integration tests on CircleCI by copying this file over from the stage server before running the tests. We would like to store credentials on CircleCI directly, but for that to work the integration tests must be able to run without a `.env.integration` file. This pull request updates the Makefile to allow the integration tests to run with credentials from the environment, if the `OPENSTACK_USER` environment variable is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,12 +104,15 @@ test_browser: clean static_external
 	honcho -e .env.test run ./manage.py test --pattern=browser_*.py --noinput
 
 test_integration: clean
-	@if [ -a .env.integration ] ; then \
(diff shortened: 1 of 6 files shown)


omarkhan:
@pomegranited can I get your +1 on this or should I make some changes?

pomegranited:
@omarkhan :+1: Integration tests work fine.

bradenmacdonald:
Merging this now since it's blocking several people and it's ready to merge.

PR #77, opened 2016-05-09
Force SSL on deployed IM

jbzdak:
Only thing to review here is 22a6d0f882760878d0aee3528e6481f986341ea4 . Rest was reviewed by me as part in #68

### DIFF
--- a/deploy/requirements.yml
+++ b/deploy/requirements.yml
@@ -41,4 +41,4 @@
 
 - name: opencraft
   src: https://github.com/open-craft/ansible-opencraft
-  version: 'origin/smarnach/opencraft'
+  version: v0.0.1
--- a/deploy/.gitignore
+++ b/deploy/.gitignore
@@ -1,2 +1,3 @@
 private.yml
 opencraft.retry
+hosts
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -45,7 +45,9 @@ Running the playbook
 
         ansible-playbook opencraft.yml -u ubuntu --extra-vars @private.yml -i your.host.name.here,
 
-   (The trailing comma must be preserved.)
+   (The trailing comma must be preserved.). Note: you can't pass naked ip address to -i switch, long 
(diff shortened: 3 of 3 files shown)


deploy/requirements.yml
```diff
@@ -41,4 +41,4 @@
 
 - name: opencraft
   src: https://github.com/open-craft/ansible-opencraft
-  version: 'origin/smarnach/opencraft'
+  version: 'origin/jbzdak/ssl-rebased'
```
e-kolpakov: @jbzdak not sure if it is a problem or not, but shouldn't we use master here?

jbzdak: For now we plan to use commit hashes, but only for things that are merged to master, if you approve this one: https://github.com/open-craft/ansible-opencraft/pull/3 I'll update this to hash :)

deploy/requirements.yml
```diff
@@ -41,4 +41,4 @@
 
 - name: opencraft
   src: https://github.com/open-craft/ansible-opencraft
-  version: 'origin/smarnach/opencraft'
+  version: a3e9cde2968e9de94cb9e908244d05654def8dcd
```
e-kolpakov: @jbzdak same note - shouldn't it be `origin/master`?

e-kolpakov:
@jbzdak needs rebasing

deploy/requirements.yml
jbzdak: @e-kolpakov In most of the other repositories I pinned requirements to specific hash, just to have versioning. Does is make sense?

e-kolpakov: @jbzdak yes, but than it might make sense to actually do versioning, i.e. using git tags. Does it make sense?

jbzdak: @e-kolpakov I'm fine adding git tags for this purpose. I'll do this later today.

e-kolpakov:
@jbzdak :+1:

PR #78, opened 2016-05-17
Update opencraft ansible role

omarkhan:
Adds SSL support. See https://github.com/open-craft/ansible-opencraft/pull/3 and https://github.com/open-craft/ansible-opencraft/pull/5.


omarkhan:
@jbzdak can I get your +1 on this PR?

jbzdak:
:+1:

PR #79, opened 2016-05-17
Set up LMS preview subdomain for each new instance

mtyaka:
Recent edx-platform releases require that the LMS preview domain be different from the default LMS domain.

We use `preview.<instance-domain>` and set it as a CNAME record that points to `<instance-domain>`, which is very similar to what we do with the studio (sub)domain.


README.md
```diff
@@ -392,7 +392,7 @@ instance = SingleVMOpenEdXInstance.objects.create(
     sub_domain='dogwood.sandbox',
     name='Dogwood',
     fork_name='edx/edx-platform',
-    branch_name='named-release/dogwood',
+    branch_name='named-release/dogwood.rc',
```
antoviaque: @mtyaka Why switching to `dogwood.rc` here?

pylintrc
```diff
@@ -326,7 +326,7 @@ max-attributes=7
 min-public-methods=2
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=22
+max-public-methods=25
```
antoviaque: @mtyaka That's cheating : ) When we need to do that, that usually means a class has too much logic on it, and needs to be split. I haven't checked if that's going to solve this for you, but note that there has been some extensive refactoring of the instance class in https://github.com/open-craft/opencraft/pull/63 .

README.md
mtyaka: @antoviaque Because `named-release/dogwood` is not a branch, it's a tag. Using the tag didn't work for me because the github API expects a branch.

pylintrc
mtyaka: @antoviaque We already bumped that from 20 to 22 a while ago, so I thought why not bump it a bit more :) It might be better to disable the warning on the class instead, we already do that for a related warning: https://github.com/open-craft/opencraft/blob/53e65a3f3926f9f6e9ca17715eb6dea0e0cfd33d/instance/models/instance.py#L313

The three methods that I added are very simple, there's no logic in them, and they mirror similar methods for studio subdomains. I don't think code quality would be any better if they would be split out into a a mixin for no other reason than to keep the linter happy.

README.md
antoviaque: @mtyaka Do you know which call fails when using a tag? I thought it worked with tags (it's just a different ref_type).

mtyaka: @antoviaque The code failed when it was calling https://api.github.com/repos/edx/edx-platform/git/refs/heads/named-release/dogwood and the result was an array of objects instead of a single object (github api returns an array of all branches that start with the given string if it can't find the branch with the exact name, which is kind of weird).

I just checked and see that our code also supports tags, you just have to initialize the instance with `ref_type='tags`' (`ref_type` defaults to `heads`).
Would you prefer I add `ref_type='tags'` and keep `branch_name='named-release/dogwood'` in the README?

pylintrc
antoviaque: @mtyaka Yup, I know you're not causing the issue here - but when we encounter this type of situation, if we just disable pylint we are just leaving the issue for later; it still usually shows a problem - I would like to stick to the pylint rules, and make the necessary changes to get pylint happy rather than sliencing it. Sometimes it feels silly, but it makes it easier to notice/check for real issues when we disable less warnings, and it helps maintaining a more consistent code style.

In this case, just rebasing after #63 is merged might be all that is needed, but I still wanted to mention it, and ensure we don't bump this more than it already is.

On the topic of pylint, see also OC-1572, to address another part of the debt we have on pylint warning disabling.

CC @open-craft/core FYI, since that's something I mentioned in a few PRs already.

README.md
antoviaque: @mtyaka Thanks for checking! I think it might make sense to use the tag in the README, since that's what edX uses to mark its releases. That would be `named-release/dogwood.3` though at the moment, I think.

On the other side, we are going to be maintaining our own release branches to support the production instances, so it might make sense to use that here once we have it. @itsjeyd @smarnach Since you're working on related tickets (OC-1554 and OC-1556), what do you think? Maybe @mtyaka just uses the edx/edx-platform tag here, and you update it to our release branch of dogwood in one of your PRs once we have one?

mtyaka: @antoviaque I guess `branch_name` is not the best name when it can be something other than branch... Maybe we should think about renaming it to `ref_name`. Or perhaps use a separate property `tag_name` for when build is based on a tag rather than a branch?

itsjeyd: @antoviaque @mtyaka I suggest we leave the README alone for now: With the changes from #63, `branch_name` is no longer a field belonging to instances; and since that PR is going to land very soon (the only thing that's left is for @bradenmacdonald to update the README :D), there's no need to change/work on the instructions for manual provisioning in this context. (Even if this PR lands before #63, we'd only have to live with slightly outdated instructions for another day, which seems manageable.)

itsjeyd:
@mtyaka I had a look at the code, it LGTM :) Also checked preview functionality with your instance (mtyaka-dogwood), all good. Now I'm just waiting on another instance that I spun up from my own machine to finish provisioning. Final thumbs-up pending until I've tested the functionality on that instance, but will likely land today.

README.md
smarnach: The git command-line interface generally accepts a "rev" or revision specifier everywhere.  "Refs" or references are just one kind of "rev" (which is slightly confusing).  The particular endpoint of the Github API that we are using doesn't give that flexibility, but there is a [different endpoint](https://developer.github.com/v3/repos/commits/#get-a-single-commit) that does:

https://api.github.com/repos/edx/edx-platform/commits/named-release/dogwood.3
https://api.github.com/repos/edx/edx-platform/commits/named-release/dogwood.rc

If we use this endpoint, we won't need to specify the `ref_type` anymore, and any git "rev" will just work (i.e. everything documented in the [gitrevisions man page](https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html)), so I suggest doing that to avoid the confusion.  +1 on waiting until #63 is merged before any further changes.

antoviaque: @itsjeyd That would work for me to leave it alone on this PR, yep.

I don't see an update to the README in #63 though - so it will be even more out of date after #63 is merged?

antoviaque: @smarnach I posted before your answer appeared - that seem like a good suggestion, if changing the call doesn't have side effects which would add more work; we were just looking at updating the README originally, I wouldn't want this to bubble up and having to postpone the work. In which PR would this be done?

CC @bradenmacdonald FYI btw, since we are discussing your PR too at the end : )

itsjeyd: @antoviaque 

> I don't see an update to the README in #63 though - so it will be even more out of date after #63 is merged?

No, like I said

> the only thing that's left is for @bradenmacdonald to update the README

:)

cf. [my comment](https://github.com/open-craft/opencraft/pull/63/#issuecomment-219648634) on #63 about updating the README and [Braden's reply](https://github.com/open-craft/opencraft/pull/63/#issuecomment-219814653)

antoviaque: @itsjeyd Damn, sorry, I'm stupid here - I had missed that in your comment. That makes everything simpler. : ) Thanks, and sorry for the trouble.

bradenmacdonald: Hey guys,

A few points to add here:
* In #63, I have moved the `branch_name` attribute to a separate model that's just for pull requests. Pull requests are always based around branches, so there's no need to rename the attribute. What's more, I've identified [a simple TODO cleanup](https://github.com/open-craft/opencraft/pull/63/files#diff-4ccb9a3bd803e3e653eac6d28f3d5be4R103) that will eliminate the need to call any GitHub API to get the "tip commit" for a given ref, and at the same time fix the bug where OpenCraft IM breaks when a PR branch is deleted. (This is done by simple getting the sha1 hash directly from the PR details when querying the API for PRs, rather than making a separate call)
* After #63, instances and appservers have an `edx_platform_version` attribute, which can be any type of git ref - branch, commit, tag, etc. all work perfectly fine.
* The only catch is that the `named-release/dogwood` style tags are only on the `edx/edx-platform` fork, and you will get errors if you try to use those tags with the `open-craft` fork.

instance/models/instance.py
```diff
@@ -410,6 +410,27 @@ def ansible_swift_settings(self):
                                 'region': self.swift_openstack_region})
 
     @property
+    def lms_preview_sub_domain(self):
+        """
+        LMS preview sub-domain name (eg. 'preview.master')
+        """
+        return 'preview.{}'.format(self.sub_domain)
```
bradenmacdonald: I thought that we had agreed to use `preview-{}` instead of `preview.{}`? And I was hoping this would change the Studio URLs to `studio-{}` at the same time, though that's not a requirement.

See comment 3 and 4 on OC-1594.

bradenmacdonald:
If possible, I'd like to merge #63 before this one, since rebasing this on top of #63 should be straightforward, but the opposite will be more painful.

I also think we need to clarify the `preview-` vs `preview.` before this merges. (The latter is better for prod instances but the former is better for sandboxes, so we can use SSL)

itsjeyd:
@mtyaka Preview functionality worked well in my own instance. Based on @bradenmacdonald's recent comments I'll hold off on the thumbs-up until the you've rebased and the `preview-` vs `preview.` thing has been sorted out.

README.md
mtyaka: @bradenmacdonald That's excellent, thanks for the info. I'll remove the README change from this PR.

instance/models/instance.py
mtyaka: @bradenmacdonald Ah, sorry about that, it totally slipped my mind. I will change it, and I'll also change the studio subdomain in this same PR.

mtyaka:
Thanks for all the comments. I will change `preview.` to `preview-` and `studio.` to `studio-` and I will wait until #63 gets merged before merging this one.

mtyaka:
I rebased on top of master (which now includes changes from #63) and addressed comments.

instance/tests/api/test_openedx_appserver.py
```diff
@@ -176,7 +176,7 @@ def test_make_active(self, mock_set_dns_record):
         response = self.api_client.post('/api/v1/openedx_appserver/{pk}/make_active/'.format(pk=app_server.pk))
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data, {'status': 'App server updated.'})
-        self.assertEqual(mock_set_dns_record.call_count, 2)  # 2 calls: 1 for LMS, 1 for Studio
+        self.assertEqual(mock_set_dns_record.call_count, 3)  # 2 calls: 1 for LMS, 1 for LMS preview, 1 for Studio
```
itsjeyd: @mtyaka 1 + 1 + 1 = ... 2? :) The comment should say `# 3 calls: ...`

mtyaka: @itsjeyd Ooops, fixed; thanks :)

itsjeyd:
@mtyaka Verified the feature on previewtest5, all good :+1:

PR #80, opened 2016-05-17
Add login flow

omarkhan:
#### Testing instructions

1. Checkout this branch
2. Run migrations
3. `make rundev`
4. Go to `/`. Should redirect to `/registration/`
5. Fill in and submit registration form
6. You should be redirected back to the registration form. Verify that:
  - You are logged in as the new user
    - Email address and 'logout' link appear at the top right
    - Session cookie set
  - There is an 'application successfully sent' message
  - There is a 'pending email confirmation' message
  - The form is populated
  - The 'Full name', 'Your project' and 'Subscribe to updates' fields are editable, all other fields are read only
  - Read only fields cannot be modified by bypassing the form and sending a manual post request
7. Click the 'logout' link at the top right. Verify that you are logged out
8. Click the 'login' button at the top right. Verify that client-side validation of login form works
9. Login with the username + password you registered with in step 5. Verify that you see the same form as in step 6
10. Logout and click 'login' again
11. Login with the email + password you registered with in step 5. You should be able to log in with either username or email address
12. Search through the `make rundev` console output for the email verification emails. There should be one email sent to the user's email address, and another to the 'public contact email' address (or only one email if the same address was used for both). Click the links in both emails
13. Refresh the registration form. Verify that the 'pending email confirmation' message disappears


opencraft/urls.py
```diff
@@ -26,18 +26,15 @@
...
     url(r'^admin/', include(admin.site.urls)),
-    url(r'^api/', include('api.urls', namespace="api")),
-    url(r'^instance/', include('instance.urls', namespace="instance")),
-    url(r'^beta/', include('betatest.urls', namespace='beta')),
+    url(r'^api/', include('api.urls', namespace='api')),
+    url(r'^instance/', include('instance.urls', namespace='instance')),
     url(r'^email-verification/', include('email_verification.urls', namespace='email-verification')),
     url(r'^favicon\.ico$', RedirectView.as_view(url='/static/img/favicon/favicon.ico', permanent=False)),
-    url(r'^$', instance.views.index),
+    url(r'^', include('registration.urls', namespace='registration')),
```
antoviaque: Would it make sense to put the registration on its own url (like `/register`) and just redirect the home there, so we don't need to move it again when it gets a proper home page?

registration/tests/test_views.py
```diff
@@ -32,7 +32,7 @@
 from django.contrib.auth.models import User
 from django.test import TestCase
 
-from betatest.models import BetaTestApplication
+from registration.models import BetaTestApplication
```
antoviaque: I noticed that the `BetaTestApplication` and `UserProfile` models doesn't inherit from `ValidateModelMixin, TimeStampedModel` - for consistency could you add them?

omarkhan: @antoviaque done

opencraft/urls.py
omarkhan: Done

omarkhan:
@e-kolpakov this is ready for a first look. Still some work to do (tests mostly) but please have a look and I will finish this on Monday.

registration/backends.py
```diff
@@ -0,0 +1,52 @@
...
+
+class ModelBackend(backends.ModelBackend):
+    """
+    Extends the default ModelBackend to fetch users by email address as well as
+    by username.
+    """
+    def authenticate(self, username=None, password=None):
+        """
+        This is mostly copied from the default ModelBackend. Attempts to fetch
+        users by username or email address.
```
e-kolpakov: @omarkhan could be helpful to mention what's different; otherwise it is not clear why can't we use default ModelBackend (actually, can't we?)

registration/views.py
```diff
@@ -17,49 +17,60 @@
...
-from simple_email_confirmation.models import EmailAddress
+from registration.forms import BetaTestApplicationForm
 
 
 # Views #######################################################################
 
-class BetaTestApplicationView(CreateView):
+@method_decorator(sensitive_post_parameters('password',
+                                            'password_confirmation'),
+                  name='post')
```
e-kolpakov: @omarkhan nit: maybe move `sensitive_post_parameters` to next line:

    @method_decorator(
        sensitive_post_parameters('password', 'password_confirmation'), name='post'
    )

registration/backends.py
smarnach: The second sentence states what's different:  The original backend only allows login by username, while this one allows both username or email address.

registration/templates/registration/registration.html
```diff
@@ -0,0 +1,42 @@
...
+                    successfully sent! We will contact you after reviewing it.
+                </div>
+                {% if not form.instance.email_addresses_verified %}
+                    <div class="alert-box warning">
+                        Application status: pending email confirmation. You
+                        should receive verification emails at both of the
+                        email addresses you provided. Please click on the
+                        links in the emails to confirm your registration. Be
+                        sure to check your spam folder if you can't find them,
+                        or <a href="">contact us</a>.
```
antoviaque: @omarkhan You can use `contact@opencraft.com` here.

omarkhan: Oops thanks @antoviaque, fixed

registration/views.py
omarkhan: Done

registration/backends.py
omarkhan: I have updated the docstring to make this clearer.

e-kolpakov:
@omarkhan :+1:, but needs rebasing

PR #81, opened 2016-05-19
Make instance deletion more robust (part of OC-1541)

bradenmacdonald:
I noticed that after my refactor in #63, it's possible to delete an `OpenEdXInstance` but leave behind an `InstanceReference`, or vice versa. The syntax for deleting all of an instance's VMs was also more complicated.

With this changeset, deleting either an `Instance` or an `InstanceReference` always deletes the other object (unless using bulk deletion through the ORM). Also, calling `instance.delete()` will delete the instance as well as the VMs of all of its appservers.

**Author Concerns**
Is it too dangerous that one method can delete an instance and terminate all of its VMs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,11 +448,18 @@ Then use the "Launch new AppServer" button in the web UI to provision a server
 with the updated settings, and click "Activate this app server" to use the new
 server when it's ready.
 
-**To delete an instance**, ensuring that all virtual machines are terminated, run:
+**To terminate all VMs associated with an instance**, but still preserve the
+information about the AppServers and their configuration, run:
 
 ```python
 for appserver in instance.appserver_set.all():
     appserver.terminate_vm()
+```
+
+**To completely delete an instance**, which will also terminate all associated
+AppServers and their virtual machines, run:
+
+```python
 instance.delete()
 ```
 
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -24,10 +24,13 @@ OpenEdXInstance model - Tests
 
 from unittest.mock import call, patch, Mock
 
+import ddt
 from django.test import override_settings
 import yaml
 
 from instance.models.appserver import Status as AppServerStatus
+from instance.models.instance import InstanceReference
+from instance.models.openedx_appserver import OpenEdXAppServer
 from instance.models.openedx_instance import OpenEdXInstance, OpenEdXAppConfiguration
 from instance.models.server import Server
 from instance.tests.base import TestCase
@@ -40,6 +43,7 @@ from instance.tests.utils import patch_services
 # Factory boy doesn't properly support pylint+django
 #pylint: disable=no-member
 
+@ddt.ddt
 class OpenEdXInstanceTestCase(TestCase):
     """
     Test cases for OpenEdXInstance models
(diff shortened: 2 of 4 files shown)


antoviaque:
> Is it too dangerous that one method can delete an instance and terminate all of its VMs?

To the contrary - thanks for the change, it will make cleaning up old sandboxes simpler.

On the other side, I'll likely have to go track old app servers on existing sandboxes now, too : )

bradenmacdonald:
@antoviaque I should also mention that you can also terminate AppServers without deleting them. Just use `appserver.terminate_vm()` which will terminate the OpenStack VM, but preserve the AppServer object, so we can refer back to the log entries, configuration history, etc.

If we also had a way to hide Instances from the instance list (at first; there should be a button to "show hidden instances" or something), then I would say we should always just use that approach and never delete instances. It would give us a lot more data and history, which could be helpful.

I think we should add a story to add a UI button to terminate VMs, which would be a relatively easy addition.

antoviaque:
@bradenmacdonald Yup, I agree it will be better to not delete instances, especially for production instances, where keeping history is particularly important.

For the button, it could indeed be a good addition. I would hold off for now to finish the MVP features first - it only really affect me so far, when I have to clean up VMs. We'll also need to think about the UX to allow users to access it, as it's a little complex for non-technical users in its current form, so that could be something to include in the reflection (basically, my goal is to require as little intervention or reflection from the user as possible - so ideally we would find a way to not need that button at all, and deprovision automatically.)

README.md
```diff
@@ -448,11 +448,10 @@ Then use the "Launch new AppServer" button in the web UI to provision a server
 with the updated settings, and click "Activate this app server" to use the new
 server when it's ready.
 
-**To delete an instance**, ensuring that all virtual machines are terminated, run:
+**To delete an instance**, which will also terminate all associated AppServers
+and their virtual machines, run:
 
 ```python
-for appserver in instance.appserver_set.all():
-    appserver.terminate_vm()
```
itsjeyd: @bradenmacdonald I think it would be worth mentioning in the README that it's possible to terminate VMs associated with an instance, but keep its app servers (and the instance itself) around. Maybe just add another section that starts like this:

```
**To terminate all VMs associated with an instance** ...
```

... and add this code for terminating VMs back in.

instance/tests/models/test_openedx_instance.py
```diff
@@ -241,3 +245,31 @@ def test_spawn_appserver_with_external_databases(self, mocks, mock_provision):
...
+
+        # Now delete the instance, either using InstanceReference or the OpenEdXInstance class:
+        if delete_by_ref:
+            instance_ref.delete()
+        else:
+            instance.delete()
+
+        self.assertEqual(mock_terminate_vm.call_count, 1)
+        with self.assertRaises(OpenEdXInstance.DoesNotExist):
+            OpenEdXInstance.objects.get(pk=instance.pk)
```
itsjeyd: @bradenmacdonald Could you use `instance.refresh_from_db()` here as well, to be consistent with what you're doing for `instance_ref` and `appserver`?

itsjeyd:
@bradenmacdonald Reviewed, all is well functionality-wise. The code looks good as well, I just had a couple comments about the README and the tests.

instance/tests/models/test_openedx_instance.py
bradenmacdonald: @itsjeyd I did that at first, but `OpenEdXInstance` overrides `refresh_from_db()` to also refresh the `InstanceReference` at the same time, so that ended up generating the `InstanceReference.DoesNotExist` exception instead of the `OpenEdXInstance.DoesNotExist` that I wanted to test here. So I have to do it this way to be more direct.

README.md
bradenmacdonald: Sure, added. Can I get a +1 now?  :)

bradenmacdonald: Edit: added in cba5216d2ccce9d3c4da970eb3f4949d224feeb0 .

I'm going to rebase and squash now because GitHub won't let me merge as-is.

itsjeyd:
:+1: once the build is green.

PR #82, opened 2016-05-19
Add postgresql-client

bradenmacdonald:
Because otherwise you can't run `manage.py dbshell`.

### DIFF
--- a/debian_packages.lst
+++ b/debian_packages.lst
@@ -7,3 +7,4 @@ python3-pip
 redis-server
 subversion
 xvfb
+postgresql-client


antoviaque:
@bradenmacdonald Is this still something that would be useful?

bradenmacdonald:
@antoviaque Yep.

```
~/opencraft$ honcho run python3 manage.py dbshell
CommandError: You appear not to have the 'psql' program installed or on your path.
```

antoviaque:
@bradenmacdonald Thanks - sounds good, added to the scope of OC-2024

PR #83, opened 2016-05-19
Fix "Update Instance from PR" button when PR source and target forks are different (OC-1541)

bradenmacdonald:
I noticed on our stage server that this button wasn't working:
screen shot 2016-05-18 at 8 25 12 pm

It seems that I confused the PR source fork with the PR target fork at some point, which caused this bug when using the API to update an instance based on its PR settings. (This did not affect the automatic creation of instances when a new PR was detected.)

I have in mind a nicer way to refactor this code, but for now this is a quick fix.


itsjeyd:
@bradenmacdonald Tested using [this PR](https://github.com/itsjeyd/edx-platform/pull/1). I was able to update the instance settings via the UI and launch a new app server with updated settings. Code LGTM as well. :+1:

PR #84, opened 2016-05-19
Set up databases for different services when using persistent databases

itsjeyd:
cf. [OC-1610](https://tasks.opencraft.com/browse/OC-1610)

**Dependencies**

* https://github.com/edx/configuration/pull/3052 (fixes issue where `edxlocal` role would fail due to missing dependencies when using non-local databases)
* https://github.com/open-craft/configuration/pull/5 (fixes issue affecting creation of user for ecommerce database in integration tests; *should merge before this PR* so we can go back to using [open-craft:integration](https://github.com/open-craft/configuration/tree/integration) branch for integration tests)

**Test instructions**

1. Prepare your `.env` by setting `INSTANCE_MYSQL_URL` and `INSTANCE_MONGO_URL` to values from corresponding settings from stage.console.opencraft.com `.env` file. Make sure your `.env` sets `SWIFT_ENABLE=true` (if it doesn't contain the `SWIFT_ENABLE` setting, that's fine too since it defaults to `true`).

2. Log in to https://horizon.cloud.ovh.net using credentials for "OVH - OpenStack - Horizon - Databases" (stored in KeePassX database). Go to "Access & Security" and add a rule to the `databases-mysql-im-stage` and `databases-mongodb-stage` security groups that grants your IP access to the corresponding databases.

3. Create a new instance from the shell using the following settings:

   ```python

       sub_domain="databases-test.sandbox",
       name='Databases Test',
       edx_platform_repository_url="https://github.com/edx/edx-platform.git",
       configuration_source_repo_url="https://github.com/open-craft/configuration.git",
       configuration_version="jill/edx-sandbox-remote",  # branch from related PR
       edx_platform_commit="master",
       openedx_release="master",
   )
   ```

4. Run the development server, go to http://localhost:5000/instance, find the instance created in the previous step and click "Launch new AppServer".

5. Once the status of the app server switches to "Configuring VM", copy its "VM IP Address" from the "Status" section. Using this IP address, repeat step 2 to grant the VM access to the external MySQL and MongoDB databases.

6. Wait for the provisioning process to finish.

### DIFF
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -24,7 +24,6 @@ Instance app model mixins - Database
 
 from django.conf import settings
 from django.db import models
-from django.utils.crypto import get_random_string
 import MySQLdb as mysql
 import pymongo
 
@@ -111,7 +110,7 @@ class MySQLInstanceMixin(models.Model):
             cursor, connection = _get_mysql_cursor()
 
             # Create migration user
-            _create_user(cursor, self.migrate_user, get_random_string(length=32))
+            _create_user(cursor, self.migrate_user, self._get_mysql_pass(self.migrate_user))
 
             # Create default databases and users, and grant privileges
             for database in self.mysql_databases:
@@ -120,7 +119,7 @@ class MySQLInstanceMixin(models.Model):
                 database_name = connection.escape_string(database["name"]).decode()
                 _create_database(cursor, database_name)
                 user = database["user"]
-                _create_user(cursor, user, get_random_string(length=32))
+                _create_user(cursor, user, self._get_mysql_pass(user))
                 privileges = database.get("priv", "ALL")
                 _grant_privileges(cursor, database_name, user, privileges)
                 _grant_privileges(cursor, database_name, self.migrate_user, "ALL")
@@ -129,11 +128,11 @@ class MySQLInstanceMixin(models.Model):
                     _grant_privileges(cursor, database_name, additional_user["name"], additional_user["priv"])
 
             # Create read_only user with appropriate privileges
-            _create_user(cursor, self.read_only_user, get_random_string(length=32))
+            _create_user(cursor, self.read_only_user, self._get_mysql_pass(self.read_only_user))
             _grant_privileges(cursor, "*", self.read_only_user, "ALL")
 
             # Create admin user with appropriate privileges
-            _create_user(cursor, self.admin_user, get_random_string(length=32))
+            _create_user(cursor, self.admin_user, self._get_mysql_pass(self.admin_user))
             _grant_privileges(cursor, "*", self.admin_user, "CREATE USER")
 
             self.mysql_provisioned = True
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -145,8 +145,9 @@ class MySQLInstanceMixin(models.Model):
         if settings.INSTANCE_MYSQL_URL_OBJ and not self.mysql_provisioned:
             cursor = _get_mysql_cursor()
 
-            # Create migration user
+            # Create migration and read_only users
             _create_user(cursor, self.migrate_user, self._get_mysql_pass(self.migrate_user))
+            _create_user(cursor, self.read_only_user, self._get_mysql_pass(self.read_only_user))
 
             # Create default databases and users, and grant privileges
             for database in self.mysql_databases:
@@ -157,13 +158,10 @@ class MySQLInstanceMixin(models.Model):
                 privileges = database.get("priv", "ALL")
                 _grant_privileges(cursor, database_name, user, privileges)
                 _grant_privileges(cursor, database_name, self.migrate_user, "ALL")
+                _grant_privileges(cursor, database_name, self.read_only_user, "ALL")
                 additional_users = database.get("additional_users", [])
                 for additional_user in additional_users:
                     _grant_privileges(cursor, database_name, additional_user["name"], additional_user["priv"])
-
-            # Create read_only user with appropriate privileges
-            _create_user(cursor, self.read_only_user, self._get_mysql_pass(self.read_only_user))
-            _grant_privileges(cursor, "*", self.read_only_user, "ALL")
 
             # Create admin user with appropriate privileges
             _create_user(cursor, self.admin_user, self._get_mysql_pass(self.admin_user))
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -48,7 +48,7 @@ def _create_database(cursor, database):
     """
     Create MySQL database
     """
-    cursor.execute('CREATE DATABASE `{0}` DEFAULT CHARACTER SET utf8'.format(database))
+    cursor.execute('CREATE DATABASE `{database}` DEFAULT CHARACTER SET utf8'.format(database=database))
 
 
 def _create_user(cursor, user, password):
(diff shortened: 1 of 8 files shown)


instance/models/mixins/database.py
```diff
@@ -69,28 +108,55 @@ def provision_mysql(self):
...
+            # Create default databases and users, and grant privileges
+            for database in self.mysql_databases:
                 # We can't use the database name in a parameterized query, the
-                # driver doesn't escape it properly. Se we escape it here instead
-                database_name = connection.escape_string(database).decode()
-                cursor.execute('CREATE DATABASE `{0}` DEFAULT CHARACTER SET utf8'.format(database_name))
-                cursor.execute('CREATE USER %s IDENTIFIED BY %s', (self.mysql_user, self.mysql_pass,))
-                cursor.execute('GRANT ALL ON `{0}`.* TO %s'.format(database_name), (self.mysql_user,))
+                # driver doesn't escape it properly. So we escape it here instead
+                database_name = connection.escape_string(database["name"]).decode()
```
mtyaka: Now that we have database creation neatly encapsulated inside `_create_database`, we should make that method take care of escaping the database name. I would move the line above into `_create_database` and pass it the raw (unescaped) named here.

instance/models/mixins/database.py
```diff
@@ -69,28 +108,55 @@ def provision_mysql(self):
...
+
+            # Drop default databases and users
+            for database in self.mysql_databases:
                 # We can't use the database name in a parameterized query, the
-                # driver doesn't escape it properly. Se we escape it here instead
-                database_name = connection.escape_string(database).decode()
-                cursor.execute('DROP DATABASE IF EXISTS `{0}`'.format(database_name))
-            cursor.execute('DROP USER %s', (self.mysql_user,))
+                # driver doesn't escape it properly. So we escape it here instead
+                database_name = connection.escape_string(database["name"]).decode()
```
mtyaka: Just like the `_create_database` method, I think it should be the `_drop_database` method itself that should do the escaping.

instance/models/openedx_instance.py
```diff
@@ -127,7 +127,7 @@ def database_name(self):
         # Escape all non-ascii characters and truncate to 64 chars, the maximum for mysql:
```
mtyaka: This comment is now outdated, we truncate to 50 chars. I think it would be good to explain why we are truncating to 50 rather than 64 chars.

instance/models/mixins/openedx_database.py
```diff
@@ -75,12 +117,58 @@ def mongo_database_names(self):
...
+        """
+        List of MySQL users with global privileges (i.e., privileges spanning multiple databases)
+        """
+        return self.migrate_user, self.read_only_user, self.admin_user
+
+    def _get_mysql_database_name(self, suffix):
+        """
+        Build unique name for MySQL database using suffix
+        """
+        return "{0}_{1}".format(self.mysql_database_name, suffix)
```
mtyaka: Maybe we could add an `assert` here to ensure the generated database name not longer than 64 characters (maximum supported by mysql).

instance/models/mixins/openedx_database.py
```diff
@@ -75,12 +117,58 @@ def mongo_database_names(self):
...
+
+    def _get_mysql_user_name(self, suffix):
+        """
+        Build unique name for MySQL user using suffix
+        """
+        return "{0}_{1}".format(self.mysql_user, suffix)
```
mtyaka: We could also `assert` that the generated username is not longer than 16 characters.

instance/models/mixins/openedx_database.py
```diff
@@ -48,11 +49,52 @@ def mysql_database_name(self):
...
-        """
-        return [self.mysql_database_name]
+    def mysql_databases(self):
+        """
+        List of mysql databases
+        """
+        return [
+            {
+                "name": self._get_mysql_database_name("ecommerce"),
+                "user": self._get_mysql_user_name("ecomm001"),
```
mtyaka: I'm not a fan of these usernames, even though I think the inspiration comes from the defaults in edx/configuration. I don't see the point of the `001` suffix. I would prefer to remove it unless that would cause issues.

instance/models/mixins/openedx_database.py
```diff
@@ -75,12 +117,58 @@ def mongo_database_names(self):
...
+        """
+        Build unique name for MySQL user using suffix
+        """
+        return "{0}_{1}".format(self.mysql_user, suffix)
+
+    def _get_mysql_pass(self, user):
+        """
+        Build unique password for user, derived from MySQL password for this instance and user
+        """
+        return str(uuid.uuid5(uuid.NAMESPACE_DNS, self.mysql_pass + user))
```
mtyaka: I don't know enough about cryptography to know whether uuid5 is the proper thing to use here. If I had to choose a hashing strategy, I would go with a SHA HMAC with `mysql_pass` being the key and `user` being the message, which I think is somewhat more secure than uuid5, which if I understand correctly is a single-pass sha1 hash.

mtyaka:
@itsjeyd I left a couple of minor comments, but the changes look very nice overall.

I think mysql.yml should include mysql credentials for other services.

itsjeyd:
@mtyaka Thanks for the nice suggestions! I integrated them all, and mysql.yml now includes appropriate credentials for relevant services.

This is ready for another look, and for manual testing (see PR description for instructions).

instance/models/mixins/openedx_database.py
```diff
@@ -75,6 +119,100 @@ def mongo_database_names(self):
...
+        return mysql_database_name
+
+    def _get_mysql_user_name(self, suffix):
+        """
+        Build unique name for MySQL user using suffix.
+
+        To generate a unique name, this method adds an underscore and the specified suffix
+        to the mysql_user of this instance.
+
+        mysql_user is 6 characters long.
```
mtyaka: You missed a line that says something like: "The maximum length of usernames in MySQL is 16 characters."

instance/models/mixins/openedx_database.py
```diff
@@ -75,6 +119,100 @@ def mongo_database_names(self):
...
+        encoding = "utf-8"
+        key = bytes(source=self.mysql_pass, encoding=encoding)
+        msg = bytes(source=user, encoding=encoding)
+        return hmac.new(key, msg=msg, digestmod=hashlib.sha256).hexdigest()
+
+    def _get_database_suffix(self, name):
+        """
+        Return suffix that differentiates database identified by name from databases for other services.
+        """
+        return remove_prefix(self.mysql_database_name, name)
```
mtyaka: Nit: I'm not sure it's worth extracting prefix removal to a separate `remove_prefix` function, it's only two lines and not reused anywhere.

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -88,6 +102,118 @@ def check_mysql_vars_not_set(self, instance):
...
+        suffix = "test"
+        database_name = self.instance._get_mysql_database_name(suffix)
+        self.assertEqual(self.instance._get_database_suffix(database_name), suffix)
+
+        with patch(
+            "instance.models.mixins.openedx_database.remove_prefix", return_value="dummy"
+        ) as patched_remove_prefix:
+            suffix = self.instance._get_database_suffix(database_name)
+            self.assertEqual(suffix, patched_remove_prefix.return_value)
+            patched_remove_prefix.assert_called_once_with(self.instance.mysql_database_name, database_name)
```
mtyaka: Nit: this tests internal implementation of `_get_database_suffix`, which I don't think is necessary. It shouldn't matter whether `_get_database_suffix` uses `remove_prefix` or some other method internally, as long as it returns the correct result (which is already tested above).

instance/models/mixins/openedx_database.py
itsjeyd: @mtyaka At one point I was using the prefix removal logic in two places, that's why I originally put it in a separate function. But now that that's not the case anymore I agree that it makes sense to remove it :)

mtyaka:
👍

PR #85, opened 2016-05-19
Provision production instances for beta test applicants.

smarnach:
Changes in this PR:

1. Allow blacklisting subdomains to make them unavailable in the beta test application form.
  **Testing**: Set `SUBDOMAIN_BLACKLIST` and try that the blacklisted domains are rejected on http://localhost:5000/beta/register.
2. Trigger provisioning of a beta tester instance as soon as a beta tester confirms all their email addresses.
  **Testing**: File a beta test application on http://localhost:5000/beta/register and confirm all email addresses (see the terminal output for confirmation links).  Verify on the console that a new instance is being provisioned.
3. Create a LMS user account on the instance by copying the beta tester user account from the instance manager.
  **Testing**: This depends on edx/configuration#3059 and edx/edx-platform#12519, so to test this we currently have to create an instance pointing to the right release versions.  In the Django shell, this can be done using

          sub_domain='user-creation-test',
          edx_platform_repository_url='https://github.com/open-craft/edx-platform.git',
          edx_platform_commit='4a3cd60e5926dd8c360d8eb82bdd2ebddc0d77fa',
          configuration_source_repo_url='https://github.com/open-craft/configuration.git',
          configuration_version='smarnach/create-lms-users',
          openedx_release='master',
      )

    Then add an existing user and spawn an Appserver:

      instance.lms_users.add(User.objects.get(username='some_existing_user'))
      instance.spawn_appserver()

    Once the provisioning succeeds, you should be able to log into the new instance using the credentials of the local user you passed in.


instance/models/openedx_appserver.py
```diff
@@ -142,21 +149,42 @@ def set_field_defaults(self):
...
+        )
+
+    def lms_user_creation_playbook(self):
+        """
+        Return a Playbook instance for creating LMS users.
+        """
+        return Playbook(
+            source_repo=self.configuration_source_repo_url,
+            requirements_path='requirements.txt',
+            playbook_path='playbooks/{}.yml'.format(self.CONFIGURATION_PLAYBOOK),
```
bradenmacdonald: The playbooks in these methods are swapped. This should be `RUN_ROLE_PLAYBOOK`.

betatest/templates/betatest/welcome_email.txt
```diff
@@ -0,0 +1,26 @@
+{% autoescape off %}
+Hi {{ application.user.profile.full_name }},
+
+we are pleased to inform you that your application to participate in the
+OpenCraft Instance Manager beta test has been accepted.  Your Open edX instance
+has been provisioned and is available at
```
bradenmacdonald: "will be available ... soon" or "is available..." but not "is available ... soon"

instance/models/openedx_appserver.py
```diff
@@ -142,21 +149,42 @@ def set_field_defaults(self):
...
+            variables=self.configuration_settings,
+        )
+
+    def lms_user_creation_playbook(self):
+        """
+        Return a Playbook instance for creating LMS users.
+        """
+        return Playbook(
+            source_repo=self.configuration_source_repo_url,
+            requirements_path='requirements.txt',
```
bradenmacdonald: Too bad this will (presumably) set up a whole new virtualenv that's identical to the one already used for the first playbook. I guess that separation is important though.

betatest/templates/betatest/reject_email.txt
```diff
@@ -0,0 +1,10 @@
+{% autoescape off %}
+Hi {{ application.user.profile.full_name }},
+
+we are sorry to inform you that after careful consideration we were unable to
```
antoviaque: Capital "W" : ) -- "We ..."

betatest/templates/betatest/welcome_email.txt
```diff
@@ -0,0 +1,26 @@
+{% autoescape off %}
+Hi {{ application.user.profile.full_name }},
+
+we are pleased to inform you that your application to participate in the
```
antoviaque: Capital "W" : ) -- "We ..."

betatest/templates/betatest/welcome_email.txt
```diff
@@ -0,0 +1,26 @@
...
+"Forgot password?" link on the login page
+
+    {{ application.instance.url }}/login
+
+Since the OpenCraft Instance Manager is in beta state, there might still be some
+problems or downtime for your instance.  We appreciate any feedback on the
+project – just reply to this email!
+
+Happy testing,
+    the OpenCraft team
```
antoviaque: For these emails, can you switch to our standard email footer? Also it's nicer when emails are sent by a person, so you can switch to putting my email & name in the From field/signature?

These should likely be configurable btw, and default to fake emails/names, in case someone reuses this.

opencraft/settings.py
```diff
@@ -411,3 +415,7 @@
 
 # Limit the number of log entries fetched for each instance, for performance
 LOG_LIMIT = env.int('LOG_LIMIT', default=10000)
+
+# Subdomain blacklist #########################################################
+
+SUBDOMAIN_BLACKLIST = env.list('SUBDOMAIN_BLACKLIST', default=[])
```
antoviaque: Do you already know what you will set this to in our config?

opencraft/settings.py
```diff
@@ -295,6 +295,10 @@
 # for provisioning Open edX app servers.
 DEFAULT_CONFIGURATION_VERSION = env('DEFAULT_CONFIGURATION_VERSION', default=DEFAULT_OPENEDX_RELEASE)
 
+# Git tag for latest Open edX release. Used as a default refspec for
+# configuration, edx-platform, forum, notifier, xqueue, and certs when creating production instances.
+LATEST_OPENEDX_RELEASE = env('LATEST_OPENEDX_RELEASE', default='named-release/dogwood.rc')
```
antoviaque: Should we use the branch or the tag, at the end? Also, I don't see a dogwood branch on https://github.com/open-craft/edx-platform yet - what branch/ref will we be using on our fork?

Also, the latest Open edX release isn't really something which is configuration dependent - maybe this should be called `OPENEDX_RELEASE_STABLE_REF` or something like that, to reflect that it's more the release branch to use, rather than necessarily the last one edX released.

itsjeyd: @antoviaque I introduced this setting as part of https://github.com/open-craft/opencraft/pull/71. Changing the name to `OPENEDX_RELEASE_STABLE_REF` is a good idea - done in https://github.com/open-craft/opencraft/pull/71/commits/e200b556707ad8e31608b510552aa41dbb2a307a.

instance/models/openedx_appserver.py
smarnach: I had fixed that before, but emacs automatic buffer refreshing sometimes doesn't play well with git.  I haven't figured out the exact circumstances when this happens yet, and it's fortunately rare, but it's annoying anyway.

betatest/templates/betatest/welcome_email.txt
smarnach: @bradenmacdonald Feel free to suggest a better wording for these emails.  As a native speaker, you will probably be able to draft something nicer. :)

(The particular mistake you pointed out here is due to sloppy editing.  It was future before Xavier asked for appservers to be activated immediately, and I changed it to present tense, but missed the "soon".)

instance/models/openedx_appserver.py
smarnach: @bradenmacdonald I don't quite understand the last sentence in this comment.  What separation do you mean?

betatest/templates/betatest/reject_email.txt
smarnach: In German, we use a lower-case letter after the comma in the salutation, but you are right, in English it seems to be more common to use an upper-case letter here. :)

opencraft/settings.py
smarnach: @antoviaque No.  How about this list I just made up?

    www, mail, blog, web, ftp, ns1, ns2, smtp, test, opencraft, api, forum, admin

(Note that names shorter than three characters are disallowed anyway.)

instance/models/openedx_appserver.py
bradenmacdonald: @smarnach Before running the playbooks, the code in ansible.py creates a new python virtualenv and installs ansible and everything that's listed in `requirements.txt`. By separation, I mean that if we run multiple playbooks, it's possible that they come from different repos or different versions of the same repo, so the contents of requirements.txt could differ. Thus, it's important in theory that each time we run a playbook we provision a separate virtualenv.

In practice, the temporary virtualenv created for `CONFIGURATION_PLAYBOOK` here is going to be identical to the one created for `RUN_ROLE_PLAYBOOK`

betatest/templates/betatest/welcome_email.txt
bradenmacdonald: @smarnach I don't think I said this, but for the most part I really liked how the emails were written and I thought you did a great job with them. And yeah, I figured that sort of simple edit issue was what happened here but wanted to be clear what exactly the problem was I was pointing out.

instance/models/openedx_appserver.py
smarnach: @bradenmacdonald Ah, I see, you just meant the fact the different playbooks _potentially_ need different virtualenvs.  I don't think it's a big issue that the virtualenv needs to be regenerated, since it's rather small and typically takes around 30 seconds to set up.

opencraft/settings.py
antoviaque: @smarnach That would work for me as a start - we can always add to it as we think of new ones.

opencraft/settings.py
antoviaque: @itsjeyd Thanks!

opencraft/settings.py
```diff
@@ -295,6 +295,10 @@
 # for provisioning Open edX app servers.
 DEFAULT_CONFIGURATION_VERSION = env('DEFAULT_CONFIGURATION_VERSION', default=DEFAULT_OPENEDX_RELEASE)
 
+# Git ref for stable Open edX release. Used as a default refspec for
+# configuration, edx-platform, forum, notifier, xqueue, and certs when creating production instances.
+OPENEDX_RELEASE_STABLE_REF = env('OPENEDX_RELEASE_STABLE_REF', default='named-release/dogwood.rc')
```
antoviaque: @smarnach These questions remain open for you:

> Should we use the branch or the tag, at the end? Also, I don't see a dogwood branch on https://github.com/open-craft/edx-platform yet - what branch/ref will we be using on our fork?

https://github.com/open-craft/opencraft/pull/85#discussion_r64007865

opencraft/settings.py
smarnach: @antoviaque I would suggest to use our own branch name, like `opencraft/dogwood` for the Dogwood release we use on the our edx-platform fork, since we want to add custom commits to it.  I would build on the latest release tag `dogwood.3` and add whatever additional commits we need on top of that.  Same thing for the configuration repo.

smarnach: @antoviaque I just realised that this might be problematic, since we are using the same version for forum, xqueue, certs etc. as well, so we would have to fork all of them and create opencraft-specific branches.  Then it might be better to use the standard branchname `named-release/dogwood.rc` and add additional commits to it on our fork.  (I was worried that it might be confusing if the release branch on our branch is ahead of upstream, but I don't think it's too bad, and the alternative of using a new branch name seems worse.)

jbzdak:
NIT: You have added two configuration keys to settings, would be worthwile to add them to `Readme.md`

jbzdak:
@smarnach I have tested first two points, I'm waiting for full instructions for third one.

opencraft/settings.py
antoviaque: @smarnach Yup, good point - using `release/dogwood.rc` works for me.

smarnach:
@jbzdak I've rebased again on current master.  Could you please review the last three commits (assuming that you have already reviewed the earlier ones) and give me a thumbs up?

registration/forms.py
```diff
@@ -43,7 +43,7 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
     """
     class Meta:
         model = BetaTestApplication
-        exclude = ('user', 'status')
+        exclude = ('user', 'status', 'instance')
```
jbzdak: Is this a user visible change? Or just some post rebase artifact?

jbzdak:
@smarnach LGTM :+1:

registration/forms.py
smarnach: This was a change silently dropped by git during the rebase,  It already existed before rebasing.  Without this change, the form still kind of works, but behaves slightly weird, and some of the messages shown at the top of the form disappear.  I've got no idea why this happens, and it was a pain to debug, since I simply didn't expect git to silently drop changes (and I still don't no how this can have happened).

jbzdak: OK. Makes sense, go ahead and merge.

bradenmacdonald: @smarnach Yeah, having to rebase can sometimes be a bit annoying for reasons like that. What I like to do to be safe is merge master into my branch, tag the tip of the branch with a new branch name (to create an easily-accessible "backup" of the branch state), then rebase onto master. At that point, I can diff my rebased branch with the pre-rebase tip; they should be 100% identical, and if they're not, I know a problem happened in the rebase.

That's a bit more work, but is a safer pattern to use when rebases are being required IMHO.

smarnach: @bradenmacdonald In this particular case, I had quite a few merge conflicts, since an entire directory was renamed, and I would have needed to either resolve them twice or get `git rerere` to work correctly with them, both of which seemed too much hassle.

Even when rebasing, git shouldn't drop my changes just because the file isn't there anymore; and it usually doesn't, but in this particular case it somehow did.

mtyaka:
@smarnach I got an error when trying to reprovision an instance with persistent databases that might be related to the changes in this PR.
The failed appserver: https://stage.console.opencraft.com/instance/150/edx-appserver/41/

The error from the log:
```
django.core.exceptions.ValidationError: {'email': [u'It looks like verified@example.com belongs to an existing account. Try again with a different email address.']}
```

smarnach:
@mtyaka I'll look into that error as soon as possible (but in any case today).

smarnach:
@mtyaka I had a quick look now, and I think this is actually one of the errors I *fixed*.  See the PR description of https://github.com/edx/edx-platform/pull/12519 for details.  You can cherry pick the commits from that PR on your instance branch to work around this problem.

mtyaka:
@smarnach Thanks for the quick response! That makes sense. I don't actually need the instance with persistent databases now, so I switch to using another instance with ephemeral dbs.

PR #86, opened 2016-05-20
Minor UI tweaks

mtyaka:
This PR contains several UI tweaks.

### before:
screen shot 2016-05-20 at 10 15 12

### after:
screen shot 2016-05-20 at 10 14 53

### before:
screen shot 2016-05-20 at 10 39 43

### after:
screen shot 2016-05-20 at 10 39 23

### before:
screen shot 2016-05-20 at 11 41 50

### after:
screen shot 2016-05-20 at 11 38 55

### DIFF
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -352,7 +352,8 @@ class OpenStackServer(Server):
                 settings.OPENSTACK_SANDBOX_BASE_IMAGE,
                 key_name=settings.OPENSTACK_SANDBOX_SSH_KEYNAME,
             )
-        except novaclient.exceptions.ClientException:
+        except novaclient.exceptions.ClientException as e:
+            self.logger.error('Failed to start server: %s', e)
             self._status_to_build_failed()
         else:
             self.openstack_id = os_server.id
--- a/instance/static/html/instance/appserver.html
+++ b/instance/static/html/instance/appserver.html
@@ -119,15 +119,24 @@
       <i ng-attr-class="caret fa {{ $parent.isOpen ? 'fa-caret-down' : 'fa-caret-right' }}"></i>
       Log
     </accordion-heading>
-    <pre 
-      ng-if="appserver.status == 'failed'"
-      class="error-log"
-      ng-repeat="line in appserver.log_error_entries track by $index"
-      >{{ line.created | date:'yyyy-MM-dd HH:mm:ssZ' }} | {{ line.level }} | {{ line.text }}</pre><br>
-    <pre
-      class="instance-log"
-      ng-repeat="line in appserver.log_entries track by $index"
-      >{{ line.created | date:'yyyy-MM-dd HH:mm:ssZ' }} | {{ line.level }} | {{ line.text }}</pre>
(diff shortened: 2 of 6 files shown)


bradenmacdonald:
@mtyaka Nice! Yeah, I didn't have time to work on these little things myself, so I'm glad you could.

I did not test the code since I'm not the main reviewer here, but otherwise +1 from me.

My only comment is that it would be awesome if we could hide the long prefix from the log entries (`instance.models.appserver | instance=60 (Mtyaka Preview5),app_server=12 (AppServer 1) | `). That prefix makes sense when viewing the console log streams, but doesn't make sense in this context where every line has the same prefix and it's displaying information already displayed in the UI. Doing that in a clean way may be a larger change though, so we can definitely defer that for now.

antoviaque:
@mtyaka Cool! Thanks for this : )

smarnach:
@mtyaka Very nice!  I tested all changes locally and looked at the code, and all looks fine.

I particularly like the new log formatting.  Braden's improvements made the logs vastly more usable, but I find the visual improvements in this PR almost as important for imrproved readability.  It would be nice to visually distinguish the parts separated by the vertical lines each log line, but that would either require an extension of the `LogEntry` model, or some heuristic for parsing the log line on the client side, both of which would be more work than it is worth.

A few suggestions:

* Start with a closed accordion in the AppServer detail view.  This makes it easier to directly navigate to the desired section.
* Move the error log entries to a separate section in the accordion.  Currently, the log entries are shown in the "Log" section, followed by a `<br>`, followed by _all_ log entries, including a repetition of the error entries.  This is rather confusing, and the line break looks more like a bug than providing a clear visual separation between the two sections:

  image

  I'd prefer to have a section "Logged error messages" followed by a section "Full log" to make clear what the user is looking at.
* Make the shade of red used to highlight error message a bit brighter, and the foreground text a bit darker (or use black text) to improve contrast.

:+1: after considering my suggestions.

mtyaka:
@smarnach Thanks for the suggestions! I implemented all of them. I also implemented a custom angular filter to remove the redundant prefix from logs as @bradenmacdonald suggested. This makes the logs look muuuch better, so I think it was worth it; see screenshot:

screen shot 2016-05-22 at 13 54 04

The only problem is that sometimes the log contains empty lines (for example when ansible outputs multiple lines of output, some lines may be blank. I tried removing these lines from the output using `ng-init` and `ng-if`, but couldn't figure out how to do it without using some custom code in the angular controller, which I figured wasn't worth it, especially since we should probably not allow empty logs to get into the database in the first place.

It's out of the scope of this PR, but I was also wondering whether it would be possible to putting that prefix into the `LogEntry.text` column and instead save that metadata in a separate column on the `LogEntry`. I have very little experience with postgres, but I believe it supports a JSON column type which could be useful for storing this meta data.

smarnach:
@mtyaka Thanks.  I reviewed ans tested again, and everything looks good.  :+1:

mtyaka:
@smarnach Thanks for the quick response! I rebased and will merge once I get a green build.

mtyaka:
Build is green but I have to rebase again.

PR #88, opened 2016-05-21
Don't trigger an instance list update in the UI for each PR every minute (OC-1541)

bradenmacdonald:
I noticed on our stage server that whenever the GitHub watcher checks the status of a PR, the "Loading..." indicator appears in the web UI and the instance list is refreshed. This is because after #63, the PR watcher ends up updating instance properties and calling `.save()` on each time it checks a PR.

With this change, the periodic PR watcher only updates the instance when the PR is new. To update an instance from an existing PR, you have to do it manually by clicking the "Update instance settings from PR" button in the web UI.

This should reduce load on the server slightly and eliminate the "Loading..." flicker, while also making behavior more predictable.

Alternative: The alternative would be to keep the current behavior of automatically updating instance settings from the PR every minute, but only call `.save()` when the new settings are different. However that is trickier to implement. I also don't like it as much conceptually, because then it's less clear if/when settings get updated from a PR.

### DIFF
--- a/pr_watch/models.py
+++ b/pr_watch/models.py
@@ -50,16 +50,17 @@ class WatchedPullRequestQuerySet(models.QuerySet):
     Additional methods for WatchedPullRequest querysets
     Also used as the standard manager for the WatchedPullRequest model
     """
-    def update_or_create_from_pr(self, pr):
+    def get_or_create_from_pr(self, pr):
         """
-        Create or update an instance for the given pull request
+        Get or create an instance for the given pull request
         """
         watched_pr, created = self.get_or_create(
             fork_name=pr.fork_name,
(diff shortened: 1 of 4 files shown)


bradenmacdonald:
@itsjeyd is the reviewer on this but since he's away, any chance @smarnach @antoviaque or @mtyaka  can review?  The diff is only +13 -8 LOC :-)

smarnach:
@bradenmacdonald I reviewed the code and tested this locally (by enabling the PR watcher in the development server).  I could reproduce the flicker in the current master branch, and stared at the new version for one minute straight without blinking, and the flicker seems to be gone. :)

:+1:

bradenmacdonald:
Thanks @smarnach !

PR #89, opened 2016-05-21
Reinstate Jasmine tests + minor improvements (OC-1541)

bradenmacdonald:
Changes in this PR:
* Reinstates Jasmine tests + updates and expands them
* Refactor swampdragon event emitters to be part of model `save()` methods instead of using signals. This is a nice simplification suggested by @smarnach 
* Make the UI more responsive when selecting an Instance. Currently it loads the instance from the API before updating the UI. Now it updates the UI immediately and loads the instance in the background (as was already the case for AppServers). This issue wasn't noticeable when developing locally, but became quite annoying when testing on stage.


instance/tests/js/instance_app_spec.js
```diff
@@ -75,99 +106,337 @@ describe('Instance app', function () {
...
+                expect($scope.notification.message).toEqual('Update failed.');
+                expect($scope.notification.type).toEqual('alert');
+                $timeout.flush();
+                expect($scope.notification).toBe(null);
+            });
+        });
+    });
+
+
+    describe('Details controller', function() {
```
itsjeyd: @bradenmacdonald "Details" is a bit generic. Maybe go for "Instance Details controller" here? (If I understand correctly, you'd probably want/need to update the name of the actual controller too. I'm not sure how much work that would be, so feel free to skip this if it's not an easy update.)

instance/tests/fixtures/api/appserver_detail.json
```diff
@@ -0,0 +1,53 @@
...
+    "instance": {
+        "id": 50,
+        "api_url": "http://localhost:5000/api/v1/instance/50/"
+    },
+    "server": {
+        "id": 18,
+        "api_url": "http://localhost:5000/api/v1/openstackserver/18/",
+        "created": "2016-05-19T03:33:24.351037Z",
```
itsjeyd: @bradenmacdonald Is it possible for an `OpenStackServer` to be created before the parent app server? If not it would probably make sense to adjust all occurrences of this timestamp (or the `"created"`/`"modified"` data for the parent app server).

instance/tests/js/helpers/templates.js
```diff
@@ -0,0 +1,28 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(function(){
```
itsjeyd: @bradenmacdonald Nit: missing space between `)` and `{`

instance/tests/js/instance_app_spec.js
```diff
@@ -75,99 +106,337 @@ describe('Instance app', function () {
...
-                expect($scope.selected.instance.domain).toEqual('old.example.com');
+        describe('swampdragon events', function() {
+            beforeEach(function() {
+                spyOn($scope, 'updateInstanceList');
+            });
+            it('update the instance list whenever an instance is updated', function() {
+                swampdragon.sendChannelMessage({type: "instance_update"});
+                expect($scope.updateInstanceList).toHaveBeenCalled();
+            });
+            it('update the instance list whenever a AppServer is updated', function() {
```
itsjeyd: @bradenmacdonald "whenever an AppServer is updated"

itsjeyd:
@bradenmacdonald The code LGTM, I only had a few small comments. As for manual testing, are there any specific steps I should perform?

instance/tests/js/helpers/templates.js
bradenmacdonald: This format (with no space) is already used by all the other files in this folder. Normally I prefer the space but for this sort of outer scoping wrapper, I'm fine with no space.

instance/tests/js/instance_app_spec.js
bradenmacdonald: If it's OK, I'll just change it here for now. I think the controller code itself is pretty clear from its location in the source code. ("instance.js")

instance/tests/fixtures/api/appserver_detail.json
bradenmacdonald: @itsjeyd Yes, the `OpenStackServer` usually gets created by `AppServer.save()` before the actual `AppServer` subclass is saved. I took this fixture from the API of our stage server, so it should be quite realistic.

bradenmacdonald:
Thanks @itsjeyd. Changes addressed in 63f941039e20e93fb72f50ad932eb3339faab297, then I squashed and rebased.

For manual testing, just make sure the UI is still working, and test that when an instance/appserver's properties change, the UI is updated automatially. You can do this by running "make rundev" in one shell, and "make shell" in another shell, then making changes to an instance. Or just spawn a new AppServer.

itsjeyd:
@bradenmacdonald :+1:

mtyaka:
@bradenmacdonald Thanks for improving UI responsiveness, it feels faster now even though it still takes the same amount of time to load instance data.
Instances load slowly because instance logs are loaded together with other instance data. It would be great if the logs could be loaded separately, only when/if you click on the 'Log' tab. Same thing with app servers.

smarnach:
@mtyaka I suspect one reason for loading log data being slow is that the Postgres database is running in a different region than the instance manager itself.  @antoviaque suggested to only change this if it becomes a problem.  We probably don't need to try any other optimizations before we've moved the database (though what you suggest makes sense).

antoviaque:
@smarnach I would be surprised if the DB location was the bottleneck here - if I remember correctly they are still hosted in very close locations in Europe, with a very small ping and a large bandwidth between the two servers. By contrast, getting the log data to the client, and processing it, would be much slower imho.

smarnach:
@antoviaque I did a quick test, and you are right.  Retrieving the log entries of an instance is translated into a single database query, so the latency doesn't matter much.  (I couldn't measure the latency between the two regions, since neither of the two servers replies to pings, but it shouldn't be above 10ms judging from geographical distance.)

bradenmacdonald:
It looks to me like those large (some are 600KB+) log responses are not being compressed. (Is that right?) If so, I think it would really help to add the [`gzip_page()`](https://docs.djangoproject.com/en/1.9/topics/http/decorators/#django.views.decorators.gzip.gzip_page) decorator to any API calls that return log entries. (It's a bad idea to use GZIPMiddleware for all responses on the site due to things like the [BREACH attack](http://breachattack.com/) but for this sort of API response that's not user-specific and doesn't include any CRSF token in the response body it's totally fine).

PR #90, opened 2016-05-24
Don't break dns if an instance's base domain includes a subdomain part

omarkhan:
#### Background

Sandboxes provisioned on stage.console.opencraft.com cannot be accessed by url, only by ip address. On investigation, it turns out that the dns records are being set incorrectly, as they ignore any part of the subdomain that has been set as part of the instance's base domain. For example:

The `INSTANCES_BASE_DOMAIN` setting is set to `stage.opencraft.com`. When a new instance is created with a subdomain of `pr12338-2.sandbox`, 3 gandi DNS records are added:

- A: `pr12338-2.sandbox.opencraft.com` -> public IP
- CNAME: `preview-pr12338-2.sandbox.opencraft.com` -> `pr12338-2.sandbox.opencraft.com`
- CNAME: `studio-pr12338-2.sandbox.opencraft.com` -> `pr12338-2.sandbox.opencraft.com`

However, the instance expects to be reached at pr12338-2.sandbox.**stage.**opencraft.com.

#### Proposal

This pull request modifies `OpenEdXInstance.set_appserver_active` to use the [tldextract](https://github.com/john-kurkowski/tldextract) package to correctly separate the subdomain from the base domain when adding DNS records.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -25,6 +25,7 @@ from django.conf import settings
 from django.db import models, transaction
 from django.db.backends.utils import truncate_name
 from django.db.models.signals import post_save
+from tldextract import TLDExtract
 
 from instance.gandi import GandiAPI
 from instance.logging import log_exception
@@ -37,8 +38,12 @@ from .openedx_appserver import OpenEdXAppConfiguration, OpenEdXAppServer, DEFAUL
 
 gandi = GandiAPI()
 
-# Models ######################################################################
+# By default, tldextract will make an http request to fetch an updated list of
+# TLDs on first invocation. Passing suffix_list_urls=None here prevents this.
+tldextract = TLDExtract(suffix_list_urls=None)
 
+
+# Models ######################################################################
 
 # pylint: disable=too-many-instance-attributes
 class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin, OpenEdXStorageMixin):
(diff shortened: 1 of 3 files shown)


mtyaka:
Cool! Thanks for the explanation and a proper solution. 👍

smarnach:
I would have assumed that the `base_domain` field on the instance is the base domain used on Gandi – otherwise the separation into `base_domain` and `sub_domain` is rather arbitrary and meaningless, and we could merge the two fields into a single `domain_name` field.

Any opinions on this @omarkhan @mtyaka @antoviaque?

omarkhan:
It was not obvious to me that setting `INSTANCES_BASE_DOMAIN` to `stage.opencraft.com` would break DNS. If that is expected behaviour it should at lease raise `ImproperlyConfigured` or something. Then we would still need a way to specify a subdomain suffix for instances.

This is a smaller, simpler change that works. I agree that the name `sub_domain` is not completely appropriate, maybe we could rename it to `domain_prefix` or similar?

smarnach:
@omarkhan That's defintely not what I meant; I agree that `INSTANCES_BASE_DOMAIN` in the form it was implemented was kind of broken.  I was mainly pondering why we split the domain name on instances into two parts, if the two parts don't have any technical meaning.

mtyaka:
@smarnach I think having the two split will be useful for OC-1553 to be able to fetch correct zone file based on `base_domain`. I like @omarkhan's suggestion to rename `sub_domain` to `domain_prefix`.

smarnach:
@mtyaka I think it would only be useful for OC-1553 if `base_domain` really reflected the base domain to use for Gandi, but it doesn't.  You still would need to use `tldextract()` on `base_domain` and access the `registered_domain` property of the result.  The very same code would work if we only stored the full domain name on the instance instead, so I don't see any benefit of the split.

mtyaka:
@smarnach Ah yes, you are absolutely right. I actually don't see much value in storing the `base_domain` on the instance at all - more on that on OC-1553.

PR #91, opened 2016-05-25
Fix registration for existing users

omarkhan:
I noticed a bug in the beta registration after I deployed it to stage. If you are already logged in, but have not yet registered, the form will not allow you to register ('username is already taken'). This pull request fixes the bug by ignoring the logged in user when checking for existing usernames and emails.

#### Testing

1. Create a new user in the admin or the shell
2. Log in as that user
3. Go to `/registration/`. The username and email fields should be prepopulated and read only, and the password fields should not appear.
4. Fill in and submit the form

### DIFF
--- a/registration/tests/browser/browser_registration.py
+++ b/registration/tests/browser/browser_registration.py
@@ -60,18 +60,42 @@ class BetaTestBrowserTestCase(BetaTestApplicationViewTestMixin,
             path=reverse('registration:register'),
         )
 
+    @property
+    def form(self):
+        """
+        The form on the page.
+        """
+        return self.client.find_element_by_tag_name('form')
+
+    def fill_form(self, form_data):
+        """
+        Fill in the form with the given data.
+        """
+        for field, value in form_data.items():
+            element = self.form.find_element_by_name(field)
+            if element.get_attribute('type') == 'checkbox':
+                if bool(value) != element.is_selected():
+                    element.click()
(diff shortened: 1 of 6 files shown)


registration/tests/browser/browser_registration.py
```diff
@@ -60,18 +61,47 @@ def url(self):
...
+        """
+        Fill in the form with the given data.
+        """
+        for field, value in form_data.items():
+            try:
+                element = self.form.find_element_by_name(field)
+            except NoSuchElementException:
+                # Some form fields, like the password fields, only appear if
+                # the user is not already logged in.
+                continue
```
e-kolpakov: @omarkhan feels like a bad idea - shouldn't caller not pass field value in field_data in such case?

e-kolpakov:
@omarkhan All good and works as specified, but CI build fails; so :+1: conditional build passes.

One more thing - it is not related to the bug you're fixing, but instance/#/isntances have some minor UI glitch - double logout buttons:

image

Might be cool to fix it; or log a bug for later.

registration/tests/browser/browser_registration.py
omarkhan: Good point, fixed

omarkhan:
Thanks for the tip about the double logout buttons - I had removed it in the last PR but Braden added it back in another template before I merged it. I have removed it again, I will merge when the build is green.

PR #92, opened 2016-05-26
Support multiple gandi base domains

mtyaka:
This PR adds support for multiple base domains. The base domains have to be handled by the same Gandi account.

The Gandi client now accepts the base registered domain and uses the client API to automatically obtain the corresponding zone id, which makes the `GANDI_ZONE_ID` setting obsolete. The setting was removed in this PR.

This PR also renames the `INSTANCES_BASE_DOMAIN` setting to `DEFAULT_INSTANCE_BASE_DOMAIN` to better reflect its purpose.

cf. OC-1553


instance/gandi.py
```diff
@@ -52,50 +52,51 @@ def api_key(self):
...
         Client domain zone API endpoint
         """
         return self.client.domain.zone
 
-    def delete_dns_record(self, zone_version_id, record_name):
+    def get_zone_id(self, domain):
+        """
+        Gandi zone ID used by domain
+        """
+        return self.client.domain.info(self.api_key, domain)['zone_id']
```
smarnach: Since the result of this call will never change, it might make sense to cache it.  Since I don't think `functools.lru_cache` will work here, I'd simply store the zone IDs in a dictionary.

smarnach:
@mtyaka Code changes look good; I've left one minor comment.

This still needs manual testing by me, and testing on stage by you (to test that everything keeps working for the old instances).

I'll test this locally once my vagrant instance becomes available (it's currently busy with an Ansible run on a different branch for an unrelated test).  Feel free to deploy to stage in the meantime (coordinate with @antoviaque, who is currently reviewing the log-in flow in stage).

instance/gandi.py
mtyaka: Thanks @smarnach, yeah it makes sense to cache it. I pushed an update that caches the zone id in a dict.

mtyaka:
@antoviaque When I merge to master (after circleci tests pass and I get a +1 from @smarnach), is it ok to pull the changes to staging or should I wait until you finish testing the log-in flow?

antoviaque:
@mtyaka That's completely fine to update stage - I was only testing the registration/login flow anyway, so you wouldn't interrupt any provisioning or anything, and updates are very short downtimes most of the time.

smarnach:
@mtyaka I tested this locally, and all is working fine.  The changes to cache the zone ID look good, too.  :+1:

mtyaka:
Thanks!

PR #93, opened 2016-05-26
Apply SASS styles from agency

omarkhan:
#### Testing

1. `make rundev`
2. Go to `/registration/`, check that everything works and looks right
3. Try to log in, make sure everything works and looks right
4. Go to the instance manager, make sure everything works and looks right


registration/models.py
```diff
@@ -83,14 +83,14 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
     )
     instance_name = models.CharField(
         max_length=255,
-        help_text=('The name of your institution, company or project.'
-                   '\n\nExample: Hogwarts Online Learning'),
+        help_text=('The name of your institution, company or project.<br/>'
+                   'Example: Hogwarts Online Learning'),
```
antoviaque: Would it be possible to avoid using HTML in the help text?

registration/templates/registration/form_checkbox_field.html
```diff
@@ -1,13 +1,14 @@
-<div class="row">
-    <div class="large-6 large-offset-1 columns end">
+<div class="page-content__inner__wrapper">
+    <div class="page-content__inner__left">
         {{ field.errors }}
+        <span class="wpcf7-form-control-wrap">
```
antoviaque: +1 to removing the wordpress classes, although I know you already have it in mind

registration/models.py
omarkhan: For line breaks yes, we will just have to add 'white-space: pre-line' to
the css instead. But one of them contains a link, we have to use html for
that.

requirements.txt
```diff
@@ -26,7 +26,6 @@ django-grappelli==2.8.1
 django-libsass==0.6
 django-redis==4.4.2
 -e git+https://github.com/kcsry/django-simple-email-confirmation.git@2df833cb3736264aac46f661c48bc2abdcbaa18f#egg=django-simple-email-confirmation
-django-zurb-foundation==5.5.0
```
antoviaque: Why did you remove the foundation dependency?

omarkhan: Because I saw bootstrap was used in the sass and it didn't make sense to
use both. Now that I know that we are not switching to bootstrap, I will
put it back in.

static/scss/_breakpoint-slicer.sass
```diff
@@ -0,0 +1,8 @@
+// Default values
+@import breakpoint-slicer/variables
+
+// Helper functions
+@import breakpoint-slicer/helper-functions
+
+// The magic!
+@import breakpoint-slicer/mixins
```
antoviaque: This seem to be an external dependency - to download instead of commiting a copy of the files in git? https://github.com/lolmaus/breakpoint-slicer

static/scss/_normalize.scss
```diff
@@ -0,0 +1,419 @@
+/*! normalize.css v4.1.1 | MIT License | github.com/necolas/normalize.css */
+
```
antoviaque: Same remark - external dependency.

static/scss/abstracts/_animations.scss
```diff
@@ -0,0 +1,118 @@
+/**
+ * ANIMATIONS
```
antoviaque: The `abstracts` also looks like an external dep?

static/scss/base/_typography.scss
```diff
@@ -0,0 +1,83 @@
+/**
+ * TYPOGRAPHIE
+ */
+
+// ROBOTO MONO
+@import url(https://fonts.googleapis.com/css?family=Roboto+Mono:400,100,100italic,300,300italic,400italic,500,500italic,700,700italic);
+
+// PLAYFAIR DISPLAY
+@import url(https://fonts.googleapis.com/css?family=Playfair+Display:400,400italic,700,700italic,900,900italic);
```
antoviaque: Can these be loaded out of our assets, instead of Google?

static/scss/breakpoint/_context.scss
```diff
@@ -0,0 +1,95 @@
+//////////////////////////////
+// Private Breakpoint Variables
+//////////////////////////////
```
antoviaque: And `breakpoint` seem might be https://github.com/at-import/breakpoint ?

static/scss/components/_components.scss
```diff
@@ -0,0 +1,265 @@
...
+	position: relative;
+	padding-bottom: 100%;
+
+	// &::after
+	// {
+	// 	display: block;
+	// 	position: absolute;
+	// 	padding-bottom: 100%;
+	// 	content:'';
+	// }
```
antoviaque: Commented code

static/scss/vendors/_bootstrap-custom.scss
```diff
@@ -0,0 +1,68 @@
+/*!
+ * Bootstrap v3.3.6 (http://getbootstrap.com)
+ * Copyright 2011-2015 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
```
antoviaque: I'd like to avoid switching to bootstrap, and not have to add it to the dependencies - we would be adding technical debt given that the rest of the interface is using Foundation.

templates/base.html
```diff
@@ -10,13 +10,12 @@
     <title>{% block title %}{% endblock title %} - OpenCraft -</title>
     {% endblock base_title %}
 
-    {# Base Foundation CSS #}
-    {% block foundation_css %}
+    {# Base CSS #}
+    {% block base_css %}
     {% compress css %}
-    <link href='{% static "foundation/scss/normalize.scss" %}' type="text/x-scss" rel="stylesheet"/>
-    <link href='{% static "foundation/scss/foundation.scss" %}' type="text/x-scss" rel="stylesheet"/>
```
antoviaque: That would break the rest of the instance manager interface, no?

static/scss/base/_typography.scss
omarkhan: Sure, done in https://github.com/open-craft/opencraft/pull/93/commits/1a509be1e7ce747ecff2ccbe5d4b0f16c6fd955b

registration/static/js/src/registration.js
```diff
@@ -49,20 +64,22 @@ app.controller('Registration', ['$scope', '$http', 'djangoForm', function($scope
         return _.keys($scope.form).filter(key => !/^\$/.test(key));
```
mtyaka: @omarkhan The `=>` function notation does not have very good browser support yet. It doesn't work with latest Safari on OS X, for example.

mtyaka:
👍

static/external/scss/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Variables ###################################################################
+
+SVN = svn checkout
```
smarnach: Wouldn't `svn export` be the more appropriate command here?  (I only vaguely remember `svn` commands, and I always got them wrong, so don't give too much on what I'm saying.)

mtyaka: @smarnach Yeah, `svn export` would be better, but I think it only allows you to export the full repository (top-level folder) while here we want to only dowload the stylesheets subfolder.

smarnach: @mtyaka I just ran

    svn export https://github.com/at-import/breakpoint/tags/v2.7.0/stylesheets

on my machine, and it worked fine.

mtyaka: @smarnach Hmm, you're right it works for me, too. I think there was some problem with `export` though. I'll let @omarkhan answer :)

omarkhan: Thanks for the tip @smarnach. I have never used subversion before, I didn't know about `svn export`. I have updated the Makefile to use `svn export --force` (without `--force` it refuses to download into the current directory)

omarkhan:
Rebased, I will merge when the build is green.

PR #94, opened 2016-05-27
Show different message for each application status

omarkhan:
The message "Your application for the OpenCraft beta has been successfully sent! We will contact you after reviewing it." still shows even when the user application is in the "accepted" or "refused" status. This should only display while the application is pending.

This pull request adds different messages for the 'pending', 'accepted' and 'refused' statuses.

### DIFF
--- a/registration/templates/registration/registration.html
+++ b/registration/templates/registration/registration.html
@@ -7,31 +7,44 @@
 <div class="container content" ng-app="RegistrationApp">
     <div class="row">
         <div class="large-12 columns">
-            <h1>Register for Beta</h1>
-            {% if form.instance.pk %}
-                <div class="alert-box success">
-                    Your application for the OpenCraft beta has been
-                    successfully sent! We will contact you after reviewing it.
-                </div>
-                {% if not form.instance.email_addresses_verified %}
-                    <div class="alert-box warning">
-                        Application status: pending email confirmation. You
-                        should receive verification emails at both of the
-                        email addresses you provided. Please click on the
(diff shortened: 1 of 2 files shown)


registration/templates/registration/registration.html
```diff
@@ -7,31 +7,44 @@
...
+                            sure to check your spam folder if you can't find them,
+                            or <a href="mailto:contact@opencraft.com">contact us</a>.
+                        </div>
+                    {% endif %}
+                {% else %}
+                    <div class="alert-box info">
+                        This is the application form for the beta of the
+                        self-service hosting offered by OpenCraft. For enterprise
+                        or institutional offerings,
+                        <a href="mailto:contact@opencraft.com">contact us</a>.
```
antoviaque: The content of the messages LGTM - thanks!

jbzdak:
@omarkhan  Looks good! Works as advertised. 

Some NIT's: 

1. I have repased this on top of master and tested that. 
2. I'm lacking a lot of context here, so this could have been discussed, or split to follow-up task: it would be good to add some additional information to: "Thank you for applying for the OpenCraft beta. Your application has been accepted!", so it is immediately obvious to the user what to do next. like: "Please check your e-mail inbox for details regarding the instance" or "Go to {{ instance_url }} and login using the same login/password as you use here".

omarkhan:
Thanks @jbzdak. @antoviaque has already signed off on the messages so I will leave them as they are for now. We can update them in a separate task.

PR #96, opened 2016-05-29
Redirect staff users to instance manager on login

omarkhan:
Staff users are currently redirected to the beta registration page on login. This pull request redirects staff users to the instance manager instead.

### DIFF
--- a/opencraft/urls.py
+++ b/opencraft/urls.py
@@ -26,6 +26,8 @@ from django.conf.urls import include, url
 from django.contrib import admin
 from django.views.generic.base import RedirectView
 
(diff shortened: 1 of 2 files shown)


antoviaque:
@omarkhan For the permissions, cf my comment on OC-1551, this is blocked on another permission. I'd rather not add another approach in the meantime, it's just a question of going to the right URL. to access the instance list - just let people know before you deploy to prod.

omarkhan:
@bradenmacdonald FYI

bradenmacdonald:
Ok. @antoviaque's comment above makes sense to me.

PR #97, opened 2016-05-29
Don't prefix IM angular routes with /instances/

omarkhan:
The instance manager django app now lives at `/instance/`, which means that we end up with routes like `/instance/#/instances/50`. This pull request removes the `/instances/` prefix from the angular routes and turns on the html5 pushState mode, turning that url into the less redundant `/instance/50`.

### DIFF
--- a/instance/static/js/src/instance.js
+++ b/instance/static/js/src/instance.js
@@ -33,15 +33,15 @@ app.config(function($httpProvider) {
 });
 
 app.config(function($stateProvider, $urlRouterProvider, RestangularProvider) {
-    // For any unmatched url, send to /instances/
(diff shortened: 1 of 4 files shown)


bradenmacdonald:
LGTM but I'd prefer that you include [the HTML5 change](https://gist.github.com/bradenmacdonald/51795b51a82fd4a045cbd53e0c9e2771) at the same time.

omarkhan:
@bradenmacdonald done, see https://github.com/open-craft/opencraft/pull/97/commits/859357ecd971c6616fb7a8517d86f9cf98a85bf7. I would prefer not to do much more here while the tests are commented out.

PR #98, opened 2016-06-02
Add support for custom instance domains

mtyaka:
**Background**

We want customers to be able to access their instances from their own custom domains which we don't control. We have to use custom domains in edx-platform ansible configuration when provisioning new app servers, but we still need to set DNS entries for the internal (*.opencraft.hosting) domains. The customer will create a CNAME entry pointing their custom domains to the internal *.opencraft.hosting domain.

The patch adds three new fields to instance configuration:

- external_lms_domain
- external_lms_preview_domain
- external_studio_domain

When set, the external domains are used instead of the internal (opencraft-controlled) domains in ansible configuration and the UI.

Regardless of the presence of external domains, each instance is still assigned an internal domain and the DNS entry for the internal domain is always updated to point to the current active appserver.

Customers who use external domains will have to set a CNAME record that points to the main internal domain.

**Testing**

1. Create a new OpenEdXInstance (or use an existing one) and set `external_lms_domain`, `external_lms_preview_domain`, and `external_studio_domain` to custom values (for example `courses.mydomain.org`, `preview.mydomain.org`, `studio.mydomain.org`). You can do that either using the django console or the django admin UI.
2. Make sure that LMS and Studio URLs displayed in the UI point to the external domain.
3. Provision a new appserver and make sure that `COMMON_HOSTNAME`, `EDXAPP_LMS_BASE`, `EDXAPP_PREVIEW_LMS_BASE`, and `EDXAPP_CMS_BASE` are using external rather than internal domains.
4. Activate the new app server and make sure that it uses the internal domain rather than external domain when setting DNS entries.
5. Check that everything still works correctly for instances without external domains (internal domains should be used everywhere).


instance/models/openedx_instance.py
```diff
@@ -67,58 +71,77 @@ def __str__(self):
...
-        return 'studio-{}'.format(self.sub_domain)
+        return 'studio-{0.sub_domain}.{0.base_domain}'.format(self)
 
     @property
-    def studio_domain(self):
+    def domain(self):
         """
-        Studio full domain name (eg. 'studio-master.sandbox.opencraft.com')
+        LMS domain name.
+        Returns external domain if present, otherwise returns internal domain.
```
itsjeyd: @mtyaka Nit: Insert blank line between this line and the one above, to keep docstring formatting consistent. Same comment applies to docstrings for the next two methods.

It might be worth providing some info here about how the external domain of an instance is used/why we distinguish between internal and external domains.

instance/static/html/instance/details.html
```diff
@@ -71,9 +71,9 @@ <h3 ng-if="instance">
         <tr><td>Modified</td><td>{{ instance.modified | date:'yyyy-MM-dd HH:mm:ssZ' }}</td></tr>
         <tr><td>Type</td><td>{{ instance.instance_type }}</td></tr>
         <!-- The following are all specific to Open edX instances: -->
-        <tr><td>Domain</td><td>{{ instance.domain}}<br>Sub domain: {{ instance.sub_domain}}, base domain: {{ instance.base_domain }}</td></tr>
         <tr><td>LMS URL</td><td><a href="{{ instance.url}}">{{ instance.url}}</a></td></tr>
         <tr><td>Studio URL</td><td><a href="{{ instance.studio_url}}">{{ instance.studio_url}}</a></td></tr>
+        <tr><td>Internal domain</td><td>{{ instance.internal_lms_domain }}</td></tr>
```
itsjeyd: @mtyaka Based on previous discussions on OC-1555 I think we should display the external domain here?

instance/tests/models/test_openedx_instance.py
```diff
@@ -191,6 +227,26 @@ def test_spawn_appserver(self, mocks, mock_provision):
         self.assertEqual(configuration_vars['EDXAPP_CONTACT_EMAIL'], instance.email)
 
     @patch_services
+    @patch('instance.models.openedx_instance.OpenEdXAppServer.provision', return_value=True)
+    def test_spawn_appserver_with_external_domains(self, mocks, mock_provision):
+        """
+        Test that external domains are used when provisioning a new server.
```
itsjeyd: @mtyaka "Test that relevant configuration variables use external domains when provisioning a new app server." perhaps?

instance/tests/models/test_openedx_instance.py
```diff
@@ -191,6 +227,26 @@ def test_spawn_appserver(self, mocks, mock_provision):
...
+        """
+        instance = OpenEdXInstanceFactory(
+            sub_domain='test.spawn', use_ephemeral_databases=True,
+            external_lms_domain='lms.external.com', external_lms_preview_domain='lmspreview.external.com',
+            external_studio_domain='cms.external.com'
+        )
```
itsjeyd: @mtyaka Nit: This would be even more readable (and easier to edit) if you put each arg on a separate line.

instance/tests/models/test_openedx_instance.py
```diff
@@ -208,6 +264,27 @@ def test_set_appserver_active(self, mocks):
         ])
 
     @patch_services
+    def test_set_appserver_active_external_domain(self, mocks):
+        """
+        Test set_appserver_active() with custom external domains.
+        Ensure that the dns records are only created for the internal domains.
```
itsjeyd: @mtyaka Nit: DNS

instance/models/openedx_instance.py
mtyaka: @itsjeyd Thanks, I added a blank line. I also some info about external vs internal domain, but I did that in a comment above external domain declarations rather than in the docstring of this method.

instance/static/html/instance/details.html
mtyaka: @itsjeyd We display URLs that will use external domains when present in two lines above. I think we need to show the internal domain somewhere so that the customer will know what to point their CNAME record to.

instance/tests/models/test_openedx_instance.py
mtyaka: @itsjeyd Yes, that's better; thanks :)

instance/tests/models/test_openedx_instance.py
mtyaka: Done.

instance/tests/models/test_openedx_instance.py
mtyaka: Fixed.

instance/static/html/instance/details.html
itsjeyd: @mtyaka Ah, yes, my bad - I was too focused on the line you added to pay attention to the context :)

instance/migrations/0052_add_internal_domain_fiels.py
```diff
@@ -0,0 +1,47 @@
...
+            name='internal_lms_preview_domain',
+            field=models.CharField(blank=True, max_length=100),
+        ),
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='internal_studio_domain',
+            field=models.CharField(blank=True, max_length=100),
+        ),
+        migrations.RunPython(migrate_internal_domains, reverse_code=migrations.RunPython.noop),
+    ]
```
itsjeyd: @mtyaka Typo in file name: "fiels" ⟶ "fields"

instance/migrations/0053_remove_subdomain_basedomain.py
```diff
@@ -0,0 +1,42 @@
...
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='internal_lms_preview_domain',
+            field=models.CharField(max_length=100, unique=True),
+        ),
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='internal_studio_domain',
+            field=models.CharField(max_length=100, unique=True),
+        ),
```
itsjeyd: @mtyaka Would it be possible to get rid of these by adding `unique=True` to the corresponding `AddField` operations in the previous migration (0052_add_internal_domain_fiels)?

instance/migrations/0054_merge.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0051_auto_20160530_1229'),
+        ('instance', '0053_remove_subdomain_basedomain'),
+    ]
+
+    operations = [
+    ]
```
itsjeyd: @mtyaka I think it would be worth renaming the migrations to linearize them. That would allow us to get rid of this migration (I think).

```
Applying instance.0051_add_external_domains... OK
Applying instance.0052_add_internal_domain_fiels... OK
Applying instance.0053_remove_subdomain_basedomain... OK
Applying instance.0051_auto_20160530_1229... OK
Applying instance.0054_merge... OK
```

vs.

```
Applying instance.0051_auto_20160530_1229... OK
Applying instance.0052_add_external_domains... OK
Applying instance.0053_add_internal_domain_fiels... OK
Applying instance.0054_remove_subdomain_basedomain... OK
```

instance/models/openedx_instance.py
```diff
@@ -52,73 +62,105 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin, O
...
-    sub_domain = models.CharField(max_length=50)
-    base_domain = models.CharField(max_length=50, blank=True)
+
+    # Internal domains are controlled by us and their DNS records are automatically set to point to the current active
+    # appserver. They are generated from a unique prefix (given as 'sub_domain' in instance factories) and the value of
+    # DEFAULT_INSTANCE_BASE_DOMAIN at instance creation time. They cannot be blank and are normally never changed after
+    # the instance is created.
+    # External domains on the other hand are controlled by the customer and are optional. We use external domains in
+    # preference to internal domains when displaying links to the instance in the UI and when passing domain related
+    # settings to ansible vars when provisioning appservers.
```
itsjeyd: @mtyaka Nit: "ansible" ⟶ "Ansible"; "domain related" ⟶ "domain-related"

instance/models/openedx_instance.py
```diff
@@ -52,73 +62,105 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin, O
...
+    # The `domain`, `lms_preview_domain`, and `studio_domain` methods are useful when you need to access the external
+    # domain if present, otherwise fall back to the corresponding internal domain.
```
itsjeyd: @mtyaka "The `domain`, `lms_preview_domain`, and `studio_domain` properties below are useful if you need to access corresponding domains regardless of whether an instance uses external domains or not (they return the external domain if set, and fall back to the corresponding internal domain otherwise)."

Thanks for adding a detailed comment here!

instance/models/openedx_instance.py
```diff
@@ -52,73 +62,105 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin, O
...
 
+    def __init__(self, *args, **kwargs):
+        """
+        OpenEdXInstance constructor.
+
+        The constructor is overridden to optionally accept a 'sub_domain' parameter instead of a full
+        value for 'internal_lms_domain'. When 'sub_domain' is provided, the 'internal_lms_domain' field is automatically
+        generated from from the value of 'sub_domain' and the DEFAULT_INSTANCE_BASE_DOMAIN setting.
+        """
+        sub_domain = None
```
itsjeyd: @mtyaka You can remove this line I think.

instance/tests/integration/factories/instance.py
```diff
@@ -39,7 +39,16 @@ class OpenEdXInstanceFactory(DjangoModelFactory):
...
         model = OpenEdXInstance
 
-    sub_domain = factory.LazyAttribute(lambda o: '{}.integration'.format(str(uuid.uuid4())[:8]))
+    @classmethod
+    def create(cls, *args, **kwargs):
+        # OpenEdXInstance constructor accepts either a 'sub_domain' or 'instance_lms_domain' value. Only generate a
+        # random value for 'internal_lms_domain' if neither 'sub_domain' nor 'internal_lms_domain' are provided.
+        if 'sub_domain' not in kwargs and 'internal_lms_domain' not in kwargs:
+            kwargs = kwargs.copy()
+            domain = generate_internal_lms_domain('{}.integration'.format(str(uuid.uuid4())[:8]))
```
itsjeyd: @mtyaka For readability I'd probably split this up into two lines:

```python
            sub_domain = '{}.integration'.format(str(uuid.uuid4())[:8])
            domain = generate_internal_lms_domain(sub_domain)
```

Or even:

```python
            random_id = str(uuid.uuid4())[:8]
            sub_domain = '{}.integration'.format(random_id)
            domain = generate_internal_lms_domain(sub_domain)
```

That way it's immediately obvious that `generate_internal_lms_domain` expects to receive a subdomain.

instance/tests/models/factories/openedx_instance.py
```diff
@@ -37,5 +39,14 @@ class OpenEdXInstanceFactory(DjangoModelFactory):
...
         model = OpenEdXInstance
 
-    sub_domain = factory.Sequence('instance{}.test'.format)
+    @classmethod
+    def create(cls, *args, **kwargs):
+        # OpenEdXInstance constructor accepts either a 'sub_domain' or 'instance_lms_domain' value. Only generate a
+        # random value for 'internal_lms_domain' if neither 'sub_domain' nor 'internal_lms_domain' are provided.
+        if 'sub_domain' not in kwargs and 'internal_lms_domain' not in kwargs:
+            kwargs = kwargs.copy()
+            domain = generate_internal_lms_domain('instance{}.test'.format(str(uuid.uuid4())[:8]))
```
itsjeyd: @mtyaka Same comment here about splitting this up into two/three lines for readability.

pr_watch/tests/test_models.py
```diff
@@ -141,7 +144,10 @@ def test_create_from_pr(self):
         self.assertEqual(watched_pr.fork_name, pr.fork_name)
         self.assertEqual(watched_pr.branch_name, pr.branch_name)
 
-        self.assertEqual(instance.sub_domain, 'pr{}.sandbox'.format(pr.number))
+        internal_lms_domain = 'pr{}.sandbox.basedomain.com'.format(pr.number)
+        self.assertEqual(instance.internal_lms_domain, internal_lms_domain)
+        self.assertEqual(instance.internal_lms_preview_domain, 'preview-{}'.format(internal_lms_domain))
+        self.assertEqual(instance.internal_studio_domain, 'studio-{}'.format(internal_lms_domain))
```
itsjeyd: @mtyaka Instead of hardcoding default prefixes (`preview-`, `studio-`) here, use values from settings? That way this test would also ensure that `settings.DEFAULT_STUDIO_DOMAIN_PREFIX` and `settings.DEFAULT_LMS_PREVIEW_DOMAIN_PREFIX` are respected. (It might also make sense to keep the hardcoded prefixes here and adjust one of the other tests instead; just mentioning this now because I didn't notice it earlier. Overall I think it would be worth having at least one test that checks if prefixes are equal to corresponding settings.)

instance/migrations/0052_add_internal_domain_fiels.py
mtyaka: @itsjeyd ooops, fixed thanks :)

instance/migrations/0053_remove_subdomain_basedomain.py
mtyaka: @itsjeyd I don't think I can make these columns unique at creation time, because I also want them to be non-nullable and at creation time when I start with empty columns I can't have both (non-null and unique). I have to make them unique only after I have filled them with unique values.

instance/migrations/0054_merge.py
mtyaka: @itsjeyd Thanks, I will do that, that's nicer.

instance/models/openedx_instance.py
mtyaka: Thanks, fixed

instance/models/openedx_instance.py
mtyaka: @itsjeyd Much better, thank you :)

instance/models/openedx_instance.py
mtyaka: @itsjeyd Ah yes, that's a leftover from previous iteration before I refactored the code. Thanks for spotting it.

instance/tests/integration/factories/instance.py
mtyaka: @itsjeyd That is better, I will change that.

instance/tests/models/factories/openedx_instance.py
mtyaka: Done.

pr_watch/tests/test_models.py
mtyaka: @itsjeyd Good idea, not sure why I hardcoded the prefixes in that test.

instance/templates/instance/ansible/vars.yml
```diff
@@ -83,5 +83,5 @@ EDXAPP_WORKER_DEFAULT_STOPWAITSECS: 1200
 COMMON_ENABLE_NEWRELIC: {% if newrelic_license_key %}true{% else %}false{% endif %}
 COMMON_ENABLE_NEWRELIC_APP: {% if newrelic_license_key %}true{% else %}false{% endif %}
 COMMON_ENVIRONMENT: opencraft
-COMMON_DEPLOYMENT: '{{ instance.sub_domain }}'
+COMMON_DEPLOYMENT: '{{ instance.internal_lms_domain }}'
```
itsjeyd: @mtyaka So it doesn't matter if `COMMON_DEPLOYMENT` is set to `foo.sandbox` or `foo.sandbox.example.com`? I'm not familiar with that variable, so this is a genuine question :)

mtyaka: @itsjeyd It doesn't matter, as long as it's unique for each instance. As far as I know, it is only used for displaying the name of the app in New Relic dashboard.

itsjeyd:
@mtyaka Thanks for the updates! The code LGTM, I just had one more question (which likely won't result in further changes to the code).

So :+1: once you address that, squash the commits (it might make sense to keep https://github.com/open-craft/opencraft/pull/98/commits/07de7d95dc8967363c80c19c669adcfc1d400938 and https://github.com/open-craft/opencraft/pull/98/commits/2a44f36f8b4fd330e71bc6f980b5b80f7f204791 separate), and get a green build.

instance/templates/instance/ansible/vars.yml
itsjeyd: @mtyaka I see - using the full domain is even better (more informative) then. Great :)

PR #99, opened 2016-06-02
Add newrelic monitoring for instances

omarkhan:
This pull request adds New Relic monitoring for instances. Monitoring is activated for new app servers by setting the `NEWRELIC_LICENSE_KEY` and `NEWRELIC_ADMIN_USER_API_KEY` settings.

App and server monitoring is activated if `NEWRELIC_LICENSE_KEY` is set, and is configured using the following ansible vars:

- `COMMON_ENABLE_NEWRELIC` for server monitoring
- `COMMON_ENABLE_NEWRELIC_APP` for app monitoring
- `COMMON_ENVIRONMENT` global prefix for instances in the newrelic UI
- `COMMON_DEPLOYMENT` instance-specific prefix in the newrelic UI
- `NEWRELIC_LICENSE_KEY`

Availability monitoring using [Synthetics](https://newrelic.com/synthetics) is activated if the `NEWRELIC_ADMIN_USER_API_KEY` is set. Monitors are created when an app server is activated, and removed when an instance is deleted. Email alerts are set up for email addresses in the `ADMINS` list.

#### Testing

- Deploy this branch to stage
- Provision a new instance
- Open the LMS and Studio to cause data to be sent to New Relic
- Go to the [New Relic applications page](https://rpm.newrelic.com/accounts/1346955/applications). The LMS, Studio, workers and xqueue for your instance should all be monitored
- Go to the [Synthetics](https://synthetics.newrelic.com/accounts/1346955/synthetics) page. Availability monitors should be set up for the LMS, Studio, and LMS preview urls
- Open each monitor and click on 'Alert notifications'. ops@opencraft.com should be in the notifications list
- Kill the LMS and/or Studio for your instance. After a while an email alert should be sent to ops@opencraft.com
- Delete the instance
- Check that the Synthetics monitors have been removed


instance/models/mixins/openedx_monitoring.py
```diff
@@ -0,0 +1,95 @@
...
+    def enable_monitoring(self):
+        """
+        Enable monitoring on this instance.
+        """
+        if not settings.NEWRELIC_ADMIN_USER_API_KEY:
+            return
+
+        # Delete existing monitors if they don't monitor this instance's
+        # public urls
+        monitors = newrelic.get_synthetics_monitors()
```
mtyaka: @omarkhan Nit: Maybe we should add a `uri` field to `NewRelicAvailabilityMonitor` model so that we don't have to fetch the full global list of monitors every time?

instance/models/mixins/openedx_monitoring.py
```diff
@@ -0,0 +1,95 @@
...
+        # Add monitors for urls that are not already being monitored
+        already_enabled_urls = {enabled['uri'] for enabled in already_enabled}
+        for url in self._urls_to_monitor - already_enabled_urls:
+            monitor_id = newrelic.create_synthetics_monitor(url)
+            self.new_relic_availability_monitors.create(pk=monitor_id)
+
+            # Set up email alerts
+            if settings.ADMINS:
+                emails = [email for name, email in settings.ADMINS]
+                newrelic.add_synthetics_email_alerts(monitor_id, emails)
```
mtyaka: @omarkhan The NewRelic API docs aren't very clear, but from https://docs.newrelic.com/docs/apis/synthetics-rest-api/monitor-examples/payload-attributes-synthetics-rest-api it looks like the `create` endpoint might accept an array of `emails`. Did you try that?

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -0,0 +1,147 @@
...
+    Tests for OpenEdXMonitoringMixin.
+    """
+    @patch_services
+    @patch('instance.models.mixins.openedx_monitoring.OpenEdXMonitoringMixin.enable_monitoring')
+    def test_set_appserver_active(self, mocks, mock_enable_monitoring):
+        """
+        Check that monitoring is enabled when an appserver is activated.
+        """
+        instance = OpenEdXInstanceFactory()
+        appserver_id = instance.spawn_appserver()  # pylint: disable=no-member
```
mtyaka: @omarkhan I see that in other tests we set `pylint: disable=no-member` at the top of the test class, but that doesn't seem like a great idea. I didn't try this, but maybe adding `OpenEdXInstanceFactory` to [ignored-classes in pylinrc](https://github.com/open-craft/opencraft/blob/05578e947b97fcc5ce0bbd44ef53c848524e0c30/pylintrc#L290) would work to silence these annoying `no-member` warnings without having to manually do it on every line?

instance/models/mixins/openedx_monitoring.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Mixin that sets up availability monitoring for Open edX instances.
+    """
+    def enable_monitoring(self):
+        """
+        Enable monitoring on this instance.
+        """
+        if not settings.NEWRELIC_ADMIN_USER_API_KEY:
+            return
+
```
mtyaka: @omarkhan Nit: can we add some logging - `logging.info("Setting up NewRelic monitors")` or something similar? It's more fun to watch a new instance spinning up in the UI if you know what it's doing :)

instance/models/mixins/openedx_monitoring.py
```diff
@@ -0,0 +1,95 @@
...
+            # Set up email alerts
+            if settings.ADMINS:
+                emails = [email for name, email in settings.ADMINS]
+                newrelic.add_synthetics_email_alerts(monitor_id, emails)
+
+    def disable_monitoring(self):
+        """
+        Disable monitoring on this instance.
+        """
```
mtyaka: @omarkhan Nit: Can we add something like `logger.info("Removing NewRelic monitors")`?

instance/models/mixins/openedx_monitoring.py
omarkhan: @mtyaka the way I see it, New Relic is the ultimate source of truth here (someone might manually change a monitor's url in the UI, for example). If there is a clear performance/reliability benefit to storing the url locally then we can do it, but I don't think there is a compelling reason to do that now (premature optimisation and all that).

instance/models/mixins/openedx_monitoring.py
omarkhan: @mtyaka yes it does accept an array of emails. That's what I am sending here: the email addresses from the `ADMINS` list, once for each monitor.

instance/models/mixins/openedx_monitoring.py
mtyaka: ok, makes sense :)

instance/models/mixins/openedx_monitoring.py
mtyaka: @omarkhan I meant that the monitor create endpoint might accept the emails as well - so that you can supply the list of emails when you are creating the monitor instead of having to do it in a separate request (although the docs aren't clear and I didn't try that so I'm not sure).

omarkhan: Ah yes I didn't see that! Well spotted, I will try that.

omarkhan: @mtyaka I just tried that from the command line, it didn't work. The monitor was created successfully, but the email address I used was not in the notifications list.

mtyaka: Ah :/ Thanks for trying.

instance/tests/models/test_openedx_monitoring_mixins.py
omarkhan: @mtyaka good idea. I tried setting `ignored-classes` to `OpenEdXInstanceFactory` in pylintrc but it didn't work for some reason. Adding `*factories*` to `ignored-modules` seems to work though. See https://github.com/open-craft/opencraft/pull/99/commits/abbe41f4ab37f9424aae3fa533c88c3517fd9c16

omarkhan:
@mtyaka thanks for the review. I have added some logging: https://github.com/open-craft/opencraft/pull/99/commits/26860ab8882fcf488c98b10a557904e24451ee72

mtyaka:
Thanks for the updates 👍

antoviaque:
@omarkhan @mtyaka Thanks for this!

It's missing one thing though - an update of the doc / README, to add the description of the new configuration variables.

omarkhan:
Good catch @antoviaque. @mtyaka see https://github.com/open-craft/opencraft/pull/105

PR #102, opened 2016-06-03
Use js version of zxcvbn

itsjeyd:
... to improve quality of user-visible feedback from password validation.

cf. [OC-1587](https://tasks.opencraft.com/browse/OC-1587)

The form now behaves like this w/r/t password validation: When user enters password it is validated on the client, and the form displays feedback (suggestions and warnings) from `zxcvbn`. If a user goes ahead and submits the form while password strength is still too low, the server rejects the submission based on the password strength computed on the client, and the form displays a generic error message: "Please use a stronger password: avoid common patterns and make it long enough to be difficult to crack.". At this point, the input field for "password" is empty (since `django-angular` clears contents of password fields by default), so it wouldn't make sense to display feedback that is specific to a given password. Then, when the user enters another password, client side validation kicks in again and displays password-specific feedback from `zxcvbn` (until the password is strong enough).

### DIFF
--- a/registration/forms.py
+++ b/registration/forms.py
@@ -22,16 +22,22 @@ Forms for registration/login
 
 # Imports #####################################################################
 
+import logging
+
 from django import forms
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.models import User
 from django.db import transaction
 from django.utils.text import capfirst
 from djng.forms import NgDeclarativeFieldsMetaclass, NgFormValidationMixin, NgModelForm, NgModelFormMixin
-from zxcvbn import password_strength
 
 from registration.models import BetaTestApplication
 from userprofile.models import UserProfile
+
+
+# Logging #####################################################################
+
+logger = logging.getLogger(__name__)
 
 
 # Widgets #####################################################################
@@ -124,6 +130,9 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
         help_text=('This is also your account name, and where we will send '
                    'important notices.'),
     )
+    password_strength = forms.IntegerField(
+        widget=forms.HiddenInput,
+    )
     password = forms.CharField(
         strip=False,
         widget=PasswordInput,
@@ -188,6 +197,7 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
 
                 # Remove the password fields, the user already has a password
                 del self.fields['password']
+                del self.fields['password_strength']
                 del self.fields['password_confirmation']
 
                 # If the user has a profile, populate the full_name field
@@ -239,12 +249,32 @@ class BetaTestApplicationForm(NgModelFormMixin, NgFormValidationMixin, NgModelFo
             )
         return email
 
+    def clean_password_strength(self):
+        """
+        Check that client submitted password strength, and that value is in expected range.
+
+        If there are issues, produce appropriate warnings but don't raise ValidationError
+        (we don't want to display validation errors for this field to users).
+        """
+        password_strength = self.cleaned_data.get('password_strength')
+        if password_strength is None:
+            logger.warning('Did not receive password strength from client.')
+        elif password_strength not in range(5):
+            logger.warning(
+                'Received suspicious value for password strength. '
+                'Value should be in range [0, 4]. Observed value: {password_strength}',
+                **{'password_strength': password_strength}
+            )
+            password_strength = None
+        return password_strength
+
     def clean_password(self):
         """
-        Check password strength.
+        Using score computed on the client, check that password is reasonably strong.
         """
         password = self.cleaned_data.get('password')
-        if password and password_strength(password)['score'] < 2:
+        password_strength = self.cleaned_data.get('password_strength')
+        if password and password_strength is not None and password_strength < 2:
             raise forms.ValidationError(
                 ('Please use a stronger password: avoid common patterns and '
                  'make it long enough to be difficult to crack.'),
--- a/registration/static/js/src/registration.js
+++ b/registration/static/js/src/registration.js
@@ -47,15 +47,15 @@ app.config(function($httpProvider) {
 
 app.controller('Registration', ['$scope', '$http', 'djangoForm', function($scope, $http, djangoForm) {
 
+    // Reference to this controller.
+    var reg = this;
+
     // These fields will be validated on the server via ajax:
     // - Unique checks can only be done on the server.
-    // - Password strength is validated on the server as the python and js
-    //   versions of zxcvbn do not always give the same result.
     var serverValidationFields = [
         'subdomain',
         'username',
         'email',
-        'password',
         'password_confirmation'
     ];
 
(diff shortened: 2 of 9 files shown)


registration/forms.py
```diff
@@ -155,6 +164,23 @@ def __init__(self, *args, **kwargs):
...
+
+        Return password_strength as-is if there are no issues, otherwise return None.
+        """
+        if password_strength is None:
+            logger.warning('Did not receive password strength from client.')
+        elif password_strength not in range(5):
+            logger.warning(
+                'Received suspicious value for password strength. '
+                'Value should be in range [0, 4]. '
+                'Observed value: {password_strength}'.format(password_strength=password_strength)
```
omarkhan: Pylint is complaining about the use of `.format` here. It says you should use % formatting instead.

This might seem weird but it can be useful for grouping error messages, when using something like Sentry for example.

registration/forms.py
```diff
@@ -195,7 +221,8 @@ def clean_password(self):
         Check password strength.
         """
         password = self.cleaned_data.get('password')
-        if password and password_strength(password)['score'] < 2:
+        password_strength = self.cleaned_data.get('password_strength')
+        if password and self._check_password_strength(password_strength) is not None and password_strength < 3:
```
omarkhan: Do we need to be that strict here? According to the docs:

>   2 # somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
>   3 # safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)

Can we settle for 2? I feel that a lot of people's passwords will be rejected otherwise.

registration/templates/registration/registration_form.html
```diff
@@ -20,6 +20,9 @@
         {% include row with field=form.accept_terms only %}
         {% include row with field=form.subscribe_to_updates only %}
     {% endwith %}
+    {% with row="registration/hidden_field.html" %}
```
omarkhan: Do we need a separate template just for a hidden form field? Why not just include it directly?

registration/forms.py
```diff
@@ -155,6 +164,23 @@ def __init__(self, *args, **kwargs):
                     if name not in self.can_be_modified:
                         field.widget.attrs['readonly'] = True
 
+    def _check_password_strength(self, password_strength):
```
omarkhan: Pylint is complaining about 'no self use' here. Maybe make it a staticmethod?

registration/static/js/src/registration.js
```diff
@@ -54,10 +53,45 @@ app.controller('Registration', ['$scope', '$http', 'djangoForm', function($scope
         return getFormFields().filter(key => $scope.form[key].$dirty || $scope.form[key].$message);
     };
 
+    // Returns the result of validating password using zxcvbn.
+    var validatePassword = function() {
+        var password = $scope.form.password.$viewValue;
```
omarkhan: This is `null` when the user hasn't entered a password yet:

screen shot 2016-06-06 at 09 53 18

registration/static/js/src/registration.js
```diff
@@ -54,10 +53,45 @@ app.controller('Registration', ['$scope', '$http', 'djangoForm', function($scope
...
+    var getValidationFeedback = function() {
+        var validationResult = validatePassword(),
+            suggestions = validationResult.feedback.suggestions,
+            warning = validationResult.feedback.warning;
+        if (warning) {
+            suggestions.push(warning);
+        }
+        suggestions = formatSuggestions(suggestions);
+        // Concatenate suggestions to ensure that all of them are displayed
+        return [suggestions.join(" ")];
```
omarkhan: Why not just return them as an array? Is this another django-angular bug?

registration/static/js/src/registration.js
```diff
@@ -54,10 +53,45 @@ app.controller('Registration', ['$scope', '$http', 'djangoForm', function($scope
...
+    };
+
     // Display the given error messages, making sure not to clear messages
     // that are already displayed.
     var displayErrors = function(errors) {
         errors = _.pick(errors, getModifiedFields());
+        if (errors.password && $scope.form.password.$viewValue) {
```
omarkhan: You can validate this on the client directly without a roundtrip to the server:

- Remove this
- Remove server-side password strength validation (so that client-side validation messages are not replaced by the error message from the server)
- Remove 'password' from `serverValidationFields`
- Add something like this:

    ```javascript
    $scope.$watch('registration.password', function() {
        displayErrors({ password: getValidationFeedback() });
    });
    ```

omarkhan:
@itsjeyd I have tested this, it works nicely. See my comments above

registration/templates/registration/registration_form.html
itsjeyd: @omarkhan I had included the field directly at some point and felt that it made the `registration_form.html` template kind of messy. That's the only reason; it shouldn't make a difference w/r/t functionality. If you feel strongly about it, let me know and I'll change the code to include the field directly.

registration/static/js/src/registration.js
itsjeyd: @omarkhan Yes, it looks like this might be a django-angular bug. I'm not too familiar with it but I found that only a single message/suggestion was displayed when `errors` contained multiple messages for a given form field.

registration/static/js/src/registration.js
itsjeyd: @omarkhan Thanks for the pointers! Being able to avoid a roundtrip to the server is great. I updated the code, keeping the default error message from the server as a fallback for the odd case where `zxcvbn` doesn't produce any suggestions or warnings for a given password.

However, I'm now having some issues validating that the client calculated password strength: This now has to be done when the user submits the form, so the `password_strength` field has to be set to `zxcvbn(password).score` on the client. Do you know how to do that (in a way where the value actually reaches the server when submitting the form)? I found that I could set `$scope.form.password_strength.$viewValue` successfully on the client, but the server still didn't receive the value that I chose for it. Same result when setting `$modelValue` or using `$setViewValue(score)`.

registration/templates/registration/registration_form.html
bradenmacdonald: @itsjeyd  I think you should just put `{{ field }}` directly in registration_form.html here. You don't need all that extra HTML to create a row and columns just to contain a hidden field.

itsjeyd:
@omarkhan Thanks! I addressed your comments and updated the tests; please have another look.

registration/tests/browser/browser_registration.py
```diff
@@ -82,8 +82,16 @@ def fill_form(self, form_data):
...
                     element.click()
             elif not element.get_attribute('readonly'):
-                element.clear()
-                element.send_keys(value)
+                if not element.get_attribute('type') == 'hidden':
+                    element.clear()
+                    element.send_keys(value)
+                else:
+                    element_id = element.get_attribute('id')
+                    self.client.execute_script(
```
omarkhan: @itsjeyd why is this necessary? Doesn't the controller code update the value of the hidden form field?

omarkhan:
@itsjeyd looks good. Apart from my comment about the browser tests, 👍

itsjeyd:
Thanks @omarkhan!

itsjeyd:
For future reference, the reason why it shows a failing build above is because I accidentally pushed a commit that was meant for another PR to the `zxcvbn` branch. When I merged the PR, the most recent build ([#1115](https://circleci.com/gh/open-craft/opencraft/1115)) was green :)

PR #103, opened 2016-06-06
Introduce repository fork settings to be used for production instances.

smarnach:
We want to be able to set different default forks for production instances than for sandboxes.  For the sandboxes, the upstream forks should be used by default, while we want to use the modified named release branches for production instances by default.

This PR introduces two new settings to allow this.

### DIFF
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -312,6 +312,16 @@ DEFAULT_CONFIGURATION_VERSION = env('DEFAULT_CONFIGURATION_VERSION', default=DEF
 # configuration, edx-platform, forum, notifier, xqueue, and certs when creating production instances.
 OPENEDX_RELEASE_STABLE_REF = env('OPENEDX_RELEASE_STABLE_REF', default='named-release/dogwood.rc')
 
+# The edx-platform repository used by default for production instances
+STABLE_EDX_PLATFORM_REPO_URL = env(
+    'STABLE_EDX_PLATFORM_REPO_URL', default='https://github.com/{}.git'.format(DEFAULT_FORK)
+)
+
+# The configuration repository used by default for production instances
+STABLE_CONFIGURATION_REPO_URL = env(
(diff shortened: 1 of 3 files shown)


opencraft/settings.py
```diff
@@ -300,6 +300,16 @@
 # configuration, edx-platform, forum, notifier, xqueue, and certs when creating production instances.
 OPENEDX_RELEASE_STABLE_REF = env('OPENEDX_RELEASE_STABLE_REF', default='named-release/dogwood.rc')
 
+# The edx-platform repository used by default for production instances
+STABLE_EDX_PLATFORM_REPO_URL = env(
```
jbzdak: NIT: This is a nit, so no need to address it, just pointing a thing that we might clean-up: 

 ``STABLE_EDX_PLATFORM_REPO_URL`` is a setting by itself, while ``DEFAULT_EDX_PLATFORM_REPO_URL`` is a constant in instance module, generated from ``DEFAULT_FORK``. 

This is a thing that should be symmetric IMO, either introduce ``DEFAULT_EDX_PLATFORM_REPO_URL`` as a setting, or add ``STABLE_FORK`` setting.

smarnach: @jbzdak The reason we have `DEFAULT_FORK` instead of a URL is that it is used by the PR watcher.  The notion of a "fork" and the "organization/repo" syntax are Github-specific, and only needed for the PR watcher.  To clean this up, I would separate the settings for the PR watcher better, and prefix them with `PR_WATCHER`, and write the rest of the code to be completely Github-agnostic.  I thought about cleaning this up now, but figured it is more important to get this out quickly, but I consciously made the decision to use the full  URL instead of the fork name here.

jbzdak: @smarnach :+1: Makes sense !

jbzdak:
@smarnach Tested manually by creating production instance in shell and checking that it has proper `edx_platform_repository_url` and `foo.configuration_source_repo_url`. :+1:  When tests pass.

smarnach:
@jbzdak Thanks for the fast review!

PR #104, opened 2016-06-07
Restrict access to /instance and /api endpoints

pomegranited:
This change restricts access to the /instance and /api URLs to staff users, and users with `InstanceManagerPermission`.

**JIRA tickets**: Implements [OC-1550](https://tasks.opencraft.com/browse/OC-1550)

**Screenshots**: 

Non-privileged users see Permission Denied for /api URLs:

screen shot 2016-06-08 at 4 30 54 pm

**Testing instructions**:

1. Set up an opencraft devstack according to [instructions in README](https://github.com/open-craft/opencraft#install).
1. Be sure to run `make migrate`, as this change introduces a new permission model.
1. Run automatic tests: `make test`
1. Set up users for manual tests: `make shell`
```python
In [1]: from django.contrib.auth.models import User
In [2]: beta = User.objects.create_user('beta', 'beta@example.com', 'beta')
In [3]: staff = User.objects.create_user('staff', 'staff@example.com', 'edx')
In [4]: staff.is_staff = True
In [5]: staff.save()
In [6]: superuser = User.objects.create_user('superuser', 'superuser@example.com', 'super')
In [7]: superuser.is_superuser = True
In [8]: superuser.save()
```
1. With each user in [`beta`, `staff`, `superuser`]:
    * Visit [http://localhost:5000](http://localhost:5000), and click Login.
        * beta should redirect to `/registration/`
        * staff, superuser should redirect to `/instance/`
    * Visit [http://localhost:5000/instance/](http://localhost:5000/instance/):
        * beta should redirect to `/registration/`
        * staff, superuser should redirect to `/instance/`
    * Visit [http://localhost:5000/api/](http://localhost:5000/api/)
        * beta user should get 403
        * staff, superuser should get 200, and be able to navigate further
    * Click 'Logout' in between each user.
        * Should redirect to `/registration/`

1. `make test_prospector`'s `pylint` tests were returning a number of `no-member` errors for inherited members of the new `InstanceManagerPermission` and `InstanceManagerPermissionManager` classes.  Errors messages are visible in the [initial CircleCI run](https://circleci.com/gh/open-craft/opencraft/1083). 

   This appears to be a [pylint-django bug, addressed by PR #74](https://github.com/landscapeio/pylint-django/pull/74).  I've suppressed the errors with commit [efcbac2](https://github.com/open-craft/opencraft/pull/104/commits/efcbac2e30c249538e31e1e6ac87ac32f8964f04).

**Reviewers**
- [ ] @smarnach


instance/migrations/0053_instancemanagerpermission_data.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.5 on 2016-06-07 06:13
+from __future__ import unicode_literals
+
+from django.db import migrations
+from instance.permissions.models import InstanceManagerPermission
+
+def create_instance_manager_permission(apps, schema_editor):
+	"""Create the singleton InstanceManagerPermission"""
+	InstanceManagerPermission.objects.get_or_create(codename=InstanceManagerPermission._meta.verbose_name)
```
smarnach: Can't we just create this in the startup code?  I don't like the irreversable migration here.  (This is probably mostly becaus I don't like migrations in general, since I can't come up with any strong reason why it should be bad, so feel free to ignore this comment.)

instance/permissions/models.py
```diff
@@ -0,0 +1,75 @@
...
+
+    class Meta:
+        proxy = True
+        verbose_name = InstanceManagerPermissionManager.content_type_model
+
+    def save(self, *args, **kwargs):
+        """Use a proxy content type called instance.instance_manager_permission"""
+        ct, dummy = ContentType.objects.get_or_create(  # pylint: disable=no-member
+            app_label=self._meta.app_label, model=self._meta.verbose_name,
+        )
```
smarnach: When I wrote the task description, I didn't realise that creating a global permission in Django is *more* complicated than creating a permission attached to a model.  I tried to keep it as simple as possible, but I guess it isn't too bad anyway. :)

instance/permissions/models.py
```diff
@@ -0,0 +1,75 @@
...
+        Determines whether the user is an "instance manager" user.
+
+        User must:
+        * be a staff member, or
+        * have the InstanceManagerPermission
+        """
+        permission = '{}.{}'.format(cls._meta.app_label, cls._meta.verbose_name)
+        if user.is_staff or user.has_perm(permission):
+            return True
+        return False
```
smarnach: How about

    return user.is_staff or user.has_perm(permission)

instead?

smarnach:
@pomegranited This is looking really nice!  I've added a few minor comments.

You still have a test failure.  I first thought this were caused by access to the validation API used by the registration form now being blocked, but the registration part of the API is still accessible to everyone (even without being logged in), since [it uses the `AllowAny` permission](https://github.com/open-craft/opencraft/blob/master/registration/api.py#L39).  So I've got no idea what is causing the test failure.

instance/migrations/0053_instancemanagerpermission_data.py
smarnach: Here is an alternative approach: https://gist.github.com/plar/5acaf6467e327dd3e478

In that code, the permissions are created when running the migrations, but not inside any permission.  However, they use the plain `Permission` model, and don't create a proxy model.  Not sure whether this could be adapted.

Treat these comments here with low priority.  I don't even know what problem I'm trying to solve here. :)

instance/permissions/models.py
pomegranited: Fixed in [ce1116c](https://github.com/open-craft/opencraft/pull/104/commits/ce1116c6041a400a9646302d3c761c18cec45bf9).

instance/migrations/0053_instancemanagerpermission_data.py
pomegranited: @smarnach Yes, I can't tell which option is worse 😄   Mine felt like an awful lot of code just to create a simple Permission, but your alternative approach looks a bit hacky.

Is there anyone else on the team that might want to weigh in on this? @open-craft/core

pomegranited:
@smarnach [Browser tests succeeded](https://circleci.com/gh/open-craft/opencraft/1113) after this latest commit.  I'll re-run them a few times to see if it's flaky.

instance/migrations/0053_instancemanagerpermission_data.py
bradenmacdonald: @smarnach @pomegranited 
I would prefer that you define the permission on the `InstanceReference` model, which is more or less a global model, and you create the permission using that model's `Meta`: https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#custom-permissions

No migration required (it creates the permission automatically during `post_migrate`).

I don't think you should create a custom model just to have a permission.

smarnach: @bradenmacdonald I don't feel the permission really belongs on a model, since it controls access to more than just that model.  The fact that it happens to be easy to create model permissions doesn't seem a strong enough reason to put the permission in the wrong place, and I would prefer the current solution.  (What do you mean by "global model", by the way?)

The custom model is just a proxy model used to encapsulate creating the content type you need for a permission.  It would be perfectly possible to use the `Permission` model directly.

bradenmacdonald: @smarnach I know, but attaching it to `InstanceReference` requires only two lines of code and no migrations, since it works with the way django permissions were designed.

By "global model" I meant "apex model" - more specifically, I meant that "InstanceReference" is the primary model of the `instance` app, and there is only ever that one model serving that purpose. Other models in the instance app like `Instance` (which is abstract), `OpenEdXInstance`, `OpenEdXAppServer` are merely themselves instances of a general pattern and there could be many more such models like `DaliteInstance`, `DaliteAppServer`, etc. So it definitely does not make any sense to attach the permission to any other model in the `instance` app other than `InstanceReference`, if we must attach it to a model.

> The custom model is just a proxy model used to encapsulate creating the content type you need for a permission.
Ah, I see now. I found [the explanation](http://stackoverflow.com/questions/13932774/how-can-i-use-django-permissions-without-defining-a-content-type-or-model) on StackOverflow. I think I should clarify something though: it looks to me like the real purpose of the proxy model is to prevent django from thinking the content type is stale and deleting it. If all you needed was a custom content type, you could simply create one in a single line within a data migration; the proxy model is a roundabout way of doing that.

bradenmacdonald: Also re 
> I don't feel the permission really belongs on a model, since it controls access to more than just that model.

True, but everything it does control access to is either that model or conceptually a "child" of that model.

InstanceReferences own concrete Instance subclasses which own concrete AppServer subclasses which own concrete Server subclasses, and all of those owned things also own log entries. The permission controls access to all those things and their APIs.

Overall I'm fine with any of these approaches. If you want my opinion, I do prefer @smarnach 's code over the one taken in the PR (if you can confirm that django doesn't think the content type is 'stale' upon subsequent migrations and prompt the user to delete it), and I prefer the InstanceReference approach overall though.

pomegranited: @smarnach I agree with @bradenmacdonald on this one.  The only hazy part of the use of this permission is that we're using it to restrict access to all of /api, not just /api/v1/instance.

Other than that, it really is just a restriction on access to the "instance" models, and since InstanceReference is shared by all of them, it's not a stretch to use that instead of a proxy model.  Also, when we want to add new "owned instance" permissions, it will be straightforward.

Also means a significantly smaller diff.  Please see [de57bdf](https://github.com/open-craft/opencraft/pull/104/commits/de57bdfed33b71bb0e4eb1315e6e0656b7d1c2f5) and let me know what you think?

instance/models/instance.py
```diff
@@ -89,6 +92,18 @@ def save(self, *args, **kwargs):
...
+    def can_manage(cls, user):
+        """
+        Returns true if the user is allowed to manage all instances.
+
+        User must:
+        * be a staff member, or
+        * have the "manage_all" permission
+        """
+        permission = '{}.{}'.format(cls._meta.app_label, "manage_all")
+        return user.is_staff or user.has_perm(permission)
```
bradenmacdonald: Do we need `is_staff` here? I thought the goal was to have just this one new permission, for simplicity.

pomegranited: @bradenmacdonald I took this requirement from [OC-1550](https://tasks.opencraft.com/browse/OC-1550) to mean it needed to include is_staff users as well as people with the permission:

> Make sure staff users (is_staff = True) get access to everything as well.

smarnach: @bradenmacdonald We don't strictly need it, but the general idea in Django apps is that staff users get access to everything.  (It's my fault that this is here, since I explicitly requested it in the task description.)

instance/migrations/0053_instancemanagerpermission_data.py
smarnach: @pomegranited Yeah, actually the code convinces me.  It's a lot easier, and we want to keep it simple for now.

instance/models/instance.py
smarnach: To expand a bit, if you install the instance manager locally in Vagrant, you typically create an account for you using the `createsuperuser` management command.  This account should have full access.

bradenmacdonald: @smarnach Ah, ok. I seem to recall @antoviaque wanted only one way to add this permission and rejected the idea of using is_staff, but I could definitely be mis-remembering.

pomegranited: @bradenmacdonald @smarnach You're correct, the model access rights work fine without checking for `is_staff`.

However, since I'm using this `can_manage` method for the [ApiInstanceManagerPermission](https://github.com/open-craft/opencraft/pull/104/files#diff-ff05414fb9cbd08c954e674eef451fc5R32) and the [opencraft IndexView](https://github.com/open-craft/opencraft/pull/104/files#diff-8d4846bd50606dc2973def39917f5f68R33) redirect as well, then it helps me to keep it in one place.

pomegranited: And yes, superusers have full access too, as described in the test instructions above.

bradenmacdonald: > you typically create an account for you using the createsuperuser management command. This account should have full access.

@smarnach Yeah but [has_perm already checks for is_superuser](https://github.com/django/django/blob/9935f97cd203bdcc722bc3d4e96858e221d96ff8/django/contrib/auth/models.py#L261-L272) so you don't have to check for that explicitly.

smarnach: @bradenmacdonald You are right, my mistake.  In that case, I don't care whether to include `is_staff` here.

pomegranited: @bradenmacdonald @smarnach 

Currently, only Braden, Jacek, and Xavier are superusers in PRD, and the rest of us are just staff users.

As part of the deployment, I can add the new permission to the existing is_staff users, but then new people will have to remember to do that too.

I think it's simpler to just keep the `is_staff` check, and use this permission without `is_staff` for users that we want to give broad privileges too without giving them full access to /admin.

pomegranited:
@smarnach How's this PR looking now?  Is there anything else I should do to it before merging and deploying (tomorrow)?

instance/models/instance.py
antoviaque: @pomegranited I agree with @bradenmacdonald that it's better to just check for the permission. The `is_staff` is normally really just to control who can login in the Django admin as far as I know. We could just give superuser privileges to everyone in the team for now. 

Another alternative if we don't want to use the superuser privilege is to create an "admin" group in the django admin, and grant the "manage all" permission to that group, then add the whole team to that group.

smarnach:
@pomegranited Looks all good now!  I tested again, and all went fine.  Points that remain to be done before merging:

1. Remove the check for `is_staff`, as desired by @antoviaque.
2. Add a comment to the README that users need the permission to be able to mange instances.
3. Rebase and squash commits.

:+1: after these have been addressed.

PR #105, opened 2016-06-08
Document new relic settings in README

omarkhan:
I forgot to do this in https://github.com/open-craft/opencraft/pull/99.


mtyaka:
👍

PR #107, opened 2016-06-13
Fix footer spacing on smaller displays

omarkhan:
@pomegranited noticed that there is no footer spacing on smaller displays:

screen shot 2016-06-10 at 11 50 02 am

This pull request fixes the spacing.


pomegranited:
Looks great, Omar 👍 

Re that test failing - I've had that fail before, but it succeeds on re-run.

PR #108, opened 2016-06-13
Set CMS_HOSTNAME Ansible variable to the Studio domain.

smarnach:
This Ansible variable is used as the `server_name` for nginx, so it must match the Studio domain name.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -19,6 +19,7 @@
 """
 Open edX Instance models
 """
+import re
 import string
 
 from django.conf import settings
@@ -144,6 +145,21 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin,
             return self.external_studio_domain
         else:
             return self.internal_studio_domain
+
+    @property
+    def studio_domain_nginx_regex(self):
+        """
+        Regex that matches either the internal or the external Studio URL.
+
+        This is exclusively meant for the Ansible variable CMS_HOSTNAME to configure the nginx
+        server_name regex to match the Studio domains.
+        """
+        if self.external_studio_domain:
+            domains = [self.external_studio_domain, self.internal_studio_domain]
+        else:
+            domains = [self.internal_studio_domain]
+        choices = '|'.join(map(re.escape, domains))  # pylint: disable=bad-builtin
+        return '~^({})$'.format(choices)
 
     @property
     def url(self):
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -157,6 +157,7 @@ class OpenEdXInstanceTestCase(TestCase):
         self.assertEqual(instance.domain, internal_lms_domain)
         self.assertEqual(instance.lms_preview_domain, internal_lms_preview_domain)
         self.assertEqual(instance.studio_domain, internal_studio_domain)
+        self.assertEqual(instance.studio_domain_nginx_regex, r'~^(studio\-sample\.example\.org)$')
         self.assertEqual(instance.url, 'http://{}/'.format(internal_lms_domain))
         self.assertEqual(instance.lms_preview_url, 'http://{}/'.format(internal_lms_preview_domain))
         self.assertEqual(instance.studio_url, 'http://{}/'.format(internal_studio_domain))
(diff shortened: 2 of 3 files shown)


mtyaka:
@smarnach Nice find! Do you think we could somehow make it work for both the internal and external domain? The default regex usually works because both domains often start with "studio", but that might not always be the case and it would be nice if it would always work for both domains. Same goes for LMS preview, but I think that one would be harder to implement.

smarnach:
@mtyaka That's a good suggestion.  I can simply use a regular expression that matches either address.

bradenmacdonald:
+1 for a regex that matches either.

instance/tests/models/test_openedx_instance.py
```diff
@@ -157,6 +157,7 @@ def test_domain_url(self):
         self.assertEqual(instance.domain, internal_lms_domain)
         self.assertEqual(instance.lms_preview_domain, internal_lms_preview_domain)
         self.assertEqual(instance.studio_domain, internal_studio_domain)
+        self.assertEqual(instance.studio_domain_nginx_regex, r'~^(studio\-sample\.example\.org)$')
```
itsjeyd: @smarnach Why does `re.escape` escape `-`?

smarnach: @itsjeyd Inside character classes, it's sometimes necessary to escape `-` (usually you just put it as the first character of a character class, so it doesn't need escaping).

When escaping a full regular expression, the square brackets around the character class would be escaped, so the hyphen wouldn't need escaping.  I suspect they want the `re.escape()` to be generic enough to also escape _part_ of a regular expression.  Just guessing though, I didn't check the version history of that function.

itsjeyd:
@smarnach :+1:

instance/models/openedx_instance.py
```diff
@@ -146,6 +147,21 @@ def studio_domain(self):
...
+        Regex that matches either the internal or the external Studio URL.
+
+        This is exclusively meant for the Ansible variable CMS_HOSTNAME to configure the nginx
+        server_name regex to match the Studio domains.
+        """
+        if self.external_studio_domain:
+            domains = [self.external_studio_domain, self.internal_studio_domain]
+        else:
+            domains = [self.internal_studio_domain]
+        choices = '|'.join(map(re.escape, domains))  # pylint: disable=bad-builtin
```
antoviaque: @smarnach @itsjeyd Pylint was complaining about `map` here, right? Was there a reason for not replacing it with a list comprehension rather than silencing it?

itsjeyd: @antoviaque I can't speak for Sven but the reason I didn't point this out in my review is that I couldn't think of a reason to prefer a list comprehension in this case (`map` is lazy in Python 3 and the code maps a built-in function as opposed to a `lambda`; `map(re.escape, domains)` doesn't seem less readable than `[re.escape(d) for d in domains]` to me). On the other hand, I am in favor of keeping the number of `# pylint: ...` comments to a minimum, so I wouldn't be opposed to changing this when we get a chance :)

antoviaque: @itsjeyd Alright, thanks for the explanation - yes, let's try to limit the exclusions to cases where we can't do it any other way. Besides making them more manageable, it will also help with having a consistent style. For this particular one, I've added it to the list of OC-1572.

smarnach: @antoviaque You can always use a list comprehension (or in this case, a generator expression) instead of using `map()`.  However, I personally think `map()` is more readable in this case.  It's less verbose and doesn't introduce a loop variable, so I find it easier to grasp.

I'd be in favour of removing `map` from the `bad-functions` list in the pylint configuration.  It simply isn't bad at all; `map()` has been a core part of Python since at least version 1.5, and I don't see why we should stop using it just because someone decided to add this warning to pylint.

Pylint in default configuration complains about a lot of stuff it actually shouldn't.  We should adjust the configuration to our likings rather than taking arbitrary (and sometimes outright bad) defaults as the law.

antoviaque: @smarnach I personally can see their argument, and find the list comprehension clearer. Ultimately, most of these cases will be a matter of personal preference, there will often be arguments that can go in both cases. Taking the approach of just following pylint recommendations has the advantage of providing a ready-made coding style, while nullifying the need for debates on these, which in turn allows us to spend time on more important things.

PR #109, opened 2016-06-14
Update ansible vars and enable some features.

mtyaka:
This adds some ansible vars to the yaml template mainly for documentation purposes (they are set to  default values, so they should have no effect).

This commit also enables some extra feature flags, which are often useful and don't have any negative side effects (s3 grade downloads, third party auth...).

### DIFF
--- a/instance/templates/instance/ansible/s3.yml
+++ b/instance/templates/instance/ansible/s3.yml
@@ -10,3 +10,7 @@ XQUEUE_AWS_ACCESS_KEY_ID: '{{ instance.s3_access_key }}'
 XQUEUE_AWS_SECRET_ACCESS_KEY: '{{ instance.s3_secret_access_key }}'
 XQUEUE_S3_BUCKET: '{{ instance.s3_bucket_name }}'
 XQUEUE_S3_PATH_PREFIX: 'xqueue'
+
+EDXAPP_GRADE_STORAGE_TYPE: 's3'
+EDXAPP_GRADE_BUCKET: '{{ instance.s3_bucket_name }}'
+EDXAPP_GRADE_ROOT_PATH: 'grades-download'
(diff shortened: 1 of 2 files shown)


smarnach:
@mtyaka I recently noticed that we should also set `EDX_NOTES_API_VERSION` to some reasonable value, and maybe we could include this change in this PR.  (If you don't want that scope expansion, I can create a separate PR for that change.)

We currently don't set it at all, which means [we always install the master version of edx_notes_api](https://github.com/edx/configuration/blob/master/playbooks/roles/edx_notes_api/defaults/main.yml#L30).  This doesn't work when using the Dogwood version of configuration.

Unfortunately, edx_notes_api doesn't have the `named-release/dogwood.3` tags or similar (in fact, [it doesn't have any tags whatsoever](https://github.com/edx/edx-notes-api/tags)), so using `appserver.openedx_release` would further restrict valid possible values for `openedx_release`.  Do you have any good idea what to do about this?

mtyaka:
@smarnach Thanks for bringing this up, I remember running into some problems because of this in the past. I guess the easiest thing to do would be to ask edX to create some `named-release/*` tags for edx_notes_api and then use {{appserver.openedx_release}} just like we do for other services (forums, xqueue, certs, etc).

smarnach:
@mtyaka Adding the tags is easy (in principle we could just do it ourselves).  However, they would have to include the repo in their release script, so new releases get tagges automatically.  I don't know what their process is, so this would require some coordination with edX.  We should exclude this from this PR then, because otherwise we wouldn't get it merged in the current sprint.

smarnach:
@mtyaka The instance you deployed is looking good, and I couldn't find any problems with it.  The code is also looking good, so :+1:.

PR #110, opened 2016-06-16
Set ansible variables for postfix_queue role.

mtyaka:
When external SMTP host is defined in settings, populate `POSTFIX_QUEUE_*` ansible vars when  provisioning app servers.

In addition to merely passing the SMTP server settings to postfix, we also set it up to rewrite the sender address to use a fixed domain defined in the settings, that has been pre-validated and authorized as
a sending domain in AuthSMTP.
The local part of the rewritten address (part before @) is the instance domain, so the rewritten From address looks something like:
```
pr123.opencraft.hosting@opencraft.hosting
```
or when using external LMS domain:
```
courses.mydomain.com@opencraft.hosting
```
The original sender address is copied into the `Reply-To` header.

**Testing**

Add the following settings to your `.env` file (replace username and password with valid values):
```python
INSTANCE_SMTP_RELAY_HOST='mail.authsmtp.com'
INSTANCE_SMTP_RELAY_PORT=587
INSTANCE_SMTP_RELAY_USERNAME='xxx'
INSTANCE_SMTP_RELAY_PASSWORD='xxx'
INSTANCE_SMTP_RELAY_SENDER_DOMAIN='plebia.org'
```
I'm using `plebia.org` as the sender domain because it has already been validated in our AuthSMTP account.

Restart the IM and provision a new instance. The configuration playbook you will be using should include the postfix_queue role. I've been using this one: https://github.com/open-craft/configuration/blob/stgallen/playbooks/edx_sandbox.yml

Check that combined vars include the `POSTFIX_QUEUE_*` variables set to expected values.

When the instance finishes provisioning, sign up in the LMS to receive the activation email. The email should come from a `@plebia.org` email address, but if you hit "reply to", it should reploy to the original instance's email setting (as defined by `instance.email`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,11 +176,11 @@ flavor:
 
 ### Application settings
 
-* `DEBUG` Turn on debug mode. Use in development only (default: False)
-* `SECRET_KEY` Set this to something unique and keep it secret (required)
-* `DATABASE_URL` Your database, e.g. `postgres://localhost/opencraft` (required)
-* `REDIS_URL` (default: `redis://localhost:6379/`)
-* `HUEY_ALWAYS_EAGER` Set to True to run huey tasks synchronously, in the web
+* `DEBUG`: Turn on debug mode. Use in development only (default: False)
+* `SECRET_KEY`: Set this to something unique and keep it secret (required)
+* `DATABASE_URL`: Your database, e.g. `postgres://localhost/opencraft` (required)
+* `REDIS_URL`: (default: `redis://localhost:6379/`)
+* `HUEY_ALWAYS_EAGER`: Set to True to run huey tasks synchronously, in the web
   process. Use in development only (default: False)
 * `LOGGING_ROTATE_MAX_KBYTES`: The max size of each log file (in KB, default: 10MB)
 * `LOGGING_ROTATE_MAX_FILES`: The max number of log files to keep (default: 60)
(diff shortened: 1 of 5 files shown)


instance/models/openedx_appserver.py
```diff
@@ -222,6 +223,22 @@ def create_lms_user_settings(self):
         template = loader.get_template(self.LMS_USER_VARS_TEMPLATE)
         return template.render(dict(lms_users=self.lms_users.all()))
 
+    def smtp_relay_settings(self):
```
itsjeyd: @mtyaka I think we could make this a `@property` (just like `github_admin_username_list`).

instance/models/openedx_appserver.py
```diff
@@ -208,6 +208,7 @@ def create_configuration_settings(self):
             # querying the Github API twice), we pass it as a context variable.
             'github_admin_username_list': self.github_admin_username_list,
             'newrelic_license_key': settings.NEWRELIC_LICENSE_KEY,
+            'smtp_relay_settings': self.smtp_relay_settings
```
itsjeyd: @mtyaka I'm wondering if we need to pass these settings as a separate context variable here. According to the comment above, there's a specific reason for doing this for `github_admin_username_list`; but to obtain `smtp_relay_settings` we're just reading configuration variables. So I think it would make sense to remove this line and reference `smtp_relay_settings` as `appserver.smtp_relay_settings` in the template. (That would also make it immediately obvious where the settings come from when reading just the template.)

instance/tests/models/test_openedx_instance.py
```diff
@@ -240,6 +240,36 @@ def test_newrelic_configuration(self, mocks, mock_provision):
...
+        self.assertEqual(configuration_vars['POSTFIX_QUEUE_EXTERNAL_SMTP_HOST'], 'smtp.myhost.com')
+        self.assertEqual(configuration_vars['POSTFIX_QUEUE_EXTERNAL_SMTP_PORT'], '2525')
+        self.assertEqual(configuration_vars['POSTFIX_QUEUE_EXTERNAL_SMTP_USER'], 'smtpuser')
+        self.assertEqual(configuration_vars['POSTFIX_QUEUE_EXTERNAL_SMTP_PASSWORD'], 'smtppass')
+        self.assertEqual(configuration_vars['POSTFIX_QUEUE_HEADER_CHECKS'], '/^From:(.*)$/   PREPEND Reply-To:$1')
+        self.assertEqual(
+            configuration_vars['POSTFIX_QUEUE_SENDER_CANONICAL_MAPS'],
+            'test.postfix@myinstance.org  lms.myinstance.org@opencraft.hosting'
+        )
+
```
itsjeyd: @mtyaka I think you could put this test in `test_openedx_appserver.py` (since the code above treats SMTP relay settings as appserver-level settings).

README.md
```diff
@@ -243,6 +243,20 @@ flavor:
...
+### External SMTP service settings
+
+If you want to use an external SMTP service for sending email from app servers,
+set the following configuration variables:
+
+* `INSTANCE_SMTP_RELAY_HOST` External SMTP host
+* `INSTANCE_SMTP_RELAY_PORT` External SMTP port
+* `INSTANCE_SMTP_RELAY_USERNAME`: External SMTP provider username
+* `INSTANCE_SMTP_RELAY_PASSWORD`: External SMTP provider password
+* `INSTANCE_SMTP_RELAY_SENDER_DOMAIN`: When using external SMTP provider, email
```
itsjeyd: @mtyaka I think it's awesome that you started putting colons after names of settings! This has been bugging me for a while :) It's a small thing, but IMHO it does improve readability. If you could do that consistently for all settings, that would be great! (If you don't have time, that's fine as well -- just make sure to be consistent within this section in that case.)

instance/models/openedx_appserver.py
mtyaka: @itsjeyd Good idea, done.

instance/models/openedx_appserver.py
mtyaka: @itsjeyd We are accessing that dict of settings six times in the template. I know it probably has zero practical impact on performance, but constructing the dict of settings six times just to read out one entry still bothers me.

I remembered about django's `{% with %}` tag which I believe was designed for these kind of use-cases (caching variables in templates), so I used it for smtp relay settings, as well as github admin usernames. It makes the vars template more noisy, but I like the fact that it's obvious where the values are coming from just by looking at the template.

instance/tests/models/test_openedx_instance.py
mtyaka: @itsjeyd Good point, I moved the test and added another one for the case where SMTP relay is not configured.

README.md
mtyaka: @itsjeyd lol, I have no idea how I managed to start without colons and then start adding them half way through the list. I added colon to all settings, it really makes the text more readable.

instance/models/openedx_appserver.py
itsjeyd: @mtyaka That's a great solution! Makes the code in `create_configuration_settings` much nicer, I like it :)

itsjeyd:
@mtyaka :+1:

instance/models/openedx_appserver.py
```diff
@@ -204,9 +204,6 @@ def create_configuration_settings(self):
         vars_str = template.render({
             'appserver': self,
             'instance': self.instance,
-            # This property is needed twice in the template.  To avoid evaluating it twice (and
-            # querying the Github API twice), we pass it as a context variable.
-            'github_admin_username_list': self.github_admin_username_list,
```
antoviaque: @mtyaka Thanks for that cleanup! : )

PR #111, opened 2016-06-20
Use production playbook

omarkhan:
https://github.com/edx/configuration/pull/3118 adds an `edx_production.yml` playbook specifically designed to deploy production instances. This pull request modifies the Instance Manager to use the new playbook.


pomegranited:
PR irrelevant with changes to Ocim and edx:configuration that allow use of custom playbooks.

PR #124, opened 2016-07-25
Only install Django Debug Toolbar when in debugging mode.

smarnach:
The toolbar mostly disables itself when Django is not in debugging mode, but it still leaks some memory in that case.  In particular, the LoggingCollector collects all log messages in memory unconditionally, leading to a small but continuous memory leak.  The debug_toolbar app is also installed on all Huey workers, where it is essentially useless even when in debug mode.

### DIFF
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -94,8 +94,10 @@ INSTALLED_APPS = (
     'huey.contrib.djhuey',
     'swampdragon',
     'simple_email_confirmation',
-    'debug_toolbar',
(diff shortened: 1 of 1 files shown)


haikuginger:
This looks good to me. Conditional :+1: once you change it to `+= ('debug_toolbar',)`

smarnach:
@haikuginger Thanks for the quick review!  And that's an embarassing bug you found.  I only tested it with `DEBUG = False`, so I didn't notice…

PR #125, opened 2016-08-05
Use new version of the user management playbook.

smarnach:
Make use of the user management playbook updated in edx/configuration#3274 and edx/edx-platform#13124.

#### Testing instructions

From the Django shell of an instance manager installation, run these commands (replacing the subdomain and the user name with appropriate values):
```
    sub_domain='user-creation-test',
    edx_platform_repository_url='https://github.com/edx/edx-platform.git',
    configuration_version='master',
    openedx_release='master',
)
instance.lms_users.add(User.objects.get(username='some_existing_user'))
instance.spawn_appserver()
```
This will deploy a new instance and create a user on it.  Test that the user can log into the LMS and they have staff and superuser permissions.

### DIFF
--- a/instance/models/openedx_appserver.py
+++ b/instance/models/openedx_appserver.py
@@ -126,9 +126,9 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
     ))
     lms_user_settings = models.TextField(blank=True, help_text='YAML variables for LMS user creation.')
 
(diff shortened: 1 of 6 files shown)


haikuginger:
:+1:

- [x] I tested this: I spun up a local instance of OpenCraft IM, and followed the testing instructions. I was then able to log into the edX portal with the password I created. I had full privileges available in LMS and Studio.
- [x] I read through the code
- [x] No documentation necessary.

PR #127, opened 2016-08-10
Support multiple Github admin users and organizations.

smarnach:
This feature allows to configure a list of admin  Github users and organizations for each instance who will be given SSH access to the VMs belonging to the instance.

#### Testing instructions

Run the included migration using

    make migrate

In the Django console, for any `OpenEdXInstance`, say `instance`, set the `github_admin_users` field to a list containing at least one Github organization and one Github user, e.g.

    instance.github_admin_users = ['smarnach-org', 'smarnach']

The Github API token you are using needs to have the `read:org` permission, and needs to be a member of the org, and the org needs to have a team called "Owners".

Then launch an AppServer for the instance using

    instance.spawn_appserver()

Once the provisioning started, look at the Ansible configuration of that AppServer (e.g. in the web interface).  It should contain the following entry:

    COMMON_USER_INFO:
    - {github: true, name: smarnach-test, type: admin}
    - {github: true, name: smarnach, type: admin}

This list of users should include all members in the Owners team of the org as well as all users listed individually.

You could wait for the provisionig to complete and try to log in as one of these users, but I don't think that's strictly necessary, since this feature has always worked, and I only changed how the list of users is built.

You can also try reverting the migration

    honcho run ./manage.py migrate instance 0056

and applying it again

    make migrate

After that cycle, the first element of `git_admin_users` should have been preserved.


instance/migrations/0057_github_admin_users.py
```diff
@@ -0,0 +1,68 @@
...
+
+def _users_to_org_name(instance):
+    """Convert to convert the user list back to the old org name field by picking the first item."""
+    if instance.github_admin_users:
+        instance.github_admin_organization_name = instance.github_admin_users[0]
+        instance.save()
+
+
+def _process_instances(apps, schema_editor, callback):
+    """Aplly the given callback to all OpenEdXAppServer and OpenEdXInstance instances."""
```
haikuginger: Typo: `s/Aplly/Apply/`

instance/models/openedx_appserver.py
```diff
@@ -87,9 +88,12 @@ class Meta:
 
     # Misc settings:
     use_ephemeral_databases = models.BooleanField()
-    github_admin_organization_name = models.CharField(
-        max_length=200, blank=True, default=settings.DEFAULT_ADMIN_ORGANIZATION,
-        help_text="GitHub organization whose users will be given SSH access to this instance's VMs",
+    github_admin_users = JSONField(
```
haikuginger: So, I understand that organizations and users are under a common namespace, but I don't think that's necessarily self-evident to all possible users of our platform, especially considering that the field is named `github_admin_users`. I would prefer that we update the `github_admin_organization_name` field to be `github_admin_organizations` and also add `github_admin_users`, keeping them as separate fields that can be addressed individually.

Of course, `github_admin_username_list()` should still return all users, whether directly listed or derived.

instance/tests/fixtures/api/appserver_detail.json
```diff
@@ -18,7 +18,7 @@
     "edx_platform_commit": "073e9c56a8093691dea782c8b727d839260833b1",
     "openedx_release": "master",
     "use_ephemeral_databases": true,
-    "github_admin_organization_name": "open-craft",
+    "github_admin_organization_name": "[\"open-craft\"]",
```
haikuginger: Does the name of this variable need to change based on the new database field name(s)?

instance/tests/fixtures/api/instance_detail.json
```diff
@@ -9,7 +9,7 @@
     "email": "contact@example.com",
     "protocol": "http",
     "use_ephemeral_databases": true,
-    "github_admin_organization_name": "open-craft",
+    "github_admin_organization_name": "[\"open-craft\"]",
```
haikuginger: See above.

instance/models/openedx_appserver.py
smarnach: I considered that approach as well, but in the end decided to go with what I have now.

The help text should make clear that you can have either organizations or users in the list.  We currently don't have any UI for this, and if we add one, we could explain this as well.

Github itself also mixes the two concepts:  You can navigate to either `https://github.com/<user>` or `https://github.com/<org>`, and you can query both types of entities using the `/users` endpoint of the API.  On Github, they are just two different kinds of users, so I treated them the same way on our end.

I don't think the problem with the reverse migration you mention is relevant.  The reverse migration is only provided to make it easier to switch between branches during development.  It will never be run in a production environment.

That said, I don't have a strong preference either way, and since you seem to prefer the other approach, I can just as well go with it.  It actually makes the code a bit simpler, and reduces the number of queries to the Github API.

instance/tests/fixtures/api/appserver_detail.json
smarnach: Yes, definitely. :)

instance/models/openedx_appserver.py
haikuginger: > That said, I don't have a strong preference either way, and since you seem to prefer the other approach, I can just as well go with it. It actually makes the code a bit simpler, and reduces the number of queries to the Github API.

Let's do that then; I think it'll give us a cleaner, more explicit codebase that'll be easier to maintain.

smarnach:
@haikuginger Thanks for the review and the comments!  I hope to have addressed them in the latest commit.

I needed to split up the migrations in three steps: Add the new fields, perform the data migration, remove the old fields.  This was necessary since I discovered cases where Postgres would refuse to perform all the steps in a single transaction.

Please take another look.

instance/models/openedx_appserver.py
```diff
@@ -87,9 +88,18 @@ class Meta:
...
     # Misc settings:
     use_ephemeral_databases = models.BooleanField()
-    github_admin_organization_name = models.CharField(
-        max_length=200, blank=True, default=settings.DEFAULT_ADMIN_ORGANIZATION,
-        help_text="GitHub organization whose users will be given SSH access to this instance's VMs",
+    github_admin_organizations = JSONField(
+        max_length=256,
+        blank=True,
+        default=[],
+        help_text="A list of Github organizations whose members will be given SSH admin access to "
```
haikuginger: This is just members of the Owners group on that org, correct?

instance/models/openedx_appserver.py
```diff
@@ -144,6 +154,8 @@ def set_field_defaults(self):
         """
         Set default values.
         """
+        if not self.github_admin_organizations and settings.DEFAULT_ADMIN_ORGANIZATION:
```
haikuginger: Can we change this to `if not (self.github_admin_organizations or self.github_admin_users) and settings.DEFAULT_ADMIN_ORGANIZATION`? If any users have access, I think we shouldn't set a default. @antoviaque or @bradenmacdonald may have an opinion on that though.

instance/tests/fixtures/api/appserver_detail.json
```diff
@@ -18,7 +18,8 @@
     "edx_platform_commit": "073e9c56a8093691dea782c8b727d839260833b1",
     "openedx_release": "master",
     "use_ephemeral_databases": true,
-    "github_admin_organization_name": "open-craft",
+    "github_admin_organizations": "[\"open-craft\"]",
+    "github_admin_users": "[]",
```
haikuginger: Should these be quoted lists, or JSON lists?

instance/tests/fixtures/api/instance_detail.json
```diff
@@ -9,7 +9,8 @@
     "email": "contact@example.com",
     "protocol": "http",
     "use_ephemeral_databases": true,
-    "github_admin_organization_name": "open-craft",
+    "github_admin_organizations": "[\"open-craft\"]",
+    "github_admin_users": "[]",
```
haikuginger: See above

instance/models/openedx_appserver.py
bradenmacdonald: We will almost always want `DEFAULT_ADMIN_ORGANIZATION`to be included on the list no matter what, so I think the way it is now is preferable.

instance/models/openedx_appserver.py
```diff
@@ -23,6 +23,7 @@
 from django.db import models
 from django.template import loader
 from django.utils.text import slugify
+from django_extensions.db.fields.json import JSONField
```
bradenmacdonald: Can we not use the official one instead? `from django.contrib.postgres.fields import JSONField`

haikuginger:
Sorry @smarnach! In addition to the above, could you merge the three migrations so that they're processed as an atomic transaction when possible?

instance/models/openedx_appserver.py
smarnach: @bradenmacdonald The problem with the current approach is that Django wants to create a new migreation whenever `DEFAULT_ADMIN_ORGANIZATION` is changed, which is the reason why we avoid setting default values based on settings in the field constructor.  We mostly got away with this because we have all set this to the empty string in our dev environments, but on production it was always complaining about pending changes.

instance/models/openedx_appserver.py
smarnach: @bradenmacdonald No, we can't, since it requires Postgres 9.5 or above, and our production server is 9.3.  I also had to update django_extensions to the latest version, since the `JSONField` in the version we used before was broken.

instance/tests/fixtures/api/appserver_detail.json
smarnach: The `JSONField`s are internally represented as `CharField`s, and the REST framework passes them on as strings, so it's correct the way it is.

haikuginger: Thanks for clarifying!

instance/models/openedx_appserver.py
smarnach: @haikuginger That change makes sense, since the only way to get in a situation that an admin user is set, but the list of admin orgs is empty while a default org is configured is by manually deleting the org from the list of admin orgs, in which case we don't want it to reappear.

instance/models/openedx_appserver.py
smarnach: @haikuginger Yes, correct, clarified.

smarnach:
@haikuginger As mentioned in my previous comment, we can't join the migrations in a single one.  From the [docs](https://docs.djangoproject.com/en/1.9/ref/migration-operations/#django.db.migrations.operations.RunPython):

> On databases that do support DDL transactions (SQLite and PostgreSQL), `RunPython` operations do not have any transactions automatically added besides the transactions created for each migration (the `atomic` parameter has no effect on these databases). Thus, on PostgreSQL, for example, you should avoid combining schema changes and `RunPython` operations in the same migration or you may hit errors like `OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events`.

I found out by actually running into this problem, and the only solution is to split the migrations up.

I think I have addressed all comments again, most of them by saying "no, we can't". :)

instance/models/openedx_appserver.py
antoviaque: @smarnach I understood @bradenmacdonald 's comment as being better to keep it like @haikuginger wrote it above (`if not self.github_admin_organizations and settings.DEFAULT_ADMIN_ORGANIZATION`) - that wouldn't create a migration?

In any case, I think that if any users have access but no list of organization is set, we should set the default team - I agree with @bradenmacdonald that we will almost always want `DEFAULT_ADMIN_ORGANIZATION` to be included on the list .

bradenmacdonald: Yes, when I said "the way it is now", I meant to leave `if not self.github_admin_organizations and settings.DEFAULT_ADMIN_ORGANIZATION:` as @smarnach wrote (as the diff shows now). Sorry for the ambiguity.

instance/models/openedx_appserver.py
bradenmacdonald: Ah, damn - proper JSON fields would be nicer, and might also fix the fixture serialization issue that @haikuginger was asking about. Unfortunately upgrading PostgreSQL can be a bit tricky.

instance/models/openedx_appserver.py
```diff
@@ -241,12 +253,14 @@ def github_admin_username_list(self):
...
 
-        Admins are the members of the default team of the `github_admin_organization_name` org
+        Admins are all users listed in github_admin_users and the members of the default team of the
+        organizations in github_admin_organizations.
         """
-        if self.github_admin_organization_name:
-            return get_username_list_from_team(self.github_admin_organization_name)
-        else:
-            return []
+        admin_users = []
```
haikuginger: Nit: It shouldn't make a difference in usage, but it's possible for a user to be added multiple times if she exists in more than one organization, or if she's manually specified in addition to belonging to a member organization.

haikuginger:
:+1:, conditional on green unit tests.

- [x] I tested this: With this branch set up, I was able to migrate an existing instance, add organizations and users, and have them reflected in the settings when spawning an app server. I ran unit tests locally, but they failed during `make test_migrations_missing`.
- [x] I read through the code
- [x] No documentation needed.

instance/models/openedx_appserver.py
smarnach: @bradenmacdonald @antoviaque Ah, I got that wrong.  I'll change it back.

instance/models/openedx_appserver.py
smarnach: @haikuginger You are right with that.  I thought about it when I wrote the code, and decided that it's good enough as it is.  The case is rare, and it works fine as it is, you will only end up with the SSH keys of the affected users added to `authorized_keys` twice, which doesn't really hurt.

PR #128, opened 2016-08-12
Enable full Swift support for OpenCraft IM instances

smarnach:
_PR edits by @haikuginger_

We found that we weren't providing sufficient information to fully enable Swift support for OpenCraft IM instances. This PR updates our logic and playbooks to pass necessary information through appropriately. This code is deployed on our [staging instance manager.](https://stage.console.opencraft.com)

**Dependencies**: https://github.com/open-craft/configuration/commit/33a085b9d5fbcebc560e3ce5c4062f2f7a87b3e2

**Testing instructions**:

1. Deploy a new instance on Stage IM; be sure to provide necessary OpenStack settings during initialization.

2. Start a new appserver for that instance.

3. Test that instructor report CSVs can be created and downloaded on that appserver.

**Reviewers**
- [ ] @itsjeyd

### DIFF
--- a/instance/templates/instance/ansible/swift.yml
+++ b/instance/templates/instance/ansible/swift.yml
@@ -1,6 +1,9 @@
 EDXAPP_SETTINGS: 'openstack'
 XQUEUE_SETTINGS: 'openstack_settings'
-COMMON_VHOST_ROLE_NAME: 'openstack'
+
+VHOST_NAME: 'openstack'
+# COMMON_VHOST_ROLE_NAME is deprecated; keeping it around for compatibility
+COMMON_VHOST_ROLE_NAME: '{{ VHOST_NAME }}'
 
 EDXAPP_DEFAULT_FILE_STORAGE: 'swift.storage.SwiftStorage'
 EDXAPP_FILE_UPLOAD_STORAGE_BUCKET_NAME: '{{ container_name }}'
@@ -15,12 +18,11 @@ EDXAPP_GRADE_STORAGE_CLASS: 'swift.storage.SwiftStorage'
 EDXAPP_GRADE_STORAGE_KWARGS:
   name_prefix: 'grades-download/'
 
-XQUEUE_SETTINGS: 'openstack_settings'
 XQUEUE_SWIFT_USERNAME: '{{ user }}'
 XQUEUE_SWIFT_KEY: '{{ password }}'
 XQUEUE_SWIFT_TENANT_NAME: '{{ tenant }}'
 XQUEUE_SWIFT_AUTH_URL: '{{ auth_url }}'
-XQUEUE_SWIFT_AUTH_VERSION: '2'
+XQUEUE_SWIFT_AUTH_VERSION: '{{ EDXAPP_SWIFT_AUTH_VERSION }}'
 XQUEUE_SWIFT_REGION_NAME: '{{ region }}'
 XQUEUE_UPLOAD_BUCKET: '{{ container_name }}'
 XQUEUE_UPLOAD_PATH_PREFIX: 'xqueue'
--- a/instance/templates/instance/ansible/swift.yml
+++ b/instance/templates/instance/ansible/swift.yml
@@ -14,3 +14,13 @@ EDXAPP_SWIFT_REGION_NAME: '{{ region }}'
 EDXAPP_GRADE_STORAGE_CLASS: 'swift.storage.SwiftStorage'
 EDXAPP_GRADE_STORAGE_KWARGS:
   name_prefix: 'grades-download/'
+
+XQUEUE_SETTINGS: 'openstack_settings'
+XQUEUE_SWIFT_USERNAME: '{{ user }}'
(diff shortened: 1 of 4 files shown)


haikuginger:
This, in combination with our modified `configuration` repo that installs OpenStack requirements as part of platform setup, should be ready to go.

haikuginger:
@itsjeyd, let's go ahead with review. I'm investigating test failures.

instance/templates/instance/ansible/swift.yml
```diff
@@ -1,8 +1,26 @@
-# The following line is commented out until the openstack role has been merged upstream.
-# EDXAPP_DEFAULT_FILE_STORAGE: 'swift.storage.SwiftStorage'
+EDXAPP_SETTINGS: 'openstack'
+XQUEUE_SETTINGS: 'openstack_settings'
+COMMON_VHOST_ROLE_NAME: 'openstack'
```
itsjeyd: @haikuginger The configuration repo as of [eucalyptus/1rc1](https://github.com/open-craft/configuration/tree/eucalyptus/1rc1) doesn't use this variable (it's not referenced anywhere). It has been replaced by [`VHOST_NAME`](https://github.com/open-craft/configuration/blob/eucalyptus/1rc1/playbooks/roles/vhost/defaults/main.yml#L16), so that's probably what we should be using here.

We might need to keep `COMMON_VHOST_ROLE_NAME` to stay backward compatible for a little longer (until we've migrated all of our instances to Eucalyptus). Please check before removing it.

instance/templates/instance/ansible/swift.yml
```diff
@@ -1,8 +1,26 @@
...
 EDXAPP_SWIFT_KEY: '{{ password }}'
 EDXAPP_SWIFT_TENANT_NAME: '{{ tenant }}'
 EDXAPP_SWIFT_AUTH_URL: '{{ auth_url }}'
 EDXAPP_SWIFT_REGION_NAME: '{{ region }}'
+
+EDXAPP_GRADE_STORAGE_CLASS: 'swift.storage.SwiftStorage'
+EDXAPP_GRADE_STORAGE_KWARGS:
+  name_prefix: 'grades-download/'
+
+XQUEUE_SETTINGS: 'openstack_settings'
```
itsjeyd: @haikuginger Duplicate of L2

itsjeyd:
@haikuginger :+1: once comments are addressed and build is green.

- [x] I tested this: Created a new [instance](https://stage.console.opencraft.com/instance/330) on stage and downloaded CSV report from instructor dashboard; works as expected
- [x] I read through the code
- [ ] <strike>Includes documentation</strike> (PR doesn't introduce new settings)

haikuginger:
@itsjeyd, care to take a look at the testing change? Thanks!

itsjeyd:
@haikuginger Had a look; that change looks like it should fix the integration test problems for the time being.

haikuginger:
We decided to not mess with the testing; rather, we're going to pursue updating our integration tests to accommodate Dogwood so that we don't need to.

bradenmacdonald:
Is this change backwards compatible with Dogwood, or can we only deploy Eucalyptus once this merges? We may have a need to use Dogwood form time to time over the next month or two.

smarnach:
@braden: We can still deploy Dogwood instances, provided that we either use ephemeral databases or configure S3 by setting the fields `s3_access_key`, `s3_secret_access_key` and `s3_bucket_name` on the instance.  Trying to deploy an instance with permanent databases, but without S3 configuration will try to use Swift, which will fail for Dogwood.

In this case, it was relatively straightforward to keep backward compatibility.  I've encountered other cases where it is awkward to try to support multiple named releases at the same time.  In the long run, we probably need some mechanism to dispatch based on the release.

bradenmacdonald:
Makes sense. Thanks @smarnach.

haikuginger:
Rebased from 76bd6d58a4c101b7961a0e799c574ab81ce932e8.

haikuginger:
@itsjeyd, I've rebased this onto master. Integration testing _should_ pass now, so we'll just need a final 👍 before squash and merge.

itsjeyd:
@haikuginger It doesn't look like you made any additional changes to the code, so on that front we're good to go. However, the integration tests still aren't passing. The provisioning process for `test_external_databases` is failing with `ImportError: No module named swift.storage`. Normally, the `openstack` role should take care of installing that dependency, but there's no trace of it in the CircleCI logs. It looks like it hasn't been added to the [opencraft_integration.yml](https://github.com/open-craft/configuration/blob/integration/playbooks/opencraft_integration.yml) playbook. That might be why `test_external_databases` fails.

smarnach:
@itsjeyd Right, good catch.  I didn't need the role to make the tests pass, but with the additional changes in this PR it is needed.  I'll add it to the playbook.

smarnach:
I've added the role and triggered a new build.

haikuginger:
Green build; merging.

PR #129, opened 2016-08-18
Enable Let's Encrypt automatic certificate deployment

mfitzp:
Updates `vars.yml` to set LETS_ENCRYPT_DOMAINS with:
- internal_lms_domain
- external_lms_domain
- internal_lms_preview_domain
- external_lms_preview_domain
- internal_studio_domain
- external_studio_domain

In combination with PR on openedx/configuration
( edx/configuration#3313 ) this should automatically
enable Let's Encrypt SSL certs on newly deployed hosts.

### DIFF
--- a/instance/templates/instance/ansible/vars.yml
+++ b/instance/templates/instance/ansible/vars.yml
@@ -25,6 +25,13 @@ CMS_HOSTNAME: '{{ instance.studio_domain_nginx_regex }}'
 {% if instance.protocol == 'https' %}
 NGINX_ENABLE_SSL: true
 NGINX_REDIRECT_TO_HTTPS: true
+LETS_ENCRYPT_DOMAINS: 
+  - "{{ instance.internal_lms_domain }}"
+  - "{{ instance.external_lms_domain }}"
+  - "{{ instance.internal_lms_preview_domain }}"
+  - "{{ instance.external_lms_preview_domain }}"
+  - "{{ instance.internal_studio_domain }}"
+  - "{{ instance.external_studio_domain }}"
 {% else %}
 NGINX_ENABLE_SSL: false
 NGINX_REDIRECT_TO_HTTPS: false
--- a/instance/templates/instance/ansible/vars.yml
+++ b/instance/templates/instance/ansible/vars.yml
@@ -27,11 +27,11 @@ NGINX_ENABLE_SSL: true
 NGINX_REDIRECT_TO_HTTPS: true
 LETS_ENCRYPT_DOMAINS: 
   - "{{ instance.internal_lms_domain }}"
-  - "{{ instance.external_lms_domain }}"
   - "{{ instance.internal_lms_preview_domain }}"
-  - "{{ instance.external_lms_preview_domain }}"
   - "{{ instance.internal_studio_domain }}"
-  - "{{ instance.external_studio_domain }}"
+{% if instance.external_lms_domain %}  - "{{ instance.external_lms_domain }}"{% endif %}
+{% if instance.external_lms_preview_domain %}  - "{{ instance.external_lms_preview_domain }}"{% endif %}
+{% if instance.external_studio_domain %}  - "{{ instance.external_studio_domain }}"{% endif %}
 {% else %}
 NGINX_ENABLE_SSL: false
 NGINX_REDIRECT_TO_HTTPS: false


mfitzp:
This is the continuation to the work on the PR on edx/configuration. It currently sets all domains (internal/external) as it sounds as though internal domains are available on the external DNS and should work.

Unset domains (empty string) will be consumed on the command-line certbot (they are listed as a space-separated list), and an internal domain is listed first as this will (always?) be set. However, external domains could be wrapped in `{% if %}` blocks, e.g.

    - "{{ instance.internal_lms_domain }}"
    - "{{ instance.internal_lms_preview_domain }}"
    - "{{ instance.internal_studio_domain }}"
    {% if instance.external_lms_domain %}  - "{{ instance.external_lms_domain }}"{% endif %}
    {% if instance.external_lms_preview_domain %}  - "{{ instance.external_lms_preview_domain }}"{% endif %}
    {% if instance.external_studio_domain %}  - "{{ instance.external_studio_domain }}"{% endif %}

bradenmacdonald:
@mfitzp Thanks - that makes sense to me (using `{% if`). Will pass this on to @haikuginger .

antoviaque:
Cool to see this : )

This will need an integration test.

mfitzp:
Re: the integration test happy to implement, but not sure how to approach. Are the integration tests run on a machine with a publicly available domain/DNS to perform the cert check (in `env.test` the domain is set to `DEFAULT_INSTANCE_BASE_DOMAIN='example.com'`, and I've found various others e.g. `opencraft.hosting`, `pr234.sandbox.awesome.hosting.org`, `openedxhosting.com`, `pr12338.sandbox.example.com`).

Or should the tests simply ensure the vars (`LETS_ENCRYPT_DOMAINS`) is set correctly?

bradenmacdonald:
@mfitzp 

> Are the integration tests run on a machine with a publicly available domain/DNS

Yes. The vars that configure our env tests are necessarily private, but you can assume that the integration test environment has full ability to provision publically accessible OpenStack VMs and update DNS to point to them.

> Am I right in thinking this playbook is used to run the tests?

Yep, that's the one.

mfitzp:
I've pushed the configuration changes to bring up the servers with https support over in [this PR against the integration branch]. 

For the integration test, I was looking at something like this in `/instance/tests/integration`: 

    def assert_instance_https_valid_cert(self, instance):
        """
        Check that the given instance has a valid https certification
        """
        instance.refresh_from_db()
        self.assertIsNotNone(instance.active_appserver)
        self.assertEqual(instance.active_appserver.status, AppServerStatus.Running)
        self.assertEqual(instance.active_appserver.server.status, ServerStatus.Ready)
        server = instance.active_appserver.server

        try:
            requests.get('https://{0}'.format(server.public_ip)).raise_for_status()
        except Exception:  # pylint: disable=broad-except
            raise

`requests` automatically raises an SSLerror if the cert is invalid, which it will be unless the Let's Encrypt request succeeded. The question is how to implement the domain name for the request — the `server` model doesn't hold. Should I pull this direct from the config? I guess this will mean splitting to two tests, one each for lms/cms.

haikuginger:
@mfitzp, I'd go ahead and pull the relevant domain(s) from the `instance` object. And, of course, you'll want to catch a more specific exception. You may also want to catch the exception and do a `self.fail()` with a helpful error message.

haikuginger:
I actually do have a problem with this. I'll do some testing to confirm, but the problem is that appservers aren't given a domain name immediately when they're spawned - it happens when the appserver is made active. As a result, by default, provisioning new appservers will fail when they get to the letsencrypt role.

We may want to only fill in the `LETS_ENCRYPT_DOMAIN` variable when an appserver is active. Your upstream PR will only execute the role when that variable has a nonempty value, so leaving it empty will allow provisioning to proceed normally.

Alternatively, we can generate the certificates using OpenCraft IM directly, since the DNS will generically point subdomains to the IM until an appserver is made active, then copy the certificates to the instance once they're available, but that doesn't take advantage of the role you created for OpenEdX, and is pretty much out of scope for this task.

haikuginger:
It may be a good option to add an optional kwarg to `spawn_appserver()` (and the supporting backend) to immediately activate the app server as soon as we have a VM IP address.

bradenmacdonald:
@haikuginger Even though the DNS is not changed until the instance is activated, I thought that the letsencrypt role would still configure certbot properly, and schedule it with cron; at first, certbot will fail so the instance will use a self-signed cert, but once the instance is activated, certbot will run (triggered by cron) and then install the new cert and restart nginx. This obviously isn't the best, but is at least workable.

Long-term, yes, we need to set up our VMs (and the default placeholder site) to forward verification requests to OpenCraft IM, so it can manage the verifications as @mfitzp described:

> having a dedicated VM to generate the certificates and pass them to the others (or, a separate server, or the balancer...). The load balancer (or the VMs?) would need to redirect `/.well-known` traffic to this server to ensure the validation works OR use DNS-01 validation itself (it doesn't actually matter at that point).
> B could be implemented as an option in the lets-encrypt playbook — auto-downloading updated certs from a target (e.g. via `scp` path defined in a var) and then triggering a renewal-download cron. The advantage here is that this setup can run before the server goes up — the first time nginx is spun up it will be serving with a valid certificate.

bradenmacdonald:
> It may be a good option to add an optional kwarg to spawn_appserver() (and the supporting backend) to immediately activate the app server as soon as we have a VM IP address.

I don't want to do this in general, because it could result in our beta testers or users sometimes going to their site and seeing a broken VM for an extended period of time, during provisioning and if the provisioning fails.

haikuginger:
@bradenmacdonald - My understanding is that Ansible fails fast, so if the play to initially get the certs fails, we won't even get to the stage where we set up the cron task (and would, I believe, not do the rest of the standard provisioning process).

I believe that the behavior you're looking for would be the case if we were to change the `configuration` PR to set `ignore_errors` to True for the `run certbot` task (allowing the playbook to progress past an initial failure), and moved symlinking the certs into the `run_certbot.sh` script (so that the symlinks are created immediately upon the certificates actually becoming available).

mfitzp:
@haikuginger actually I think it's worse than that(!) If the certbot installation fails, then the renewal cron *will not work* because it requires a previously successful installation (the renewal is run off configuration files generated during the creation process).

So running with `ignore_errors` to `True` will result in a broken setup — with nginx looking in the wrong place for the certificates, and a cron running renew that will fail. The good news is that running the `certbot` role after that would fix the installation. We would also need nginx to be installed with `NGINX_REDIRECT_TO_HTTPS=False` during this time or the site will be inaccessible.

Another option is to install nginx with SSL enabled, but postpone the `certbot` run. That leaves the server with the self-signed certificate. However, this causes `requests` to throw an `SSLError` exception which I believe would break the `assert_instance_up` test unless `NGINX_REDIRECT_TO_HTTPS=False` or `verify=False` is used with `request.get` to skip the certificate check. This would allow us to install nginx with SSL, and to test that this has worked... but not test the Let's Encrypt deployment.

Is it possible to trigger the letsencrypt role once the domain is in place?

If not — Your suggestion about the symlinks gave me an idea:   
Could we restructure the letsencrypt role to also put copying the nginx cert config into the shell script? These 3 steps are tightly linked. Running the role would create this script, attempt to get the certs, fail, but continue to install the cron. At this point nginx would be up, but still be using it's self-signed cert, and we'd have a script available to complete the install at some point in the future.

haikuginger:
> Could we restructure the letsencrypt role to also put copying the nginx cert config into the shell script? These 3 steps are tightly linked. Running the role would create this script, attempt to get the certs, fail, but continue to install the cron. At this point nginx would be up, but still be using it's self-signed cert, and we'd have a script available to complete the install at some point in the future.

That seems reasonable to me. I don't think we can really have a surefire way to trigger the role when the domain starts pointing to the server, so the best route is likely to set up a resilient script that can keep failing until it succeeds.

mfitzp:
@haikuginger I've implemented this as described in https://github.com/mfitzp/configuration/tree/lets-encrypt-sh (I put this in a separate branch for now, as I didn't want to push new code to the opened PR while it's being tested).

This creates the script as a shell script (in `/opt`) to allow it to be re-run at a future time to complete the installation. But it still doesn't run automatically (i.e. on the cron).

If we want to implement that it occurred to me that we could use the --pre-hook of `certbot-auto`, together with a test to confirm if the installation was complete (e.g. test for presence of the symlinked certs). One it was set up that script would effectively become a no-op.

The script should probably also contain the nginx service status checking (for the reload|stop that is currently in the ansible script), but I wanted to check for a general 👍 before doing that. The result would be that the ansible script itself wouldn't do very much at all (!) save for creating the `.well-known` folder and starting up the cron.

haikuginger:
Thanks for this change, @mfitzp! That sounds like a good way to go. It is nice to be able to use Ansible for more things rather than less, but functionality needs to come first.

mfitzp:
@haikuginger am back working on this (have been busy PhD thesis writing). 

I created a working implementation of the semaphore to [letsencrypt-sh](https://github.com/mfitzp/configuration/tree/lets-encrypt-sh) branch (HEAD~1). This used a semaphore file to flag the installation as incomplete/complete. If the install fails (certificate not retrieved) it can be re-run at a later date. The idea mentioned above to put this into the `--pre-hook` won't work unfortunately, as that is *only* run if there are updates (and there won't be until the thing is installed). 

Doing this I realised that the `certbot-auto` script was (now) smart enough to run a renewal if the domains were all already configured. So this means we can use the same `run_certbot.sh` script for the renewal run. The latest commit updates to this, which simplifies things a lot. The only niggle is that there are now rendundant link/copy of files even when there is no update, but this will have no effect (the config is only reloaded if the certs change).

I think this should solve all the problems we've identified so far.

haikuginger:
@mfitzp, this is looking really good. Two small nits before you push this to your existing PR branch. 

First, can we fail when an email isn't provided? It looks like that's essential; I know LE requires it when the domain hasn't already been registered.

Second, can we move the `service nginx restart` from certbot's `post-hook` arguments to its own line after the symlinks? I believe we'd need to restart/reload after modifying the configuration.

Once you've done that,  go ahead and push the commits to your existing PR and tag me; I'll explain why the changes, re-test, and give a new thumbs-up; it doesn't look like upstream review has started yet, so it shouldn't be a big deal.

I appreciate your patience on this; this task is probably turning out to be a bit bigger than we had intended.

mfitzp:
@haikuginger Thanks for the suggestions, I've made those changes and pushed them back across to https://github.com/edx/configuration/pull/3313 

No worries, sometimes difficult to spot these things before you get really familiar with the problem!

bradenmacdonald:
@mfitzp @haikuginger  With this PR, will it still be possible to use custom SSL certs? I assume it can be done, by overriding `LETS_ENCRYPT_DOMAINS` to be empty in the instance `configuration_extra_settings`, but I just wanted to confirm. 

(We have some HTTPS instances that are managed with OpenCraft IM now that we might want to keep using with their current certs rather than forcing a change to Let's Encrypt.)

mfitzp:
@bradenmacdonald That's right. If `LETS_ENCRYPT_DOMAINS` is empty the role won't run, and the nginx config will be left as-is.

mfitzp:
@bradenmacdonald @haikuginger Once the other PR is complete I'll complete the integration test here to wrap it up. With the new setup there is no guarantee that the cert will be installed before the server goes up. Should the integration test for a valid certificate, or just test for HTTPS being set up correctly (i.e. should it fail until the Let's Encrypt cert is installed)?

haikuginger:
@mfitzp, requests will throw different errors for a server not being available at a domain vs. having an invalid certificate. An invalid certificate should fail; if the domain isn't reachable, then have the test sleep and try again in a few minutes.

Be sure to update the integration testing vars to include a `LETS_ENCRYPT_EMAIL` value. You might also want to figure out if there's an existing variable we can use to get a value for that field in production, since the provisioning will fail without it.

haikuginger:
@mfitzp, actually, ignore my last comment in re: testing methods. The problem is that the cron task delay is 12 hours, so if the cert didn't get provisioned during the deploy, it wouldn't get provisioned for quite a while.

@bradenmacdonald, I'm not sure the role, as written, lends itself to integration testing - can you comment?

bradenmacdonald:
@haikuginger  Can we use an ansible var to shorten the cron time for integration tests? To every 5 min or so?

haikuginger:
@bradenmacdonald: With some tweaks, that should be doable.

antoviaque:
Superseeded by #145, so closing this. Thank you for the work @mfitzp, and sorry we couldn't merge it.

PR #130, opened 2016-08-24
Configure integration tests to use the Eucalyptus release candidate.

smarnach:
This configures the integration tests to use the current Eucalyptus release candidate instead of the Cypress release.

For edx-platform and configuration, custom branches are used with some extra commits on top of the release candidate:
* [Additional commits for configuration](https://github.com/edx/configuration/compare/open-release/eucalyptus.1rc2...open-craft:integration)
* [Additional commits for edx-platform](https://github.com/edx/edx-platform/compare/open-release/eucalyptus.1rc2...open-craft:integration)

Manual testing isn't necessary for this PR; all we need is the integration tests pass.


haikuginger:
Much quicker review than I would have thought. Tests pass no problem, and you've added some good documentation, so I'm :+1:.

PR #131, opened 2016-08-30
Allow to set separate stable versions for edx-platform and configuration.

smarnach:
Currently, we only have one setting `OPENEDX_RELEASE_STABLE_REF` to set the stable release ref used by production instances.  This setting is used as `openedx_release` and `configuration_version` for production instances, which means it needs to be a valid ref for edx-platform, configuration, notifier, certs, forum, xqueue and possibly others, so usually only official release names will work.  However, we often want to commit custom changes to the edx-platform and configuration repos.  So far, we managed to do this by having a branch with the same name as an official release tag in the open-craft forks of configuration and edx-platform.  This solution is a confusing hack, so this PR introduces proper settings to set the stable versions of edx-platform and configuration explicitly.  The change is backwards-compatible since these settings default to `OPENEDX_RELEASE_STABLE_REF`.

**Testing instructions**

I'm not sure manual testing is strictly necessary, given how simple the change is.  In case you wish to test it, here is how.

Add the new settings `STABLE_EDX_PLATFORM_COMMIT` and `STABLE_CONFIGURATION_VERSION` to your `.env` file with some values.  In the Django shell, run

    instance = production_instance_factory(sub_domain="some_subdomain")
    vars(instance)

and verify that the fields `edx_platform_commit` and `configuration_version` have the desired values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,9 +272,17 @@ set the following configuration variables:
   not set, the value of `DEFAULT_OPENEDX_RELEASE` will be used.
 * `DEFAULT_FORK`: The fork of `edx-platform` to use by default. Defaults to the
   main repository, `edx/edx-platform`.
-* `OPENEDX_RELEASE_STABLE_REF`: Set this to a tag or branch for a stable
-  Open edX release. It is used as a default value for `configuration_version`
-  and `openedx_release` fields when creating production instances.
+* `OPENEDX_RELEASE_STABLE_REF`: Set this to a tag or branch for a stable Open
+  edX release. It is used as a default value for the `openedx_release` field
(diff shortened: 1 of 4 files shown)


itsjeyd:
@smarnach The code looks good to me, and I verified that it works according to your instructions (thanks!). Aside from fixing the build failure, this is only missing a README update (since it introduces new settings).

itsjeyd:
@smarnach Looking good, thanks for the updates. :+1: once build is green.

- [x] I tested this: Works as advertised in PR description
- [x] I read through the code
- [x] Includes documentation

PR #132, opened 2016-08-31
Add helper functions to upgrade production instances to Eucalyptus.

smarnach:
These functions automate selecting the instances that need an upgrade to Eucalyptus, performing the upgrade and cleaning up the settings after the upgrade.

During the upgrade, the oauth2_provider migration is performed by Ansible using the custom code in [this commit](https://github.com/open-craft/configuration/commit/6c2967d3d6e191806d2628dcfa722cbe6e647cb2).

Here are the custom modifications we use on top of the Eucalyptus.1 release:

* https://github.com/edx/configuration/compare/open-release/eucalyptus.1...open-craft:opencraft-release/eucalyptus.1
* https://github.com/edx/edx-platform/compare/open-release/eucalyptus.1...open-craft:opencraft-release/eucalyptus.1

I've tested the process for [this instance](https://console.opencraft.com/instance/1190/edx-appserver/316/), and the upgrade worked without any problems.

### DIFF
--- a/instance/eucalyptus_upgrade.py
+++ b/instance/eucalyptus_upgrade.py
@@ -17,7 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-Helper function to upgrade production instances from Dogwood to Eucalyptus.
+Helper functions to upgrade production instances from Dogwood to Eucalyptus.
 """
 
 import logging
@@ -49,8 +49,6 @@ def upgrade_to_eucalyptus(instances):
         instance.configuration_version = "opencraft-release/eucalyptus.1"
         instance.openedx_release = "open-release/eucalyptus.1"
         instance.save()
-        for appserver in instance.appserver_set.iterator():
-            appserver.terminate_vm()
         spawn_appserver(instance.ref.pk, mark_active_on_success=True, num_attempts=1)
 
 
--- a/instance/eucalyptus_upgrade.py
+++ b/instance/eucalyptus_upgrade.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
Code LGTM except for terminating the existing VMs, which I've asked you about separately.

I wish we could run an additional playbook instead of setting `COMMON_EUCALYPTUS_UPGRADE`, but I guess that's more work and the oauth fix has to be run at a specific point in time during the existing playbook.

smarnach:
@bradenmacdonald I thought about solutions to put that in a separate playbook, but couldn't come up with anything robust that I'd be able to complete in a day.  The fix indeed needs to be run in somewhere in the middle of the playbook, so simply inserting it is by far the easiest approach.

I'll remove the code to terminate the old appservers (see my comments on the ticket).

itsjeyd:
@smarnach The code looks good to me as well. My only nit is that I think it would be nice to have a dedicated directory for upgrade modules such as `eucalyptus_upgrade.py` -- assuming that future upgrades will require introducing additional modules that serve a similar purpose. But I'm totally fine with crossing that bridge when we get there :)

One question before I give this a thumbs-up: I noticed that the upgraded instance is not uploading files generated via the instructor dashboard to Swift (it seems to be using local storage instead). I think that's because the new app server still has this in its `configuration_storage_settings`:

```
# The following line is commented out until the openstack role has been merged upstream.
# EDXAPP_DEFAULT_FILE_STORAGE: 'swift.storage.SwiftStorage'
```

This doesn't match the settings from [`swift.yml`](https://github.com/open-craft/opencraft/blob/master/instance/templates/instance/ansible/swift.yml) that we have on `master`. Should newly spawned app servers use the most recent version of the Swift settings?

Addressing this might be out-of-scope for this task, but perhaps it would be a good idea to fully migrate production instances to Swift now, while we're still in beta?

smarnach:
@itsjeyd I'm not sure we are going to keep the `eucalyptus_upgrade.py` file around for long.  To me, it's more like a one-off until we have a better way to deal with such issues, and I'd like to remove it again from the repo once it has served its purpos, so I don't think it makes sense to create a dedicated directory for it.

Regarding your comment that the settings didn't get updated, this is because I didn't have the Swift fix installed to the production instance manager yet when I tried the upgrade for Braden's test instance (which I presume is the instance you were looking at).  In general, new appservers will pick up the latest version of the settings, as you can see e.g. for [this instance](https://console.opencraft.com/instance/830/edx-appserver/324/).  (I'm still investigating the deployment failure for this appserver.)

itsjeyd:
@smarnach OK, thanks for the explanation. That makes sense.

:+1:

- [x] I tested this: Successfully registered and activated a new user account on the [upgraded instance](https://console.opencraft.com/instance/1190/edx-appserver/316/) mentioned in the PR description. Checked the storage settings of the [app server](https://console.opencraft.com/instance/830/edx-appserver/324/) mentioned in @smarnach's latest [comment](https://github.com/open-craft/opencraft/pull/132#issuecomment-244039683).
- [x] I read through the code
- [ ] ~~Includes documentation~~ New code is specific to the process of upgrading existing Dogwood instances to Eucalyptus

PR #133, opened 2016-09-01
Pin the Insights and Analytics API versions for instances.

smarnach:
Set the Insights and Analytics API versions to `openedx_release`.  The default value for these versions is `master`, which means that these services won't get pinned to the selected release, but follow master instead.  This is clearly not intended (and probably won't work in many cases).

### DIFF
--- a/instance/templates/instance/ansible/vars.yml
+++ b/instance/templates/instance/ansible/vars.yml
@@ -64,6 +64,8 @@ forum_version: '{{ appserver.openedx_release }}'
 notifier_version: '{{ appserver.openedx_release }}'
 xqueue_version: '{{ appserver.openedx_release }}'
 certs_version: '{{ appserver.openedx_release }}'
+ANALYTICS_API_VERSION: '{{ appserver.openedx_release }}'
+INSIGHTS_VERSION: '{{ appserver.openedx_release }}'
 
 # Misc
 EDXAPP_LANG: 'en_US.UTF-8'


itsjeyd:
@smarnach The code looks good to me. Did you already deploy an app server on stage or prod to test these settings?

smarnach:
@itsjeyd No, I didn't actually use the settings yet.  For Eucalyptus, we decided to simply remove these roles from the playbook, so this fix is to only to prevent someone else will be bitten by the same issue in the future.

smarnach:
@itsjeyd This doesn't change any code, just settings.  I think this change is very low risk, and would be fine merging it without manual testing.  If this change were part of a larger PR, we definitely wouldn't test it.  It's up to you to judge whether you agree with this. :)

itsjeyd:
@smarnach Thanks for updating the PR description.

> If this change were part of a larger PR, we definitely wouldn't test it.

I don't agree with this unless you meant to say "would" instead of wouldn't :P :)

But yeah, I also consider this a low-risk change. :+1:

bradenmacdonald:
@smarnach Do you have time to rebase and merge this before you go? (If not, can you do that, @itsjeyd ?)

smarnach:
@bradenmacdonald Rebased.  The build failure for the old version was due to OVH issues, if I remember correctly, so I'd expect the build to pass now.  Once it's green, it can be merged by whoever gets here first.

PR #134, opened 2016-09-08
Compress application log files

pomegranited:
Adds `opencraft.logging.GzipRotatingFileHandler`, which gzips the instance manager's rotated log files.

*Testing Instructions*

1. Set `LOGGING_ROTATE_MAX_KBYTES=1` in your local OpenCraft IM .env file - this is to trigger log rotation without needing to generate huge logs.
1. Run `make rundev`, and refresh http://localhost:5000/ several times.
1. Ensure that `.gz` log files are created in under `logs/`
1. Ensure that you can `gunzip` and read them.

### DIFF
--- a/opencraft/logging.py
+++ b/opencraft/logging.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


opencraft/logging.py
```diff
@@ -0,0 +1,56 @@
...
+
+    https://docs.python.org/3/howto/logging-cookbook.html#using-a-rotator-and-namer-to-customize-log-rotation-processing
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(GzipRotatingFileHandler, self).__init__(*args, **kwargs)
+        self.namer = GzipRotatingFileHandler._namer
+        self.rotator = GzipRotatingFileHandler._rotator
+
+    @staticmethod
```
haikuginger: I would just define these as the standard instance methods `namer` and `rotator`; that's how they're called from `RotatingFileHandler`. Since we're subclassing, we don't need to dynamically patch them in.

haikuginger:
:+1: conditional on the above change!

pomegranited:
@haikuginger I totally agree that `namer` and `rotator` should be standard instance methods, but it doesn't work!  I think because [these lines](https://hg.python.org/cpython/file/3.5/Lib/logging/handlers.py#l53) set `self.namer=None` and `self.rotator=None`, so the instance methods get dissociated.

With 6249299, removes the `@staticmethod` bit, which is slightly cleaner.

haikuginger:
Looks good to me! 👍

pomegranited:
Sorry, @haikuginger I had to revert my `@staticmethod` remove; prospector didn't like it:

```
Messages
========

opencraft/logging.py
  Line: 44
    pylint: no-self-use / Method could be a function (col 4)
  Line: 48
    pylint: no-self-use / Method could be a function (col 4)
```

PR #138, opened 2016-10-02
Allow user playbook failure in some cases

haikuginger:
When merged, this pull request will allow OCIM to ignore any user provisioning errors that occur if the instance in question has persistent databases and at least one appserver has been provisioned successfully in the past.

**Dependencies**: edx/configuration#3398

**Testing instructions**:

1. Using the branch from edx/configuration#3398 as your `configuration_version` variable, with `configuration_source_repo_url` also set to track that branch, create a new instance with non-ephemeral databases.
2. Spawn an appserver; be sure that your copy of OCIM has at least one user that can be automatically set up on the appserver.
3. Verify that you can login with that user on the appserver; change the email address for that user on the appserver.
4. Spawn another appserver; verify that the process succeeds.

**Reviewers**
- [ ] @pomegranited


haikuginger:
@pomegranited, try using this branch instead - it brings in some Swift fixes that are pending upstream:

https://github.com/open-craft/configuration/tree/haikuginger/allow-user-provisioning-failures-swift

PR #142, opened 2016-10-17
Use the new Huey release that includes the fix.

smarnach:
Upstream already merged the fix and released a new version.


itsjeyd:
@smarnach That was quick! :) :+1:

PR #143, opened 2016-10-19
Add task for cleaning up VMs

itsjeyd:
This PR adds a task to the scheduler which runs every day and (1) shuts down instances whose PRs were merged (more than) 7 days ago and (2) terminates app servers that were created (more than) two days prior to the active app server of the parent instance. It also makes sure that shut-down instances are hidden from the GUI.

cf. [OC-1998](https://tasks.opencraft.com/browse/OC-1998)

**Test instructions**

1. Create an instance that is not associated with a PR:

    * Run `make shell`.

    * Create instance via:

        ```python
        regular_instance = instance_factory(name="Regular instance, no PR", sub_domain="regular-instance-zohoRek6")
        ```
    * Spawn four app servers for the instance (to speed this up you can use separate terminal windows):

        ```python
        first_appserver_id = regular_instance.spawn_appserver()
        second_appserver_id = regular_instance.spawn_appserver()
        third_appserver_id = regular_instance.spawn_appserver()
        fourth_appserver_id = regular_instance.spawn_appserver()
        ```

    * Update the `created` field of the app servers to fake their age:

        ```python
        from datetime import timedelta

        reference_date = timezone.now()
        
        # Obsolete app server:
        first_appserver = regular_instance.appserver_set.get(pk=first_appserver_id)
        first_appserver.created = reference_date - timedelta(days=5)
        first_appserver.save()
        # Recent app server:
        second_appserver = regular_instance.appserver_set.get(pk=second_appserver_id)
        second_appserver.created = reference_date - timedelta(days=4)
        second_appserver.save()
        # Active app server:
        third_appserver = regular_instance.appserver_set.get(pk=third_appserver_id)
        third_appserver.created = reference_date - timedelta(days=3)
        third_appserver.save()
        # Newest app server:
        fourth_appserver = regular_instance.appserver_set.get(pk=fourth_appserver_id)
        fourth_appserver.created = reference_date - timedelta(days=2)
        fourth_appserver.save()
        ```

    * Run the new `clean_up` task:

        ```python
        from instance.tasks import clean_up
        clean_up()
        ```

    * Verify that `regular_instance` is unchanged (i.e., none of its app servers have been terminated). You can use the GUI or do it directly in the shell via something like:

        ```python
        for appserver in regular_instance.appserver_set.all():
            print(appserver)
            print("Status: ", appserver.status)
            print("Status (server): ", appserver.server.status)
        ```

        This should produce the following output:

        ```
        AppServer 1
        Status:  Running [running]
        Status (server):  Ready [ready]
        AppServer 2
        Status:  Running [running]
        Status (server):  Ready [ready]
        AppServer 3
        Status:  Running [running]
        Status (server):  Ready [ready]
        AppServer 4
        Status:  Running [running]
        Status (server):  Ready [ready]
        ```

    * Activate the third app server:

        ```python
        regular_instance.set_appserver_active(third_appserver_id)
        ```

2. Create an instance that is associated with a PR that was merged *more than 7 days ago*:

    * Create instance:

        ```python
        from pr_watch.github import get_pr_by_number
        merged_pr = get_pr_by_number('edx/edx-platform', 13443)
        merged_pr_instance, created = WatchedPullRequest.objects.get_or_create_from_pr(merged_pr)
        ```

    * Spawn app server:

        ```python
        appserver_id = merged_pr_instance.spawn_appserver()
        ```

    * Set app server active:

        ```python
        merged_pr_instance.set_appserver_active(appserver_id)
        ```

3. Create an instance that is associated with a PR that was merged *less than 7 days ago*:

    * Create instance:

        ```python
        from pr_watch.github import get_pr_by_number
        recently_merged_pr = get_pr_by_number('edx/edx-platform', 13611)
        recently_merged_pr_instance, created = WatchedPullRequest.objects.get_or_create_from_pr(recently_merged_pr)
        ```

    * Spawn app server:

        ```python
        appserver_id = recently_merged_pr_instance.spawn_appserver()
        ```

    * Set app server active:

        ```python
        recently_merged_pr_instance.set_appserver_active(appserver_id)
        ```

4. Create an instance that is associated with an open PR (if the PR referenced below is no longer open, you'll need to select a different one):

    * Create instance:

        ```python
        from pr_watch.github import get_pr_by_number
        open_pr = get_pr_by_number('edx/edx-platform', 13771)
        open_pr_instance, created = WatchedPullRequest.objects.get_or_create_from_pr(open_pr)
        ```

    * Spawn app server:

        ```python
        appserver_id = open_pr_instance.spawn_appserver()
        ```

    * Set app server active:

        ```python
        open_pr_instance.set_appserver_active(appserver_id)
        ```

5. Run the `clean_up` task again:

    ```python
    from instance.tasks import clean_up
    clean_up()
    ```

6. Make sure that: 

    * Instance not associated with PR has had it's oldest app server terminated, is unchanged otherwise, and visible in the GUI.
    * Instance associated with PR merged more than 7 days ago has been shut down, and is no longer visible in the GUI.
    * Instance associated with PR merged less than 7 days ago is unchanged, and visible in the GUI.
    * Instance associated with open PR is unchanged, and visible in the GUI.

7. Double-check "OVH - OpenStack - Horizon - OpenCraft IM - Dev" account to see if correct set of app servers left.

**Reviewers**

- [ ] @haikuginger

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -29,7 +29,8 @@ from tldextract import TLDExtract
 
 from instance.gandi import GandiAPI
 from instance.logging import log_exception
-from instance.models.appserver import Status
+from instance.models.appserver import Status as AppServerStatus
+from instance.models.server import Status as ServerStatus
 from instance.utils import sufficient_time_passed
 from .instance import Instance
 from .mixins.openedx_database import OpenEdXDatabaseMixin
@@ -214,7 +215,10 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin,
         if self.appserver_set.count() == 0:
             return False
         all_appservers_terminated = all(
-            appserver.status == Status.Terminated for appserver in self.appserver_set.all()
+            appserver.status == AppServerStatus.Terminated or
+            (appserver.status == AppServerStatus.ConfigurationFailed and
+             appserver.server.status == ServerStatus.Terminated)
+            for appserver in self.appserver_set.all()
         )
         monitoring_turned_off = self.new_relic_availability_monitors.count() == 0
         return all_appservers_terminated and monitoring_turned_off
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -29,6 +29,7 @@ from tldextract import TLDExtract
 
 from instance.gandi import GandiAPI
 from instance.logging import log_exception
+from instance.models.appserver import Status
 from .instance import Instance
 from .mixins.openedx_database import OpenEdXDatabaseMixin
 from .mixins.openedx_monitoring import OpenEdXMonitoringMixin
@@ -197,6 +198,25 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin,
         allowed = string.ascii_letters + string.digits + '_'
         escaped = ''.join(char for char in name if char in allowed)
         return truncate_name(escaped, length=50)
+
+    @property
+    def shut_down(self):
+        """
+        Return True if this instance has been shut down, else False.
+
+        An instance has been shut down if all of its app servers have been terminated,
+        and monitoring has been turned off.
+
+        If an instance has no app servers, we assume that it has *not* been shut down.
+        This ensures that the GUI lists newly created instances without app servers.
+        """
+        if self.appserver_set.count() == 0:
+            return False
+        all_appservers_terminated = all(
+            appserver.status == Status.Terminated for appserver in self.appserver_set.all()
+        )
+        monitoring_turned_off = self.new_relic_availability_monitors.count() == 0
+        return all_appservers_terminated and monitoring_turned_off
 
     def set_field_defaults(self):
         """
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -30,6 +30,7 @@ from tldextract import TLDExtract
 from instance.gandi import GandiAPI
 from instance.logging import log_exception
 from instance.models.appserver import Status
+from instance.utils import sufficient_time_passed
 from .instance import Instance
 from .mixins.openedx_database import OpenEdXDatabaseMixin
 from .mixins.openedx_monitoring import OpenEdXMonitoringMixin
@@ -200,7 +201,7 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin,
         return truncate_name(escaped, length=50)
 
     @property
-    def shut_down(self):
+    def shutdown(self):
         """
         Return True if this instance has been shut down, else False.
 
@@ -372,3 +373,29 @@ class OpenEdXInstance(Instance, OpenEdXAppConfiguration, OpenEdXDatabaseMixin,
         an appserver (read: database) for this instance in the past.
         """
         return not self.successfully_provisioned or self.use_ephemeral_databases
+
+    def terminate_obsolete_appservers(self, days=2):
+        """
+        Terminate app servers that were created (more than) `days`
+        before the currently-active app server of this instance.
+
+        Do nothing if this instance doesn't have an active app server.
+        """
+        active_appserver = self.active_appserver
+        if active_appserver:
+            for appserver in self.appserver_set.all():
+                if sufficient_time_passed(appserver.created, active_appserver.created, days):
+                    appserver.terminate_vm()
+
+    def shut_down(self):
+        """
+        Shut down this instance.
+
+        This process consists of two steps:
+
+        1) Disable New Relic monitors.
+        2) Terminate all app servers belonging to this instance.
+        """
+        self.disable_monitoring()
+        for appserver in self.appserver_set.all():
+            appserver.terminate_vm()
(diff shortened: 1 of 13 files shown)


instance/tasks.py
```diff
@@ -60,3 +63,51 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
...
+            instance.watchedpullrequest.target_fork_name,
+            instance.watchedpullrequest.github_pr_number
+        )
+        if pr['state'] == 'closed':
+            closed_at = github.parse_date(pr['closed_at'])
+            now = datetime.now()
+            if sufficient_time_passed(closed_at, now, 7):
+                instance.disable_monitoring()
+                for appserver in instance.appserver_set.all():
+                    appserver.terminate_vm()
```
antoviaque: @itsjeyd Could these three lines be moved to a `instance.shutdown()` method, to keep this logic internal? This task doesn't need to know how the shutdown happens.

instance/tasks.py
```diff
@@ -60,3 +63,51 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
...
+    of each individual instance.
+
+    Do nothing for instances that don't have an active app server.
+    """
+    for instance in OpenEdXInstance.objects.all():
+        active_appserver = instance.active_appserver
+        if active_appserver:
+            for appserver in instance.appserver_set.all():
+                if sufficient_time_passed(appserver.created, active_appserver.created, 2):
+                    appserver.terminate_vm()
```
antoviaque: @itsjeyd Could lines 98-102 be moved to a `instance.terminate_obsolete_servers(days=2)` method? (And maybe rename this task into `terminate_all_instances_obsolete_appservers()`?

instance/models/openedx_instance.py
```diff
@@ -198,6 +199,25 @@ def database_name(self):
         escaped = ''.join(char for char in name if char in allowed)
         return truncate_name(escaped, length=50)
 
+    @property
+    def shut_down(self):
```
haikuginger: The naming of this property could use some clarification, especially if, as per @antoviaque's request, we're adding a `shutdown()` method to the instance.

instance/tasks.py
haikuginger: I agree.

haikuginger:
@itsjeyd, you're looking pretty good in general. One more change I want to request that @antoviaque might want to weigh in on; should we possibly be terminating appservers with failed configurations? My understanding is that they're still keeping a VM running, and they're not much use, especially if we've got an active appserver that's been successfully provisioned. If not, then we want to modify the "shut down" logic to hide instances that have failed appservers in addition to at least one terminated appserver.

itsjeyd:
Thanks @antoviaque and @haikuginger for the feedback, I'll address your comments tomorrow.

@haikuginger 

> ... should we possibly be terminating appservers with failed configurations? My understanding is that they're still keeping a VM running, ...

I'm not sure what contexts/circumstances you're referring to, but app servers that have status `ConfigurationFailed` do get terminated as part of the task(s) that this PR adds. They just don't get their status set to `Terminated`. Back when we first implemented app server states, we decided to reserve the `Terminated` status for instances that had been running successfully at some point. I don't remember the details of that discussion right now; but changing the way we manage states is something that multiple people (including Braden) would likely want to weigh in on :) So I think it would make sense to discuss that outside of the scope of this PR.

> If not, then we want to modify the "shut down" logic to hide instances that have failed appservers in addition to at least one terminated appserver.

There's a chance that this condition would return too many false positives (since many long running instances have some failed app servers and one or more terminated app servers). But we could change the logic to take into account server (VM) status: The `shut_down` property could return `True` for any instance that only has app servers with status `Terminated` and/or app servers with status `ConfigurationFailed` and server status `Terminated`. That would end up hiding more instances than the current logic, while making it less likely for instances that are still in use to be hidden.

haikuginger:
> The shut_down property could return True for any instance that only has app servers with status Terminated and/or app servers with status ConfigurationFailed and server status Terminated. 

That sounds great! Glad to be wrong on this topic! 😉

itsjeyd:
@antoviaque @haikuginger Could you have another look please?

instance/models/openedx_instance.py
```diff
@@ -198,6 +201,28 @@ def database_name(self):
...
+
+        An instance has been shut down if all of its app servers have been terminated,
+        and monitoring has been turned off.
+
+        If an instance has no app servers, we assume that it has *not* been shut down.
+        This ensures that the GUI lists newly created instances without app servers.
+        """
+        if self.appserver_set.count() == 0:
+            return False
+        all_appservers_terminated = all(
```
itsjeyd: @haikuginger I think that mixing app server states with properties that provide state(-related) information has the potential to be confusing, so I didn't add an `is_shutdown` property to the app server. But I did move the logic from this block to a separate function to be able to simplify the generator expression (which now looks almost identical to what you were suggesting). See 6e9255b7457431601977d65faea992788e8ec7d0.

itsjeyd:
@haikuginger I addressed your latest comments. Another look please? :)

instance/tasks.py
```diff
@@ -60,3 +63,40 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
                 # finish and replace the second as the active server. We are not really worried about that for now.
                 instance.set_appserver_active(appserver_id)
             break
+
+
+@db_task()
+def shut_down_instances():
```
antoviaque: Nit: Worth having a more descriptive method name? It makes it sound more generic than it is.

itsjeyd:
Thanks @antoviaque, I addressed your nit via 71d02288f2c9e8774236455c09a76be5e5207d30.

itsjeyd:
Thanks @haikuginger! Merging now.

PR #147, opened 2016-10-26
Add secret key generation to OpenCraft IM

haikuginger:
This PR adds logic to automatically generate secret keys for a variety of services populated by OpenCraft IM.

**Dependencies**: None

**Testing instructions**:
1. Using a copy of OCIM as it exists right now, set up an instance and spawn an appserver. There's no need to wait for the appserver to provision; just look at the appserver's vars and observe that items like `EDXAPP_EDXAPP_SECRET_KEY` are not present.
2. Pull this branch into your existing OCIM instance and migrate it.
3. Spawn a new appserver on the existing instance. Again, no need to wait for it to provision; just look at the vars in OCIM and observe that items like `EDXAPP_EDXAPP_SECRET_KEY` are still missing - for instances that don't have a master secret key set, we don't want to change things.
4. Create a new instance and spawn an appserver. Observe that variables like `EDXAPP_EDXAPP_SECRET_KEY` and `FORUM_API_KEY` are populated with seemingly-random values.
5. Spawn a new appserver on the same instance; observe that its secret key vars have the same values as the first one.

1. The current setup is a little awkward. Because the secret keys are derived based on the variable names, we need to have those names in source, not just in a template. I explored other options, but the current setup is robust and allowed me to get the code deployed. I am very open to suggestions on how to improve it - one thought I had was by feeding in the secret keys we're interested in filling as environment variables and building the YAML string entirely in Python, but that seems a bit obfuscated.

**Reviewers**
- [ ] @bradenmacdonald


instance/models/mixins/secret_keys.py
```diff
@@ -0,0 +1,111 @@
...
+                "EDXAPP_PROFILE_IMAGE_SECRET_KEY": self.generate_key('EDXAPP_PROFILE_IMAGE_SECRET_KEY'),
+                "INSIGHTS_SECRET_KEY": self.generate_key('INSIGHTS_SECRET_KEY'),
+                "NOTIFIER_LMS_SECRET_KEY": self.generate_key('NOTIFIER_LMS_SECRET_KEY'),
+                "PROGRAMS_SECRET_KEY": self.generate_key('PROGRAMS_SECRET_KEY'),
+                "ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_KEY": self.generate_key('ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_KEY'),
+                "ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_SECRET": self.generate_key('ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_SECRET'),
+                "EDXAPP_ANALYTICS_API_KEY": self.generate_key('EDXAPP_ANALYTICS_API_KEY'),
+                "FORUM_API_KEY": self.generate_key('FORUM_API_KEY'),
+                "NOTIFIER_COMMENT_SERVICE_API_KEY": self.generate_key('NOTIFIER_COMMENT_SERVICE_API_KEY'),
+                "NOTIFIER_USER_SERVICE_API_KEY": self.generate_key('NOTIFIER_USER_SERVICE_API_KEY')
```
bradenmacdonald: I don't like having these hard-coded in the app either... I was going to suggest that you could write an ansible lookup plugin, so that the final YML sent to ansible would look like
```
ANALYTICS_API_SECRET_KEY: "{{ lookup('hmac_secret', 'ANALYTICS_API_SECRET_KEY', key=secret_key_base) }}"
```

But honestly that's a lot of work, and I realized there's a much simpler solution since we're actually pre-rendering the ansible vars using Django's template engine.

If you create a class like this:

```python
class SecretKeyProvider:
    def __init__(self, secret_key):
        self.secret_key = secret_key

    def __getattr__(self, var_name):
        return hmac.new(self.secret_key, msg=var_name.encode('utf-8'), digestmod=hashlib.sha256).hexdigest()
```

Then you can write your template like this:
```
ANALYTICS_API_SECRET_KEY: '{{ secret_generator.ANALYTICS_API_SECRET_KEY }}'
CREDENTIALS_SECRET_KEY: '{{ secret_generator.CREDENTIALS_SECRET_KEY }}'
...
```

and call it like this: `template.render({"secret_generator": SecretKeyProvider(b64decode(self.secret_key))})`

I tried that to confirm it should work and produce the result you want without needing to hard-code var names.

instance/models/mixins/secret_keys.py
```diff
@@ -0,0 +1,111 @@
...
+    return random_string
+
+
+# Classes #####################################################################
+
+class SecretKeyInstanceMixin(models.Model):
+    """
+    An instance that needs to generate a set of secret keys for different purposes
+    """
+    secret_key = models.CharField(max_length=48, blank=True)
```
bradenmacdonald: I think this is a tad confusing... this is not actually the secret key used for the HMAC digests; it's the base64-encoding of the secret key. So if you want to stick with a base64-encoded secret key field, I think you should rename it to `secret_key_b64` to clarify that the value of this field is not itself the secret key. Alternately, skip the base64-encoding and just store the real secret key in a `BinaryField`. Or, as a final alternative, you could just use the base 64 encoded string as the key; even though some bits would be invariant due to the encoding scheme, it should still have [more than enough entropy](http://security.stackexchange.com/questions/95972/what-are-requirements-for-hmac-secret-key)

instance/models/mixins/secret_keys.py
haikuginger: I swear, I tried this exact construction and it didn't work - got blank vars. Will try again, though.

instance/models/mixins/secret_keys.py
haikuginger: I'll rename the existing field to `secret_key_b64`.

instance/models/mixins/secret_keys.py
haikuginger: Okay, that's odd. It's working now. I wonder if I fatfingered the variable name in the template. Will have a new commit in a few minutes.

haikuginger:
@bradenmacdonald, changes have been made. I've also added some unit tests.

instance/models/mixins/secret_keys.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        self.secret_key_b64encoded = generate_secret_key(48)
+
+    def get_secret_key_settings(self):
+        """
+        Render the secret keys template as YAML and return it for use
+        in the appserver.
+        """
+        if not self.secret_key_b64encoded:
+            return ''
```
bradenmacdonald: Question (not necessarily asking for changes:) 

Should this be an error instead of silently using insecure defaults? And do we want to consider a data migration to create a secret key for existing instances? Or were you thinking it's better not to break existing instances, which is likely what changing their secret key would do?

bradenmacdonald:
Thanks @haikuginger. Looking at the new tests, I feel we also need a higher-level test that checks that these secret vars actually appear in the final, combined ansible vars. The tests that you currently have nicely confirm that the keys are generated correctly, but not that they're actually included in the playbook nor that they're not overridden by other vars/playbooks.

instance/models/mixins/secret_keys.py
haikuginger: >Or were you thinking it's better not to break existing instances, which is likely what changing their secret key would do?

This is exactly what my thinking was. We have tests to ensure that new instances get a secure base key, but we don't want to break existing instances by either failing to create new appservers or by messing with the existing (missing) keys.

bradenmacdonald: Ah, yeah, I see that now in your test instructions, which I had misunderstood. How about raising an exception if no AppServers currently exist for this instance (in other words, if this is a new instance with no secret key, raise an exception), and logging a warning otherwise?

Ideally I'd also like to see a warning in the UI for any instance without secret_key set in this manner, but I'll consider that out of scope here for now.

instance/models/mixins/secret_keys.py
```diff
@@ -0,0 +1,105 @@
...
+    return random_string
+
+
+# Classes #####################################################################
+
+class SecretKeyInstanceMixin(models.Model):
+    """
+    An instance that needs to generate a set of secret keys for different purposes
+    """
+    secret_key_b64encoded = models.CharField(max_length=48, blank=True)
```
bradenmacdonald: Nit: add `verbose_name` and `help_text` to improve the appearance of this field in the django admin?

screen shot 2016-10-27 at 1 39 53 pm

bradenmacdonald:
Looking good. Please make the additional test(s) and consider the other nits I mentioned, and then I'll approve. I tested the code and it seems to be working nicely.

haikuginger:
@bradenmacdonald, back to you once CircleCI passes.

instance/models/mixins/secret_keys.py
```diff
@@ -83,6 +98,10 @@ def get_key_provider(self):
         Get a proxy we can pass to a template that'll generate secret keys
         for any attribute requested from it.
         """
+        if not self.apserver_set.count() and not self.secret_key_b64encoded:
```
bradenmacdonald: Nit (no need to change, but for future reference): better to write `if not self.appserver_set.exists()` (more efficient) or `if count() == 0` (more clear)

instance/models/mixins/secret_keys.py
```diff
@@ -56,7 +62,16 @@ class SecretKeyInstanceMixin(models.Model):
     """
     An instance that needs to generate a set of secret keys for different purposes
     """
-    secret_key_b64encoded = models.CharField(max_length=48, blank=True)
+    secret_key_b64encoded = models.CharField(
+        max_length=48,
+        blank=True,
+        verbose_name='Instance-specific base-64-encoded secret key',
```
bradenmacdonald: Now *that* is a verbose name :-p

haikuginger:
@bradenmacdonald, code has been reorganized. Note that when we don't have a secret key, we'll still return an empty string for the secret key varset rather than empty individual variables, which would override lower-level variables.

instance/models/mixins/secret_keys.py
```diff
@@ -88,17 +88,19 @@ def get_secret_key_settings(self):
...
+        return template.render({"secret_generator": key_provider})
 
     def get_key_provider(self):
         """
         Get a proxy we can pass to a template that'll generate secret keys
         for any attribute requested from it.
         """
-        if not self.appserver_set.count() and not self.secret_key_b64encoded:
+        if not self.appserver_set.exists() and not self.secret_key_b64encoded:
             raise ValueError(
```
bradenmacdonald: There should also be a new unit test like `test_secret_key_creation` that creates an instance, updates it to have `secret_key_b64encoded = ""`, then asserts that provisioning an AppServer using `instance.spawn_appserver()` (or just `_create_owned_appserver` if that's easier to reduce the amount of mocking required) fails with this specific `ValueError`.

instance/models/mixins/secret_keys.py
```diff
@@ -90,7 +85,7 @@ def get_secret_key_settings(self):
         """
         key_provider = self.get_key_provider()
         if not self.secret_key_b64encoded:
-            logger.warning('Instance %s does not have a secret key; not prefilling variables.', self)
+            self.logger.warning('Instance %s does not have a secret key; not prefilling variables.', self)
```
bradenmacdonald: FYI, you don't need to add `%s..., self` here, because the instance's `logger` includes that info. It's not doing any harm though :)

bradenmacdonald:
Thanks @haikuginger! Good to go now from my perspective, once tests are green and branch is updated.

haikuginger:
Rebased and squashed from 2ef71a22c494afa414cbb3311e5cf40bf558f8ba, which had a green build. @bradenmacdonald, formal 👍 please.

pomegranited:
@haikuginger There was [1 Jasmine test failure](https://circleci.com/gh/open-craft/opencraft/1603?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) which didn't look related to this change.

So I've restarted the tests, hoping that maybe this can be merged and deployed before the end of the sprint?

haikuginger:
I appreciate it, @pomegranited! Looks good for approval, @bradenmacdonald.

bradenmacdonald:
@haikuginger  Sorry, I said "bradenmacdonald approved these changes" and "Good to go now from my perspective, once tests are green and branch is updated." so I thought this was formally approved. Here you go though: 👍

PR #148, opened 2016-10-27
Integration test improvements

haikuginger:
This pull request adds tests to verify that, when provisioning an instance with non-ephemeral databases, the expected MySQL and Mongo databases are correctly created. 

**Dependencies**: None

**Testing instructions**:
1. Read code.
2. Run tests.

1. We were hoping to do more than this for this particular task, but I ran into some confusing Mongo behavior that took some time to figure out, and didn't have time to research what other tests might be useful in the context of this task.

**Reviewers**
- [ ] @bradenmacdonald

### DIFF
--- a/instance/tests/integration/integration_instance.py
+++ b/instance/tests/integration/integration_instance.py
@@ -23,11 +23,14 @@ Instance - Integration Tests
 
 import os
 from unittest.mock import patch
+from urllib.parse import urlparse
 
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.management import call_command
 from django.utils.six import StringIO
+import MySQLdb as mysql
+import pymongo
 
 from instance.models.appserver import Status as AppServerStatus
 from instance.models.openedx_appserver import OpenEdXAppServer
@@ -88,6 +91,48 @@ class InstanceIntegrationTestCase(IntegrationTestCase):
         for expected_key in self.EXPECTED_SECRET_KEYS:
             self.assertIn(getattr(key_generator, expected_key), appserver.configuration_settings)
 
+    def assert_mysql_db_provisioned(self, instance):
+        """
+        Verify that the MySQL database for the instance has been provisioned and can be
+        connected to with the credentials the instance provides.
+        """
+        mysql_url_obj = urlparse(settings.DEFAULT_INSTANCE_MYSQL_URL)
+        for database in instance.mysql_databases:
+            connection = mysql.connect(
+                host=mysql_url_obj.hostname,
+                user=database['user'],
+                passwd=instance._get_mysql_pass(database['user']),
+                port=mysql_url_obj.port or 3306,
+            )
+            database_name = connection.escape_string(database['name']).decode()
+            cur = connection.cursor()
+            result = cur.execute(
+                "SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = %s",
+                (database_name,),
+            )
+            self.assertEqual(result, 1)
+
+    def assert_mongo_db_provisioned(self, instance):
+        """
+        Verify that the databases ostensibly created by the provisioning process actually were.
+        """
+        mongo_url_obj = urlparse(settings.DEFAULT_INSTANCE_MONGO_URL)
+        for db_name in instance.mongo_database_names:
+            mongo = pymongo.MongoClient(
+                host=mongo_url_obj.hostname,
+                port=mongo_url_obj.port or 27017
+            )
+            # Verify that we can log into a particular database, using the instance's saved
+            # mongo username and password
+            db = getattr(mongo, db_name)
+            # Successful authentication will return True; failure will raise an exception.
+            result = db.authenticate(
+                name=instance.mongo_user,
+                password=instance.mongo_pass,
+            )
+            self.assertTrue(result)
+            db.logout()
+
     @shard(1)
     def test_spawn_appserver(self):
         """
@@ -119,6 +164,8 @@ class InstanceIntegrationTestCase(IntegrationTestCase):
         self.assertFalse(instance.require_user_creation_success())
         for appserver in instance.appserver_set.all():
             self.assert_secret_keys(instance, appserver)
+        self.assert_mysql_db_provisioned(instance)
+        self.assert_mongo_db_provisioned(instance)
 
     @shard(3)
     def test_activity_csv(self):
--- a/registration/tests/browser/browser_login.py
+++ b/registration/tests/browser/browser_login.py
@@ -23,6 +23,7 @@ Login browser tests
 # Imports #####################################################################
 
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
+from django.core.urlresolvers import reverse
 
 from registration.tests.utils import BrowserTestMixin, UserMixin
 
@@ -48,3 +49,21 @@ class LoginBrowserTestCase(BrowserTestMixin, UserMixin,
         self.fill_form({'username': self.username})
         self.assertIn('Please enter a correct username and password',
                       self.form.text)
+
+    def test_valid_credentials(self):
+        """
+        Check that we're logged in successfully when entering correct credentials.
+        """
+        self._login(
+            username=self.username,
+            password=self.password
+        )
+        # Check that after logging in, we're directed to the registration URL (shared
+        # between registration and updating details)
+        self.assertEqual(
+            self.client.current_url,
+            '{host}{path}'.format(
+                host=self.live_server_url,
+                path=reverse('registration:register'),
+            ),
+        )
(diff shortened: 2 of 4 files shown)


instance/tests/integration/integration_instance.py
```diff
@@ -79,6 +81,41 @@ def assert_swift_container_provisioned(self, instance):
...
+        Verify that the databases ostensibly created by the provisioning process actually were.
+        """
+        mongo = pymongo.MongoClient(settings.INSTANCE_MONGO_URL)
+        db_names = mongo.database_names()
+        # This looks a little odd - basically, Mongo's `database_names` command doesn't list
+        # databases that don't have any documents in them. And, we have at least one database
+        # (forums) that doesn't have any documents inserted during the provisioning. So, as a
+        # result, if we require every db_name to be present, we'll fail, even though the db
+        # is, in actuality, there. So, we're checking that at least one of the databases that
+        # we're supposed to create shows up.
```
bradenmacdonald: Yeah, the "current database name" in mongo is treated more like an implicit label that gets attached to documents than some sort of structure that exists independently, so it doesn't really make sense to speak of a database name existing or not; from a behavioral perspective, there are only either documents annotated with a particular "database name" or not, and arbitrary new database names can be used at any time.

So I think all that we really need to do is test that the connection string/URL is valid; this seems like a reasonable way of doing that, as long as we are sure that the provisioning process always creates at least one mongo document.

Question: If we provision an instance without any demo courses, do mongo documents still get created? Is there any chance some future type of instance may start with no mongo documents at all (blank slate)?

instance/tests/integration/integration_instance.py
```diff
@@ -106,6 +143,8 @@ def test_external_databases(self):
         self.assert_instance_up(instance)
         self.assertTrue(instance.successfully_provisioned)
         self.assertFalse(instance.require_user_creation_success())
+        self.assert_mysql_db_provisioned(instance)
+        self.assert_mongo_db_provisioned(instance)
```
bradenmacdonald: Is the assumption here that if these DB credentials use e.g. a `localhost` hostname, then these assertions will fail because they'll be unable to connect?

instance/tests/integration/integration_instance.py
```diff
@@ -79,6 +81,41 @@ def assert_swift_container_provisioned(self, instance):
...
+                user=database['user'],
+                passwd=instance._get_mysql_pass(database['user']),
+                port=settings.INSTANCE_MYSQL_URL_OBJ.port or 3306,
+            )
+            database_name = connection.escape_string(database['name']).decode()
+            cur = connection.cursor()
+            result = cur.execute(
+                "SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = '{db_name}'".format(
+                    db_name=database_name
+                )
```
bradenmacdonald: Instead of using `connection.escape_string` together with `format` (any use of `format` to pass data via SQL should be suspect), can't you just do this more simple/secure version?

```
result = cur.execute("SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = %s", (db_name, ))
```

instance/tests/integration/integration_instance.py
haikuginger: > If we provision an instance without any demo courses, do mongo documents still get created? Is there any chance some future type of instance may start with no mongo documents at all (blank slate)?

I'm not sure.

instance/tests/integration/integration_instance.py
haikuginger: That is correct. We're making the connections from the integration-spawned OCIM, so we shouldn't be able to make connections to appserver-internal databases.

instance/tests/integration/integration_instance.py
bradenmacdonald: Then a better test may be to just insert a simple document `{"foo": "bar"}` into a database like "temp_test_db"; if that succeeds, everything should be working well, and that test should work even if the assumption I mentioned above changes.

haikuginger:
Rebased from ab36f08, where we had a green build. @bradenmacdonald, I think we've got what we want, so once I've got a fresh ✅ and a 👍 from you, I can squash and merge.

haikuginger:
@bradenmacdonald, I've added several tests that will need review once Circle verifies that they're working.

haikuginger:
Okay, @bradenmacdonald, once you give the go-ahead, I can rebase, squash, and merge. Thank goodness we're not likely to run out of MD5 hashes any time soon.

instance/tests/integration/integration_instance.py
```diff
@@ -81,6 +83,44 @@ def assert_swift_container_provisioned(self, instance):
...
+    def assert_mongo_db_provisioned(self, instance):
+        """
+        Verify that the databases ostensibly created by the provisioning process actually were.
+        """
+        for db_name in instance.mongo_database_names:
+            mongo = pymongo.MongoClient(
+                host=settings.INSTANCE_MONGO_URL_OBJ.hostname,
+                port=settings.INSTANCE_MONGO_URL_OBJ.port or 27017
+            )
+            # Insert a document into the database, verifying that you can.
```
bradenmacdonald: This comment doesn't seem to reflect what the code is doing... It looks to me like it's only calling `db.authenticate()` but not doing any insertion?

registration/tests/browser/browser_login.py
```diff
@@ -48,3 +48,18 @@ def test_invalid_credentials(self):
...
+
+    def test_valid_credentials(self):
+        """
+        Check that we're logged in successfully when entering correct credentials.
+        """
+        self._login(
+            username=self.username,
+            password=self.password
+        )
+        # Logging in can direct a number of different ways, so we're checking that
```
bradenmacdonald: Did you mean to use the word "direct" here? I'm not clear what you mean.

registration/tests/browser/browser_login.py
```diff
@@ -48,3 +48,18 @@ def test_invalid_credentials(self):
...
+        # we _don't_ present an error message.
```
bradenmacdonald: Logging in a newly-created test user should always redirect to a consistent result page, so I don't see why you can't test that here? IMHO it would be better to test for the expected result than to test for the absence of one particular error message.

bradenmacdonald:
@haikuginger Left some minor comments; I don't think they'll take you long to address.

instance/tests/integration/integration_instance.py
haikuginger: Ah, yup. I had done a change to try object insertion, but ran into some issues that resulted in `db.authenticate()` being a better bet. Will update the comment.

registration/tests/browser/browser_login.py
haikuginger: I meant an HTTP redirect of the client browser.

haikuginger:
@bradenmacdonald, made some changes. For whatever reason, I was thinking in context of "a brand new OpenEdX instance" rather than "a brand new OpenCraft IM instance".

bradenmacdonald:
:+1: when rebased/squashed and tests pass. Thanks @haikuginger.

- [x] I tested this: via CircleCI only
- [x] I read through the code
- [ ] ~~Includes documentation~~: n/a

haikuginger:
@bradenmacdonald, this should be passing now. However, I did remove one test, so you may want to reconfirm that things look good. Specifically, I was checking that the submit button text changed; however, we're not creating an Application object as part of the login test environment, so the text of the button displayed as it would for an initial registration.

haikuginger:
Rebased from 529e1ae.

registration/tests/test_views.py
```diff
@@ -88,10 +86,15 @@ def _assert_registration_succeeds(self, form_data):
...
-            self.client.get(verify_url)
+        with patch('registration.provision._provision_instance', autospec=True) as mock_handler:
+            expected_call_count = 0
+            for verification_email in mail.outbox:  # fix flaky pylint: disable=no-member,useless-suppression
+                verify_url = re.search(r'https?://[^\s]+',
+                                       verification_email.body).group(0)
+                self.client.get(verify_url)
+                expected_call_count += 1
+            # Check to make sure we called _provision_instance when emails were verified.
+            self.assertEqual(mock_handler.call_count, expected_call_count)
```
bradenmacdonald: @haikuginger Argh, I just noticed this is testing the wrong thing. We don't really care if `_provision_instance` is called once or twice, or whatever. What we really want to know is (1) that the emails get marked as verified, and (2) that once *all* emails have been verified (but not before), does an instance get created and is [`spawn_appserver `](https://github.com/open-craft/opencraft/blob/f3f7e20/registration/provision.py#L67) called?

We might also want to test, say, that if an email address is re-verified multiple times, that no additional app servers get spawned.

(If these things tested elsewhere, then that's fine and if so, I'm not sure what value this change adds)

haikuginger:
Rebased, squashed, and variable names fixed from b98d7ba.

PR #149, opened 2016-11-01
When deleting New Relic monitors, don't fail if monitor can't be found.

itsjeyd:
Follow-up to #143.

Makes sure that the process for shutting down an instance does not fail if the New Relic monitors associated with the instance no longer exist.

**Test instructions**

1. Temporarily set `NEWRELIC_LICENSE_KEY` and `NEWRELIC_ADMIN_USER_API_KEY` in your `.env` (check OC IM prod config for appropriate values). This is to make sure that the `DELETE` request performed by `delete_synthetics_monitor` does not fail with authentication errors.

2. Start the shell via `make shell`.

3. Create a new instance:

    ```python
    instance = instance_factory(name="Instance with fake monitors", sub_domain="instance-fake-monitors")
    ```

4. Associate a single running app server with the instance. This is to make sure that `is_shut_down` does not short-circuit later on:

    ```python
    appserver = instance._create_owned_appserver()
    appserver._status_to_waiting_for_server()
    appserver._status_to_configuring_server()
    appserver._status_to_running()
    ```

5. Associate fake New Relics monitors with it:

    ```python
    monitor_ids = ["fake-monitor-id-{}".format(i) for i in range(3)]
    for monitor_id in monitor_ids:
        instance.new_relic_availability_monitors.create(pk=monitor_id)
    # Precondition check
    assert instance.new_relic_availability_monitors.count() == 3
    assert instance.is_shut_down == False
    ```

6. Shut down instance:

    ```python
    instance.shut_down()
    ```

7. Verify that monitors have been deleted and instance manager considers instance to be shut down:

    ```python
    assert instance.new_relic_availability_monitors.count() == 0
    assert instance.is_shut_down == True
    ```

**Reviewers**

- [ ] @haikuginger

### DIFF
--- a/instance/newrelic.py
+++ b/instance/newrelic.py
@@ -87,11 +87,20 @@ def add_synthetics_email_alerts(monitor_id, emails):
 def delete_synthetics_monitor(monitor_id):
     """
     Delete the Synthetics monitor with the given id.
+
+    If the monitor can't be found (DELETE request comes back with 404),
+    treat it as if it has already been deleted; do not raise an exception in that case.
     """
     url = '{0}/monitors/{1}'.format(SYNTHETICS_API_URL, monitor_id)
     logger.info('DELETE %s', url)
-    r = requests.delete(url, headers=_request_headers())
-    r.raise_for_status()
+    try:
+        r = requests.delete(url, headers=_request_headers())
+        r.raise_for_status()
+    except requests.exceptions.HTTPError:
+        if r.status_code == requests.codes.not_found:  # pylint: disable=no-member
+            logger.info('Monitor for %s has already been deleted. Proceeding.')
+        else:
+            raise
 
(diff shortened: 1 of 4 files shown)


haikuginger:
@itsjeyd, code looks good. I'll check behavior tonight when I have a reliable Internet connection.

haikuginger:
:+1:

- [x] I tested this by following the testing instructions and didn't run into anything unexpected
- [x] I read through the code, and it seems reasonable and standard-compliant
- [x] ~~I checked for accessibility issues~~ N/A; no UI
- [x] ~~Includes documentation~~ N/A; bugfix

Rebase, and it should be set to merge.

itsjeyd:
@haikuginger Thanks, PR rebased and build is green. Merging now.

PR #151, opened 2016-11-02
Add support for multiple databases.

itsjeyd:
This PR adds support for defining multiple MySQL and MongoDB servers in the OC IM database. When creating a new instance, a single MySQL and MongoDB database server is assigned to it. If the instance is configured to use persistent databases, it will create necessary databases on its MySQL and MongoDB servers as part of the provisioning process for a new app server. If no database servers are present in the database when creating an instance, default MySQL and MongoDB servers will automatically be created using information from the `DEFAULT_INSTANCE_MYSQL_URL` and `DEFAULT_INSTANCE_MONGO_URL` environment variables.

cf. [OC-1966](https://tasks.opencraft.com/browse/OC-1966)

**Test instructions**

1. Checkout this branch and run new migrations via `make migrate`.

2. Set `DEFAULT_INSTANCE_MYSQL_URL` and `DEFAULT_INSTANCE_MONGO_URL` in your `.env` file. These settings replace `INSTANCE_MYSQL_URL` and `INSTANCE_MONGO_URL`, respectively; use values from OC IM stage `.env` for testing.

3. As a precondition check, verify that there are no `MySQLServer` or `MongoDBServer` objects in the database. You can do this by going to the Django admin (http://localhost:5000/admin/instance/mysqlserver/, http://localhost:5000/admin/instance/mongodbserver/) or via the shell:

    ```python
    assert MySQLServer.objects.count() == 0
    assert MongoDBServer.objects.count() == 0
    ```

4. Create a new instance that uses persistent databases:

    ```python
    production_instance = production_instance_factory(name='Production instance', sub_domain='production-instance-IiZei2se')
    ```

5. Verify that the database contains a single `MySQLServer` and a single `MongoDBServer`, and that the fields of these servers match the info from `DEFAULT_INSTANCE_MYSQL_URL` and `DEFAULT_INSTANCE_MONGO_URL`. You can do this by going to the Django admin or via the shell:

    ```python
    assert MySQLServer.objects.count() == 1
    assert MongoDBServer.objects.count() == 1
    
    from urllib.parse import urlparse

    mysql_url_obj = urlparse(settings.DEFAULT_INSTANCE_MYSQL_URL)
    mongo_url_obj = urlparse(settings.DEFAULT_INSTANCE_MONGO_URL)

    mysql_server = MySQLServer.objects.get()
    mongodb_server = MongoDBServer.objects.get()

    assert mysql_server.hostname == mysql_url_obj.hostname
    assert mysql_server.port == 3306
    assert mysql_server.username == mysql_url_obj.username
    assert mysql_server.password == mysql_url_obj.password

    assert mongodb_server.hostname == mongo_url_obj.hostname
    assert mongodb_server.port == 27017
    assert mongodb_server.username == mongo_url_obj.username
    assert mongodb_server.password == mongo_url_obj.password
    ```

6. Verify that the instance is configured to use the newly created database servers:

    ```python
    assert production_instance.mysql_server == mysql_server
    assert production_instance.mongodb_server == mongodb_server
    ```

7. Spawn a new app server for the instance:

    ```python
    production_instance.spawn_appserver()
    ```

    Watch the output for a few seconds; it should contain (and not error out at) the following lines:

    ```
    instance.models.instance  | instance=90 (Production inst) | Provisioning MySQL database...
    instance.models.instance  | instance=90 (Production inst) | Provisioning MongoDB databases...
    ```

    If you get Swift-related errors, just ignore them: Using Swift containers requires additional setup, and this PR does not touch Swift-related functionality/code, so it's not necessary to get Swift to work locally for the purpose of testing this PR.

8. To further test database server selection, create a few additional database servers and instances, then verify that each of the newly created instances is associated with one of the existing `MySQLServer` and `MongoDBServer` objects:

    ```python
    for i in range(3):
        MySQLServer.objects.create(
            hostname='mysql-server-{}'.format(i),
            username='mysql-user-{}'.format(i),
            password='mysql-pass-{}'.format(i),
            port=i,
        )
        MongoDBServer.objects.create(
            hostname='mongodb-server-{}'.format(i),
            username='mongodb-user-{}'.format(i),
            password='mongodb-pass-{}'.format(i),
            port=i,
        )

    for i in range(3):
        production_instance = production_instance_factory(
            name='Production instance {}'.format(i), 
            sub_domain='production-instance-{}'.format(i),
        )
        assert production_instance.mysql_server in MySQLServer.objects.all()
        assert production_instance.mongodb_server in MongoDBServer.objects.all()
    ```

**Reviewers**

- [x] @pomegranited


instance/models/mixins/openedx_database.py
```diff
@@ -228,6 +229,12 @@ def set_field_defaults(self):
         But since we add suffixes to mysql_user to generate unique user names
         for different services (e.g. xqueue) we don't want to use the maximum length here.
         """
+        if not self.use_ephemeral_databases:
+            if not self.mysql_server and MySQLServer.objects.exists():
+                self.mysql_server = random.choice(MySQLServer.objects.all())  # FIXME: Proper selection algorithm
```
smarnach: @itsjeyd You can take a look how I implemented this for the load balancers here: https://github.com/open-craft/opencraft/pull/145/files#diff-b5582c15d549b6740502b12f1b5fd04bR66

The manager class also contains logic that allows to configure a single load-balancer via the Django settings, which is needed for easy configuration of the integration tests.  I suggest you use the same logic here.  It should be possible to factor out the common parts.

instance/models/database_server.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Imports #####################################################################
+
+from django.db import models
+
+from .utils import ValidateModelMixin
+
+# Models ######################################################################
+
+class DatabaseServer(ValidateModelMixin, models.Model):  # FIXME: Does this need to subclass ValidateModelMixin?
```
antoviaque: yes, it does : )

Actually since everyone is tripping over this, it might be worth factoring it into a BaseModel object somewhere, which would then just need to be imported by all other models.

antoviaque: And TimeStampedModel, too

itsjeyd: @antoviaque Thanks, added `TimeStampedModel` as a superclass.

bradenmacdonald:
@itsjeyd Is it possible to administer these via the django admin?

instance/models/mixins/openedx_database.py
```diff
@@ -228,6 +228,13 @@ def set_field_defaults(self):
         But since we add suffixes to mysql_user to generate unique user names
         for different services (e.g. xqueue) we don't want to use the maximum length here.
         """
+        # FIXME: Remove first check to make sure any instance can switch to using persistent databases?
```
itsjeyd: @pomegranited 

> It looks like this this code already does allow us to switch from ephemeral to persistent,  by just setting `use_ephemeral_databases = False`

Yes it does, but `set_field_defaults` is currently only called if an instance has not been saved to the database, see `ValidateModelMixin` > `clean_fields` > [L94](https://github.com/open-craft/opencraft/blob/master/instance/models/utils.py#L94). This means that if a field doesn't get set via one of the `set_field_defaults` methods, we have to set it manually afterwards. And that's the reason why the code sets `self.mysql_user`, `self.mongo_user`, etc. below without checking the value of `self.use_ephemeral_databases` first -- an instance that uses ephemeral databases doesn't use/need these credentials, but we want them to be available right away in case we decide to "flip the switch" (set `use_ephemeral_databases` to `False`) on the instance.

I think it makes sense to have the same behavior for MySQL and MongoDB servers, so I'm going to remove the first check.

README.md
```diff
@@ -606,8 +616,17 @@ Databases
...
+mysql/mongo database that you want to use. When creating a new instance,
+it will automatically be assigned an existing `MySQLServer` and an existing `MongoDBServer`.
+If there are no `MySQLServer` or `MongoDBServer` objects to choose from
+when creating a new instance, the instance manager will automatically
+create a single `MySQLServer` and a single `MongoDBServer` using the values
+of the `DEFAULT_INSTANCE_MYSQL_URL` and `DEFAULT_INSTANCE_MONGO_URL` settings.
+So, if you want to skip creating a `MySQLServer` and a `MongoDBServer` manually,
+just make sure to add (or update) these settings in your `.env` file
+once you have obtained hostname and authentication information (username, password)
+for your external mysql and mongo databases.
```
itsjeyd: @pomegranited Thanks for the suggestions, I incorporated most of them into the README.

> Could you rephrase this paragraph to iterate the recommended steps to:

> * Provision a new VM to host MySQL or MongoDB - I found instructions for deploying a new MySQL server, but nothing for deploying a new MongoDB instance. Do we have instructions anywhere?

I'm not sure if we have instructions for deploying a new MongoDB instance anywhere. But I don't want to tie the instructions in the instance manager README to a specific workflow for setting up MySQL and MongoDB servers, so I didn't specify how to do this in the updated instructions.

itsjeyd:
@pomegranited I addressed your comments, added support for editing database servers via the Django admin, and finished the tests. This is now ready for manual testing. The new commits are 7b061bf386a4354d26509273d47f4bf59c47d29f through 85fd145e3cbc567dda0ccee9cfc4d240dd8764ca. (The "Update tests" commit also shows up as new, but that's because I made a few small adjustments to the changes, so no need to look at that commit again.)

Note that the integration tests are currently failing because I haven't added the new `` and `` variables to the CircleCI environment yet. I'll do that tomorrow.

README.md
```diff
@@ -603,15 +613,75 @@ output file can be specified by using the `--out` flag.
...
+2. Set up external mysql and mongo databases, making a note of hostname
+   and authentication information (username, password) for each one of them.
+
+3. In your `.env` file, set `DEFAULT_INSTANCE_MYSQL_URL` and `DEFAULT_INSTANCE_MONGO_URL`
+   to URLs that point to the MySQL and MongoDB servers created in the previous step:
+
+   ```
+   DEFAULT_INSTANCE_MYSQL_URL='mysql://<user>:<password>@<hostname>'
+   DEFAULT_INSTANCE_MONGO_URL='mongodb://<user>:<password>@<hostname>'
+   ```
```
pomegranited: To be complete, could you add `:<port>`, and note the default values below?

README.md
```diff
@@ -603,15 +613,75 @@ output file can be specified by using the `--out` flag.
...
+   You can create as many `MySQLServer` and `MongoDBServer` objects as you like.
+   If there are multiple servers of a given type to choose from, the instance manager
+   will randomly select and assign one of them when you create a new instance.
+
+When an app server for an instance that is configured to use persistent databases is provisioned,
+the instance manager will automatically add necessary settings to enable the app server to read
+and store application data on the external database servers that the instance is associated with.
+Each instance controls its own set of databases on the external database servers,
+so it is fine for multiple instances to use the same MySQL and MongoDB database servers.
+Setting up multiple database servers if the instance manager controls a large number of instances,
```
pomegranited: Nit typo: Change "Setting up" to "Set up".

README.md
```diff
@@ -603,15 +613,75 @@ output file can be specified by using the `--out` flag.
...
+   )
+   ```
+
+   You can create as many `MySQLServer` and `MongoDBServer` objects as you like.
+   If there are multiple servers of a given type to choose from, the instance manager
+   will randomly select and assign one of them when you create a new instance.
+
+When an app server for an instance that is configured to use persistent databases is provisioned,
+the instance manager will automatically add necessary settings to enable the app server to read
+and store application data on the external database servers that the instance is associated with.
```
pomegranited: This sentence is a bit convoluted..  How about this instead?

When the instance manager provisions an app server for an instance that uses persistent databases, it will automatically add the necessary settings on the associated external database server to enable the app server to read and store application data.

instance/models/database_server.py
```diff
@@ -0,0 +1,232 @@
...
+        return url
+
+    def __str__(self):
+        return self.hostname
+
+    def settings_match(self, username, password, port):
+        """
+        Return True if settings of this database server match settings passed to this method, else False.
+        """
+        return self.username == username and self.password == password and self.port == port
```
pomegranited: There's a lot of finagling due to the default port number (lines [187-188](https://github.com/open-craft/opencraft/pull/151/files#diff-9c5bd8a78df5d6dfd0289a644e03cc8eR187), [197-198](https://github.com/open-craft/opencraft/pull/151/files#diff-9c5bd8a78df5d6dfd0289a644e03cc8eR197), [220-221](https://github.com/open-craft/opencraft/pull/151/files#diff-9c5bd8a78df5d6dfd0289a644e03cc8eR220), [230-231](https://github.com/open-craft/opencraft/pull/151/files#diff-9c5bd8a78df5d6dfd0289a644e03cc8eR230)).

Could you instead have a `default_port` property that each `DatabaseServer` subclass must override?  Then, you could move `set_field_defaults()` into the base class, and update`DatabaseServer.settings_match()` to call it too, removing the need for the subclass overrides:

```
def settings_match(self, username, password, port):
    if port is None:
        port = self.default_port
    return self.username == username and self.password == password and self.port == port 
```

README.md
itsjeyd: @pomegranited Yes, that sounds better :) It's just missing the info that the app server will also be configured to be able to connect to the database server (via `*.auth.json` files). I'll add that in a separate sentence.

README.md
itsjeyd: @pomegranited Yup, consider it done :)

README.md
itsjeyd: @pomegranited Thanks for catching that. What I meant to say was,

> Setting up multiple database servers **is recommended** if the instance manager controls a large number of instances, ...

But your version is short and to the point, so I'm going to go with just changing "Setting up" to "Set up".

itsjeyd:
@pomegranited Thanks for the feedback! This is ready for another look.

itsjeyd:
Rebased from a2df0162d5b5c8d87859c195751ce823f84a9338.

itsjeyd:
Rebased from 18629505cf5af3fd3bffd66d2452bc201846e2c8 (conflicts: `.env.test`).

pomegranited:
Thanks for making those changes, @itsjeyd !  

Looks like CircleCI is failing with this error?  Don't know where it got that hostname from though:
```
_mysql_exceptions.OperationalError: (2005, "Unknown MySQL server host 'opencraft-sandbox-db.cjqzsg5yz40z.eu-central-1.rds.amazonaws.com' (0)")
```

Also, I tried changing `use_ephemeral_databases=True` for an existing, ephemeral instance, and it didn't update the database settings to use my default databases.  Was that supposed to work? If it wasn't, then it's not a blocker for this PR.

:+1:  once tests pass.

- [x] I tested this using the given instructions.  I also tested that the new models could be edited via the Admin.
- [x] I read through the code
- [x] Includes documentation

itsjeyd:
Thanks @pomegranited!

> Looks like CircleCI is failing with this error? Don't know where it got that hostname from though: ...

Turns out I set `DEFAULT_INSTANCE_MYSQL_URL` to an outdated URL when updating CircleCI settings. Figured out the correct URL now, so next test run should succeed.

> Also, I tried changing use_ephemeral_databases=True for an existing, ephemeral instance, and it didn't update the database settings to use my default databases. Was that supposed to work?

IIUC, this wasn't supposed to work -- existing instances created prior to these changes don't get a `mysq_server` and `mongodb_server` automatically (`set_field_defaults` only gets called when an instance is created). I'll do the necessary data migration manually when deploying this to stage and prod:

```python
# Update .env

# Create default database servers:
MySQLServer.objects._create_default()
MongoDBServer.objects._create_default()

# Assign default servers to instances
mysql_server = MySQLServer.objects.get()
mongodb_server = MongoDBServer.objects.get()

for instance in OpenEdXInstance.objects.all():
    instance.mysql_server = mysql_server
    instance.mongodb_server = mongodb_server
    instance.save()
```

To wrap up, can I get official (GitHub-PR-review) approval from you? :)

pomegranited:
@itsjeyd Oh yep!  Forgot that we have two 👍  to give now.

itsjeyd:
@pomegranited No worries, and thanks!

itsjeyd:
@pomegranited Got a green build! \o/ Merging now.

PR #154, opened 2016-11-10
Fixes for load balancer issues

smarnach:
This PR includes three fixes:

1. Don't include appservers that have been turned down in the LB configuration.  This has caused `None` appearing as an IP address in the configuration, resulting in haproxy refusing to reload the config.

2. Convert all domain names to lower case for the LB configuration.  I've already manually patched this up on the LB server, but we should do it in the IM as well.  Actually, all domain names should be converted to lower case when entered, and only stored in lower case, but this fix makes sure we at least never send upper-case names to the load balancer.

3. Add a new Huey task for activating an appserver.  This takes much longer than it used to, since it now has to run a playbook to reconfigure the LB server, so we can't do it as part of a normal request response anymore.

### DIFF
--- a/instance/api/openedx_appserver.py
+++ b/instance/api/openedx_appserver.py
@@ -33,7 +33,7 @@ from instance.models.openedx_appserver import OpenEdXAppServer
 from instance.models.openedx_instance import OpenEdXInstance
 from instance.serializers.appserver import AppServerBasicSerializer
 from instance.serializers.openedx_appserver import OpenEdXAppServerSerializer, SpawnAppServerSerializer
-from instance.tasks import spawn_appserver
+from instance.tasks import set_appserver_active, spawn_appserver
 
 
 # Views - API #################################################################
@@ -98,5 +98,5 @@ class OpenEdXAppServerViewSet(viewsets.ReadOnlyModelViewSet):
             return Response(
                 {"error": "Cannot make an unhealthy app server active."}, status=status.HTTP_400_BAD_REQUEST
             )
-        app_server.instance.set_appserver_active(app_server.pk)
-        return Response({'status': 'App server updated.'})
+        set_appserver_active(app_server.pk)
+        return Response({'status': 'App server activation initiated.'})
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -274,9 +274,14 @@ class OpenEdXInstance(LoadBalancedInstance, OpenEdXAppConfiguration, OpenEdXData
         """
         if not self.active_appserver:
             return self.get_preliminary_page_config(self.ref.pk)
+        ip_address = self.active_appserver.server.public_ip
+        if not ip_address:
+            # The active appserver doesn't have a public IP address.  This means that the server
+            # has been terminated, so we don't show the preliminary page in this case, and simply
+            # deconfigure the backend instead.
+            return [], []
         backend_name = "be-{}".format(self.active_appserver.server.name)
         server_name = "appserver-{}".format(self.active_appserver.pk)
-        ip_address = self.active_appserver.server.public_ip
         template = loader.get_template("instance/haproxy/openedx.conf")
         config = template.render(dict(
             domain=self.domain,
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -27,6 +27,7 @@ import logging
 
 from huey.contrib.djhuey import crontab, db_task, db_periodic_task
 
+from instance.models.openedx_appserver import OpenEdXAppServer
 from instance.models.openedx_instance import OpenEdXInstance
 from instance.utils import sufficient_time_passed
 from pr_watch import github
@@ -63,6 +64,16 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
                 # finish and replace the second as the active server. We are not really worried about that for now.
                 instance.set_appserver_active(appserver_id)
             break
+
+
+@db_task()
+def set_appserver_active(appserver_id):
+    """
+    Mark an AppServer as active.
+    """
+    logger.info('Retrieving AppServer: ID=%s', appserver_id)
+    appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
+    appserver.instance.set_appserver_active(appserver_id)
 
 
 @db_task()
--- a/instance/models/load_balancer.py
+++ b/instance/models/load_balancer.py
@@ -149,7 +149,7 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
         for instance in self.get_instances():
             map_entries, conf_entries = instance.get_load_balancer_configuration()
             backend_map.extend(
-                " ".join([domain, backend + self.fragment_name_postfix])
+                " ".join([domain.lower(), backend + self.fragment_name_postfix])
                 for domain, backend in map_entries
             )
             backend_conf.extend(
--- a/instance/tests/api/test_openedx_appserver.py
+++ b/instance/tests/api/test_openedx_appserver.py
@@ -209,7 +209,7 @@ class OpenEdXAppServerAPITestCase(APITestCase):
 
         response = self.api_client.post('/api/v1/openedx_appserver/{pk}/make_active/'.format(pk=app_server.pk))
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(response.data, {'status': 'App server updated.'})
+        self.assertEqual(response.data, {'status': 'App server activation initiated.'})
         self.assertEqual(mock_run_playbook.call_count, 1)
 
         instance.refresh_from_db()


bdero:
:+1:

- [x] I tested this: I pulled this locally and confirmed that active appserver switching is now executing in a task. I also confirmed that instances with terminated active appservers are not included in the rendered backend fragment.
- [x] I read through the code
- [ ] ~~Includes documentation~~ Not necessary

instance/models/openedx_instance.py
```diff
@@ -274,9 +274,14 @@ def get_load_balancer_configuration(self):
         """
         if not self.active_appserver:
             return self.get_preliminary_page_config(self.ref.pk)
+        ip_address = self.active_appserver.server.public_ip
+        if not ip_address:
+            # The active appserver doesn't have a public IP address.  This means that the server
+            # has been terminated, so we don't show the preliminary page in this case, and simply
+            # deconfigure the backend instead.
+            return [], []
```
antoviaque: What would the user see in the UI in that case? Is there be a way to signal this to the user? It would feel wrong to just silently deconfigure the backen, without giving any clue to the user that something was off.

instance/api/openedx_appserver.py
```diff
@@ -98,5 +98,5 @@ def make_active(self, request, pk):
             return Response(
                 {"error": "Cannot make an unhealthy app server active."}, status=status.HTTP_400_BAD_REQUEST
             )
-        app_server.instance.set_appserver_active(app_server.pk)
-        return Response({'status': 'App server updated.'})
+        set_appserver_active(app_server.pk)
+        return Response({'status': 'App server activation initiated.'})
```
antoviaque: Nice language tweak : )

Does the UI correctly refresh once the the server is actually activated? I'm not talking about the feedback from the LB that I was mentioning in the ticket (although, if you have an idea of how to approach that, I would be curious too), but just the UI <-> Django backend part. It seemed to be the case in OpenCraft.Hosting prod when I just tested now, but since I'm not sure which version is live at the moment, I'm double-checking, it seem to have the old "App server updated" copy there.

I have also noticed that some of the haproxy tasks seemed to be failing in the huey logs - I haven't dug though:

```
18:33:24 worker.1    | 11/Nov/2016 18:33:24 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |   Traceback (most recent call last):
18:33:24 worker.1    | 11/Nov/2016 18:33:24 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "<string>", line 1, in <module>
18:33:24 worker.1    | 11/Nov/2016 18:33:24 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/pip-build-_S9Fqw/cryptography/setup.py", line 334, in <module>
18:33:24 worker.1    | 11/Nov/2016 18:33:24 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       **keywords_with_side_effects(sys.argv)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/usr/lib/python2.7/distutils/core.py", line 111, in setup
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       _setup_distribution = dist = klass(attrs)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/dist.py", line 315, in __init__
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       self.fetch_build_eggs(attrs['setup_requires'])
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/dist.py", line 361, in fetch_build_eggs
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       replace_conflicting=True,
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 850, in resolve
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       dist = best[req.key] = env.best_match(req, ws, installer)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1122, in best_match
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       return self.obtain(req, installer)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1134, in obtain
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       return installer(requirement)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/dist.py", line 429, in fetch_build_egg
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       return cmd.easy_install(req)
18:33:25 worker.1    | 11/Nov/2016 18:33:25 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 665, in easy_install
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       return self.install_item(spec, dist.location, tmpdir, deps)
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 695, in install_item
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       dists = self.install_eggs(spec, download, tmpdir)
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 876, in install_eggs
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       return self.build_and_install(setup_script, setup_base)
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 1115, in build_and_install
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       self.run_setup(setup_script, setup_base, args)
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |     File "/tmp/tmp2f52a1co/venv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 1103, in run_setup
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |       raise DistutilsError("Setup script exited with %s" % (v.args[0],))
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |   distutils.errors.DistutilsError: Setup script exited with error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |   ----------------------------------------
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |   Running setup.py clean for cryptography
18:33:26 worker.1    | 11/Nov/2016 18:33:26 |     INFO | process=6584  | instance.models.load_bala | load_balancer=1 (haproxy.opencra) |   Complete output from command /tmp/tmp2f52a1co/venv/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/tmp2f52a1co/pip-build-_S9Fqw/cryptography/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" clean --all:
```

I tried to go to the instance log in the UI, but I couldn't find the logs of the ansible task run there: https://console.opencraft.com/instance/1890/edx-appserver/601/ - are we attaching these logs to their instance?

instance/tasks.py
```diff
@@ -66,6 +67,16 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
 
 
 @db_task()
+def set_appserver_active(appserver_id):
+    """
+    Mark an AppServer as active.
+    """
+    logger.info('Retrieving AppServer: ID=%s', appserver_id)
```
antoviaque: This log should be attached to the instance object? Like in https://github.com/open-craft/opencraft/blob/97a3b3a726c35b927706c2286b3b795e2869a262/instance/tasks.py#L58

instance/models/openedx_instance.py
smarnach: @antoviaque The active AppServer will be shown as "terminated" in the UI, so it shouldn't be surprising it can't be reached.  If you prefer, we could also show the "please wait" screen again in this situation, but in general if the active AppServer is terminated this means that the instance has been shut down.

antoviaque:
Also missing tests - the rule of thumb is usually 1 bug = 1 test (except maybe for the move to an asynchronous task, which I don't think you can meaningfully test). Also, it's actually one of the areas where adding the test before the code/fix can be very useful, as you can use the test to reproduce the bug first, and then fix it by making the test pass.

instance/api/openedx_appserver.py
smarnach: These fixes are already live, and the active AppServer should be updated correctly.  I've got no idea why you are still seeing "App server updated", since this string doesn't occur anywhere in the currently running code (I used grep on console.opencraft.com to double check).

The traceback you included is happening while installing the requirements, but the installation succeeds anyway.  I haven't investigated what exactly is causing the traceback, but since pip returns with a success exit code I presume that the error is immaterial.

The log is attached to the load balancer, and not to a particular instance.  Since load balancers aren't shown in the UI yet, this means the messages aren't visible in the UI for now.  It doesn't make sense to attach the messages to an instance, since the reconfiguration of the load balancer happens for all attached instances at once, so there isn't a single instance to attach it to.

instance/tasks.py
smarnach: No, this line is analogous to https://github.com/open-craft/opencraft/blob/97a3b3a726c35b927706c2286b3b795e2869a262/instance/tasks.py#L54.  Since retrieving the AppServer might fail, we log which AppServer we are trying to retrieve.  At this point, we don't have any instance or AppServer to log to.

Once the AppServer was successfully retrieved, we call `OpenEdXInstance.set_instance_active()`, and the first line inside this message logs to the instance log.  For this reason, I didn't think it's necessary to log to the instance log from here.

smarnach:
@antoviaque Since this is already deployed to production, I'm merging this PR now.  I don't have time to add the missing tests this evening (and yesterday I focused on getting the fixes in to unblock others).  If someone else merges another PR before this one gets merged, deployment will get messy, so I think we are better off addressing the remaining issues in a separate PR.

smarnach:
@antoviaque The follow-up PR to address the comments in this PR is #156.  I've inlcuded test modifications for two of the fixes as requested, and I will wait for your reply on the remaining commments.

instance/models/openedx_instance.py
antoviaque: @smarnach I meant that we should log an error here, and show that error to the user, not just silently act like if this was a normal situation. If the load balancer is instructed to point to a server which doesn't have an IP, something went wrong somewhere, and the more proactive we are in showing that error, the easier it will be to figure it out.

instance/tasks.py
antoviaque: @smarnach Ah, yes - you are right.

instance/api/openedx_appserver.py
antoviaque: @smarnach Not sure either for the "App server updated" - if I see it again I'll look into it more closely, now I know what I should expect.

> The traceback you included is happening while installing the requirements, but the installation succeeds anyway. I haven't investigated what exactly is causing the traceback, but since pip returns with a success exit code I presume that the error is immaterial.

What does "immaterial" mean? Even if it was not causing an issue (but are we sure this is the case?), I'd rather not leave errors like this in the logs - otherwise we'll just grow used to see traceback there, and start missing issues.

> The log is attached to the load balancer, and not to a particular instance. Since load balancers aren't shown in the UI yet, this means the messages aren't visible in the UI for now. It doesn't make sense to attach the messages to an instance, since the reconfiguration of the load balancer happens for all attached instances at once, so there isn't a single instance to attach it to.

The load balancer is shared by multiple instances/clients, so I'm not sure we will ever want to show it there. And even if we did, it would still make sense to tie logs for actions related to a specific instance to that instance, in addition to the load balancer. 

Could you at least log, at the instance level:
* The output of the `get_load_balancer_configuration()` method
* When this configuration has been applied by the load balancer

instance/models/openedx_instance.py
smarnach: But this _isn't_ an error.  It just means the active appserver has been terminated, which is what happens if an instance gets shut down.

instance/api/openedx_appserver.py
smarnach: @antoviaque 

> What does "immaterial" mean? Even if it was not causing an issue (but are we sure this is the case?), I'd rather not leave errors like this in the logs - otherwise we'll just grow used to see traceback there, and start missing issues.

I've seen quite a few times that installing a package via pip shows tracebacks during the installation, without the installation failing. (E.g. pylint has done this for a while, but several other packages as well.)  I've looked at these cases in the past only to learn that this is normal, and as long as the installation succeeded it has been fine.  What's more, in these cases there was nothing that could be done to prevent the traceback from being shown.  Since pip succeeded in this case, I assumed that's one of these cases again.

The error is gone now.  Yesterday, we had a different error instead that actually did cause the installation to fail.  I've looked into it and fixed it.

Regarding the load balancer logs, the idea is that load balancers will eventually become managed by the instance manager itself.  Each instance will then link to the associated load balancer, and the load balancer will have all logs attached.  That's what I had in mind when I said "Since load balancers aren't shown in the UI yet, this means the messages aren't visible in the UI for now."

Your suggestions to add more logging at the instance level make sense.  They are not easy to implement, though.  If I add logging to the `get_load_balancer_configuration()` method, this will be executed everytime the load balancer gets reconfigured.  I'll think about this tomorrow and figure something out.

instance/models/openedx_instance.py
smarnach: Well, I guess I can log a message here anyway, even if it isn't an error.

antoviaque: @smarnach I don't think it makes sense to consider an appserver as active if it is deprovisioned? Either there is a newer appserver available and it should have been made active before the active appserver got deprovisioned, or the instance itself has been shutdown, and in this case we shouldn't be calling this method at all?

instance/api/openedx_appserver.py
antoviaque: OK - I had not encountered these cases personally; pip doesn't catch failing installation scripts? It's hard to imagine the installation must have worked correctly if one of the scripts in the middle failed? Anyway, if the error is gone, we don't need to care about this now, but if we get something like that again, I'd like to understand why this happens and is still ok, before ignoring it. I would still be concerned about missing valid errors in the middle, if we start getting used to see tracebacks in the logs.

Thanks for the additional logging btw!

instance/models/openedx_instance.py
smarnach: @antoviaque You are right.  This _is_ and error.  I was assuming that this method would still be called for instances that have been turned down, but this isn't the case.  The clean-up code I added unsets the `load_balancing_server` field when an instance is turned down.  Moreover, the new `set_appserver_inactive()` method gets implicitly called when a VM is terminated, so the Appserver won't be active any more after being terminated.  This behaviour seems to make sense, and it means we should never get her.

I had a look at production, and the only instance this applies to is [this one](https://console.opencraft.com/instance/1860/edx-appserver/584/).  It is indeed in a weird state, since it is marked as running, but doesn't have an IP address.  It might be worthwhile to investigate how it got into that state.

I will change the code to explicitly log an error here.

instance/api/openedx_appserver.py
antoviaque: @smarnach I think a side effect of this would be to lose the ability to run playbooks from different version of ansible, right? Given how incompatible version of ansible can be, that might be useful to preserve - or at least, be conscious of the choice if we don't. 

Another possibility would be to only build each environment once, and only cleaning them up on restarts of the instance manager? Identifying them by venv name would allow to easily group identical environments, and have a default one, without losing the flexibility of specifying a specific venv on a command execution?

instance/models/openedx_instance.py
antoviaque: @smarnach For that instance, that might have been a manual deprovisioning from the Horizon interface - I've done this to quickly cleanup some of the unused VMs on some of the sandboxes, while waiting for https://tasks.opencraft.com/browse/OC-2041 to make it to a sprint. Since it was just sandboxes I didn't bother doing it from the console, but I missed this specific VM was actually its own instance, and it's true that it can create uncertainty in cases like this, where we might wonder if there could be an issue in our code - so I'll do it from the console next time.

PR #156, opened 2016-11-13
Some further load balancer fixes.

smarnach:
This PR addresses the open comments in #154.  See that PR for discussions.

The PR also contains a further fix:

* Install the libffi-dev library, which for some reason is needed to install Ansible version 2.1.2.0 as of today.  (Yesterday it was still possible to install it without that lib being available.)  This lib has already been installed on stage and prod.  See also the mail from Matjaz.

### DIFF
--- a/instance/models/load_balancer.py
+++ b/instance/models/load_balancer.py
@@ -137,18 +137,21 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
             if field.one_to_many and issubclass(field.related_model, LoadBalancedInstance):
                 yield from getattr(self, field.get_accessor_name()).iterator()
 
-    def get_configuration(self, triggering_id=None):
+    def get_configuration(self, triggering_instance_id=None):
         """
         Collect the backend maps and configuration fragments from all associated instances.
 
         This function also appends fragment_name_postfix to all backend names to avoid name clashes
         between multiple instance managers using the same load balancer (e.g. for the integration
         tests).
+
+        The triggering_instance_id indicates the id of the instance reference that initiated the
+        reconfiguration of the load balancer.
         """
         backend_map = []
         backend_conf = []
         for instance in self.get_instances():
-            triggered_by_instance = instance.ref.pk == triggering_id
+            triggered_by_instance = instance.ref.pk == triggering_instance_id
             map_entries, conf_entries = instance.get_load_balancer_configuration(
                 triggered_by_instance
             )
@@ -162,11 +165,14 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
             )
         return "\n".join(backend_map), "\n".join(backend_conf)
 
-    def get_ansible_vars(self, triggering_id=None):
+    def get_ansible_vars(self, triggering_instance_id=None):
         """
         Render the configuration script to be executed on the load balancer.
+
+        The triggering_instance_id indicates the id of the instance reference that initiated the
+        reconfiguration of the load balancer.
         """
-        backend_map, backend_conf = self.get_configuration(triggering_id)
+        backend_map, backend_conf = self.get_configuration(triggering_instance_id)
         fragment_name = settings.LOAD_BALANCER_FRAGMENT_NAME_PREFIX + self.fragment_name_postfix
         return (
             "FRAGMENT_NAME: {fragment_name}\n"
@@ -200,12 +206,15 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
             self.logger.error("Playbook to reconfigure load-balancing server %s failed.", self)
             raise ReconfigurationFailed
 
-    def reconfigure(self, triggering_id=None):
+    def reconfigure(self, triggering_instance_id=None):
         """
         Regenerate the configuration fragments on the load-balancing server.
+
+        The triggering_instance_id indicates the id of the instance reference that initiated the
+        reconfiguration of the load balancer.
         """
         self.logger.info("Reconfiguring load-balancing server %s", self.domain)
-        self.run_playbook(self.get_ansible_vars(triggering_id))
+        self.run_playbook(self.get_ansible_vars(triggering_instance_id))
 
     def deconfigure(self):
         """
--- a/instance/models/load_balancer.py
+++ b/instance/models/load_balancer.py
@@ -136,7 +136,7 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
             if field.one_to_many and issubclass(field.related_model, LoadBalancedInstance):
                 yield from getattr(self, field.get_accessor_name()).iterator()
 
-    def get_configuration(self):
+    def get_configuration(self, triggering_id=None):
         """
         Collect the backend maps and configuration fragments from all associated instances.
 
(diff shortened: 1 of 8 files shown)


pomegranited:
@smarnach This is looking good, but it's taking a while for my devstack to provision a valid appserver.

In the meantime, could you please:

* rebase your PR on `master`?  I had to do this to get rid of an error: `debug_toolbar.middleware.DebugToolbarMiddleware is missing from MIDDLEWARE_CLASSES.`
* Add an admin model for `LoadBalancingServer`, so that we can view and edit these servers from the GUI.  @bradenmacdonald requested a similar change for the DatabaseServer objects Tim added.

smarnach:
@pomegranited Rebased and `LoadBalancingServer` added to the admin.  (The error you got looks like you already had the new requirements from master installed, by the way.)

instance/tests/models/test_openedx_instance.py
```diff
@@ -467,6 +467,10 @@ def test_get_load_balancer_configuration(self, mocks):
             backend_map, config, domain_names, instance.active_appserver.server.public_ip
         )
 
+        # Test configuration after terminating appserver
+        with patch('instance.openstack.get_server_public_address', return_value=None):
+            self.assertEqual(instance.get_load_balancer_configuration(), ([], []))
```
antoviaque: @smarnach Could you split this into a dedicated test? It's easy to miss what is being tested when several different tests are concatenated in one.

instance/tests/models/test_load_balancer.py
```diff
@@ -47,7 +47,7 @@ def mock_instances():
     """
     instances = [
         _make_mock_instance(
-            ["test1.lb.opencraft.hosting", "test2.lb.opencraft.hosting"],
+            ["test1.lb.opencraft.hosting", "TEST2.lb.opencraft.hosting"],
```
antoviaque: @smarnach Here I see why it's useful to put the test in a shared method, since it will try it on all the tests, which is better than trying it in one. To remember that this caused a bug at one point, it could be worth adding a comment, so the test isn't removed inadvertently.

instance/tests/models/test_openedx_instance.py
smarnach: @antoviaque This is not concatenating different test cases.  The steps have to appear in this order:
1. Create appserver, check LB config
2. Activate appserver, check LB config
3. Terminate appserver, check LB config
I could turn this into three tests, the first one just performing the first step, the second test performing steps one and two, skipping the LB config check in step 1, and the third one performing all three steps.  However, this would result in code duplication and longer running times.  Do you really want me to do this?  Can you suggest a way to improve the comments to make clearer what's being tested?

smarnach: @antoviaque I'm still waiting for your answer here.  I personally wouldn't make that change, since I don't think the tests are simply concatenated, but if you prefer, I'm happy to change it.

smarnach:
The integration tests are currently failing since we currently can't start VMs on OVH.

pomegranited:
@smarnach I've restarted your tests to see if OVH is healthier now..  For reference, the [previous test run is here](https://circleci.com/gh/open-craft/opencraft/1771?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

instance/models/load_balancer.py
```diff
@@ -136,7 +137,7 @@ def get_instances(self):
             if field.one_to_many and issubclass(field.related_model, LoadBalancedInstance):
                 yield from getattr(self, field.get_accessor_name()).iterator()
 
-    def get_configuration(self):
+    def get_configuration(self, triggering_id=None):
         """
         Collect the backend maps and configuration fragments from all associated instances.
```
antoviaque: @smarnach Could you document the role of the `triggering_id` in the docstrings? If I understand correctly it is compared to the instance id to let `instance.get_load_balancer_configuration()` know when the call has been triggered by a change on that instance, right? Maybe `triggering_instance_id` would be more explicit, too.

instance/models/openedx_instance.py
```diff
@@ -279,6 +279,11 @@ def get_load_balancer_configuration(self):
             # The active appserver doesn't have a public IP address.  This means that the server
             # has been terminated, so we don't show the preliminary page in this case, and simply
             # deconfigure the backend instead.
+            if triggered_by_instance:
+                self.logger.info(
```
antoviaque: `self.logger.error()`? Cf my comment at https://github.com/open-craft/opencraft/pull/154#discussion_r89264460

instance/models/openedx_instance.py
```diff
@@ -279,6 +279,11 @@ def get_load_balancer_configuration(self):
             # The active appserver doesn't have a public IP address.  This means that the server
```
antoviaque: @smarnach Btw, doesn't the call to `ip_address = self.active_appserver.server.public_ip` mean that every change to a single instance will lead to sending a request per instance for all instances to the Nova API? That's a lot of unnecessary requests. It could be something to remember if we find out one day that our calls to the Nova API are being throttled due to the amount that we make.

instance/tests/models/test_openedx_instance.py
antoviaque: @smarnach No, you're right, it's not worth it here. Maybe describe the bug we had though, with a ticket/comment URL, so we don't end up removing inadvertently, and can easily get the context if needed?

instance/models/load_balancer.py
```diff
@@ -147,7 +148,10 @@ def get_configuration(self):
         backend_map = []
         backend_conf = []
         for instance in self.get_instances():
-            map_entries, conf_entries = instance.get_load_balancer_configuration()
+            triggered_by_instance = instance.ref.pk == triggering_id
+            map_entries, conf_entries = instance.get_load_balancer_configuration(
+                triggered_by_instance
+            )
```
antoviaque: @smarnach Would we be needing this complexity if the instances were responsible for calling the load balancer to add/update/remove their configuration?

instance/models/openedx_instance.py
smarnach: Yes, that's what's happening.  The only alternative to this approach is to cache the information in some form (e.g. by remembering the configuration snippets).

Note that reconfigurations of the load balancer aren't that frequent as you might think.  They currently happen about a dozen times per day.

instance/models/openedx_instance.py
smarnach: @antoviaque Yep, updated.  I also removed the `if triggered_by_instance` to make sure this gets always logged, since it shouldn't be happening.

antoviaque:
@smarnach Thanks for the fixes - no more objections on my side. (Just note that this isn't a :+1: as I haven't done a full review or tested, just that the outcome of the discussions are good on my side)

smarnach:
@antoviaque I already had a thumbs-up from Jill, but made some chagnes to address your comments since then, mostly related to logging.  I'll see if Ican find someone to review again.

instance/models/load_balancer.py
```diff
@@ -183,24 +193,28 @@ def run_playbook(self, ansible_vars):
...
         playbook_path = pathlib.Path(settings.SITE_ROOT) / "playbooks/load_balancer_conf/load_balancer_conf.yml"
-        returncode = ansible.capture_playbook_output(
-            requirements_path=str(playbook_path.parent / "requirements.txt"),
-            inventory_str=self.domain,
-            vars_str=ansible_vars,
-            playbook_path=str(playbook_path),
-            username=self.ssh_username,
-            logger_=self.logger,
-        )
+        with cache.lock(self.domain):
```
mtyaka: nit: maybe use a more specific lock name, such as `"{}-configure-load-balancer".format(self.domain)`.

Do we know what happens if the process crashes while holding the lock? Does the lock ever get released?

smarnach: @mtyaka Since we are using a context manager here, the process would have to die abruptly to prevent the clean-up from being run, but of course this can happen (e.g. power outage).  I've added a timeout of 15 minutes to the lock, so it will get released in any case.  (The reconfiguration typically takes less than one minute, and we might further improve this by caching the virtualenv.)

I also used the more specific name for the lock.

PR #157, opened 2016-11-14
Adds EDXAPP_COMMENTS_SERVICE_KEY

pomegranited:
Adds an additional secret key to match `FORUM_API_KEY`, which allows the LMS to communicate with the forum API.

Note that in `edx/configuration`, these two keys just happen to have the same default value, instead of referencing one another:

[edxapp/defaults/main.yml:120](https://github.com/edx/configuration/blob/47dd7dee991b9ee85f7cf6164fc1587db0f652a1/playbooks/roles/edxapp/defaults/main.yml#L120):
```yaml
EDXAPP_COMMENTS_SERVICE_KEY:  'password'
```

[forum/defaults/main.yml:L27](https://github.com/edx/configuration/blob/47dd7dee991b9ee85f7cf6164fc1587db0f652a1/playbooks/roles/forum/defaults/main.yml#L27):
```yaml
FORUM_API_KEY: "password"
```

**Testing Instructions**

1. Provision a new instance using the `master` branch.
1. Note that the Discussion tab on the Demo course returns a 500 error.
1. Provision a new instance using this branch, `jill/comments_service_key`.
1. Note that the Discussion tab on the Demo Course returns 200, and works as expected.

1. I haven't added any tests for this change.  The tests added for [#147](https://github.com/open-craft/opencraft/pull/147/files#diff-ea586dca3f707d2466b9230bc4c84f0dR52) expect the generated secret key name to match the configuration key name, which isn't the case for this `EDXAPP_COMMENTS_SERVICE_KEY`.
1. Submitted https://github.com/edx/configuration/pull/3465 to address dependency issue between `FORUM_API_KEY` and `EDXAPP_COMMENTS_SERVICE_KEY`.  Once this change is merged, then we no longer need to set both of these variables.

**Reviewer**
- [ ] @mtyaka


mtyaka:
@pomegranited Maybe we can add a unit test to  opencraft/instance/tests/models/test_openedx_instance.py, something like:

```python
@patch_services
@patch('instance.models.openedx_instance.OpenEdXAppServer.provision', return_value=True)
def test_forum_api_key(self, mocks, mock_provision):
    instance = OpenEdXInstanceFactory(sub_domain='test.forum_api_key', user_ephemeral_databases=True)
    appserver_id = instance.spawn_appserver()
    appserver = instance.appserver_set.get(pk=appserver_id)
    configuration_vars = yaml.load(appserver.configuration_settings)
    api_key = configuration_vars['EDXAPP_COMMENTS_SERVICE_KEY']
    self.assertTrue(api_key)  # Is there a better way to check variable is set and not an empty string?
    self.assertEqual(configuration_vars['FORUM_API_KEY'], api_key)
```

Other than that this looks great, I'm just waiting for the new app servers that I've spawned to successfully provision.

pomegranited:
Thanks, @mtyaka!  719c83f adds that test.

mtyaka:
Thanks @pomegranited!

:+1: (once tests pass)

- [x] I tested this: spawned a new instance & verified forums work correctly
- [x] I read through the code

PR #158, opened 2016-11-22
Added human readable description and name fields for DBs

bdero:
This PR adds a human readable name and description field to database entries. The name fields are bootstrapped with the hostname field values by the migrations.

**Testing instructions:**

1. Checkout master on a local OCIM vagrant instance.
1. Login to the admin panel and create a new MySQL and MongoDB database entry. Fill in whatever hostnames you want, it doesn't have to be real for the purpose of testing this change.
1. Navigate to the [Open edX instance "add" page](https://console.opencraft.com/admin/instance/openedxinstance/add/).
1. Check the `Mysql server` and `Mongodb server` dropdowns and confirm that the `hostname` is used as the display for both.
1. Checkout this PR's branch and run `make migrate`.
1. Navigate to the [Open edX instance "add" page](https://console.opencraft.com/admin/instance/openedxinstance/add/) again.
1. Confirm that the dropdowns still show the same content as before (the hostname).
1. Navigate to the MySQL and MongoDB DB entries in OCIM and edit them - fill out new names and descriptions.
1. Navigate to the [Open edX instance "add" page](https://console.opencraft.com/admin/instance/openedxinstance/add/) again.
1. Verify that the MySQL and MongoDB server names are listed in the format **"Name (Description)"**. Note that if the description field is left blank, then the format will just be **"Name"**.

### DIFF
--- a/instance/models/database_server.py
+++ b/instance/models/database_server.py
@@ -73,6 +73,7 @@ class DatabaseServerManager(models.Manager):
                 )
             logger.info("Creating DatabaseServer %s", hostname)
             database_server, created = self.get_or_create(  # pylint: disable=no-member
+                name=hostname,
                 hostname=hostname,
                 defaults=dict(
                     username=username,
@@ -120,6 +121,10 @@ class DatabaseServer(ValidateModelMixin, TimeStampedModel):
     class Meta:
         abstract = True
 
+    # Human readable identifier for database instances
+    name = models.CharField(max_length=250, blank=False)
+    description = models.CharField(max_length=250, blank=True)
+
     # Host name or IP address of this database server
     hostname = models.CharField(max_length=128, unique=True)
 
@@ -166,7 +171,13 @@ class DatabaseServer(ValidateModelMixin, TimeStampedModel):
         return url
 
     def __str__(self):
-        return self.hostname
+        description = ''
+        if self.description:
+            description = ' ({description})'.format(description=self.description)
+        return '{name}{description}'.format(
+            name=self.name,
+            description=description
+        )
 
     def settings_match(self, username, password, port):
         """
--- a/instance/admin.py
+++ b/instance/admin.py
@@ -59,11 +59,11 @@ class OpenEdXAppServerAdmin(admin.ModelAdmin): # pylint: disable=missing-docstri
 
 
 class MySQLServerAdmin(admin.ModelAdmin): # pylint: disable=missing-docstring
-    list_display = ('hostname', 'port', 'username', 'password')
+    list_display = ('name', 'description', 'hostname', 'port', 'username', 'password')
 
 
 class MongoDBServerAdmin(admin.ModelAdmin): # pylint: disable=missing-docstring
-    list_display = ('hostname', 'port', 'username', 'password')
+    list_display = ('name', 'description', 'hostname', 'port', 'username', 'password')
 
 
 admin.site.register(LogEntry, LogEntryAdmin)
--- a/instance/migrations/0065_create_description.py
+++ b/instance/migrations/0065_create_description.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0064_merge'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='mongodbserver',
+            name='description',
+            field=models.CharField(blank=True, max_length=250),
+        ),
+        migrations.AddField(
+            model_name='mongodbserver',
+            name='name',
+            field=models.CharField(max_length=250, blank=True),
+        ),
+        migrations.AddField(
+            model_name='mysqlserver',
+            name='description',
+            field=models.CharField(blank=True, max_length=250),
+        ),
+        migrations.AddField(
+            model_name='mysqlserver',
+            name='name',
+            field=models.CharField(max_length=250, blank=True),
+        ),
+    ]
--- a/instance/migrations/0066_populate_name.py
+++ b/instance/migrations/0066_populate_name.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def populate_name(apps, schema_editor):
+    dbs = (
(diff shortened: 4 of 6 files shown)


smarnach:
@bdero The changes look good for the admin interface.  However, the [creation of database servers from the default settings][1] fails now because no name is provided.  This is probably what is causing the unit tests to fail (I didn't look too closely).

I wonder how much value the name field adds anyway.  I'd probably never set it to anything else than the hostname.  (I'm not opposed to keeping the field in any way.)

[1]: https://github.com/open-craft/opencraft/blob/bdero/db-descriptions/instance/models/database_server.py#L75

bdero:
@smarnach Thanks! That did it, I was a bit confused on first glance at the stack traces. :)

instance/models/database_server.py
```diff
@@ -120,6 +121,10 @@ class DatabaseServer(ValidateModelMixin, TimeStampedModel):
     class Meta:
         abstract = True
 
+    # Human readable identifier for database instances
+    name = models.CharField(max_length=128, unique=True)
+    description = models.CharField(max_length=128, blank=True)
```
antoviaque: @bdero Since we already have similar names on other server models, it would be good to use the opportunity to standardize it a bit - for example, have the same lengths, non-empty requirements, etc. Maybe these fields could go on a `NamedHostMixin` or maybe `NamedDomainMixin`? See for example:
* https://github.com/open-craft/opencraft/blob/08995ce31953f5b140b2b09063617b54cd5c1341/instance/models/appserver.py#L161
* https://github.com/open-craft/opencraft/blob/08995ce31953f5b140b2b09063617b54cd5c1341/instance/models/instance.py#L59

It's less related to the work at hand, so feel free to ignore this, but while we look at the standard mixins, that could also be a good opportunity to group classes which should be inherited by all our models on a single base class - we often forget these when creating new models; ie put `ValidateModelMixin` and `TimeStampedModel` on a `BaseModel`?

instance/models/database_server.py
```diff
@@ -120,6 +121,10 @@ class DatabaseServer(ValidateModelMixin, TimeStampedModel):
...
+
     # Host name or IP address of this database server
     hostname = models.CharField(max_length=128, unique=True)
```
antoviaque: @bdero Also not your fault, but looking at the field in other places made me realize that we use `hostname` in some places, and `domain` in other places (instances, load balancer), with varying lengths, so it would likely make sense to standardize this too. Maybe that could go into the `NamedHostMixin` or `NamedDomainMixin` (depending on the one you chose - although given that instances use `domain`, it would like be a lot less trouble to go with that one).

bdero:
@antoviaque I think this could be a good idea, but those two other name fields linked don't have a unique constraint, and if we want to add default behaviors (which one of them has) then we'd need to supply overrides anyhow. I think the uniqueness in the case of the DB models is in line with the field's purpose for this name field. I'm happy to change the length to 250 if appropriate - I just chose 128 since it's the length of the hostname field and I'm bootstrapping the new name field with the value of the hostname field in the migrations.

For the general mixin refactorings & the hostname refactor, I definitely agree that it would be good to have these bread & butter mixins on a single base mixin, etc. I think for this particular piece it would be best to do this in a separate ticket + PR to avoid scope increase on this one. Maybe such a ticket could also include other refactorings we can think of to reduce redundancy. I'd be more than happy to write up and take on such an issue (it's starting to look like I might have time toward the end of the sprint). Either way I think it'd be good to have a different PR.

instance/models/database_server.py
smarnach: I think it would make sense to somehow unify the database server models and the load balancing server model, since they are almost the same.  They also use custom managers that are very similar.

smarnach:
@bdero I tested this, and all seems to be working fine now.  Can you please update the [list views in the admin interface](https://github.com/open-craft/opencraft/blob/master/instance/admin.py#L61-L66) to include the name and the description field?

bdero:
@smarnach Ah, thanks. I added the new fields in the list view.

@antoviaque What are your thoughts about my response comment? Do you think this is good to go?

smarnach:
@bdero Looks all good to me now, :+1:.

bdero:
Thanks @smarnach !

instance/models/database_server.py
bradenmacdonald: Keep in mind that we are approaching some pylint limits on how many mixin classes we have, and it can get hard to follow the code paths when there are too many mixins. So a more thoughtful refactor may be required.

antoviaque: @bradenmacdonald Do you have an approach in mind for the refactor? I agree your objections are valid ones, but I'm not sure how we could go at standardizing this differently, without keeping the fields/classes duplicated and having to keep them in sync manually? Having to remember to check the other similar classes elsewhere for consistency is also a form of complexity.

antoviaque: @bdero Replying to you here because it's easier to follow conversation in inline threads:
 
> I think this could be a good idea, but those two other name fields linked don't have a unique constraint, and if we want to add default behaviors (which one of them has) then we'd need to supply overrides anyhow. I think the uniqueness in the case of the DB models is in line with the field's purpose for this name field. I'm happy to change the length to 250 if appropriate - I just chose 128 since it's the length of the hostname field and I'm bootstrapping the new name field with the value of the hostname field in the migrations.

Yes, at the minimum changing to 250 would be good. For the uniqueness, the consistency goes both way - ie if the right choice is to have a unique name, we could add it to the other fields?

> For the general mixin refactorings & the hostname refactor, I definitely agree that it would be good to have these bread & butter mixins on a single base mixin, etc. I think for this particular piece it would be best to do this in a separate ticket + PR to avoid scope increase on this one. Maybe such a ticket could also include other refactorings we can think of to reduce redundancy. I'd be more than happy to write up and take on such an issue (it's starting to look like I might have time toward the end of the sprint). Either way I think it'd be good to have a different PR.

That would work for me - I agree that it's good to not increase the scope of this PR and risk a spillover on the task. And since this has prompted discussions, it might be good to follow-up on it on a proper ticket. I would take you on your offer to write it, thank you : )

bdero:
@antoviaque Thanks, I've made the requested changes and I will merge this PR once the tests pass again.

instance/models/database_server.py
bradenmacdonald: > Do you have an approach in mind for the refactor?

Not off the top of my head. Opening a new ticket is a good idea.

PR #159, opened 2016-11-24
Correctly set the client API keys for notifier.

smarnach:
The keys `NOTIFIER_COMMENT_SERVICE_API_KEY` and `NOTIFIER_USER_SERVICE_API_KEY` need to match the corresponding API keys on the server side, i.e. the forums and the LMS.  Otherwise notifier would stop working (and it probably doesn't work for instances that already have secret keys set).

I also refactored the template slightly to make it clearer which keys are randomly generated and which ones match some other key.

The second fix in this PR changes the order in which the settings fields are processed to make it possible to manually override secret keys in `configuration_extra_settings`.

### Open questions

For a few of the variables, I don't really understand what they should be set to; e.g. `EDXAPP_ANALYTICS_API_KEY` sounds like it should get the same value as `ANALYTICS_API_SECRET_KEY`.  However, since it doesn't seem to be used at all in the edx-platform repo, I can't tell whether this is true, and it probably doesn't matter, so I didn't touch it in this PR.

While browsing the configuration repo, I found a few additional variables that we might need to set, like `INSIGHTS_OAUTH2_SECRET`, `INSIGHT_OAUTH2_KEY` or `CREDENTIALS_PROGRAMS_API_JWT_SECRET_KEY`.  For now I focused on the services the instance manager officially supports, so I didn't look into those more closely.

### Testing

For testing, deploy a new instance in your devstack with this branch checked out and verify that everything looks fine.  You can also look at the secret key settings in the Django shell using

    print(instance.get_secret_key_settings())

These settings are currently not shown in the UI (or only as part of the merged settings).

Reviewer: @bdero

CC @haikuginger


instance/templates/instance/ansible/secret-keys.yml
```diff
@@ -14,9 +15,14 @@ NOTIFIER_LMS_SECRET_KEY: '{{ secret_generator.NOTIFIER_LMS_SECRET_KEY }}'
...
 EDXAPP_ANALYTICS_API_KEY: '{{ secret_generator.EDXAPP_ANALYTICS_API_KEY }}'
 FORUM_API_KEY: '{{ secret_generator.FORUM_API_KEY }}'
-EDXAPP_COMMENTS_SERVICE_KEY: '{{ secret_generator.FORUM_API_KEY }}'
-NOTIFIER_COMMENT_SERVICE_API_KEY: '{{ secret_generator.NOTIFIER_COMMENT_SERVICE_API_KEY }}'
-NOTIFIER_USER_SERVICE_API_KEY: '{{ secret_generator.NOTIFIER_USER_SERVICE_API_KEY }}'
+
+# Keys that need to match other keys.  The variables will be filled in by Jinja2 template
+# processing in Ansible rather than by the Django template processor, hence the verbatim block.
+{% verbatim %}
+ECOMMERCE_EDX_API_KEY: '{{ EDXAPP_EDX_API_KEY }}'
```
antoviaque: @smarnach So far we had been setting all the variables directly from the Django side, it avoid having to care about who is rendering what - why would this value from ansible differ from `secret_generator.EDXAPP_EDX_API_KEY` which is the value set above for `EDXAPP_EDX_API_KEY`? https://github.com/open-craft/opencraft/blob/e1825da5438a0cad47767c1743d4e0f506243620/instance/templates/instance/ansible/secret-keys.yml#L11

smarnach: @antoviaque In the old version, it was very hard to spot that some of the values are fundamentally different from the others:  They aren't individually generated keys, but are rather set to match other values.  The new version makes this intent much clearer than the old one by explicitly configuring them to match.  If you want two values to have the same computed value, you generally would not write

    a = f(x)
    b = f(x)

but rather

    a = f(x)
    b = a

to make clear that `a` and `b` are supposed to have the same value.  This change is made in the same spirit.

haikuginger: I'm in favor of this change.

antoviaque: @smarnach Except that it's not just "b = a", otherwise I wouldn't be raising this issue - you actually end up with a file that uses two different template processors and contexts. I don't think 

```
{% verbatim %}
ECOMMERCE_EDX_API_KEY: '{{ EDXAPP_EDX_API_KEY }}'
{% endverbatim %}
```

is that simpler than

```
ECOMMERCE_EDX_API_KEY: '{{ secret_generator.EDXAPP_EDX_API_KEY }}'
```

You have cognitive overhead in both cases - in the first to know what the verbatim representation changes to the meaning of  `'{{ EDXAPP_EDX_API_KEY }}'`, and in the second case to know that the `secret_generator` is deterministic. Since you already need to know the second one to understand the other values, and there is arguably less implicit knowledge in this approach, I would prefer to keep using the second form.

antoviaque: @smarnach Alternatively, you could also move the reference to the `secret_generator` out of the template, and simply set the values in the context - that might actually be the most readable way.

smarnach: @antoviaque I never claimed it's simpler.  I said it makes the _intent_ clearer, which is one of the most important properties of good code, and I still think this is true.

Well, be it.

smarnach: One final note on this.  By reverting this change, we are only _hiding_ the fact that we are using two different template processors.  We are not making the complexit go away.  There are already [bugs caused by this][1], and I don't think simply pretending the complexity is not there will help the situation.

[1]: https://github.com/open-craft/opencraft/blob/08995ce31953f5b140b2b09063617b54cd5c1341/instance/templates/instance/ansible/vars.yml#L195-L213

smarnach:
@haikuginger Could you please take another look?

haikuginger:
@smarnach, it looks good to me, but @bdero is your official reviewer.

smarnach:
@haikuginger Thanks, you are right.  I pinged you because you chimed in earlier.

bradenmacdonald:
@smarnach Nice refactor. I don't see any test coverage for the `OPENEDX_SHARED_KEYS` feature, and I don't know if our current integration tests would catch any problems if that stopped setting paired values identically. Could you please consider adding that?

smarnach:
@bradenmacdonald I added a new unit test to improve coverage.  The tests now verify that all keys are hex digests, and that the ones listed in `OPENEDX_SECRET_KEYS` are all different.

antoviaque:
Agreed, nice refactor! Thanks @smarnach .

bdero:
:+1:

- [x] I tested this: On a local OCIM setup, I checked out this branch and verified that appservers on *new* instances are being passed unique keys for each of the vars listed under OPENEDX_SECRET_KEYS
- [x] I read through the code

PR #160, opened 2016-11-25
Configure nginx to behave correctly behind the load balancer.

smarnach:
This PR fixes the nginx configuration to not overwrite the X-Forwarded-* headers, and adds HTTP basic authentication to direct HTTP requests to the backend.  The load balancer is configured to authenticate to the backends, and the appserver link in OCIM is updated to include username and password.

Furthermore, the `protocol` field is removed, since all instances use HTTPS now.

### Testing instructions

I recommend testing this on stage, which will make it easier to successfully deploy an instance.  Note that this PR involves a migration which can't be reverted.  (Well, it could be reverted, since the field that is removed contains the same value for all instances.)

New instances should
* require HTTP authentication when trying to contact the appserver directly by IP address.
* allow HTTP access to an appserver by IP address using the username and password given in the "Settings" tab of the instance.
* allow HTTP access to the appserver using the "Authenticated Link" in the "Status" section of an appserver.
* work normally when activated and contacted via the domain name, and not require authentication.
* use HTTPS when addressed by domain name, and browsers should show a green lock symbol in the address bar (this is at least what Chrome and Firefox should do; other browser might indicate secure content in a different way).

All links to the instance in OCIM should be HTTPS links, while the direct link to the backend should be an HTTP link.


mtyaka:
This needs a rebase. The change makes total sense, I'm just waiting for a new instance to finish provisioning to verify the it.

smarnach:
@bdero This is for your review.  The comment from Matjaz was about an almost totatlly unrelated version of this PR.

instance/templates/instance/ansible/vars.yml
```diff
@@ -26,19 +31,11 @@ CMS_HOSTNAME: '{{ instance.studio_domain_nginx_regex }}'
...
 
 # Nginx
-{% if instance.protocol == 'https' %}
-NGINX_ENABLE_SSL: true
-NGINX_REDIRECT_TO_HTTPS: true
-{% else %}
+NGINX_SET_X_FORWARDED_HEADERS: false
+
+# SSL is handled on the load balancer, and the appservers are HTTP only.
 NGINX_ENABLE_SSL: false
```
bradenmacdonald: Is there any reason not to set this to true so we can optionally connect to the VM using HTTPS and test it? (Which would use the snakeoil cert, but an SSL warning is fine for internal testing). Use case is to debug issues or verify fixes that could be related to HTTPS (e.g. cross-domain asset loading) without activating the VM.

instance/tests/integration/integration_instance.py
```diff
@@ -78,7 +78,7 @@ def assert_appserver_firewalled(self, instance):
         instance.refresh_from_db()
         self.assertIsNotNone(instance.active_appserver)
         server_ip = instance.active_appserver.server.public_ip
-        ports_should_be_open = [22, 80]  # 443 may or may not be open, depending on instance.protocol
+        ports_should_be_open = [22, 80]  # 443 may or may not be open
```
bradenmacdonald: Per previous comment, maybe this should always be open? If you don't make that change then this comment is wrong - it's always closed.

bdero:
@smarnach I placed this change on stage and made a new instance + appserver.

- HTTP auth works when visiting the instance directly.
  - Host: http://149.202.165.243/
  - Credentials: `12f48c18d4ed12cf:7006c1d928cb9358`
- The base64 Authorization header in the backend matches: `http-request set-header Authorization 'Basic MTJmNDhjMThkNGVkMTJjZjo3MDA2YzFkOTI4Y2I5MzU4'`

However, when I try to visit [the instance URL](https://ocim-160.stage.opencraft.hosting/), chrome returns an ERR_SSL_UNRECOGNIZED_NAME_ALERT, which I suspect means haproxy-stage can't find a matching backend mapping for the SNI header.

At a glance the mapping looks fine and there's nothing in the syslog that looks incriminating. Can you take a look @smarnach ?

instance/templates/instance/ansible/vars.yml
smarnach: @bradenmacdonald At some point in the past, setting this to `true` without providing a certificate resulted in a deployment failure, since the snakeoil cert wasn't available at the path given by default.  I'm not sure whether this has been fixed in the meantime.  I'll test it on stage.

smarnach:
@bdero Thanks for testing this.  It turned out the weird `NXDOMAIN` error was back for haproxy-stage (but none of the other instances).  Jill removed the magic to determine the IP address in https://github.com/open-craft/ansible-load-balancer/pull/10, which I now deployed to haproxy-stage, and the site is accessible now.

However, I noticed that you won't be able to fully test that browsers consider the site secure now, since testing uses staging certificates.  Two possible work-arounds:
* Look at the details why the browser does not consider the site secure.  The only reason should be the invalid certificate, everything else should be fine.
* Trust the Let's Encrypt staging root certificate *before* accessing the test site for the first time.  In Chrome, this can be done in Settings -> Show advanced settings -> HTTP/SSL -> Manage certificates… -> Authorities -> Import…, then select `ansible-load-balancer/files/lets-encrypt-fake-cert.crt` and trust the certificate for idenitifying websites.  In other browsers, this can be done as well, but you will have to figure it out yourself.

instance/templates/instance/ansible/vars.yml
smarnach: @bradenmacdonald I tried this on a Ficus test instance, and apparently this still does not work, see https://console.opencraft.com/instance/2520/edx-appserver/947/.  We would have to either fix configuration to provide snakeoil certificates if no certificate is provided, or generate our own snakeoil certs and pass them in (the former probably makes more sense).  Can we exclude that from the current scope?  Do you want me to create a separate ticket for this?

bdero:
Thanks @smarnach. I checked it out again and it looks good. :+1:

- [x] I tested this: I tested that HTTP auth works when visiting the instance directly, checked that the authorization header in the backend is correct, and that requests to the instance URLs pass all other Chrome security checks when trusting the Letsencrypt certificate.
- [x] I read through the code
- [x] ~I checked for accessibility issues~ Not necessary.
- [x] ~Includes documentation~ Not necessary.

PR #161, opened 2016-11-26
Use multiple gunicorn workers in production

bdero:
This PR configures gunicorn to spawn 4 workers in production mode.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: gunicorn opencraft.wsgi --log-file -
+web: gunicorn opencraft.wsgi --timeout 60 --workers 4 --log-file -
 websocket: python3 websocket.py
 worker: python3 manage.py run_huey --no-periodic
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -1,4 +1,4 @@
-web: gunicorn opencraft.wsgi --log-file -
+web: gunicorn opencraft.wsgi --timeout 60 -w 4 --log-file -
```
smarnach: Please use the long form of the flag here.  This way, it's immediately obvious what all the options do, and it's also consistent with the other flags in this file.

smarnach:
:+1: after addressing the comment.

- [x] I tested this: I ran this locally, and verified that gunicorn spawns four workers and the web app still works fine.
- [x] I read through the code

bdero:
Thanks @smarnach! I just fixed up this PR and I'm waiting for tests to pass again.

PR #162, opened 2016-11-30
Fix bug occurring when shutting down instance with no active appserver.

smarnach:
This bug was encountered by @bdero when shutting down unused beta test instances.

The bug occurs when an instance without any active appserver is shut down.  This case triggered a special code path, which was not covered by the tests.  I added a test to cover it.

I also removed the `if` clause that triggered that special code path, since it's not needed anymore in the current version.  (It was needed at some point.)  I leave the test in anyway.

For testing, create an instance, spawn an appserver, and shut the instance down without activating the appserver.

### DIFF
--- a/instance/models/mixins/load_balanced.py
+++ b/instance/models/mixins/load_balanced.py
@@ -93,7 +93,7 @@ class LoadBalancedInstance(models.Model):
             if load_balancing_server is None:
                 return
         self.logger.info("Triggering reconfiguration of the load balancing server...")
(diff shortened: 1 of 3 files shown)


bdero:
@smarnach Thanks! :+1:

- [x] I tested this: I tested this locally with an instance containing only non-active appservers - then I ran the code against the instances on OCIM prod to make sure they're fully deprovisioned.
- [x] I read through the code

smarnach:
@bdero Thanks!

PR #163, opened 2016-12-13
Use the Gandi API to determine the registered domain.

smarnach:
This PR removes the dependency on the `tldextract` library and uses information from the Gandi API instead to determine the registered domain name.  This is necessary since `tldextract` depends on the Public Suffix List, so it will return incorrect results for `opencraft.hosting`, which is now on the Public Suffix List.

The new version requests the list of all domains registered with the Gandi account when first needed, and builds a dictionary of all zone IDs.  For setting a DNS record, we now pass in the fully qualified domain name and search the dictionary for a matching suffix.  If none is found, we raise an exception, since this means the domain name can't be set using the configured Gandi account.

This PR also introduces a fake implementation of the Gandi API which behaves similar to the actual Gandi API (at least similar enough).  It provides stricter tests that can concentrate more on desired outcomes at the same time.  (I actually didn't expect this to become as complex as it did.)

This is tested by the integration tests.  For manual testing, you can try this on stage, which uses the domain `stage.opencraft.hosting`, so it's a good test candidate.  You only need to _start_ provisioning an instance, since DNS records are set early in the process.

### DIFF
--- a/instance/gandi.py
+++ b/instance/gandi.py
@@ -42,10 +42,13 @@ class GandiAPI():
     Gandi API proxy object
     """
 
-    def __init__(self, api_url='https://rpc.gandi.net/xmlrpc/'):
+    def __init__(self, api_url='https://rpc.gandi.net/xmlrpc/', client=None):
         # This is a map of domain_name => zone_id key-value pairs.
-        self._zone_id_cache = {}
-        self.client = xmlrpc.client.ServerProxy(api_url)
+        self._zone_id_cache = None
+        if client:
+            self.client = client
+        else:
+            self.client = xmlrpc.client.ServerProxy(api_url)
 
     @property
     def api_key(self):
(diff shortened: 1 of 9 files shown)


instance/tests/test_gandi.py
```diff
@@ -37,69 +38,88 @@ class GandiTestCase(TestCase):
...
+            self.api.client.domain.mock_calls,
             [create_new_zone_version_call] * attempts +
             [delete_old_record_call, create_new_record_call, set_new_zone_version_call]
         )
 
+    def test_populate_zone_id_cache(self):
+        """
+        Test that populating the zone ID cache results in the expected API calls.
+        """
+        self.api._populate_zone_id_cache()
```
pomegranited: Your test coverage is awesome 😄   Only [gandi.py#L81](https://github.com/open-craft/opencraft/blob/a19ed4fecd2733c900407d5e5995385a61bcf296/instance/gandi.py#L81) is missing from the unit tests.

You could bump `gandi.py`'s coverage from 96% to 98% by simply changing this line to:

    self.api.get_zone_id('test.com')

instance/gandi.py
```diff
@@ -61,15 +64,38 @@ def client_zone(self):
...
+            self._populate_zone_id_cache()
+        labels = domain.lower().split('.')
+        for split_index in range(len(labels) - 1):
+            registered_domain = '.'.join(labels[split_index:])
+            if registered_domain in self._zone_id_cache:
+                subdomain = '.'.join(labels[:split_index]) or '@'
+                return subdomain, registered_domain
+        raise ValueError(
+            'The given domain name does not match any domain registered in the Gandi account.'
+        )
```
pomegranited: Can this `ValueError` be caught and displayed to the GUI?

instance/gandi.py
```diff
@@ -42,10 +42,13 @@ class GandiAPI():
     Gandi API proxy object
     """
 
-    def __init__(self, api_url='https://rpc.gandi.net/xmlrpc/'):
+    def __init__(self, api_url='https://rpc.gandi.net/xmlrpc/', client=None):
         # This is a map of domain_name => zone_id key-value pairs.
-        self._zone_id_cache = {}
-        self.client = xmlrpc.client.ServerProxy(api_url)
+        self._zone_id_cache = None
```
smarnach: @pomegranited I don't think this is a problem.  Every configuration change requires a restart of the IM.  Gandi zone IDs can't change, so only adding and removing domains is relevant, and this isn't going to happen often.

instance/gandi.py
smarnach: It should show up in the instance log already.  (Not in the appserver log, since the DNS record is set before a new appserver is even created, so there is no appserver to log it to.)

instance/tests/test_gandi.py
smarnach: Good suggestion, I've changed this and added a comment that `_populate_zone_id_cache()` is implicitly called.

In practice, the line should never be executed, since we will always call `split_domain_name()` before `get_zone_id()`, and the former call will already populate the cache.  This is the reason the line was not covered.

PR #165, opened 2016-12-24
Treat Errored appservers the same as ones with ConfigurationFailed when determining if an instance is shut down

bdero:
This fixes a small bug in OCIM where instances with only terminated appservers will still show up in the OCIM instance menu if there are any "Errored" appservers associated with the instance.

A simple way to test this is to make it live on OCIM and verify that the [Edraak](https://console.opencraft.com/instance/670/) instance stops showing up on the left menu.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -227,7 +227,10 @@ class OpenEdXInstance(LoadBalancedInstance, OpenEdXAppConfiguration, OpenEdXData
             """
             if appserver.status == AppServerStatus.Terminated:
                 return True
-            configuration_failed = appserver.status == AppServerStatus.ConfigurationFailed
+            configuration_failed = (
+                appserver.status == AppServerStatus.ConfigurationFailed or
+                appserver.status == AppServerStatus.Error
+            )
             vm_terminated = appserver.server.status == ServerStatus.Terminated
             return configuration_failed and vm_terminated
 
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -571,6 +571,14 @@ class OpenEdXInstanceTestCase(TestCase):
         appserver._status_to_configuration_failed()
 
     @staticmethod
+    def _set_appserver_errored(appserver):
+        """
+        Transition `appserver` to AppServerStatus.Error.
+        """
+        appserver._status_to_waiting_for_server()
+        appserver._status_to_error()
+
+    @staticmethod
     def _set_server_ready(server):
         """
         Transition `server` to Status.Ready.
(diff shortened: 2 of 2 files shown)


instance/models/openedx_instance.py
```diff
@@ -227,7 +227,10 @@ def appserver_is_shut_down(appserver):
             """
             if appserver.status == AppServerStatus.Terminated:
                 return True
-            configuration_failed = appserver.status == AppServerStatus.ConfigurationFailed
+            configuration_failed = (
+                appserver.status == AppServerStatus.ConfigurationFailed or
+                appserver.status == AppServerStatus.Error
+            )
             vm_terminated = appserver.server.status == ServerStatus.Terminated
             return configuration_failed and vm_terminated
```
pomegranited: I'm confused.. Why isn't it enough for all `appserver.server.status == ServerStatus.Terminated`?  If the server itself has been terminated, then does it matter what the `appserver.status` was/is?

bdero: > If the server itself has been terminated, then does it matter what the appserver.status was/is?
@pomegranited Hmm, I suspect it doesn't, but I'm not 100% sure. We still need to check if the config is failed/errored for running VMs, but maybe it should be written to always return `True` if either the appserver or VM statuses are set to "Terminated", and do the more specific checks on the AppServerStatus otherwise. I feel like I'm more confused about the difference between `AppServerStatus.Terminated` and `ServerStatus.Terminated` than anything here.

@itsjeyd since you wrote this function originally, could you chime in?

itsjeyd: @bdero @pomegranited You're right, to determine whether a single `appserver_is_shut_down`, we could just check whether the corresponding VM has been terminated, and stop taking the status of the app server itself into account. I think the reason we're not doing that is that we (initially) wanted to make sure that the UI would only hide an instance if we can be sure it has been "properly" shutdown. If an instance had one or more app servers that got stuck, e.g., in `ConfiguringServer`, we wanted to be aware of it, to be able to investigate.

> I feel like I'm more confused about the difference between `AppServerStatus.Terminated` and `ServerStatus.Terminated` than anything here.

* `AppServerStatus.Terminated` indicates that an app server was running successfully at some point (i.e., it never entered `AppServerStatus.ConfigurationFailed` or `AppServerStatus.Error`) and has since been shut down, cf. https://github.com/open-craft/opencraft/blob/master/instance/models/appserver.py#L111.
* `ServerStatus.Terminated` indicates that a VM has been terminated/shut down, cf. https://github.com/open-craft/opencraft/blob/master/instance/models/server.py#L98. It might have been in any of the available [server states](https://github.com/open-craft/opencraft/blob/master/instance/models/server.py#L74) before that.

bdero: @itsjeyd Ah, that clears things up - in that case, this PR does seem to make the logic change we want. Can you take another look @pomegranited ?

bdero:
@pomegranited Ah, I looked at this too hastily - I'll fix the remaining issues and let you know when this is ready again, thanks.

bdero:
@pomegranited Alright, the modified test passed locally, so this should be ready for review again in the absence of test flakiness. :)

PR #166, opened 2016-12-30
Deprovision NewRelic monitoring on instance.shut_down, and improve logging

pomegranited:
Adds `instance.auto_deprovision_monitoring`, and uses it to decide whether to disable New Relic monitoring when terminating an active appserver.

Also improves logging to help diagnose major server events:
* Adds the server ID to all server log messages
* Logs all state transitions
* Logs automated task events `terminate_obsolete_appservers_all_instances` and `shut_down_obsolete_pr_sandboxes`.

**Testing instructions**:

1. Run migrations for this branch in your devstack.
1. Create a new instance.  Ensure that the instance's `auto_deprovision_monitoring` flag is set to True by default.
1. Provision a new appserver for that instance.  
1. Ensure that one of the following messages is logged:
    * "Skipping monitoring setup," if you don't have a valid `NEWRELIC_ADMIN_USER_API_KEY` key configured.
    * "Setting up New Relic Synthetics monitors," otherwise.
1. Activate the new appserver, and then then terminate it.  Ensure that the following message is logged on termination: "Removing New Relic Synthetics monitors".
1. Unset the `auto_deprovision_monitoring` flag, and provision a new appserver for that instance.
1. Activate the new appserver, and then terminate it.  Ensure that you do _not_ see the following message logged during termination: "Removing New Relic Synthetics monitors".  Note that if you have a valid `NEWRELIC_ADMIN_USER_API_KEY` key set, your configured monitors will be failing, and sending messages to ops@opencraft.com.

1. Log message changes are tested in unit tests.
1. We wanted to also log the currently logged-in user for important events, like terminating an appserver.  However, currently, we can only terminate an appserver from the shell or via a Huey task, when there's no username available to log.

**Reviewers**
- [ ] @bradenmacdonald

### DIFF
--- a/instance/tests/api/test_openedx_appserver.py
+++ b/instance/tests/api/test_openedx_appserver.py
@@ -236,26 +236,31 @@ class OpenEdXAppServerAPITestCase(APITestCase):
             },
             {
                 'level': 'INFO',
-                'text': 'instance.models.server    | server={server_name} | info',
+                'text': 'instance.models.server    | server={server_id} ({server_name}) | info',
             },
             {
                 'level': 'ERROR',
-                'text': 'instance.models.server    | server={server_name} | error',
+                'text': 'instance.models.server    | server={server_id} ({server_name}) | error',
             },
             {
                 'level': 'INFO',
-                'text': 'instance.models.server    | server={server_name} |'
+                'text': 'instance.models.server    | server={server_id} ({server_name}) |'
                         ' Starting server (status=Pending [pending])...'
             },
             {
                 'level': 'ERROR',
-                'text': 'instance.models.server    | server={server_name} |'
+                'text': 'instance.models.server    | server={server_id} ({server_name}) |'
                         ' Failed to start server: Not found (HTTP 404)'
+            },
+            {
+                'level': 'INFO',
+                'text': 'instance.models.server    | server={server_id} ({server_name}) |'
+                        ' Transition from "Pending" to "Failed"'
             },
         ]
         self.check_log_list(
             expected_list, response.data['log_entries'],
-            inst_id=instance.ref.id, as_id=app_server.pk, server_name=server.name,
+            inst_id=instance.ref.id, as_id=app_server.pk, server_id=server.pk, server_name=server.name,
         )
 
     def check_log_list(self, expected_list, log_list, **kwargs):
@@ -297,15 +302,17 @@ class OpenEdXAppServerAPITestCase(APITestCase):
             },
             {
                 'level': 'ERROR',
-                'text': 'instance.models.server    | server={server_name} | error',
-            },
-            {
-                'level': 'ERROR',
-                'text': 'instance.models.server    | server={server_name} |'
+                'text': 'instance.models.server    | server={server_id} ({server_name}) |'
                         ' Failed to start server: Not found (HTTP 404)'
             },
+            {
+                'level': 'INFO',
+                'text': 'instance.models.server    | server={server_id} ({server_name}) |'
+                        ' Transition from "Pending" to "Failed"'
+            },
         ]
+
         self.check_log_list(
             expected_list, response.data['log_error_entries'],
-            inst_id=instance.ref.id, as_id=app_server.pk, server_name=server.name,
+            inst_id=instance.ref.id, as_id=app_server.pk, server_id=server.pk, server_name=server.name,
         )
--- a/instance/tests/api/test_openedx_appserver.py
+++ b/instance/tests/api/test_openedx_appserver.py
@@ -26,11 +26,12 @@ from unittest.mock import patch
 
 import ddt
 from rest_framework import status
+from django.conf import settings
 
 from instance.tests.api.base import APITestCase
 from instance.tests.models.factories.openedx_appserver import make_test_appserver
 from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
-from instance.tests.utils import patch_gandi
+from instance.tests.utils import patch_gandi, patch_url
 
 
 # Tests #######################################################################
@@ -111,6 +112,7 @@ class OpenEdXAppServerAPITestCase(APITestCase):
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
         self.assertEqual(response.data, {'detail': message})
 
+    @patch_url(settings.OPENSTACK_AUTH_URL)
     def test_get_details(self):
         """
         GET - Detailed attributes - instance manager allowed access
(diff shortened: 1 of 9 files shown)


antoviaque:
@pomegranited @bradenmacdonald Instead of adding custom logic for this specific case, maybe we are currently removing the monitoring at the wrong place? What if we only did so when an instance is explicitly `shut_down()` https://github.com/open-craft/opencraft/blob/b5fbf05239e1c48357481a7fa30e7e2c20fe7dfc/instance/models/openedx_instance.py#L419-L430 rather than when appservers are activated/deactivated. https://github.com/open-craft/opencraft/blob/43961f5c6c52babd3d6c5b94558701c006db817a/instance/models/openedx_instance.py#L312-L330 - we monitor instances rather than appservers, after all.

Btw, wouldn't the current code try to reset the monitoring every time we activate an appserver? That would reset the configuration done in New Relic to get the alerts to the paging system.

bradenmacdonald:
@antoviaque Hmm. Yep, I agree that it makes more sense to move the monitoring removal to the `shut_down()` method. The creation of the monitor should remain where it is, since we don't want to enable monitoring before an instance is online.

> Btw, wouldn't the current code try to reset the monitoring every time we activate an appserver?

No. Normally `set_appserver_inactive` is not supposed to be called except when an instance is shut down, or in very unusual circumstances. So the place where the monitoring removal is now does make some sense. Activating a new AppServer does not call `set_appserver_inactive`, and would not reset the monitoring.

pomegranited:
I was unsure from your comment whether to move the `disable_monitoring()` call to `instance.shut_down()` as @antoviaque suggested, or whether I should keep the new `instance.auto_deprovision_monitoring` flag, as @bradenmacdonald  suggested in the JIRA issue.

So I've kept the flag for now, but let me know which way you'd prefer?

bradenmacdonald:
I think we should forget the flag and just move the disable_monitoring() call to instance.shut_down() as @antoviaque suggested.

instance/tests/api/test_openedx_appserver.py
```diff
@@ -145,7 +145,7 @@ def test_get_details(self):
         self.assertIn(('public_ip', None), server_data)
         # The API call will try to start the server, which will fail, since we're
         # not actually talking to an OpenStack instance when unit tests are running
-        self.assertIn(('status', 'failed'), server_data)
+        self.assertIn(('status', 'unknown'), server_data)
```
bradenmacdonald: I'm not clear on what caused this to change, or why the tests below no longer get "Failed to start server: Not found (HTTP 404)" when attempting to start the server?

bradenmacdonald: Actually, it looks like these tests were failing on `master` and this is a drive-by fix. Is that right? If so, I think we need to actually determine the cause instead of just changing the test. I think that the test code was correct as written, and that `failed` is more useful than `unknown`. Same down below for the 404 error.

pomegranited: @bradenmacdonald  Ah yes, thanks for catching that.  I couldn't figure out what had changed, and so thought it was something with OpenStack, or nova client, or elsewhere.  My apologies!

I pinned it down to this exception:
```
  File "/home/vagrant/opencraft/instance/models/server.py", line 330, in update_status
    os_server = self.os_server
  File "/home/vagrant/opencraft/instance/models/server.py", line 285, in os_server
    self.start()
  File "/home/vagrant/opencraft/instance/models/utils.py", line 295, in wrapped_method
    return method(resource, *args, **kwargs)
  File "/home/vagrant/opencraft/instance/models/server.py", line 372, in start
    key_name=settings.OPENSTACK_SANDBOX_SSH_KEYNAME,
  File "/home/vagrant/opencraft/instance/openstack.py", line 74, in create_server
    flavor = nova.flavors.find(**flavor_selector)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/base.py", line 407, in find
    matches = self.findall(**kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/base.py", line 463, in findall
    listing = self.list(**list_kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/v2/flavors.py", line 132, in list
    return self._list("/flavors%s%s" % (detail, query_string), "flavors")
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/base.py", line 249, in _list
    resp, body = self.api.client.get(url)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 480, in get
    return self._cs_request(url, 'GET', **kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 436, in _cs_request
    self.authenticate()
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 612, in authenticate
    auth_url = self._v1_auth(auth_url)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 648, in _v1_auth
    resp, body = self._time_request(url, 'GET', headers=headers)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 431, in _time_request
    resp, body = self.request(url, method, **kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/novaclient/client.py", line 396, in request
    **kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/requests/sessions.py", line 596, in send
    r = adapter.send(request, **kwargs)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/requests/adapters.py", line 482, in send
    raise RetryError(e, request=request)
requests.exceptions.RetryError: HTTPConnectionPool(host='example.com', port=80): Max retries exceeded with url: /auth/url (Caused by ResponseError('too many 500 error responses',))
```

In `.env.test`, `OPENSTACK_AUTH_URL='http://example.com/auth/url'`.  The failing tests are hitting that URL, and balking at the 500 error it returns.  I don't know if `example.com` used to placidly return status code 200, but now it returns 500, so we have to mock it.

e079ae2 reverts my naughty test data changes, and 67cb886 mocks the `OPENSTACK_AUTH_URL` response.  The tests are succeeding on my local devstack; hopefully CircleCI will agree.

Is it ok to include this fix in this PR, or do I need to split it into a separate one?

bradenmacdonald: @pomegranited Thanks for figuring that out and fixing it - we definitely should not have been making HTTP requests to example.com and depending on the return code :-/ Keeping it in this PR is totally fine.

bradenmacdonald:
:+1:

- [x] I tested this: Tested on my OC IM devstack, though without a valid New Relic key
- [x] I read through the code
- [x] Includes documentation: yep, in the form of docstrings + more verbose logging

PR #167, opened 2017-01-11
OC-2178 Automatically assign edxapp appserver VMs to a network security group

bradenmacdonald:
With this PR, every AppServer deployed by OpenCraft IM will automatically be added to an OpenStack "security group" in order to provide an additional layer of protection against unwanted exposure of services that may be running on the AppServer.

The security group is created automatically by OpenCraft IM, and OpenCraft IM will define the rules of the security group. (Specifically, it will ensure the the security group always matches the following configuration: allow all egress, and allow ingress only on ports 22, 80, and 443.)

No new configuration is required, since everything is managed automatically.

This PR also introduces the ability for specific instances to be given additional security groups, e.g. if one instance needs to connect to some secure/dedicated database server, then that can be done more securely using security groups to grant access only to that instance's appservers.

Testing instructions:
* Run an OC IM devstack, and ensure it is configured with a valid OpenStack account (e.g. our OVH dev account) - an example .env file is in our internal private documentation repo if you need it.
* Optional: In your `.env` file, set `EDXAPP_APPSERVER_SECURITY_GROUP_NAME='edxapp-appserver-YOUR_NAME_HERE'` so that you can tell which rules were created for you specifically. Then restart OC IM.
* Go to https://horizon.cloud.ovh.net/project/access_and_security/ (dev account) and confirm that no security group with the name set in EDXAPP_APPSERVER_SECURITY_GROUP_NAME (e.g. "edxapp-appserver" or "edxapp-appserver-yourname") exists yet.
* Provision a new AppServer (it doesn't have to finish; it just has to get to the stage where ansible is running)
* Go to https://horizon.cloud.ovh.net/project/instances/ and confirm that the new appserver has been added to the security group, and is not a member of the "default" security group.
  screen shot 2017-01-11 at 7 21 04 pm
* Go to https://horizon.cloud.ovh.net/project/access_and_security/ and verify that the new security group has the correct rules

Notes:
* I fixed the bug that prevented editing sandbox instances from the django admin
* The `nova` python API for interacting with security groups is deprecated. So I added the `python-openstacksdk` package and used that. We should plan to Remove the `python-novaclient` and `python-swiftclient` in the future, and just use `python-openstacksdk`, which can do everything, is newer, nicer, and has very few dependencies.
* I rename "openstack.py" to "openstack_utils" so it doesn't conflict with the new "openstack" import in people's minds
* I removed `delete_servers_by_name` as it didn't seem to be used.

Things I noticed while working on this PR (to do in other tickets):
* We should remove “protocol” from OpenCraft IM OpenEdXAppConfiguration
* We should change the popup help text for "Activate AppServer" so it doesn't talk about DNS entries


antoviaque:
@bradenmacdonald I should have paid more attention to this ticket - sorry to be posting that that late, it took some time for that context to come back to mind. I'm copying my comment from OC-2178:

> I'm not sure why we didn't end up setting up any firewall rules, but I think this had been discussed during the discovery for the production instances epic. I think we had elected to have OpenCraft IM set the IPs of the appservers in the DBs security groups - this way it doesn't matter that the VMs are on a different OpenStack project (or even a different OpenStack provider, or even on AWS for that matter).
> If the OpenStack security groups are troublesome btw [Jill had mentioned reload issues], another approach, which also would have the advantage to be cross-clouds, would be to use ufw. http://docs.ansible.com/ansible/ufw_module.html

mtyaka:
@bradenmacdonald Should we add some info about security groups and the new setting variables to the README?

Also, I'm not sure if this is easy to change, but in the django admin the field for additional security groups could be wider. After I fill in additional comma-separated security groups and refresh the page, the groups get concatenated with just a comma, but no separating space.
If these are easy to fix, that would be great, but otherwise it's not worth spending time on it.

screen shot 2017-01-13 at 11 14 46

instance/models/openedx_appserver.py
```diff
@@ -279,6 +298,45 @@ def github_admin_username_list(self):
...
+        description = 'Security group for Open EdX AppServers. Managed automatically by OpenCraft IM.'
+        if main_security_group.description != description:
+            network.update_security_group(main_security_group, description=description)
+
+        # We manage this security group - update its rules to match the configured list of rules
+        sync_security_group_rules(main_security_group, EDXAPP_APPSERVER_SECURITY_GROUP_RULES, network=network)
+
+        # For any additional security groups, just verify that the group exists:
+        groups = self.security_groups
+        groups.remove(main_security_group.name) # We already checked this group
```
bradenmacdonald: So that subclasses/mixins can add other security groups by adding their custom groups by overriding `self.security_groups()`, without needing to change this method.

instance/tests/integration/integration_instance.py
```diff
@@ -71,6 +71,35 @@ def assert_instance_up(self, instance):
...
             check_url_accessible(url)
 
+    def assert_appserver_firewalled(self, instance):
+        """
+        Ensure the instance's appserver is not exposing any services it shouldn't be
+        """
+        instance.refresh_from_db()
+        self.assertIsNotNone(instance.active_appserver)
+        server_ip = instance.active_appserver.server.public_ip
+        ports_should_be_open = [22, 80]  # 443 may or may not be open, depending on instance.protocol
```
bradenmacdonald: I believe that right now, if you set "protocol=https" on an Instance, then its appservers will be listening on both port 80 and 443. If you leave the default "protocol=http", then the appservers only listen on port 80. It's based on the ["NGINX_ENABLE_SSL" nginx setting](https://github.com/open-craft/opencraft/blob/0117950b766481f120e663f572ef22824f2f6a2d/instance/templates/instance/ansible/vars.yml#L30). Note that even if the AppServer is listening on port 443, it doesn't have a valid SSL certificate.

instance/models/openedx_appserver.py
```diff
@@ -43,6 +45,12 @@
 
 DEFAULT_EDX_PLATFORM_REPO_URL = 'https://github.com/{}.git'.format(settings.DEFAULT_FORK)
 
+# OpenStack firewall rules (security group rules) to apply to the main security group of each AppServer:
+EDXAPP_APPSERVER_SECURITY_GROUP_RULES = [
```
bradenmacdonald: Sure, good point.

bradenmacdonald:
Thanks @mtyaka! Comments addressed. I updated the README - thanks for the reminder.

The django admin field is something built into django; I don't want to add that in to the scope here, as I don't think it's a simple fix.

instance/models/openedx_appserver.py
```diff
@@ -288,7 +346,17 @@ def provision(self):
...
         """
         self.logger.info('Starting provisioning')
-        # Start by requesting a new server/VM:
+
+        # Check firewall rules:
+        try:
+            self.check_security_groups()
+        except:  # pylint: disable=bare-except
+            message = "Unable to check/update the network security groups for the new VM"
+            self.logger.exception(message)
```
bradenmacdonald: `logger.exception(...)` will also log the traceback of the exception :)

PR #152, opened 2016-11-04
Managed RabbitMQ

bdero:
Implements support for a centrally managed RabbitMQ instance when not using the "ephemeral databases" option.

This PR relies on/uses a bunch of other PRs [linked from the ticket](https://tasks.opencraft.com/browse/OC-1719). Also, this code depends on changes made in #145.

**Testing instructions**:

1. Checkout this branch.
1. Add the following to your .env file:
   
   ```
   RABBITMQ_API_URL='https://rabbitmq-dev.opencraft.hosting:15671'
   RABBITMQ_ADMIN_USERNAME='[username]'
   RABBITMQ_ADMIN_PASSWORD='[password]'
   INSTANCE_RABBITMQ_HOST='rabbitmq-dev.opencraft.hosting'
   INSTANCE_RABBITMQ_PORT='5671'
   ```

1. Run `make migrate`.

1. Create a new instance and spawn an appserver. Set the following settings:

   ```
   Configuration source repo url: https://github.com/open-craft/configuration.git
   Configuration version: bdero/opencraft-rabbitmq-vhost-config

   Edx platform repository url: https://github.com/open-craft/edx-platform.git
   Edx platform commit: bdero/opencraft-celery-broker-tls
   ```

   **The following 4 steps can be done before the appserver finishes provisioning.**

1. Go to the [OpenEdXInstance admin panel](http://localhost:5000/admin/instance/openedxinstance/) and check if the vhost and user data was properly generated - this can be verified a couple of minutes after kicking off the appserver spawn task:

   screenshot

1. Navigate to https://rabbitmq-dev.opencraft.hosting:15671 in your browser and login to the administration account. The administration account credentials can be found by decrypting the [rabbitmq-dev host vars](https://github.com/open-craft/deployment-secrets-services/pull/13/files#diff-0b5b5280556d5a19a034420a0c90b5feR1).

1. Select the "Admin" tab and verify that the two new users were added to the new vhost.

1. Now, go grab some coffee/tea/beer. It's time to wait for the appserver to finish provisioning.

1. Once the appserver finishes provisioning, login to studio on the new Open edX instance and create a rerun of the default course.

1. Return to the RabbitMQ admin panel and verify that there was activity on the instance's new vhost.

**Reviewers**
- [ ] @haikuginger

PR #168, opened 2017-01-15
Adds the programs database and ansible template vars

pomegranited:
The `programs` app has been added to `open-release/ficus.master`'s [`edx_sandbox.yml`](https://github.com/edx/configuration/blob/open-release/ficus.master/playbooks/edx_sandbox.yml#L49).  To support this role, we need to add the programs database and user, and set the corresponding ansible variables.

**Testing instructions**:

1. Create a new instance with settings similar to [ficus-xenial AppServer 7](https://console.opencraft.com/instance/2420/edx-appserver/782/):

    ```
    configuration_source_repo_url:	https://github.com/open-craft/configuration 
    configuration_version:	jill/ficus.master
    edx_platform_repository_url:	https://github.com/open-craft/edx-platform
    edx_platform_commit:	jill/ficus.master
    openedx_release:	open-release/ficus.master

    configuration_extra_settings: |	
      # configuration PR ##3522 changes openstack role variables
      COMMON_ENABLE_OPENSTACK_INTEGRATION: true
      COMMON_SWIFT_LOG_SYNC: true

      # Have to specify settings as openstack, not aws
      COMMON_EDXAPP_SETTINGS: 'openstack'
    ```
    Note that AppServer 7 failed to provision due to an error in `TASK [programs : migrate]`.
1. Spawn an appserver.  Provisioning should succeed beyond the `[programs]` role.
1. Spawn a second appserver, and ensure that the 3 "Provisioning" steps runs (again) without error.

1. The `programs` role has not yet been added to [`edx-stateless.yml`](https://github.com/edx/configuration/blob/open-release/ficus.master/playbooks/edx-stateless.yml), but since our clients might want the role, it's worth adding support for it into OCIM.

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/instance/models/mixins/openedx_database.py
+++ b/instance/models/mixins/openedx_database.py
@@ -99,6 +99,10 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
                     }
                 ],
             },
+            {
+                "name": self._get_mysql_database_name("programs"),
+                "user": self._get_mysql_user_name("program"),
+            },
         ]
 
     @property
--- a/instance/tests/models/test_openedx_database_mixins.py
+++ b/instance/tests/models/test_openedx_database_mixins.py
@@ -337,6 +337,13 @@ class MySQLInstanceTestCase(TestCase):
                     "CONN_MAX_AGE": 60,
                 }}]
             ),
+            "PROGRAMS_": make_nested_group_info(
+                ["DEFAULT_DB_NAME", "DATABASES"],
+                [{"name": "programs", "user": "program", "additional_settings": {
+                    "ATOMIC_REQUESTS": True,
+                    "CONN_MAX_AGE": 60,
+                }}]
+            ),
             "INSIGHTS_": make_nested_group_info(
                 ["DATABASE_NAME", "DATABASES"],
                 [{"name": "dashboard", "user": "dashboard"}]
--- a/circle.yml
+++ b/circle.yml
@@ -32,4 +32,5 @@ test:
         timeout: 1800
         parallel: true
   post:
-    - make test_integration_cleanup
+    - make test_integration_cleanup:
+        timeout: 1200
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -27,6 +27,7 @@ import inspect
 from django.db import models
 import MySQLdb as mysql
 import pymongo
+import warnings
 
 from instance.models.database_server import MySQLServer, MongoDBServer
 
@@ -69,9 +70,11 @@ def _get_mysql_cursor(mysql_server):
 @database_name_escaped
 def _create_database(cursor, database):
     """
-    Create MySQL database
+    Create MySQL database, if it doesn't already exist.
     """
-    cursor.execute('CREATE DATABASE `{database}` DEFAULT CHARACTER SET utf8'.format(database=database))
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        cursor.execute('CREATE DATABASE IF NOT EXISTS `{db}` DEFAULT CHARACTER SET utf8'.format(db=database))
 
 
 def _create_user(cursor, user, password):
@@ -102,7 +105,9 @@ def _drop_database(cursor, database):
     """
     Drop MySQL database
     """
-    cursor.execute('DROP DATABASE IF EXISTS `{database}`'.format(database=database))
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        cursor.execute('DROP DATABASE IF EXISTS `{database}`'.format(database=database))
 
 
 def _drop_user(cursor, user):
@@ -145,7 +150,7 @@ class MySQLInstanceMixin(models.Model):
         """
         Create mysql user and databases
         """
-        if self.mysql_server and not self.mysql_provisioned:
+        if self.mysql_server:
             cursor = _get_mysql_cursor(self.mysql_server)
 
             # Create migration and read_only users
@@ -220,9 +225,10 @@ class MongoDBInstanceMixin(models.Model):
         """
         Create mongo user and databases
         """
-        if self.mongodb_server and not self.mongo_provisioned:
+        if self.mongodb_server:
             mongo = pymongo.MongoClient(self.mongodb_server.url)
             for database in self.mongo_database_names:
+                # May update the password if the user already exists
                 mongo[database].add_user(self.mongo_user, self.mongo_pass)
             self.mongo_provisioned = True
             self.save()
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -56,7 +56,7 @@ class SwiftContainerInstanceMixin(models.Model):
         """
         Create the Swift containers if necessary.
         """
-        if settings.SWIFT_ENABLE and not self.swift_provisioned:
+        if settings.SWIFT_ENABLE:
             for container_name in self.swift_container_names:
                 openstack_utils.create_swift_container(
                     container_name,
--- a/instance/templates/instance/ansible/mysql.yml
+++ b/instance/templates/instance/ansible/mysql.yml
@@ -49,6 +49,19 @@ EDX_NOTES_API_MYSQL_DB_USER: '{{ edx_notes_api_user }}'
 EDX_NOTES_API_MYSQL_DB_PASS: '{{ edx_notes_api_pass }}'
 EDX_NOTES_API_MYSQL_HOST: '{{ host }}'
 
+# programs
+PROGRAMS_DEFAULT_DB_NAME: '{{ programs_database }}'
+PROGRAMS_DATABASES:
+  default:
+    ENGINE: 'django.db.backends.mysql'
+    NAME: '{{ programs_database }}'
+    USER: '{{ programs_user }}'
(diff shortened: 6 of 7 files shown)


instance/models/mixins/openedx_database.py
```diff
@@ -95,6 +95,10 @@ def mysql_databases(self):
                     }
                 ],
             },
+            {
+                "name": self._get_mysql_database_name("programs"),
+                "user": self._get_mysql_user_name("program"),
+            },
```
antoviaque: Would that automatically create the new database for existing instances?

pomegranited: Ah, good point.  Instances with previously provisioned databases have their [`mysql_provisioned` flag set, and so the databases aren't re-created](https://github.com/open-craft/opencraft/blob/f94660e9e267f85af61ceb50c6600003998b6276/instance/models/mixins/database.py#L144-L148).

It's easy to fix, though.  `provision_mysql()` calls three methods, and only `_create_database` would need to be changed:
* [`_create_database`](https://github.com/open-craft/opencraft/blob/f94660e9e267f85af61ceb50c6600003998b6276/instance/models/mixins/database.py#L69-L74) - could be modified to be [`CREATE DATABASE <database> IF NOT EXISTS`](http://dev.mysql.com/doc/refman/5.7/en/create-database.html). 
* [`create_user`](https://github.com/open-craft/opencraft/blob/f94660e9e267f85af61ceb50c6600003998b6276/instance/models/mixins/database.py#L77-L86) - creates the given user only if it does not already exist.
* [`_grant_privileges`](https://github.com/open-craft/opencraft/blob/f94660e9e267f85af61ceb50c6600003998b6276/instance/models/mixins/database.py#L89-L97) - GRANT is a no-op if the privileges are already granted.

66c50f0 makes this change, but it raises the question: do we need the `mysql_provisioned` flag anymore?

mtyaka: Nice improvement, will make adding new databases in the future easier.

> do we need the mysql_provisioned flag anymore?

I don't think we do. The methods that deprovision databases already seem to be designed so that they won't fail if the database or user does not exist, so `mysql_provisoined` flag doesn't make much sense anymore.

pomegranited: Also looks like we can remove `mongo_provisioned`, because if [`mongo.add_user`](https://github.com/open-craft/opencraft/blob/f94660e9e267f85af61ceb50c6600003998b6276/instance/models/mixins/database.py#L223) is called on an existing user, all it does is potentially [update the password](http://api.mongodb.com/python/current/api/pymongo/database.html#pymongo.database.Database.add_user).

And [`swift_provisioned`](https://github.com/open-craft/opencraft/blob/077e1a71dcfc574581b4c11581794d9b8436d45b/instance/models/mixins/storage.py#L59) too!  If the container already exists, it might only [update its metadata](http://docs.openstack.org/developer/python-swiftclient/swiftclient.html#swiftclient.client.post_container).

Does that sound ok to you @mtyaka ?

pomegranited:
@mtyaka I've updated the PR instructions to ensure the 2nd instance provisions correctly.

pomegranited:
@mtyaka my [previous test failure](https://circleci.com/gh/open-craft/opencraft/1878) was due to a strange error:

```
psycopg2.OperationalError: database "test_opencraft" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

I've never seen that before.. There were no other builds running with mine.  Do you think it could be a side effect of this change?  I'm [re-running now to see if it goes away](https://circleci.com/gh/open-craft/opencraft/1879), but would hate to introduce any flakiness.

mtyaka:
:+1: The instance that I was trying to spawn failed due some kind of timeout, but it was past the `programs` role, so that shouldn't block this PR from getting merged.

- [x] I tested this: Spawned a new instance using settings from the PR description. Without this patch, ansible failed on the programs step; with this patch it performed the programs steps successfully. Spawned another app server to ensure reprovisioning mysql, mongo, swift works on existing instance.
- [x] I read through the code

bradenmacdonald:
@pomegranited Nit: Can we silence all these things which kind of look like errors in the log? Or catch this exception and make it nicer like `Checking database... already exists, good.`

```
/home/ubuntu/opencraft/instance/models/mixins/database.py:74: Warning: (1007, "Can't create database 'instance0de2cc7f_test_example_com_ecommerce'; database exists")
  cursor.execute('CREATE DATABASE IF NOT EXISTS `{database}` DEFAULT CHARACTER SET utf8'.format(database=database))
/home/ubuntu/opencraft/instance/models/mixins/database.py:74: Warning: (1007, "Can't create database 'instance0de2cc7f_test_example_com_dashboard'; database exists")
  cursor.execute('CREATE DATABASE IF NOT EXISTS `{database}` DEFAULT CHARACTER SET utf8'.format(database=database))

...
18/Jan/2017 04:45:19 |     INFO | process=26320 | swiftclient               | REQ: curl -i https://storage.gra1.cloud.ovh.net/v1/AUTH_2b26f56d1c394ca88ee2b412720d73de/1ec9cc4c_integration_plebia_net?format=json -X GET -H "X-Auth-Token: 660fc0aca05342c9..." -H "Accept-Encoding: gzip"
18/Jan/2017 04:45:19 |     INFO | process=26320 | swiftclient               | RESP STATUS: 404 Not Found
18/Jan/2017 04:45:19 |     INFO | process=26320 | swiftclient               | RESP HEADERS: {'Date': 'Wed, 18 Jan 2017 04:45:19 GMT', 'X-Trans-Id': 'tx722d9e04a92a437d9dfd2-00587ef2de', 'Content-Length': '70', 'Content-Type': 'text/html; charset=UTF-8'}
18/Jan/2017 04:45:19 |     INFO | process=26320 | swiftclient               | RESP BODY: b'<html><h1>Not Found</h1><p>The resource could not be found.</p></html>'
18/Jan/2017 04:45:19 |    ERROR | process=26320 | swiftclient               | Container GET failed: https://storage.gra1.cloud.ovh.net/v1/AUTH_2b26f56d1c394ca88ee2b412720d73de/1ec9cc4c_integration_plebia_net?format=json 404 Not Found  [first 60 chars of response] b'<html><h1>Not Found</h1><p>The resource could not be found.<'
  File "/home/ubuntu/virtualenvs/venv-3.5.1/lib/python3.5/site-packages/swiftclient/client.py", line 1647, in _retry
    service_token=self.service_token, **kwargs)
  File "/home/ubuntu/virtualenvs/venv-3.5.1/lib/python3.5/site-packages/swiftclient/client.py", line 917, in get_container
    raise ClientException.from_response(resp, 'Container GET failed', body)
swiftclient.exceptions.ClientException: Container GET failed: https://storage.gra1.cloud.ovh.net/v1/AUTH_2b26f56d1c394ca88ee2b412720d73de/1ec9cc4c_integration_plebia_net?format=json 404 Not Found  [first 60 chars of response] b'<html><h1>Not Found</h1><p>The resource could not be found.<'
```

Also this is an optional nit, but it would be nice to combine these three migrations into one before merging this.

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -470,6 +479,9 @@ def test_ansible_settings_mongo(self):
         self.assertIn('FORUM_MONGO_PORT: {0}'.format(MONGODB_SERVER_DEFAULT_PORT), ansible_vars)
         self.assertIn('FORUM_MONGO_DATABASE: {0}'.format(self.instance.forum_database_name), ansible_vars)
 
+        # Unset self.instance to avoid tearDown failures trying to deprovisioning databases that were not created
+        self.instance = None
```
bradenmacdonald: Why are these lines needed? I thought Matjaz said that "The methods that deprovision databases already seem to be designed so that they won't fail if the database or user does not exist"

If I go into the shell and I go:
```python
instance = production_instance_factory(name="Production instance", sub_domain="production")
instance.delete()
```
Does that still work without error (no actual provisioning or deprovisioning happens)?

pomegranited:
@mtyaka  Apologies for how long this PR has taken me to do..  I was still getting those weird errors about being unable to drop the `test_opencraft` database, even after removing my test changes.  So I decided to simplify the change, to see if that fixes it.  *fingers crossed*

You'll need to revert the migrations on your devstack to re-test this, I'm afraid.

```
make manage migrate instance 0067_remove_blank_name
make migrate
```

* Left the `*_provisioned` flags in place.  I'm not checking them on provisioning, so added databases, tables, and storage can be added after initial appserver creation, but I'm still checking them on deprovisioning.
* Using `IF NOT EXISTS` during MySQL database creation, but am silencing warnings that occur during that call. (CC @bradenmacdonald).
* I haven't been able to silence the red herring `ERROR` logs generated by the swift library though.  They've been happening all along, because the swift library checks for the existence of the container before trying to create it.

I've also rebased on the latest master, to assist with the merge.  If you're running the tests in your devstack, note the changes to your .env that are required since https://github.com/open-craft/opencraft/pull/152 merged, and the temporary change which may be required to your .env.test: https://github.com/open-craft/opencraft/pull/152#issuecomment-274225119.

mtyaka:
👍 I re-tested this and the changes make sense. You'll have to rebase again, though.

pomegranited:
Thanks, @mtyaka , but oh no!  I know that if I rebase, I'll never get my tests to pass again..  ref https://github.com/open-craft/opencraft/pull/169 😢

PR #169, opened 2017-01-17
Allows the OpenStack base image, flavor, and ssh_keyname to be configured per instance

pomegranited:
This change allows the OpenStack base image, flavor, and ssh keyname settings to be made on the instance and its appservers, instead of globally for the app.  The default values are pulled from the global .env settings.

**Testing instructions**:

1. Snapshot your opencraft VM, since there are database migrations you may want to be able to revert.
1. Checkout this branch, and `make migrate` to update the database schema.
2. Create an instance using `production_instance_factory`.
3. Set the instance's openstack settings to use a different base image, e.g. this one which is available on our Dev OVH:

    ```
    instance.openstack_server_base_image = {"name": "xenial-server-cloudimg-amd64-disk1-unmodified-20161214"}
   ```
4. Spawn a new appserver, and note that it uses the instance's base image.  Provisioning will fail because `/usr/bin/python` is missing from the xenial base image (ref https://github.com/edx/configuration/pull/3606).
5. Unset `instance.openstack_server_base_image = None`, and spawn another appserver.  That server's settings should default to your `.env` file's `OPENSTACK_SANDBOX_BASE_IMAGE` settings.

**Reviewers**
- [ ] @haikuginger


instance/models/server.py
```diff
@@ -256,7 +257,59 @@ def update_status(self):
...
+        null=True, blank=True, default=None,
+        help_text='JSON openstack base image selector, e.g. {"name": "ubuntu-12.04-ref-ul"}'
+                  ' Defaults to settings.OPENSTACK_SANDBOX_BASE_IMAGE on server creation.',
+    )
+    openstack_server_ssh_keyname = models.CharField(
+        max_length=256,
+        null=True, blank=True, default='',
+        help_text='SSH key name used when setting up access to the openstack project.'
+                  ' Defaults to settings.OPENSTACK_SANDBOX_SSH_KEYNAME on server creation.',
+    )
```
antoviaque: @pomegranited Since these settings are only set once per VM/AppServer, and can then be retrieved from the Nova API, is it useful to store it at the individual AppServer model level, vs setting it at the instance level and then using the current instance value when provisioning a new AppServer?

pomegranited: @antoviaque I thought we needed to be able to store it at the instance level, so that appservers created for that instance use the correct settings?

antoviaque: @pomegranited I haven't looked into it, but it would be good to find a way to not duplicate data already stored? There is no way to pass on the instance/appserver objects during the server object creation?

bradenmacdonald: We could just store in on the instance. In the appserver code, when you create a server, you can say `openstack_utils.create_server(flavor=self.instance.openstack_server_flavor)`.

And as Xavier mentions, we could retrieve the resulting value from the OpenStack API, rather than storing it on the server - unless there's some value in knowing what the requested flavor was at the time a historical appserver was provisioned.

instance/models/server.py
```diff
@@ -29,6 +29,7 @@
 from django.db import models
 from django.db.models import Q
 from django_extensions.db.models import TimeStampedModel
+from django_extensions.db.fields.json import JSONField
```
bradenmacdonald: For new fields, we should be using `from django.contrib.postgres.fields import JSONField`. The `django_extensions` version is deprecated in favor of the django one, and the native django one used the postgres JSONB type, which is much more efficient and powerful than the TEXT type used by this. I intend to migrate all the code in opencraft IM to use the new JSON field type at some point.

pomegranited: Thanks for the tip, @bradenmacdonald !  Also fixed in [8f17ecf](https://github.com/open-craft/opencraft/pull/169/commits/8f17ecf57f20fa4e4a0641d4c2746cac5a6ffefd#diff-b66ef4710ed40f1b980e3ab3a34ec812R26).

instance/models/server.py
pomegranited: Thanks for the tip, @bradenmacdonald !  I should have seen that the appserver has access to its own instance.  That simplifies the migration heaps: 8f17ecf

instance/migrations/0069_openstack_settings.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('instance', '0068_security_groups'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='openstack_server_base_image',
+            field=django.contrib.postgres.fields.JSONField(blank=True, default=None, help_text='JSON openstack base image selector, e.g. {"name": "ubuntu-12.04-ref-ul"} Defaults to settings.OPENSTACK_SANDBOX_BASE_IMAGE on server creation.', null=True),
```
bradenmacdonald: @haikuginger The best way to do that IMHO would be keeping the default=None in the migration and adding [a data migration](https://docs.djangoproject.com/en/1.10/topics/migrations/#data-migrations) that updates any existing rows to use the current setting value. You could safely include the data migration in this same migration file, but it's probably best to put it as an additional migration file.

instance/models/openedx_instance.py
```diff
@@ -87,6 +88,23 @@ class OpenEdXInstance(LoadBalancedInstance, OpenEdXAppConfiguration, OpenEdXData
 
     successfully_provisioned = models.BooleanField(default=False)
 
+    openstack_server_flavor = JSONField(
+        null=True, blank=True, default=None,
```
pomegranited: Ah, gotcha @haikuginger .  Fixed with 375ae9c.

instance/migrations/0069_openstack_settings.py
pomegranited: Thanks for reminding me, @haikuginger, and for the documentation pointers, @bradenmacdonald.  I've added the data migration with 32a1a76.

haikuginger:
This is looking awesome, @pomegranited! All set to go once you rebase, rename your migrations, squash, and do all that stuff!

pomegranited:
Thanks, @haikuginger !

antoviaque:
@pomegranited For the buidls timeouts -- I think the 2h are a hard limit from CircleCI, last time I checked you couldn't extend them more than that. It might be worth looking into why they got slower though?

bradenmacdonald:
OC-2043 could help reduce that :)

pomegranited:
@antoviaque  Well, we've added a new service dependency with RabbitMQ..  These tests were passing before I rebased 😦   And OVH is giving us trouble, so it could be a combination of things.

bradenmacdonald:
@pomegranited I would expect that the RabbitMQ change sped up the tests, since they can connect to an existing stage/integration RabbitMQ server, and we no longer have to deploy RabbitMQ on the VMs during integration tests??

pomegranited:
@bradenmacdonald  @antoviaque  It's not just me that's having issues: https://circleci.com/gh/open-craft/opencraft

instance/models/server.py
smarnach: I just saw this discussion.  While it's not necessary to store the information on the server object, it would be good to copy the information to appservers as they are created.  If we don't, the configuration history will be incomplete, which is undesirable and inconsistent with how we handle other configuration settings.

bradenmacdonald: @smarnach Yep, that makes sense.

smarnach: @bradenmacdonald So should we make the suggested changes, i.e. move the settings to `OpenEdXAppConfiguration`, so they get included in both appservers and instances?  If so, we need a new ticket to actually do this.  It's an easy and straightforward change, but someone has to do it.

bradenmacdonald: @smarnach  I created OC-2389.

PR #170, opened 2017-01-17
AppServers Cleanup v2 - Only keep active, fallback and RC servers

antoviaque:
Terminate app servers that were created more than `days` before now, except:
* the active appserver if there is one,
* a "release candidate" (rc) appserver, to allow testing before the next appserver activation (we keep the most recent running appserver)
* a fallback appserver, for `days` after activating an appserver, to allow reverts (we keep the most recent running appserver created before the current activation)

See OC-2041 . Follow-up to #143 

**Test Instructions**

1. Create an instance that is not associated with a PR:
   - Run `make shell`.
   - Create instance via:
     
     ``` python
     regular_instance = instance_factory(name="Regular instance, no PR", sub_domain="regular-instance-zohoRek7")
     ```
1. Spawn four app servers for the instance, faking their age:
     
     ``` python
     from datetime import timedelta
     from freezegun import freeze_time

     with freeze_time(timezone.now() - timedelta(days=5)):
         oldest_appserver_id = regular_instance.spawn_appserver()

     with freeze_time(timezone.now() - timedelta(days=4)):
         fallback_appserver_id = regular_instance.spawn_appserver()

     with freeze_time(timezone.now() - timedelta(days=1)):
         active_appserver_id = regular_instance.spawn_appserver()

     with freeze_time(timezone.now() + timedelta(days=4)):
         rc_appserver_id = regular_instance.spawn_appserver()
     ```

1. Activate `active_appserver`:

     ``` python
     regular_instance.set_appserver_active(active_appserver_id)
     ```

1. Check the state of the appservers. You can use the GUI or do it directly in the shell via something like:
     
     ``` python
     for appserver in regular_instance.appserver_set.all():
         print(appserver)
         print("Status: ", appserver.status)
         print("Status (server): ", appserver.server.status)
     ```
     
     This should produce the following output:
     
     ```
     AppServer 1
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 2
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 3
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 4
     Status:  Running [running]
     Status (server):  Ready [ready]
     ```
1. Run the `clean_up` task:
     
     ``` python
     from instance.tasks import clean_up
     clean_up()
     ```
1. Verify that only the oldest appserver has been shut down. This should produce the following output:
     ```
     AppServer 1
     Status:  Terminated [terminated]
     Status (server):  Terminated [terminated]
     AppServer 2
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 3
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 4
     Status:  Running [running]
     Status (server):  Ready [ready]
     ```

1. Re-run the `cleanup_task` from different dates to check that the behavior correctly matches the description at the top of this PR. Especially:
     - The active appserver and the most recent running appserver (rc) should never be terminated
     - The fallback appserver should only be terminated after `days` have passed
     - Far in the future, only the active appserver and the most recent running appserver should remain

     To run the `cleanup_task` from different dates, use (make sure you use `HUEY_ALWAYS_EAGER=true`):
    
     ```
     with freeze_time(timezone.now() + timedelta(days=3)):
         clean_up()
     ```

     At D+3, you should get:
    
     ```
     AppServer 1
     Status:  Terminated [terminated]
     Status (server):  Terminated [terminated]
     AppServer 2
     Status:  Terminated [terminated]
     Status (server):  Terminated [terminated]
     AppServer 3
     Status:  Running [running]
     Status (server):  Ready [ready]
     AppServer 4
     Status:  Running [running]
     Status (server):  Ready [ready]
     ```

1. Ensure AppServers which aren't in the 'running' state always get deprovisioned after two days.

1. Double-check "OVH - OpenStack - Horizon - OpenCraft IM - Dev" account to see if correct set of app servers left.

**Reviewers**

- [ ] @itsjeyd


instance/tests/models/test_openedx_instance.py
```diff
@@ -912,45 +929,91 @@ def test_terminate_obsolete_appservers(self, days):
...
+            (oldest_appserver, AppServerStatus.Terminated, ServerStatus.Terminated),
+            (oldest_appserver_failed, AppServerStatus.ConfigurationFailed, ServerStatus.Terminated),
+            (fallback_appserver, AppServerStatus.Terminated, ServerStatus.Terminated),
+            (fallback_appserver_failed, AppServerStatus.ConfigurationFailed, ServerStatus.Terminated),
+            (active_appserver, AppServerStatus.Running, ServerStatus.Pending),
+            (rc_appserver, AppServerStatus.Running, ServerStatus.Pending),
+            (rc_appserver_failed, AppServerStatus.ConfigurationFailed, ServerStatus.Pending),
+        ])
+
+        # Terminated app servers again, much later - this time only the active appserver and the most recent
```
itsjeyd: @antoviaque Typo: "Terminated" should be "Terminate"

README.md
```diff
@@ -121,14 +121,15 @@ First, **as the user that the instance manager runs as (e.g. `vagrant`)**, run:
 
 This will create an ssh key pair for that user, saving it at `~/.ssh/id_rsa` and
 `~/.ssh/id_rsa.pub` by default. Next, we need to upload the public key to
-OpenStack. Make sure the nova command line client is installed:
+OpenStack. Make sure the nova command line client is installed (it should already
+be the case from the OpenCraft IM python environment):
 
-    pip install python-novaclient
+    pip install python-openstackclient
```
bradenmacdonald: Thanks for updating this :)

itsjeyd:
@antoviaque A quick question regarding the test instructions, just to clarify:

> Ensure AppServers which aren't in the 'running' state always get deprovisioned after two days.

"After two days" -- are you referring to the (fake) date at which I'd be running the `clean_up` task here?

antoviaque:
@itsjeyd Thanks for the review! I'll go through your comments to address them, but I wanted to answer your questions quickly:

> I had a look at the code and left some comments. Nothing major, the approach looks good in general. Did not get around to testing this yet; do you think it would make sense to deploy the changes to stage? That might speed things up, especially w/r/t being able to provision multiple app servers at once. (I don't have a working instance manager configuration on my local machine at the moment; the last time I had to spin up instances locally was before Sven finished the load balancer implementation.)

Yup, we could definitely deploy on stage - I'll try to do this after addressing your current comments if I have time. Or feel free to do it directly if you want to test this before I get around to do it.

> "After two days" -- are you referring to the (fake) date at which I'd be running the clean_up task here?

Yes, exactly.

antoviaque:
@itsjeyd Addressed your comments - all good ones, thank you!

I also looked into deploying on stage, but it is currently on the rabbitmq branch, so I wasn't sure if it was free to use now. @bdero do you still need stage?

README.md
antoviaque: : ) yw!

antoviaque:
@itsjeyd I've updated stage with the current branch, @bdero confirmed on IRC that he does't need it. That actually allowed me to spot an issue - it was missing a data migration for the existing active appservers, which need a value for `last_activated` - I've set it to the time of the data migration.

instance/models/openedx_instance.py
```diff
@@ -409,15 +412,36 @@ def require_user_creation_success(self):
...
+        rc_appserver = None
+        now = timezone.now()
+
+        for appserver in self.appserver_set.all().order_by('-created'):
+            if appserver == active_appserver:
+                continue
+
+            # Keep a running appserver as fallback for `days` after activation, to allow reverts
+            if active_appserver and appserver.created < active_appserver.last_activated:
+                if not sufficient_time_passed(active_appserver.last_activated, now, days) \
```
itsjeyd: @antoviaque If there is a negative delta between the first and the second date passed to `sufficient_time_passed` (i.e., if the first date is more recent than the second one), the function will return `False`. If that situation ever came up (highly unlikely), `not sufficient_time_passed(...)` would evaluate to `True` and we'd end up keeping an additional `fallback_appserver`.

Just wanted to mention this; no need to change anything. The "dangerous" code that terminates VMs below will only run if `sufficient_time_passed` returns `True`, so a negative delta between the first and second date would not cause VMs to get terminated.

itsjeyd:
@antoviaque Thanks for the updates, and for creating the data migration! I ended up bringing my local installation of OC IM up to date yesterday, and completed a first round of testing this morning (including verifying that the data migration correctly updates `last_activated` for active app servers).

The code is looking good, and I didn't find any issues during the first round of testing. But I'll still need to verify that we're getting desired behavior when there is no active app server. I'll use stage to do that, and let you know when I'm done.

antoviaque:
@itsjeyd Thank you for the prompt review and testing! Merging and deploying this now.

PR #172, opened 2017-01-23
Added script to reprovision all instances in waves

bdero:
This is the script that was written and used to roll out #152 to all of our instances this week.

CC @pomegranited

### DIFF
--- a/bin/provision-instances
+++ b/bin/provision-instances
@@ -1,0 +1,227 @@
+#!/usr/bin/python3
+#
+# This script runs a full instance provision over all non-ephemeral instances
+# on OCIM. Running this script requires Django to be initialized - the easiest
+# way to do this is by evaluating the script from within the Django shell by
+# running:
+#
+#   > eval(open('bin/provision-instances').open())
+#
+# The provision process happens in "waves", where all instances that fail to
+# provision in a given wave are recorded and reprovisioned during the next wave.
+#
+# This provision script is designed to withstand restarts of either OCIM or
+# this script itself; it accomplishes this by always saving and loading the
+# state of the provision process to/from `ocim_update_waves.ini`.
+
+import ast
+from configparser import ConfigParser
+from datetime import datetime, timedelta
+import os
+import pytz
+from time import sleep
+
+from instance.models.openedx_instance import OpenEdXInstance
+from instance.tasks import spawn_appserver
+
+WAVES_FILE = 'ocim_update_waves.ini'
+SLEEP_TIME = 60
+MAX_PROCESSING = 6
+MAX_PROCESSING_TIME = 240  # 4 hours
+
+waves_config = ConfigParser()
+
+
+def bootstrap_instance_list():
+    # Collect all of the instances that should be included in the OCIM upgrades.
+    #
+    # Instances with any of the following attributes at the start of the upgrade
+    # are avoided for the entire upgrade:
+    #   - Any instance considered "shut down" (all VMs terminated)
+    #   - Any instance without any appservers
+    #   - Any instance set to use ephemeral databases
+    #   - Any instance where the latest appserver isn't in a healthy stable state (e.g. running)
+
+    failed_results = []
+    ready_instances = []
+    for inst in OpenEdXInstance.objects.all():
+        # Set field defaults for ALL instances, even ones we're not going to reprovision.
+        inst.set_field_defaults()
+        inst.save()
+
+        latest_appserver = inst.appserver_set.last()
+
+        if inst.is_shut_down:
+            print('-', end='', flush=True)
+            continue
+
+        if not latest_appserver:
+            failed_results.append((inst, 'is has no appservers'))
+            print('n', end='', flush=True)
+            continue
+
+        if inst.use_ephemeral_databases:
+            failed_results.append((inst, 'it has "use_ephemeral_databases" enabled'))
+            print('e', end='', flush=True)
+            continue
+
+        if not (latest_appserver.status.is_steady_state and latest_appserver.status.is_healthy_state):
+            failed_results.append((inst, 'it\'s latest appserver is in the "{}" state'.format(
+                latest_appserver.status.name
+            )))
+            print('x', end='', flush=True)
+            continue
+
+        ready_instances.append(inst.pk)
+        print('.', end='', flush=True)
+
+    print('\n')
+    for inst, message in failed_results:
+        print('* Instance {} will be excluded because {}.'.format(inst.internal_lms_domain, message))
+    print('\n** TOTAL INCLUDED: {}\n** TOTAL EXCLUDED: {}'.format(
+        len(ready_instances), len(failed_results)
+    ))
+
+    return ready_instances
+
+
+def save_waves_config():
+    with open(WAVES_FILE, 'w') as waves_config_file:
+        waves_config.write(waves_config_file)
+
+
+def get_current_wave():
+    if not os.path.isfile(WAVES_FILE):
+        initial_instance_list = bootstrap_instance_list()
+
+        print ('Waiting for 10 seconds... quit now if you don\'t want to start spawning appservers.')
+        sleep(10)
+
+        current_wave = '0'
+        print('** INITIALIZING WAVE 0 **')
+
+        waves_config[current_wave] = {
+            'pending': str(initial_instance_list),
+            'processing': '[]',
+            'successful': '[]',
+            'failed': '[]',
+        }
+        # This section stores the values of the latest appserver (before a new
+        # one was spawned). This is useful for sanity checking.
+        waves_config['!latest_appservers'] = {}
+        save_waves_config()
+    else:
+        waves_config.read(WAVES_FILE)
+
+        last_wave = sorted(waves_config.sections())[-1]
+        last_wave_config = waves_config[last_wave]
+
+        last_wave_parsed = {}
+        for category, instance_list in last_wave_config.items():
+            last_wave_parsed[category] = ast.literal_eval(instance_list)
+
+        if len(last_wave_parsed['pending']) == 0 and len(last_wave_parsed['processing']) == 0:
+            # There are no instances in the current wave that are pending or processing.
+            if len(last_wave_parsed['failed']) == 0:
+                # If there aren't any failed instances or if we've reached the
+                # wave limit, cease fire.
+                return None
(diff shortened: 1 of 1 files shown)


bin/provision-instances
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/python3
+#
```
antoviaque: Missing the standard header

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+# This script runs a full instance provision over all non-ephemeral instances
+# on OCIM. Running this script requires Django to be initialized - the easiest
+# way to do this is by evaluating the script from within the Django shell by
+# running:
```
antoviaque: Should this be a Django command then?

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+#
+#   > eval(open('bin/provision-instances').open())
+#
+# The provision process happens in "waves", where all instances that fail to
+# provision in a given wave are recorded and reprovisioned during the next wave.
+#
+# This provision script is designed to withstand restarts of either OCIM or
+# this script itself; it accomplishes this by always saving and loading the
+# state of the provision process to/from `ocim_update_waves.ini`.
```
antoviaque: That means keeping state locally on the hard-drive - can this be saved in the DB? Ideally using only instance/appserver fields, and not duplicating existing information about state (ie only adding fields generically when the information isn't available at all).

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+    ready_instances = []
+    for inst in OpenEdXInstance.objects.all():
+        # Set field defaults for ALL instances, even ones we're not going to reprovision.
+        inst.set_field_defaults()
+        inst.save()
+
+        latest_appserver = inst.appserver_set.last()
+
+        if inst.is_shut_down:
+            print('-', end='', flush=True)
```
antoviaque: Use standard logging to record this rather than print statements

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+
+        if inst.use_ephemeral_databases:
+            failed_results.append((inst, 'it has "use_ephemeral_databases" enabled'))
+            print('e', end='', flush=True)
+            continue
+
+        if not (latest_appserver.status.is_steady_state and latest_appserver.status.is_healthy_state):
+            failed_results.append((inst, 'it\'s latest appserver is in the "{}" state'.format(
+                latest_appserver.status.name
+            )))
```
antoviaque: Why are you excluding these instances, exactly?

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+    #   - Any instance where the latest appserver isn't in a healthy stable state (e.g. running)
+
+    failed_results = []
+    ready_instances = []
+    for inst in OpenEdXInstance.objects.all():
+        # Set field defaults for ALL instances, even ones we're not going to reprovision.
+        inst.set_field_defaults()
+        inst.save()
+
+        latest_appserver = inst.appserver_set.last()
```
antoviaque: Why do you need the latest appserver here? 

Rather than a separate logic to identify the latest server, could you check the rc appserver instead (and/or the active appserver, depending on why you need this)? Ie reuse and refactor/split out the logic from https://github.com/open-craft/opencraft/blob/c924013f603408c2a295fe9734bf86cfcb6ec0b0/instance/models/openedx_instance.py#L413-L421

bin/provision-instances
```diff
@@ -0,0 +1,227 @@
...
+    return ready_instances
+
+
+def save_waves_config():
+    with open(WAVES_FILE, 'w') as waves_config_file:
+        waves_config.write(waves_config_file)
+
+
+def get_current_wave():
+    if not os.path.isfile(WAVES_FILE):
```
antoviaque: Rather than managing waves manually, which seem to be a way to not overload the workers, we had discussed adding support for multiple queues priorities in Huey, which seemed already largely implemented. I can't find the thread we discussed this though, there was links to the code there if I remember correctly.

@bradenmacdonald I think I was discussing this with you? If so, do you remember where we discussed this?

bin/provision-instances
pomegranited: Ah yes.. a django management command would work better for this, instead of just a standalone script.  Sorry, that was my bad advice to put this into `bin`!

bdero: This is one possibility - another possibility would be to initialize Django via `django.setup()`. I wasn't sure if it would require more work to make sure Huey works normally, though.

Either way, since I was testing the bdero/rabbitmq branch separately on staging while writing this code, I didn't really have the time to fiddle with either of those options as I had to start provisioning the appservers overnight ASAP.

Maybe we can do this in a follow up ticket?

bin/provision-instances
bdero: This is an ad-hoc script - the other scripts don't have a standard looking header. E.g. https://github.com/open-craft/opencraft/blob/master/bin/bootstrap.

Is there a special reason this one should have it, or are the other scripts just missing it as well?

bin/provision-instances
bdero: Where are thread queues in Huey stored? If they're not stored in a non-ephemeral way that can survive e.g. a full OS restart, then it doesn't seem like a practical way to store state for a process that will take place over the course of more than 16 hours.

bin/provision-instances
bdero: If the latest appserver isn't healthy, that means it doesn't have a configuration which will succeed when we try to redeploy it. Later on if/when someone fixes its configuration and redeploys/activates an appserver, it'll be redeployed at that point.

pomegranited:
We may also want to add a way to upgrade instances before we spawn new appservers, a la https://github.com/open-craft/opencraft/blob/master/instance/instance_upgrade.py

bin/provision-instances
bdero: It's for checking whether or not the latest appserver is healthy [below](https://github.com/open-craft/opencraft/pull/172/files#diff-86f6807c62ba5f2ef8f70873c286e648R68).

bin/provision-instances
bdero: Storing this info in the DB is definitely possible, but getting this script done in a trustworthy way on time for rolling out RabbitMQ definitely wouldn't have been possible, especially since that would require migrations that would make it less of an ad-hoc script, and require extensive code review/merging into the code base before use. :)

Keep in mind that this really needed to be a quick external tool to facilitate appserver spawning in order to ensure no spillover. I would have actually much preferred to use an HTTP API to spin up/check on the appservers instead of running `spawn_appserver` directly. In fact, I would have preferred to run this tool on a server external to OCIM altogether, so that it would be able to withstand OCIM outages in a way that wouldn't steer anyone away from restarting/upgrading OCIM.

What are your thoughts about this? Do you think this tool is better suited running alongside OCIM, or do you think we should run it external to OCIM?
I think that either way, we should tease away the need to login to VMs altogether to make ourselves less vulnerable - so if we make this a management command that uses the database for storage, should we also consider having this task be launchable via a web interface instead, and make sure OCIM will resume it in a worker post-restart?

Regardless of the approach, I think we should do it in a separate ticket. :)

bin/provision-instances
antoviaque: Yup - as mentioned above, all my comments can be handled in a follow-up iteration. And imho these iterations should be every time we need to redeploy the instances, rather than separate ones, to keep things concrete. The next iteration doesn't need to address all the comments either - again it can be iterative, and probably should be. Maybe just link to this PR in other tickets that will need to redeploy the instances, and to this comment for describing the approach?

CC @bradenmacdonald

bin/provision-instances
antoviaque: @bdero 

>  if we make this a management command that uses the database for storage, should we also consider having this task be launchable via a web interface instead, and make sure OCIM will resume it in a worker post-restart?

Exactly - I think eventually this should be a standard feature of OpenCraft IM, and making it launchable from the web interface, with automated resumes in workers post-restart seem great. The more automated and transparent we make this, the easier it will make our lives, and the more likely we are to be using it and be keeping our instances up to date. 

We might even find ways to automate the triggering of these reprovisionings along the way, to make it even more transparent (like we did for deprovisioning - we avoided having to care about it by automating deprovisioning of old VMs). Here for example, we could possibly detect when there has been code changes, either to the edX repositories the instances point to, or to the code of OpenCraft IM.

And yes, again, this can be done progressively/iteratively over the next tickets that need to reprovision instances.

bin/provision-instances
antoviaque: OK - I would still want to refactor this logic with the one identifying RC servers, so we can consistently identify RC servers, and whether the instance is ready to upgrade. It will also be helpful from the UI end to have this standardized.

bin/provision-instances
antoviaque: It's ephemeral, but worker queues are more to manage concurrency than state - the state itself could still be derived from the instances/appserver status. Ie if the last provisioning didn't complete and non is running at the moment, it means it needs to be restarted - I don't think there is a need to store this separately or at the worker queue level.

antoviaque:
> We may also want to add a way to upgrade instances before we spawn new appservers, a la https://github.com/open-craft/opencraft/blob/master/instance/instance_upgrade.py

+1 - this script also needs some love, and better integration with the rest of OpenCraft IM (ie the discussion above in https://github.com/open-craft/opencraft/pull/172#discussion_r97469041 about iteratively making this a core feature would also apply). The combination of both will eventually allow us to use OpenCraft IM to do continuous deployment, of both new edX releases/security patches, and individual custom developments - which would be really cool and useful to have : )

bin/provision-instances
antoviaque: @bdero All code files should have it - so the other scripts are to be updated too : )

bin/provision-instances
bradenmacdonald: I've linked to this PR from OC-2115, OC-2232, OC-2092.

bin/provision-instances
bdero: Ah, gotcha. Yeah, the whole reason most of this was necessary was to leave a couple of Huey consumers free as you mention. If there's a way to do task priority, I think that would remove the need to keep track of this stuff.
Bonus points if there's a way to have a low priority task not run until there's more than one consumer free.

smarnach:
I'm closing this PR in favour of the approach started in #186.

PR #174, opened 2017-01-25
Revert "Allows the OpenStack base image, flavor, and ssh_keyname to be configured per instance"

pomegranited:
Reverts open-craft/opencraft#169

Got error during migration on stage: 
```
django.db.utils.ProgrammingError: type "jsonb" does not exist
LINE 1: ...nstance" ADD COLUMN "openstack_server_base_image" jsonb NULL
```

Will investigate and fix tomorrow.


bradenmacdonald:
Argh, perhaps my advice led you astray - are we running an old version of postgres on stage/prod? Weird that it would be different than on the devstack...

pomegranited:
@bradenmacdonald Yep, stage and production are only running 9.3 😦 
```
ubuntu@postgres-stage:~$ psql --version
psql (PostgreSQL) 9.3.14

ubuntu@postgres:~$ psql --version
psql (PostgreSQL) 9.3.14
```

But the [Django documentation says](https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#django.contrib.postgres.fields.JSONField): "this field requires PostgreSQL ≥ 9.4 and Psycopg2 ≥ 2.5.4.".

I'm going to revert this change and change the JSON field type back.  Will upgrade postgres as a separate task.

PR #176, opened 2017-01-27
Mock interactions with the RabbitMQ API in tests

bdero:
On top of mocking `provision_rabbitmq`, I also went ahead and removed the `test_rabbitmq_access` test since it has no purpose if the API is mocked. 

**Testing instructions:**

1. Checkout this branch
1. Run the [previously failing tests](https://github.com/open-craft/opencraft/pull/152#issuecomment-274225119) to verify that they're passing now:

   ```
   honcho -e .env.test run coverage run --source='.' --omit='*/tests/*' ./manage.py test instance.tests.models.test_openedx_instance.OpenEdXInstanceTestCase.test_spawn_appserver_with_external_databases --noinput
   honcho -e .env.test run coverage run --source='.' --omit='*/tests/*' ./manage.py test instance.tests.models.test_openedx_database_mixins.RabbitMQInstanceTestCase --noinput
   ```

**Reviewers:**

- [ ] @pomegranited

### DIFF
--- a/instance/tests/models/test_openedx_database_mixins.py
+++ b/instance/tests/models/test_openedx_database_mixins.py
@@ -23,9 +23,11 @@ OpenEdXInstance Database Mixins - Tests
 # Imports #####################################################################
 
 import subprocess
+import urllib
 
+import ddt
 import pymongo
-import requests
+import responses
 import yaml
 from django.conf import settings
 from django.test.utils import override_settings
@@ -499,6 +501,7 @@ class MongoDBInstanceTestCase(TestCase):
         self.check_mongo_vars_not_set(appserver)
 
 
+@ddt.ddt
 class RabbitMQInstanceTestCase(TestCase):
     """
     Test cases for RabbitMQInstanceMixin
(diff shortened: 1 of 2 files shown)


instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -524,18 +525,16 @@ def request_overview():
...
+    @responses.activate
     def test_rabbitmq_request(self):
         """
         Test a simple get request with _rabbitmq_request
         """
+        responses.add(
+            method=responses.GET,
+            url='{}/api/overview'.format(settings.RABBITMQ_API_URL),
+            body='{}'
+        )
```
pomegranited: This test is no longer verifying anything useful, since it's just testing that the mocked empty `{}` body is the same as the other mocked empty `{}` body.

bdero: The test is intended to verify that _rabbitmq_request forms a request to `{{service}}/api/overview`. Do you think it would be more useful if I added something more identifying to the mocked body?

bdero: I'll refactor this test to better serve it's purpose of making sure `_rabbitmq_request` forms the correct URLs based on different parameters.

bdero:
Thanks for the review @pomegranited and for checking out the coverage! I refactored one of the tests to make it more meaningful and wrote a new one to bump the coverage up.

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -507,39 +511,78 @@ def setUp(self):
...
+            '{}/api/{}'.format(settings.RABBITMQ_API_URL, url)
+            for url in vhosts_calls + users_calls + permissions_calls
+        ]
+        deprovision_calls = [
+            '{}/api/{}'.format(settings.RABBITMQ_API_URL, url)
+            for url in vhosts_calls + users_calls
+        ]
+
+        # Spec the provisioning calls
+        with responses.RequestsMock() as rsps:
```
bdero: A note about how this works: The `responses.RequestsMock` raises an exception during context deconstruction if any of the URLs added to the responses object weren't ever called. Also, if any RabbitMQ API URLs are called that haven't been mocked, a `RabbitMQAPIError` will should be raised.

So, this test will only pass *if and only if* all of the mocked URLs are called during both provision and deprovision.

pomegranited: Nit: That's good to know, @bdero , and worth putting into a code comment.

instance/tests/models/test_openedx_database_mixins.py
pomegranited: 👍

instance/tests/models/test_openedx_database_mixins.py
bdero: Done!

bdero:
Thank you @pomegranited !

PR #178, opened 2017-02-12
Support setting the OpenStack region per appserver.

smarnach:
This change allows launching appservers in other regions than the default region.

### Testing instructions

Set `openstack_region` to a different region on the instance, then spawn a new appserver.  The region you choose needs to have the base image and SSH key needed for the instance.

### Author concerns

1. In line with the other OpenStack settings, I avoided storing the region on individual appservers.  See [my comment here](https://github.com/open-craft/opencraft/pull/169#discussion_r100689887).
1. The  OpenStack region is needed to create the Nova client, which is done in the `OpenStackServer` constructor.  This only works if the region is stored on the server as well.  The `OpenStackServer` is instantiated in the abstract base class `AppServer`, so the region needs to be accessible there.  This means I need to add the region to `Instance` instead of `OpenEdXInstance`.
1. I had to remove the declaration of base classes from the migration that creates the `OpenEdXInstance` model.  Django imports the current version of the base class and implicitly creates all fields in that current version, which defies the whole purpose of migrations as a record of historic versions of the model.  the `bases` parameter to `CreateModel` can take a string of the form `app_name.model_name` to retrieve a historic version of a model.  However, this only works for actual models, not for abstract models, so we can't do this in this case.  Simply removing the `bases` parameter seems to work fine though, so this is what I'm doing for now.  (This was rather difficult to debug, by the way, since there was no indication what was wrong.  The new migration applies fine to the current state of the database, but if you try to run all migrations on a pristine database, the last one fails with the error that the `openstack_region` field already exists.  But none of the earlier migrations contains a reference to this field, so it was completely unclear how that happened.)
1. Even though all lines of code I added are tested, the total coverage had dropped for reasons I don't understand, so I had to add some tests for completely unrelated code (and I found a minor bug when doing so).

### Reviewers

* [ ] @pomegranited


instance/migrations/0048_appserver_refactor1.py
```diff
@@ -101,7 +100,6 @@ class Migration(migrations.Migration):
             options={
                 'verbose_name': 'Open edX Instance',
             },
-            bases=(instance.models.instance.Instance, models.Model),
```
pomegranited: That's a gnarly one.. I had to do [something similar for a previous data migration](https://github.com/open-craft/opencraft/pull/175/commits/71e04d092512522279f7d838395ca4354cc4bdd5) in PR #175.

smarnach: @pomegranited Yeah, migrations getting retconned by later commits. :)

I wasn't surprised about your change, since it's documented that you need to access historical models in data migrations, and it was simply a bug in the migration code you fixed.  This case here, however, is a more fundamental design problem in Django migrations.  The original code was actually correct according to the documentation, but it didn't work anyway.

instance/migrations/0074_openstack_region.py
```diff
@@ -0,0 +1,28 @@
...
+
+    dependencies = [
+        ('instance', '0073_openstack_settings_data_defaults'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='openstack_region',
+            field=models.CharField(default=settings.OPENSTACK_REGION, max_length=16),
```
smarnach: @pomegranited The important bit is that you need to pass the `preserve_default=False` parameter to `AddField` to tell Django that the default value is only stored for the purpose of this migration.

instance/tests/models/test_openedx_secret_key_mixins.py
```diff
@@ -80,6 +81,16 @@ def test_secret_key_settings(self):
         for to_var, from_var in OPENEDX_SHARED_KEYS.items():
             self.assertEqual(secret_key_settings[to_var], secret_key_settings[from_var])
 
+    def test_secret_key_settings_no_key(self):
+        """
+        Test that secret key settings are empty if not master key is set.
```
smarnach: @pomegranited I must have been tired by the time I wrote this (or my English got impacted by writing too much code).

smarnach:
@pomegranited Thanks for testing this!  I'll fix it up as quickly as possible.

instance/models/server.py
```diff
@@ -260,14 +260,20 @@ class OpenStackServer(Server):
     """
     A Server VM hosted on an OpenStack cloud
     """
+    openstack_region = models.CharField(max_length=16, blank=False)
```
smarnach: @pomegranited We generally follow the pattern that we use the settings on the instance to configure the next appserver that gets spawned.  All settings get copied to the appserver and become immutable (at least in theory), so we have a history of the settings we used in the past.  So the obvious way to select the region for the next appserver would be to set the region on the instance and then spawn an appserver.  However, we departed from the usual pattern by not copying the OpenStack parameters to the appserver; hence the comment I linked in the author concerns.

instance/migrations/0048_appserver_refactor1.py
pomegranited: Yikes.. glad you found it!

instance/migrations/0074_openstack_region.py
pomegranited: Oh, that sounds familiar.. I think @haikuginger mentioned it on my original issue.

Thanks for the clarification!

instance/models/server.py
pomegranited: @smarnach We could mitigate that concern by showing the `server.openstack_region` in the GUI..  Though that was marked as "extra credit" for this issue, so is out of the required scope.

Once we get around to doing more GUI-related functionality updates, we can make these (and the other OpenStack) settings visible.

instance/openstack_utils.py
```diff
@@ -89,15 +89,14 @@ def get_openstack_connection():
...
-def sync_security_group_rules(security_group, rule_definitions, network=None):
+def sync_security_group_rules(security_group, rule_definitions, network):
     """
     Given an OpenStack 'SecurityGroup' instance and a list of rules (in the form
     of SecurityGroupRuleDefinition tuples), ensure that the security group's
     rules match the provided rules. Add/delete rules from the remote security
     group until it matches 'rules'.
     """
     assert all(isinstance(rule, SecurityGroupRuleDefinition) for rule in rule_definitions)
-    network = network or get_openstack_connection().network
```
pomegranited: Good way to deal with the potential ambiguity here.  Thanks @smarnach !

PR #181, opened 2017-02-16
Adds support for multiple active appservers

pomegranited:
Replaces the `OpenEdXInstance.active_appserver` field with a `OpenEdXAppServer.is_active` boolean, allowing more than one active appserver on an instance.

**Testing Instructions**

1. Using the current master branch, create a new instance, and spawn 3 appservers.
1. Activate the first appserver, and wait for the load balancer to be updated.  Ensure that requests are reaching the first appserver.
1. Switch to this branch, and run the migrations.
1. Via the OCIM GUI, ensure that only first appserver is still active.  Also check that the green tick mark is shown for the instance.
1. Activate the second appserver, and check that the GUI updates to add the second active server.
1. Wait for the load balancer to be updated.  Ensure that requests are reaching both appservers.
1. Activate the third appserver, and wait for the load balancer to be updated. 
1. Via the OCIM GUI, ensure that the green tick mark is shown for the instance.
1. Via the OCIM shell, run `instance.terminate_obsolete_appserver(days=0)`.  Ensure that no active appservers are terminated.
1. Via the OCIM GUI, deactivate one of the active appservers.  Ensure that the green tick mark is still shown for the instance.
1. Via the OCIM shell, run `instance.terminate_obsolete_appserver(days=0)`.  Ensure that only the inactive appserver is terminated.
1. Via the OCIM GUI, deactivate all but one of the active appservers.  Ensure that the "Deactivate appserver" button is disabled now that there's only one remaining active appserver.

**Reviewer**
- [ ] @smarnach


instance/models/openedx_instance.py
```diff
@@ -449,24 +451,27 @@ def require_user_creation_success(self):
...
         for appserver in self.appserver_set.all().order_by('-created'):
-            if appserver == active_appserver:
+            # Skip active appservers
+            if appserver.is_active:
                 continue
 
-            # Keep a running appserver as fallback for `days` after activation, to allow reverts
-            if active_appserver and appserver.created < active_appserver.last_activated:
-                if not sufficient_time_passed(active_appserver.last_activated, now, days) \
+            # Keep a running appserver as fallback for `days` after latest activation, to allow reverts
```
antoviaque: Don't we need to keep the previous set of appservers to allow reverts, rather than just one? Otherwise we'll end up in a situation where we can't revert, because there would only be one of the previous apservers left running, which couldn't be enough to handle the traffic and/or ensure the level of fault tolerance we are after.

Same remark about the rc server -- we'll probably want to provision multiple ones, and switch them all at once.

It sounds like appservers need to be grouped by release/version, ie we are missing a logical layer. This specific logic shouldn't have to change imho, and simply apply to a set of appservers meant to be activated together, rather to individual appservers.

bradenmacdonald: We could change the UI to have a "Spawn ___ AppServers" button, and every time you use that, it spawns N appservers and groups them together with an auto-generated "generation ID" or something?  Or the "N" here could be a property of the instance, which would be simpler.

More complex but better is to introduce an AppServerSet as another layer - likely out of scope for this though?

instance/migrations/0078_remove_openedxinstance_active_appserver.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('instance', '0077_populate_is_active'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='openedxinstance',
+            name='active_appserver',
+        ),
```
bradenmacdonald: Nit: If you want, you can combine these three migrations into a single migration with three `operations`

instance/static/html/instance/appserver.html
```diff
@@ -46,12 +46,19 @@
       </tbody>
     </table>
 
-    <button ng-disabled="is_active || !appserver.is_healthy"
-            tooltip="This will update the DNS entries so that this app server is used for the instance. Only one app server can be active at any time."
-            tooltip-placement="right"
+    <button ng-disabled="appserver.is_active || !appserver.is_healthy"
+            tooltip="This will add this app server to the list of the DNS entries for this instance."
```
bradenmacdonald: "This will add this app server to the load balancer pool"
(Generally, activating appservers should not involve DNS changes)

Same change below.

instance/static/js/src/openedx_appserver.js
```diff
@@ -66,6 +65,20 @@ app.controller("OpenEdXAppServerDetails", ['$scope', '$state', '$stateParams', '
...
         };
 
+        $scope.make_appserver_inactive = function() {
+            $scope.appserver.is_active = false; // Disable the button optimistically
+            OpenCraftAPI.one("openedx_appserver", $stateParams.appserverId).post('make_inactive').then(function() {
+                // Refresh the list of app servers in the instance scope, then refresh this appserver
+                $scope.$parent.refresh().then(function() {
+                    $scope.refresh();
+                });
+                $scope.notify($scope.appserver.name + ' is now inactive. The DNS changes may take a while to propagate.');
```
bradenmacdonald: There should not be DNS changes associated with inactivating appservers.

instance/models/openedx_instance.py
pomegranited: Or we could have a numeric setting on each instance indicating how many active servers are expected to be running at any given time.  Then we could use that as the maximum number of fallback/release candidate servers to be kept running.

instance/migrations/0078_remove_openedxinstance_active_appserver.py
pomegranited: Oo that's nicer..  07fc4e7 does this, and adds a data rollback migration.

instance/static/html/instance/appserver.html
pomegranited: Good point.  Addressed with dfc60d3.

instance/static/js/src/openedx_appserver.js
pomegranited: Addressed with dfc60d3.

instance/models/openedx_instance.py
antoviaque: > More complex but better is to introduce an AppServerSet as another layer - likely out of scope for this though?

Yes, this seem like the right approach to me.

> We could change the UI to have a "Spawn ___ AppServers" button, and every time you use that, it spawns N appservers and groups them together with an auto-generated "generation ID" or something? Or the "N" here could be a property of the instance, which would be simpler.

It has the advantage of still grouping the servers logically and might be smaller. But the gain in implementation time now might be repaid by making the logic to handle groups of servers more complex, since it wouldn't be encapsulated by a dedicated class. So I'm not a huge fan, but given the urgency context, that might still be a reasonable middle ground - but imho we would need a proper follow-up story to clean it up and implement a proper AppServerSet class.

> Or we could have a numeric setting on each instance indicating how many active servers are expected to be running at any given time. Then we could use that as the maximum number of fallback/release candidate servers to be kept running.

I'd rather have an explicit grouping of servers. This would likely make the logic even more complex, and increase the odds of subtle issues - for example if multiple server creations are requested in a short amount of time, or when the number of servers on the instance changes.

instance/models/openedx_instance.py
```diff
@@ -314,24 +325,36 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
         The triggered_by_instance flag indicates whether the reconfiguration was initiated by this
         instance, in which case we log additional information.
         """
-        if not self.active_appserver:
+        active_appservers = self.get_active_appservers()
+        if not active_appservers.count():
```
smarnach: The most efficient way to find out whether a queryset is empty is to use `.exists()`.  However, since we need to retrieve all appservers further down anyway, we should force the evaluation of the queryset above:

    active_appservers = list(self.get_active_appservers())

This way, we can make do with a single query.  We can check whether the list is empty using `if not active_appserver`, and neither this nor the itertation further down requires any further database queries.

instance/models/openedx_instance.py
```diff
@@ -314,24 +325,36 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
             self.logger.error(
-                "Active appserver does not have a public IP address. This should not happen. "
+                "No active appservers found with public IP addresses.  This should not happen. "
                 "Deconfiguring the load balancer backend."
             )
             return [], []
-        backend_name = "be-{}".format(self.active_appserver.server.name)
-        server_name = "appserver-{}".format(self.active_appserver.pk)
+
+        backend_name = "be-{}".format(self.domain_slug)
```
smarnach: Thanks, this looks like a more reasonable choice for the backend name. :)

instance/models/openedx_instance.py
```diff
@@ -314,24 +325,36 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
         template = loader.get_template("instance/haproxy/openedx.conf")
-        config = template.render(dict(
-            domain=self.domain,
-            http_auth_info_base64=self.http_auth_info_base64(),
-            server_name=server_name,
-            ip_address=ip_address,
-        ))
+        config = template.render(template_vars)
```
smarnach: An alternative here is to only build the list of appservers in a separate variable, and then pass all context variables in here without first storing them in a separate variable `template_vars`.  I don't know how to write this in less awkward English, so I'll try Python instead:

    appserver_vars = []
    for appserver in active_appservers:
         # populate appserver_vars
    […]
    config = template.render(dict(
        domain = self.domain,
        http_auth_info_base64=self.http_auth_info_base64(),
        appservers=appserver_vars,
    ))

I tend to think of context vars as parameters passed to the template, so I like to list them explicitly at the point where the template is rendered.

This is just a style nit, of course, so it's up to you what you prefer.

instance/models/openedx_instance.py
smarnach: I don't really like the `AppServerSet` approach, since it is likely to add a lot of complexity and seems rather inflexible.  We will need the ability to choose the appserver set when launching a new appserver and to move appservers between appserver sets, among other things.  I would like to think about this a bit more before making a decision here.

smarnach:
@pomegranited With these changes applied, activating a new appserver and deactivating the old one become two separate steps, each triggering a reconfiguration of the LB, which seems unnecessary.  This problem is related to the discussion on "appserver sets", and might go away once we have agreed on a solution there.

instance/migrations/0076_openedxappserver__is_active.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
bradenmacdonald: @smarnach Ah, good to know. @pomegranited Sorry for the bad advice then!

instance/migrations/0076_openedxappserver__is_active.py
```diff
@@ -0,0 +1,56 @@
...
+        ('instance', '0075_remove_protocol'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='openedxappserver',
+            name='_is_active',
+            field=models.BooleanField(db_column='is_active', default=False),
+        ),
+        # ref http://stackoverflow.com/a/39541048/4302112
```
pomegranited: Yep, I got the dreaded `OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events`, but flanking the data migration with those `CONSTRAINTS` commands made it run fine, forwards and backwards (though of course, you can only keep the most recently activated appserver when undoing the migration.)

instance/migrations/0076_openedxappserver__is_active.py
pomegranited: Is ok.. I worked around it (see [comments](https://github.com/open-craft/opencraft/pull/181/files/d27df1bfc8b646eaf47644d7b8af9bc6b467257f#r101799806)).

instance/migrations/0076_openedxappserver__is_active.py
```diff
@@ -0,0 +1,56 @@
...
+from django.db import migrations, models
+
+
+def populate_is_active(apps, schema_editor):
+    '''
+    Mark all OpenEdXInstance.active_appserver with OpenEdXAppServer.is_active = True
+    '''
+
+    OpenEdXInstance = apps.get_model("instance", "OpenEdXInstance")
+    for instance in OpenEdXInstance.objects.exclude(active_appserver__isnull=True):
```
pomegranited: There's so many gotchas with data migrations, I didn't think about the multiple database issue.  Thanks for pointing this out, @smarnach! 

But I'm inclined to leave it be. When we add multiple databases, we'll have to fix all our data migrations.

instance/management/commands/activity_csv.py
```diff
@@ -70,13 +70,19 @@ def handle(self, *args, **options):
...
+        active_appservers = {}
+        for appserver in OpenEdXAppServer.objects.filter(_is_active=True):
+            public_ip = appserver.server.public_ip
+            if public_ip is not None:
+                active_appservers[public_ip] = appserver.instance
+        return active_appservers
+
+    def activity_csv(self, out):
+        """Generate the activity CSV."""
+        active_appservers = self.get_active_appservers()
```
pomegranited: @antoviaque We do output the "Internal LMS Domain", which if you sort on that column, will group your appservers.

Alternately, I can sort the `active_appserver` list by `instance.id` when [iterating through the appserver `public_ip` address](https://github.com/open-craft/opencraft/pull/181/files/d27df1bfc8b646eaf47644d7b8af9bc6b467257f#diff-827bdb29a85e840ec5396164d71be8ddR128).

instance/models/openedx_instance.py
pomegranited: Your way is clearer, and makes for a smaller diff, always to be desired 😄 

Done with b958d89.

instance/models/openedx_instance.py
pomegranited: Thanks @smarnach , addressed by 6b22fb8.

instance/models/openedx_instance.py
pomegranited: @smarnach Nice that backend_name was the only bit of appserver-specific information in the haproxy config too.. easy to factor out!

pomegranited:
@smarnach Yes, I don't like that the load balancer is getting reconfigured all the time either..  Especially since it's so slow to happen.

I was wondering if instead, we could batch the de-/activations by having a cron'd huey task that checks a flag to see if the load balancer needs to be reconfigured, then unsets the flag and runs the playbook if it does.

pomegranited:
@antoviaque  Re grouping appserver sets: can we achieve the same goal with something like what @smarnach added for tagging instances in https://github.com/open-craft/opencraft/pull/183 ?

pomegranited:
Rebased on master (91a8fbf) from [b958d89](https://github.com/open-craft/opencraft/commits/69f558b980c9092f14c18aa5bf802cc89ee3e0eb).

bradenmacdonald:
> @antoviaque Re grouping appserver sets: can we achieve the same goal with something like what @smarnach added for tagging instances in #183 ?

That sounds kind of like my suggestion:

> We could change the UI to have a "Spawn N AppServers" button, and every time you use that, it spawns N appservers and groups them together with an auto-generated "generation ID" or something? 

BTW, I would love if we could have a "reason" field, where users (or bots) must enter a string describing the reason that they are spawning appservers before they click the button, and the appservers would be annotated accordingly - would make it much easier to know what's going on.

instance/tests/integration/integration_instance.py
```diff
@@ -163,6 +163,22 @@ def assert_mongo_db_provisioned(self, instance):
             self.assertTrue(result)
             db.logout()
 
+    @shard(0)
```
pomegranited: I'm using `shard 0` here, but if we can add another container, then I can use `shard 4` instead.

Am a little worried about the timing of this one though..  if we end up having to provision 4 appservers total, then it may timeout.

instance/tests/integration/integration_instance.py
```diff
@@ -263,6 +247,24 @@ def test_activity_csv(self):
         # stdout should contain 3 lines (as opposed to 2) to account for the last newline.
         self.assertEqual(len(out_lines), 3)
 
+    ''' FIXME JV : Need to add an additional test container?
```
pomegranited: I tried using `shard 0` here instead, but it disrupted the other integration tests.

So 68bb9a8 comments out this test - @smarnach Do I have to need another container to the CircleCI project?

instance/tests/integration/integration_instance.py
```diff
@@ -244,6 +247,24 @@ def test_activity_csv(self):
...
+    @shard(4)
+    def test_activate_multiple_appservers(self):
+        """
+        Provision an instance, spawn and activate two AppServers
+        """
+        OpenEdXInstanceFactory(name='Integration - test_activate_multiple_appservers')
```
bradenmacdonald: We should be using `, use_ephemeral_databases=False` for all the integration tests other than the ephemeral test (the first one).

instance/tests/integration/integration_instance.py
bradenmacdonald: One option to free up space is to remove the code for ephemeral databases and remove the ephemeral integration test? Otherwise yeah for now we'd need to add another container I think.

instance/tests/integration/integration_instance.py
pomegranited: Ah, good point, thanks @bradenmacdonald !  Fixed with 66b4f33.

instance/tests/integration/integration_instance.py
pomegranited: @bradenmacdonald I took your suggestion with 66b4f33 and replaced the single VM ephemeral database test with a multi-vm, persistent database test.

instance/management/commands/activity_csv.py
antoviaque: That's fine yep.

I also note that this is again an area where we actually want to look at an AppServerSet globally, ie whether there is one available. The number of active AppServers within the set doesn't really matter for that report.

antoviaque:
For the AppServerSet discussion, I replied in OC-1965 since the discussions were continuing there.

instance/management/commands/activity_csv.py
smarnach: @antoviaque I don't think an appserver set matters here in any way.  We just want to look at all the active appservers.  We don't care whether they are grouped in any way.

instance/instance_upgrade.py
```diff
@@ -55,11 +56,12 @@ def get_instances_to_upgrade(self):
...
         """
-        return OpenEdXInstance.objects.filter(  # Select instances
-            active_appserver___status="running",  # that are running
+        return OpenEdXInstance.objects.filter(Q(  # Select distinct instances
+            Q(ref_set__openedxappserver_set___is_active=True,  # with active appserver(s)
+              ref_set__openedxappserver_set___status="running"),  # that are running
             openedx_release__contains=self.INITIAL_RELEASE,  # on {INITIAL_INSTANCE}
             use_ephemeral_databases=False,  # and use persistent databases.
-        )
+        )).distinct()
```
smarnach: Did you have a look at the resulting SQL?  I can't tell from looking at this code whether this would match instances that have a running appserver and an active appserver, but the active appserver is not running.  I'm also not sure it matters. :)

instance/models/openedx_instance.py
```diff
@@ -314,23 +325,35 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+        appserver_vars = []
+        for appserver in active_appservers:
+            server_name = "appserver-{}".format(appserver.pk)
+            ip_address = appserver.server.public_ip
+            if ip_address:
+                appserver_vars.append(dict(ip_address=ip_address, name=server_name))
+            else:
+                appserver.logger.error(
+                    "Active appserver does not have a public IP address. This should not happen."
+                )
```
smarnach: We are will have to change how this case is handled to fix the problem we experienced yesterday, but this is out of scope for the current PR.

instance/static/html/instance/appserver.html
```diff
@@ -47,11 +47,18 @@
...
     <button ng-disabled="is_active || !appserver.is_healthy"
-            tooltip="This will update the DNS entries so that this app server is used for the instance. Only one app server can be active at any time."
-            tooltip-placement="right"
+            tooltip="This will add this app server to the load balancer pool for this instance."
+            tooltip-placement="below"
             ng-click="make_appserver_active()">
       Activate this app server
     </button>
+
+    <button ng-disabled="!is_active || instance.active_appservers.length < 2"
```
smarnach: We could allow disabling the last appserver as well to effectively take an instance offline, which is sometimes useful.  However, the potential damage from allowing this is probably bigger than what we gain, since we can always use the Django console to deactivate the last appserver.

instance/static/html/instance/index.html
```diff
@@ -13,25 +13,25 @@
       <li ng-repeat="instance in instanceList | filter: {is_shut_down: false}"
           ng-class="{active: instance.id == state.params.instanceId}"
           ng-init="
-              active_appserver = instance.active_appserver;
-              newer_appserver = (instance.newest_appserver && instance.newest_appserver.id != active_appserver.id ? instance.newest_appserver : null);
+              newer_appserver = (instance.newest_appserver && !instance.newest_appserver.is_active ?  instance.newest_appserver : null);
           ">
-        <!-- Is there an app server newer than the active app server? -->
+        <!-- Is there an app server newer, inactive app server? -->
```
smarnach: I don't understand the new version of this comment.

instance/static/js/src/openedx_appserver.js
```diff
@@ -59,13 +59,27 @@ app.controller("OpenEdXAppServerDetails", ['$scope', '$state', '$stateParams', '
...
                 });
-                $scope.notify($scope.appserver.name + ' is now active. The DNS changes may take a while to propagate.');
+                $scope.notify($scope.appserver.name + ' is now active. The load balancer changes will take a short while to propagate.');
             }, function() {
                 $scope.refresh();
                 $scope.notify('An error occurred. ' + $scope.appserver.name + ' could not be made active.', 'alert');
             });
         };
 
+        $scope.make_appserver_inactive = function() {
```
smarnach: This function is almost identical to `make_appserver_active`.  Not sure whether it makes sense to factor out the common bits or whether the code becomes too awkward if we do.  You probably already thought about this, so what do you think?

instance/tasks.py
```diff
@@ -59,21 +59,19 @@ def spawn_appserver(instance_ref_id, mark_active_on_success=False, num_attempts=
         appserver_id = instance.spawn_appserver()
         if appserver_id:
             if mark_active_on_success:
-                # If the AppServer provisioned successfully, make it the active one:
-                # Note: if I call spawn_appserver() twice, and the second one provisions sooner, the first one may then
-                # finish and replace the second as the active server. We are not really worried about that for now.
-                instance.set_appserver_active(appserver_id)
+                # If the AppServer provisioned successfully, make it active.
+                appserver_make_active(appserver_id)
```
smarnach: This changes the semantics of `mark_active_on_success`; the new appserver will become active in addition to the old one.  This doesn't seem particularly useful, so we might want to remove the `mark_active_on_success` flag.

instance/static/html/instance/appserver.html
pomegranited: @smarnach  Yep, and I think we should use `instance.shut_down()` anyway, so that the monitoring is removed.  That's worth having as a button on the instance, with a big "ARE YOU SURE?" alert.

instance/static/html/instance/index.html
pomegranited: Total typo on my part.. fixed with 49796ed

instance/static/js/src/openedx_appserver.js
pomegranited: The code didn't end up being too complex when I combined them, so have committed that with 49796ed.

instance/tasks.py
pomegranited: @smarnach  I disagree..  Currently, we only use the `mark_active_on_success` flag when creating appservers for PRs, and it's useful for that.

For GUI actions, we have to explicitly activate the new appserver, and so will have to remember to deactivate the old one too, once this change goes through.

And anyone who activates an appserver from the shell will have to do the same, so..  Is it really that confusing?  If it is, then we should probably add a field to the instance indicating whether multiple VMs are even possible for this instance, and maintain the current "activating an appserver means deactivating the previously active appserver" behaviour.

instance/instance_upgrade.py
pomegranited: Good question... it looks like we're ok:

```sql
SELECT DISTINCT "instance_openedxinstance"."id", "instance_openedxinstance"."openstack_region", "instance_openedxinstance"."load_balancing_server_id", "instance_openedxinstance"."mysql_server_id", "instance_openedxinstance"."mysql_user", "instance_openedxinstance"."mysql_pass", "instance_openedxinstance"."mysql_provisioned", "instance_openedxinstance"."mongodb_server_id", "instance_openedxinstance"."mongo_user", "instance_openedxinstance"."mongo_pass", "instance_openedxinstance"."mongo_provisioned", "instance_openedxinstance"."rabbitmq_vhost", "instance_openedxinstance"."rabbitmq_provider_user_id", "instance_openedxinstance"."rabbitmq_consumer_user_id", "instance_openedxinstance"."rabbitmq_provisioned", "instance_openedxinstance"."swift_openstack_user", "instance_openedxinstance"."swift_openstack_password", "instance_openedxinstance"."swift_openstack_tenant", "instance_openedxinstance"."swift_openstack_auth_url", "instance_openedxinstance"."swift_openstack_region", "instance_openedxinstance"."swift_provisioned", "instance_openedxinstance"."s3_access_key", "instance_openedxinstance"."s3_secret_access_key", "instance_openedxinstance"."s3_bucket_name", "instance_openedxinstance"."secret_key_b64encoded", "instance_openedxinstance"."email", "instance_openedxinstance"."configuration_source_repo_url", "instance_openedxinstance"."configuration_version", "instance_openedxinstance"."configuration_extra_settings", "instance_openedxinstance"."edx_platform_repository_url", "instance_openedxinstance"."edx_platform_commit", "instance_openedxinstance"."openedx_release", "instance_openedxinstance"."use_ephemeral_databases", "instance_openedxinstance"."github_admin_organizations", "instance_openedxinstance"."github_admin_users", "instance_openedxinstance"."additional_security_groups", "instance_openedxinstance"."internal_lms_domain", "instance_openedxinstance"."internal_lms_preview_domain", "instance_openedxinstance"."internal_studio_domain", "instance_openedxinstance"."external_lms_domain", "instance_openedxinstance"."external_lms_preview_domain", "instance_openedxinstance"."external_studio_domain", "instance_openedxinstance"."successfully_provisioned", "instance_openedxinstance"."openstack_server_flavor", "instance_openedxinstance"."openstack_server_base_image", "instance_openedxinstance"."openstack_server_ssh_keyname" 
FROM "instance_openedxinstance" 
INNER JOIN "instance_instancereference" ON ("instance_openedxinstance"."id" = "instance_instancereference"."instance_id" AND ("instance_instancereference"."instance_type_id" = 12)) 
INNER JOIN "instance_openedxappserver" ON ("instance_instancereference"."id" = "instance_openedxappserver"."owner_id") 
WHERE ("instance_openedxappserver"."status" = running 
  AND "instance_openedxappserver"."is_active" = True 
  AND "instance_openedxinstance"."openedx_release"::text LIKE %QWERTY% AND "instance_openedxinstance"."use_ephemeral_databases" = False)
```

instance/tests/integration/integration_instance.py
```diff
@@ -164,15 +164,14 @@ def assert_mongo_db_provisioned(self, instance):
             db.logout()
 
     @shard(1)
-    def test_activate_multiple_appservers(self):
```
pomegranited: Had to remove the multi-vm integration tests, because [CircleCI was timing out](https://circleci.com/gh/open-craft/opencraft/2050?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

pomegranited:
@smarnach  Rebased from 45287cd

instance/instance_upgrade.py
smarnach: @pomegranited Thanks, looks good.

instance/tasks.py
smarnach: @pomegranited You are right, for the first appserver of an instance this still makes sense.

smarnach:
@pomegranited I managed to spawn some appservers from my devstack so I could test this.  Everything went fine after I managed to convinve Chrome to load the new versions of the HTML templates.  I first didn't notice Chrome still had the old templates, which resulted in really weird behaviour (e.g. the "Activate this appserver" button actually deactivated appservers).

I only noticed a minor UI issue.  Instances that don't have an active appserver currently show the first of the two circles next to the instance name as white and empty.  In the new version, the first circle is always shown in read in this case, even for new instances that don't have any appservers.  I find the old behaviour more intuitive in this case.

Since you already addressed all other issues, this is the only point that remains, so hopefully we can get this merged today. :)

pomegranited:
@smarnach Ah gotcha!  Previously, the left icon represented the health of the "active appserver", and was `None` if there wasn't one.  Now that we can have more than one active appserver, I changed the health icon to mark the health of the the "instance", i.e. *all* active appservers.  So if one is unhealthy, then the whole instance is flagged red.

But I'd neglected the null case, where there are no active appservers.  [Fixed here](https://github.com/open-craft/opencraft/pull/181/commits/ea9c7dffd0e8594cc9f1b5e10f4139fdae471a5c#diff-3f368ccab30f190362c50833335da158R74) (and [tests fixed here](https://github.com/open-craft/opencraft/pull/181/commits/ea9c7dffd0e8594cc9f1b5e10f4139fdae471a5c#diff-f4e70bc775b5f6220042336a641255cdR55)).

I've squashed my commits in preparation for merge..  does it look ok now?

smarnach:
@pomegranited I've tested the fix for the status icons, and all looks good now, thanks!

:+1:

- [x] I tested this: I followed the testing instructions, and additionally verified the state in the database manually after each step.
- [x] I read through the code
- [x] I don't think this requires an update to the README file.

PR #183, opened 2017-02-16
Add ability to add custom tags to instances.

smarnach:
This change add the ability to attach custom tags to instances.

The tags could be used in various ways:
* Marking certain commonly used subsets of instances, e.g. a tag "beta-test" or "sandbox" or "maintained".  This would make it easy to find all instances of certain types.
* Temporarily tagging instances for redeployment.  E.g. if we want to migrate a set of instances to a different database server, we could apply a tag "to-be-migrated-to-xyz" to them and remove the tag for each instance that has been successfully migrated.  This is the motivating use case (based on [this discussion](https://github.com/open-craft/opencraft/pull/172/files#r97255606)).

Instances tags should eventually be shown in the UI, and it should be possible to filter the instance list by tag.

The PR is very basic in its current form, but sufficient to be useful in the Django shell.  It's also minimally invasive; if we decide that the tags aren't that useful after all at some point, we can remove them again without any side effects.

Reviewers:
- @itsjeyd 
- @antoviaque 
- @bradenmacdonald


bradenmacdonald:
Big 👍 from me for this, especially for use with reprovisioning. I'd also like to see a description/notes field on the Instance or AppServer that can be updated from the UI, but that can come later.

antoviaque:
@smarnach Nice! Thanks for doing this, and great idea. :+1:

instance/models/instance.py
```diff
@@ -218,3 +223,14 @@ def delete(self, *args, **kwargs):
         if not kwargs.pop('ref_already_deleted', False):
             self.ref.delete(instance_already_deleted=True)
         super().delete(*args, **kwargs)
+
+
+class InstanceTag(ValidateModelMixin, models.Model):
+    """
+    Custom tags that can be applied to instances, for filtering and annotation.
+    """
+    name = models.SlugField(max_length=100, blank=False, unique=True)
```
itsjeyd: @smarnach Nit: Unless `SlugField` has a different [default value](https://docs.djangoproject.com/en/1.10/ref/models/fields/#blank) for `blank`, you can remove `blank=False`. (The [migration code](https://github.com/open-craft/opencraft/pull/183/files#diff-a6517275f4d31d4b15a295862ee21a86R20) doesn't have it, so I don't think the default is different.)

smarnach: @itsjeyd I just couldn't remember the default value, and couldn't be bothered to look it up.  I'll remove it.

PR #185, opened 2017-02-23
Use standard Django field default values.

smarnach:
This PR removes all `set_field_default()` methods and uses standard Django ways of setting default values instead.  This fixes the unintentionally changing secret keys, but also makes similar bugs less likely in the future, since the Django logic to set default values is less likely to be buggy.

Another advantage of the change is that the code to set default values moves closer to the place where the fields are actually defined.

The main reason we introduced the `set_field_default()` methods is to be able to use default value from settings.  Simply using `default=settings.WHATEVER` does not work, since Django generates a new migration everytime the setting changes.  This PR works around this problem by using a callable returning the desired setting as default value.  Even if the default value changes, the callable will still be the same, so Django does not create bogus migrations.  Since Django needs to be able to serialize the callable, there are certain restrictions on what callables we can use, so we need to define a few top-level  functions, but overall it's not too bad.

In cases where default values are computed based on the values of other fields, I've moved the initialization code to the `save()` method.  We were using this pattern in a few places already.


smarnach:
@antoviaque @bradenmacdonald @itsjeyd This PR is a bit more comprehensive than the fix we discussed on the ticket.  I haven't fixed the tests yet and haven't tested the code, but I'd like to get some early feedback on whether you agree with this approach before investing more time in it.  I personally think it's an improvement over the current situation.

instance/models/load_balancer.py
```diff
@@ -103,20 +111,16 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
     accepts_new_backends = models.BooleanField(default=True)
     # A random postfix appended to the haproxy configuration file names to make collisions
     # impossible.
-    fragment_name_postfix = models.CharField(max_length=8, blank=True)
+    fragment_name_postfix = models.CharField(
+        max_length=8,
+        blank=True,
+        default=functools.partial(generate_fragment_name, length=8),
```
bradenmacdonald: Nit: Since `generate_fragment_name` isn't used elsewhere you could just make `length=8` the default parameter value, and then you don't need to import functools nor apply a partial.

instance/models/load_balancer.py
```diff
@@ -103,20 +111,16 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
...
+    )
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.logger = ModelLoggerAdapter(logger, {'obj': self})
 
-    def set_field_defaults(self):
-        if not self.fragment_name_postfix:
-            # Set a unique fragment_name_postfix to avoid clashes between multiple instance
-            # managers sharing the same load balancer.
```
bradenmacdonald: This deleted comment explicitly says "between multiple instance managers sharing the same load balancer." which explains the rationale for this code more clearly than the comments you left in, which just say e.g. "A random postfix appended to the haproxy configuration file names to make collisions impossible."

Please edit the comment above `fragment_name_postfix = models.CharField` and/or the docstring of `generate_fragment_name` to include that more explicit rationale (unless it's no longer relevant).

instance/models/mixins/database.py
```diff
@@ -122,18 +125,47 @@ def _drop_user(cursor, user):
...
     )
 
-    mysql_user = models.CharField(max_length=16, blank=True)  # 16 chars is mysql maximum
-    mysql_pass = models.CharField(max_length=32, blank=True)
+    mysql_user = models.CharField(
+        max_length=16,  # 16 chars is mysql maximum
+        # Note that the maximum length for the name of a MySQL user is 16 characters.
+        # But since we add suffixes to mysql_user to generate unique user names
+        # for different services (e.g. xqueue) we don't want to use the maximum length here.
+        default=functools.partial(get_random_string, length=6, allowed_chars=string.ascii_lowercase),
```
bradenmacdonald: Is there a reason that a partial works better or is preferred to 
```python
default=lambda: get_random_string(length=6, allowed_chars=string.ascii_lowercase)
````
? I personally find the latter more readable.

File: instance/models/utils.py [hunk 4]
```diff
@@ -52,6 +55,25 @@ def format_help_text(help_text):
...
+    return getattr(settings, name)
+
+
+def default_setting(name):
+    """
+    Return a callable that returns the Django setting with the given name.
+
+    This is useful for default values in model fields.
+    """
+    return functools.partial(_get_setting, name)
```
bradenmacdonald: Within the body of this function, what do you think about adding
```python
assert hasattr(settings, name)
```
? 

This seems like a nice opportunity to add a check at app init time that all the referenced settings strings are valid.

bradenmacdonald: Also I would personally prefer just `return lambda: getattr(settings, name)` but I know pylint disagrees.

instance/models/load_balancer.py
```diff
@@ -90,6 +91,13 @@ class ReconfigurationFailed(Exception):
     """Exception indicating that reconfiguring the load balancer failed."""
 
 
+def generate_fragment_name(length):
+    """
+    Helper function to the the default value of the field `fragment_name_postfix`.
```
itsjeyd: @smarnach Typo: `the the`, and I think this is missing a verb: `Helper function to <verb> the default value`.

instance/models/mixins/rabbitmq.py
```diff
@@ -23,15 +23,45 @@
...
 from django.conf import settings
-from django.db import models
+from django.db import IntegrityError, models
+from django.utils.crypto import get_random_string
+import requests
 
 from instance.models.rabbitmq import RabbitMQUser
 
 
+# Fucntions ###################################################################
```
itsjeyd: @smarnach Typo: `Fucntions`

opencraft/settings.py
```diff
@@ -292,6 +292,7 @@
 
 # Default github repository to pull code from
 DEFAULT_FORK = env('DEFAULT_FORK', default='edx/edx-platform')
+DEFAULT_EDX_PLATFORM_REPO_URL = 'https://github.com/{}.git'.format(DEFAULT_FORK)
```
itsjeyd: @smarnach Happy to see this definition move to the settings file! It felt out of place in the previous location.

instance/models/load_balancer.py
smarnach: @bradenmacdonald The 8 passed in as `length` parameter should match the `max_length` above.  I intentionally did it this way so that the repetition is at least contained in three lines of code.

instance/models/mixins/database.py
smarnach: @bradenmacdonald Yes, there is a reason.  The partial object can be serialised by Django for the migration, while the lambda can't.  Look at the generated migration to better understand this.

(That said, I personally prefer `partial()` here, but we don't have a choice anyway.)

instance/models/utils.py
smarnach: @bradenmacdonald I've added the assertion, but can't use a lambda for the same reason as above.  We could make pylint happy by returning a named closure instead of a lambda, but Django would not be able to serialize this.

antoviaque:
@smarnach Thanks for this - and yup, this also looks like a better approach to me.

That actually makes me wonder if there is a standard best practice for this? This seem like such a common pattern.

smarnach:
@antoviaque I also thought this should be a pretty common problem, but I couldn't find any satisfactory solution with Google, so I had to make up one myself.

Django can only serialise `functools.partial()` objects since version 1.9.  Before that, you needed to create one callable for each setting you wanted to use as a field default value.

smarnach:
@itsjeyd I've finished this PR up and tested it a bit more.  There are no specific testing instructions, since this PR does not modify any specific functionality.  It's important to test that the migration works as expected.  I also ran the development server and deployed and instance.  You can also try to smoke test some parts of the IM, as you deem necessary.

instance/models/openedx_appserver.py
```diff
@@ -35,22 +35,33 @@
...
 
-# Models ######################################################################
 
+# Functions ###################################################################
+
+def default_admin_organizations():
+    """
+    Helper function to set the default value for the field `github_admin_organizations`.
+    """
+    default = settings.DEFAULT_ADMIN_ORGANIZATION
```
smarnach: @itsjeyd We set default values for all settings in `opencraft/settings.py` (and even if we didn't, adding and removing settings from the environment can't really affect whether a Django setting exists).

instance/tests/models/test_database_server.py
```diff
@@ -333,20 +340,6 @@ def test__create_default(self):
...
-    def test__create_default_no_default_settings(self):
-        """
-        Test that `_create_default` does not create database server if default settings missing.
-        """
-        MySQLServer.objects._create_default()
-        MongoDBServer.objects._create_default()
-
-        self.assertEqual(MySQLServer.objects.count(), 0)
-        self.assertEqual(MongoDBServer.objects.count(), 0)
-
```
smarnach: You are right, this is a bit half-baked.  I initially planned to get rid of the option to set the DB servers to `None` altogether, since the default function is already evaluated during the migrations, before you get a chance to manually add an entry.  I then changed the default value function to handle the case that the DB server is unset gracefully, but forgot to add this test back.

However, thinking about it, the current approach is still wrong, since it never errors out, even when it should.  I think I will remove support for not setting default DB servers again, since we eventually want to get rid of ephemeral DB support anyway.  It's the easiest way forward, and I don't think we lose much.

instance/models/openedx_appserver.py
itsjeyd: @smarnach Ah, right -- duh :)

instance/tests/models/test_database_server.py
itsjeyd: @smarnach That makes sense, thanks for the explanation!

itsjeyd:
@smarnach The most recent commit looks good to me 👍

PR #186, opened 2017-03-01
Facilitate bulk redeployments from the Django shell.

smarnach:
In order to be able to run bulk deployments of appservers from the Django shell without congesting the Huey queue for periodic tasks and deployments triggered from the UI, this PR introduces a separate Huey queue for tasks started from the Django shell.  This makes it possible to redeploy instances in bulk using simple loops in the shell.

Eventually we would like to be able to trigger bulk deployments from the GUI.  Currently, the Django integration of Huey only supports a single queue to be configured per process, but it would be rather easy to change this.

To be able to mark the instances that failed or succeeded, the `spawn_appserver()` task got additional arguments for tags to apply in either case.

With these small changes applied, the redeployment of all instances tagged by `tag` could be executed like this:

    success_tag = InstanceTag.objects.create(name=tag.name + "-succeeded")
    failure_tag = InstanceTag.objects.create(name=tag.name + "-failed")
    for instance in tag.openedxinstance_set.iterator():
        spawn_appserver(instance.ref.pk, success_tag=success_tag, failure_tag=failure_tag)

The failed instances can be retried using this code

    for instance in failure_tag.openedxinstance_set.iterator():
        spawn_appserver(instance.ref.pk, success_tag=success_tag, failure_tag=failure_tag)

We can or course add convenience functions to make running this code even more convenient.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,5 @@
 web: gunicorn opencraft.wsgi --timeout 60 --workers 4 --log-file -
 websocket: python3 websocket.py
 worker: python3 manage.py run_huey --no-periodic
+worker_low_priority: HUEY_QUEUE_NAME=opencraft_low_priority python3 manage.py run_huey --no-periodic
 periodic: python3 manage.py run_huey --workers=0
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@
 
 # Config ######################################################################
 
-WORKERS = 4
+WORKERS = 3
+WORKERS_LOW_PRIORITY = 3
 SHELL = /bin/bash
 HONCHO_MANAGE := honcho run python3 manage.py
 RUN_JS_TESTS := xvfb-run --auto-servernum jasmine-ci --logs --browser firefox
@@ -78,13 +79,13 @@ migration_autogen: clean
 	$(HONCHO_MANAGE) makemigrations
 
 run: clean migration_check collectstatic
-	honcho start --concurrency "worker=$(WORKERS)"
+	honcho start --concurrency "worker=$(WORKERS),worker_low_priority=$(WORKERS_LOW_PRIORITY)"
 
 rundev: clean migration_check static_external
 	honcho start -f Procfile.dev
 
 shell:
-	$(HONCHO_MANAGE) shell_plus
+	HUEY_QUEUE_NAME=opencraft_low_priority $(HONCHO_MANAGE) shell_plus
 
 upgrade_dependencies:
 	pip freeze --local | grep -v '^\-e' | cut -d = -f 1  | xargs -n1 pip install -U
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,4 @@
 web: python3 manage.py runserver_plus
 websocket: python3 websocket.py
 worker: python3 manage.py run_huey --no-periodic
+worker_low_priority: HUEY_QUEUE_NAME=opencraft_low_priority python3 manage.py run_huey --no-periodic
--- a/README.md
+++ b/README.md
@@ -201,6 +201,9 @@ edit its security group rules to only allow access to VMs in the
 * `REDIS_URL`: (default: `redis://localhost:6379/`)
 * `HUEY_ALWAYS_EAGER`: Set to True to run huey tasks synchronously, in the web
   process. Use in development only (default: False)
+* `HUEY_QUEUE_NAME`: The name of the Huey task queue.  This setting can be used
+  to run multiple separate worker queues, e.g. one for the web server and one
+  for batch jobs started from the Django shell.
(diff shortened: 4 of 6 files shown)


smarnach:
@bdero @bradenmacdonald I'd appreciate an early review of this PR.  It provides an alternative to #172, using a much simpler approach.  Do you think this approach covers our use cases?  Is there any feature in #172 that can't be easily implemented based on this approach?

Procfile
```diff
@@ -1,4 +1,5 @@
 web: gunicorn opencraft.wsgi --timeout 60 --workers 4 --log-file -
 websocket: python3 websocket.py
 worker: python3 manage.py run_huey --no-periodic
+worker_shell: HUEY_QUEUE_NAME=opencraft_django_shell python3 manage.py run_huey --no-periodic
```
bradenmacdonald: How about just calling this "worker_low_priority"? Because you've said that you want to add the ability to trigger this from the GUI.

bradenmacdonald:
@smarnach This approach is great, and is a nice example of composing existing features to get the functionality we want. +1 for the approach.

bdero:
@smarnach Yeah, I very much prefer this kind of approach!.

A couple of thoughts/questions:
- I can imagine calculating progress would be as easy as "instances that have been tagged so far / total size of instance set", right?
- Is it possible to fetch which instances are currently queued for being provisioned in a huey queue? I think this will come in handy, but I had trouble figuring it out. For example, we'd probably want to persist this information between OCIM restarts so that the provisioning queue will resume correctly; I don't think we'd be able to just really rely on something like `set(all_instances).symmetric_difference(set(tagged_instances))` since some ***new*** instances may have been provisioned since the start of it. This was one of the big reasons I ended up tracking it externally.
  - Actually, I think a pretty simple solution to this (when we come to it) would just be to add instance tags that indicate they're "queued" - if there's no easy way to grab this info from huey, that is. :)
- Would it be possible to fetch which instances are currently being processed by huey workers? I also couldn't find a way to reasonably detect when a task died, so I had to rely on when an appserver was last provisioned + a timeout based on timestamps.
  - Again, a simple solution for this might be, if we ever come to it, to make something like an `instance_task` decorator that:
    - Before calling the function, adds an instance tag to indicate "processing".
    - Applies passes the function to db_task and runs.
    - After calling the function, remove the instance tag.

smarnach:
@bdero Thanks for your comments.
> Currently, it doesn't look like this will survive OCIM/huey restarts in general, so we'll still need to address that at some point since these full reprovisions can take over a day (to make it fully unattended + not interrupt other people's work).

Restarting the IM while a provisioning is running currently isn't handled gracefully, but this seems to be a separate issue (that should be addressed).  The queued tasks are stored in Redis, and that queue is persistent, so after a restart any remaining tasks will be picked up.

As a side note, since the instance manager is running on a bigger machine now, we have 24 workers, and I was able to run 20 deployments in parallel, which significantly reduced the time it took.

> I can imagine calculating progress would be as easy as "instances that have been tagged so far / total size of instance set", right?

Yes, something along these lines.  We can wrap the whole thing in a class with a method to query the progress.

> Is it possible to fetch which instances are currently queued for being provisioned in a huey queue?

It's certainly possible to use the Huey API to retrieve a list of queued tasks from Redis, but I agree with you:

> Actually, I think a pretty simple solution to this (when we come to it) would just be to add instance tags that indicate they're "queued". :)

> Would it be possible to fetch which instances are currently being processed by huey workers?

It should again be possible to get this information from the Redis queue, but I don't think it's necessary.  I'd simply iterate until all instance have the "succeeded" tag to keep things simple.

smarnach:
@bdero I've addressed Braden's comment, squashed the commits and tested this manually.  Can you please take another look?  To test, you can run any of the commands given in the description, and verify that tasks scheduled from the shell are executed by the low-priority worker.

I don't have time to address any of your comments in the scope of the current task, but we can improve on this basis in later iterations.  Your comments will be a helpful guide for improvements.

bdero:
Sounds good, @smarnach. Let's get this merged once the tests pass! :+1:

- [x] I tested this: I placed this change on stage, restarted the console + django shell, and ran spawn_appserver over a new instance, which started on `worker_low_priority.3`. 
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ N/A.
- [x] Includes documentation - Good doc strings, but I'm not sure anything else is necessary for this right now.

smarnach:
After rebasing, it took almost six hours until the CircleCI tests passed.  The first two test runs experienced some internal errors (probably problems on CircleCI's part) and were automatically retried.  This isn't the first time this happens.  We definitely need to bring the build time down.

antoviaque:
@smarnach Nice approach : )

PR #187, opened 2017-03-09
Move OpenStack server settings to OpenEdXAppConfiguration.

smarnach:
This PR moves the OpenStack server settings from `OpenEdXInstance` to `OpenEdXAppConfiguration`.  The latter is an abstract model that both `OpenEdXInstance` and `OpenEdXAppServer` derive from, and the [code to create a new app server automatically copies over all settings from the instance](https://github.com/open-craft/opencraft/blob/48ff5b6511a473c44c405b1df9216aa2d91bc1df/instance/models/openedx_instance.py#L403-L416).

This change makes sure that each appserver keeps a record of the OpenStack settings it actually used.

The migration has been handcrafted to first introduce the fields in `OpenEdXAppServer` with empty default values, and then immediately modify the fields with the actual default from the settings.  This makes sure that existing app servers get empty default values instead of using the current value from the settings, which might not match the settings that were in effect at the time the appserver was created.  It is better to have no value at all for historical appservers than to have wrong values.

For testing, run the migration and make sure old app servers don't get any values set for the new fields.  Than create a new app server for any instance and make sure the settings get copied over correctly.


mtyaka:
This looks good! There are two pylint errors and several tests failed, but I think that's just OVH being unreliable. Once that's fixed this is good to go.

:+1:

- [x] I tested this: Ran the migration & checked that old appservers' settings remained empty. Spawned a new appserver and verified that the instance's settings were correctly copied to the new appserver.
- [x] I read through the code

mtyaka:
At some point it might be helpful to show the values of this settings in the UI.

smarnach:
@mtyaka Thanks for the review, and for reminding me to show the settings in the UI.

I wonder how I missed the pylint violations, since I did run prospector locally and it passed.  It's still in my terminal history.

smarnach:
@mtyaka I've fixed pylint and added the settings to the UI.  Can you please take another quick look?  I've verified that the settings show up in the UI correctly in my devstack.

mtyaka:
@smarnach I checked, it looks good! 👍

PR #189, opened 2017-03-11
Optimizations to instance list (OC-2122)

bradenmacdonald:
This implements [OC-2122](https://tasks.opencraft.com/browse/OC-2122) to improve the usability of the instance manager UI and API.

Major changes:
1. Before this PR, the `/api/v1/instances/` endpoint would require (5 + 10*n) SQL queries to return the instance list when there were n instances. After this PR, that has been reduced to 9 queries regardless of the number of instances. See the new `test_instance_list_efficiency()` test for how that is verified.
1. As discussed in the OC-2122 ticket, I have converted the `is_shut_down` method (which was fairly expensive) to a database field called `is_archived`. The meaning of archived is meant to be more clear as well: an archived instance is no longer meant to be online, and all associated resources should be freed, but its data and history should be preserved.
1. When clicking on an AppServer in the UI, the logs will now not be loaded unless you open the panel that contains the AppServer logs. In addition, the "Status" panel is now expanded by default.

screen shot 2017-03-10 at 10 27 26 pm

Minor changes:
1. I had an AppServer on my vagrant instance where the VM was `terminated` but the AppServer was stuck in `ConfiguringServer`. When I called `terminate_vm()` on it, I was surprised that nothing happened - so I fixed that method to update the AppServer status appropriately.
1. The UI did not show whether or not RabbitMQ was provisioned for an instance, as it does for MySQL, MongoDB, and Swift. Now it shows the RabbitMQ provisioning status too.

Notes:

Since `is_archived` is now a database field and `is_shut_down()` has been deleted, the data migration uses a heuristic technique to fill in `is_archived` values which may not perfectly match the old behavior of `is_shut_down()`. As a result, it's possible there are some corner cases where instances that were previously shown as shut down do not appear as archived or vice versa. I do not anticipate that being a significant problem though.


smarnach:
@bradenmacdonald Oh, great, I'm glad you did this!  This should hit production as soon as possible.

instance/models/openedx_instance.py
```diff
@@ -164,6 +163,12 @@ def get_active_appservers(self):
         """
         Returns a queryset containing the active appservers.
         """
+        if hasattr(self.ref, '_cached_active_appservers'):
+            # A database optimization like prefetch_related() hascomputed the active
+            # appservers for a large number of instances and cached that result on
+            # the InstanceReference of each one.
+            # (This is used to optimize the /api/v1/instances/ endpoint query for example)
+            return self.ref._cached_active_appservers
```
smarnach: Do we need to delete the cached appservers in any situatiion?  E.g. in the `OpenEdXAppServer.make_active()` method.

instance/serializers/openedx_instance.py
```diff
@@ -75,8 +75,17 @@ def to_representation(self, obj):
...
             output['is_steady'] = None
 
-        newest_appserver = obj.appserver_set.order_by('-created').first()
-        output['newest_appserver'] = AppServerBasicSerializer(newest_appserver, context=self.context).data
+        try:
+            # Note that appservers are ordered by '-created' by default.
+            # We don't change or check the ordering of the .appserver_set.all() queryset here
+            # because that causes the django ORM to force a new database query to be made
+            # for each instance here, even if appserver_set was previously cached.
+            assert OpenEdXAppServer._meta.ordering[0] == '-created'
```
smarnach: It would feel more natural to me to check this i a test rather than in the actual code.

instance/serializers/openedx_instance.py
```diff
@@ -75,8 +75,17 @@ def to_representation(self, obj):
...
+            newest_appserver = obj.appserver_set.all()[0]
```
smarnach: Any reason not to use `first()` here?

Edit: I see, it's probably to get an error if it does not exist.  Fair enough.  (You could also use `first()` and check the result.)

instance/static/html/instance/index.html
```diff
@@ -10,7 +10,7 @@
   <!-- Instances List -->
   <div class="large-3 columns">
     <ul class="side-nav side-instance-list">
-      <li ng-repeat="instance in instanceList | filter: {is_shut_down: false}"
+      <li ng-repeat="instance in instanceList | filter: {is_archived: false}"
```
smarnach: I think we should filter on the server side.  Archived instance aren't used in any way on the client side, so there is no point in querying and transferring them.

instance/models/instance.py
```diff
@@ -119,11 +123,13 @@ class Instance(ValidateModelMixin, models.Model):
     # in a query, e.g. to do .select_related('ref_set')
     ref_set = GenericRelation(InstanceReference, content_type_field='instance_type', object_id_field='instance_id')
     openstack_region = models.CharField(
+        # TODO: Move this to OpenedXInstance
         max_length=16,
         blank=False,
         default=default_setting('OPENSTACK_REGION'),
     )
     tags = models.ManyToManyField(
+        # TODO: Move this to InstanceReference (common fields should be in non-abstract tables where possible)
```
smarnach: Why?  I thought about whether to put this on `InstanceReference` or `Instance`, and I couldn't find strong reasons either way.  In the end I chose the way that requires less typing and indirection in the Django shell.

bradenmacdonald: To me, it's a question of data modelling and normalization. If you put `tags` on the abstract model, and we introduce other `Instance` subclasses (e.g. "DaliteInstance"), then each subclass will also need to specify `tags = models.ManyToManyField` which is not DRY, and there will be additional many-to-many intermediate tables in the data model. So if we have additional Instance types and you wanted to query something like "how many Instances are tagged as `sandbox` vs `production`?", you would have to write some code to iterate over all the concrete `Instance` subclasses, and then to separately query the `tags` intermediate table for each one, and then combine the results.

On the other hand, if we put `tags` on the `InstanceReference`, then there is only ever one intermediate table created in the DB to implement the many-to-many relation, and it can be queried efficiently regardless of the number of Instance variations we may have in the future.

To achieve less typing and indirection in the django shell, you can just add an accessor to the `Instance` abstract class:

```python
@property
def tags(self):
    return self.ref.tags
```

instance/static/html/instance/index.html
bradenmacdonald: I considered that but decided against it for now because I think it's better to have a more flexible API, and we may have some other API consumer in the future that wants to be able to know about archived instances.

On second thought, now that `is_archived` is a database table, my preference would be to modify the query used by our frontend UI to append `?is_archived=true` to the query string (or make that the default, but preserve the possibility of using the API to list archived instances, just as you can use the API to get the details of an archived instance if you want to).

instance/serializers/openedx_instance.py
bradenmacdonald: Yes, there is a reason: If we use `first()`, then the SQL of the query would be changed to include `LIMIT 1`, and django's ORM would issue a new query, rather than re-using the cached results already available in `obj.appserver_set`

instance/models/openedx_instance.py
bradenmacdonald: Cache invalidation is certainly a potential issue here, but so far the only time this cache is used is when retrieving the list of appservers, which is not an operation that modifies any models. I also think this is not a big "landmine" because `_cached_active_appservers` can only ever be populated very explicitly and deliberately by assigning some result to that specific field using `prefetch_related` with extra arguments.

instance/models/openedx_instance.py
```diff
@@ -164,6 +163,12 @@ def get_active_appservers(self):
         """
         Returns a queryset containing the active appservers.
         """
+        if hasattr(self.ref, '_cached_active_appservers'):
+            # A database optimization like prefetch_related() hascomputed the active
```
bradenmacdonald: Note to self: Fix "hascomputed"

instance/tests/api/test_openedx_instance.py
```diff
@@ -98,6 +98,31 @@ def assert_active_appserver(self, instance_id, appserver_id):
...
+
+        queries_per_api_call = 9
+
+        for num_instances in range(1, 4):
+            self.add_active_appserver(sub_domain='api{}'.format(num_instances))
+            try:
+                with self.assertNumQueries(queries_per_api_call):
+                    response = self.api_client.get('/api/v1/instance/')
+            except AssertionError:
+                # The above assertion error alone won't indicate the value of i at the time the assertion was
```
bradenmacdonald: Note to self: `i` is no longer used in this loop.

instance/models/openedx_appserver.py
```diff
@@ -222,7 +222,7 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
         'configuration_extra_settings',
     ]
 
-    class Meta:
+    class Meta(AppServer.Meta):
```
bradenmacdonald: There is magic to prevent that.

This does inherit the `ordering` though, which was the point of this change.

instance/models/instance.py
```diff
@@ -119,11 +123,13 @@ class Instance(ValidateModelMixin, models.Model):
     # in a query, e.g. to do .select_related('ref_set')
     ref_set = GenericRelation(InstanceReference, content_type_field='instance_type', object_id_field='instance_id')
     openstack_region = models.CharField(
+        # TODO: Move this to OpenedXInstance
```
bradenmacdonald: I think some larger refactoring is in order. e.g. Move server creation elsewhere (it must be made more generic anyways in the future to support multiple OpenStack accounts and/or non-OpenStack clouds like AWS/GCP).

In any case, even with the current design it could be possible for the `Instance` class to do something like `create_server(..., **self.get_subclass_instance().get_server_type_arguments())`

instance/api/instance.py
```diff
@@ -56,11 +60,20 @@ class InstanceViewSet(viewsets.ReadOnlyModelViewSet):
...
+    def list(self, request):
         """
-        Return the basic serializer for the list action, and the detailed serializer otherwise.
+        List all instances.
         """
-        if self.action == 'list':
-            return InstanceReferenceBasicSerializer
-        return InstanceReferenceDetailedSerializer
+        queryset = self.queryset.prefetch_related(
+            # Use prefetching to make the instance listing always an O(1) database query.
```
bradenmacdonald: Right, fair point.

instance/models/instance.py
smarnach: @bradenmacdonald I can follow some of the arguments, but not all of them.  We would not need to repeat the field on different instance types, since they will all inherit it from `Instance`, so I don't see a DRY violation.  Querying tags across different instance types feels rather contrived to me; I don't think we will ever use tags this way.  And regarding the indirection, I was mostly worried about the reverse relation:  Retrieving all instances with a given tag.  Currently, that's simply

    tag.openedxinstance_set.all()

If we move the field to the instance reference, the only reasonable way to do this is this

    OpenEdXInstance.objects.filter(ref_set__tags=tag)

That's not too bad, but it will actually become less efficient due to the additional join, and this is in my opinion the most common use case.

That said, I'm not really opposed to moving the tags to the instance reference.  I just don't see strong arguments either way.

bradenmacdonald: Argh, you're right, disregard what I said about the `DRY` issue - in my haste to reply I confused which model this was being added to.

> Querying tags across different instance types feels rather contrived to me

I actually considered making `is_archived` simply an `archived` tag, which would be an example of this kind of thing that applies to all instance types.

> That said, I'm not really opposed to moving the tags to the instance reference. I just don't see strong arguments either way.

OK, I'm fine to leave this for now and see how we tend to use these things in practice. Later on, we'll know more about what our priorities are. I'll remove my opinionated comment.

smarnach: @bradenmacdonald Yeah, let's see how the refactorings proposed in the app server set design doc pan out, and move the tags when needed.  Shouldn't be a big deal anyway.

I would refrain from uisng tags for any kind of core functionality (whatever "core" means, but you might get the idea).

instance/static/html/instance/index.html
bradenmacdonald: I have changed the API to return only non-archived instances by default, but to still allow listing archived instances if desired.

bradenmacdonald:
@smarnach I've just pushed two commits which address all your comments and also add that index to the logs table. Ready for another review :)

instance/static/html/instance/index.html
smarnach: Thanks for this change.  I think it will make a significantly speed things up.  The serialised data for our production instance is more than 1K per instance, so including archived instances results in a 300K response everytime the list is updated.  The database queries will also be significantly faster if we first filter out everything we don't need.

smarnach:
:+1: once the build is green (one Jasmine test failed).

- [x] I tested this: looked at the SQL queries for retrieving the instance list, tested the lazy log loading, ran the migrations and verified they did something reasonable, terminated an app server stuck in "configuring" state and a few other things
- [x] I read through the code
- [x] Includes documentation

bradenmacdonald:
Thanks @smarnach! Merging now. I'm not going to squash because my commits are nicely organized to clarify each change, and getting another green build will take a long time.

instance/tests/api/test_openedx_instance.py
```diff
@@ -98,6 +98,48 @@ def assert_active_appserver(self, instance_id, appserver_id):
             )
         return response
 
+    def test_instance_list_efficiency(self):
+        """
+        The number of database queries required to fetch /api/v1/instance/
```
e-kolpakov: @bradenmacdonald Let me be an annoyance a little bit and state that this is not the right way to test that:

* We actually don't care how much queries are exactly performed.
* The test is likely quite fragile, since any change in the Django/OCIM implementation might cause it to break.

What we actually care is the speed of loading - to assert that properly we'd better use  proper load test instead.

This is just me being pedantic - as it is written now is definitely better than having nothing at all and implementing proper load testing might be too expensive at this point (but we might consider this in future)

bradenmacdonald: @e-kolpakov The main point of this test is *not* to test how many queries are performed when the API is used. The main point of this test is within the `for` loop, which tests that the number of queries stays the same even as the number of instances increases. That is something that we do care about, and it is proportional to the loading speed, as Brandon determined during testing. If the number of queries increases with the number of instances, this API will get slower unnecessarily as the number of instances increases*.

Also, if you look at the queries that are being performed, they are very specific to the overall model we are using, rather than the details that change more often - it's something like 1 query for `InstanceReference`, 1 query for `OpenEdXInstance`, 1 query for `OpenEdXAppServer`, and 1 query for active `OpenEdXAppServer`s (the other queries are standard django things for the current user, session, and/or `content_types`). That basic structure doesn't change frequently.

If we do make changes to the core model in the future, it's totally fine to change `queries_per_api_call = 9` here to some new number - the exact value doesn't matter so much (although I do think it's good to know when it does change). What's most important is that we don't change any other parts of this test.

If we do find that we have to change `queries_per_api_call` a lot, it probably means we have some weird implementation of something that we should fix - alternatively we can modify this test to detect `queries_per_api_call` automatically on the first iteration and then ensure it doesn't change on the 2-3 iterations of the loop - without caring what the exact number is.


\* (unless of course we had some particularly bad data model and queries where the few "optimized" queries have like an O(n³) complexity and the many "unoptimized" queries that get made per instance are only O(√n))

e-kolpakov: @bradenmacdonald Again, I'm just being pedantic here, so I agree this is one way to test it. My point is kind of two fold:

* I'd rather say it asserts that no matter how many instances there are, querying for list of instances takes 9 queries.
* How much queries certain code path performs requires a bit too deep knowledge about implementation than test should have. (in my opinion)

> modify this test to detect queries_per_api_call automatically on the first iteration and then ensure it doesn't change on the 2-3 iterations of the loop

Might actually be a better way.

PR #190, opened 2017-03-24
Add option to set DNS records on individual active VMs.

smarnach:
For testing, spawn two app servers, activate them one by one and verify the created DNS records.


instance/models/openedx_instance.py
```diff
@@ -80,7 +81,6 @@ class OpenEdXInstance(LoadBalancedInstance, OpenEdXAppConfiguration, OpenEdXData
     external_lms_domain = models.CharField(max_length=100, blank=True)
     external_lms_preview_domain = models.CharField(max_length=100, blank=True)
     external_studio_domain = models.CharField(max_length=100, blank=True)
-
```
smarnach: This change was unintentional.

instance/tests/models/test_openedx_instance.py
```diff
@@ -847,3 +848,38 @@ def test_shut_down_reminder(self):
         msg = r"Use archive\(\) to shut down all of an instances app servers and remove it from the instance list."
         with self.assertRaisesRegex(AttributeError, msg):
             instance.shut_down()
+
+    def _verify_vm_dns_records(self, domain, active_vm_ips):
+        """
+        Helper function to verify a set of DNS records for active VMs.
+        """
+        dns_ips = set()
+        vm_indexes = set()
```
smarnach: @itsjeyd Both indexes and indices are standard plural forms, but for different meanings of the word.  In this case, it's definitely indices, but for e.g. a database index it's indexes.

instance/tests/models/test_openedx_instance.py
```diff
@@ -847,3 +848,38 @@ def test_shut_down_reminder(self):
...
+        Test that DNS records for active app servers get set.
+        """
+        instance = OpenEdXInstanceFactory(internal_lms_domain='vm-dns.test.com',
+                                          enable_vm_dns_records=True)
+        instance.spawn_appserver()
+        appserver1 = instance.appserver_set.first()
+        self._verify_vm_dns_records('test.com', [])
+        appserver1.make_active()
+
+        mocks.os_server_manager.set_os_server_attributes('server2', _loaded=True, status='ACTIVE')
```
itsjeyd: @smarnach I see, thanks for the explanation.

instance/models/openedx_appserver.py
```diff
@@ -176,6 +176,10 @@ class Meta:
             "behind a firewall. (In the django admin, separate group names with a comma.)"
         )
     )
+    enable_vm_dns_records = models.BooleanField(
+        default=False,
+        help_text="Set additional DNS records pointing directly to the active app servers."
+    )
```
antoviaque: @smarnach Why making it an option, especially defaulting to off? This seem like a useful feature for all our instances?

smarnach: @antoviaque Two reasons.  First, as a security precaution.  Currently, we don't leak the IP addresses of the backends to the general public, and we are still sorting out some possible security issues around that.  There is no vulnerability we are aware of since we enable HTTP auth, but we are going to disable it again and replace it by something else.  It felt safer to me not to leak the all backend IPs until this is sorted out.  And second, there currently is no clean-up code, so with the current process of first activating a new app server and then deactivating the old one, we will always end up with _two_ DNS records instead of just one.

Once those two issues are sorted out, we can enable this for all instances and remove the flag.

In general, I like to introduce new features gently to reduce the risk.

antoviaque: @smarnach The reason I ask is that when there is an issue with an instance, it would gain time to be able to directly SSH a predictible domain name, rather than having to go through the Instance manager to find the right appserver and its IP. When I was trying to figure out why Cloudera was down, on my phone in the bus at 5am last week-end, I would have really appreciated it : )

smarnach: @antoviaque I completely understand, and as I said above, I expect that we will enable this by default in the future.  However, given the tight timebox of the ticket this was developed on, I preferred to play it safe for now.

antoviaque: @smarnach OK - can you put a follow-up story to remove the option and make this the standard behavior? Also worth checking with me in the future before putting features behind configuration options or flag - having this on all instances was part of the expected outcomes of that task, and one of the reasons the support request from Harvard was addressed this way.

Thanks!

antoviaque: @smarnach True, re-reading the story I see how the fact that it was discussing the specific case of Harvard's instance could have made you consider it as a special case. I was describing a generic feature using their instance as an example, but this wasn't explicit.

Thanks for the PR! And I think we can leave the DNS records setup as a TODO in the code for now, no need for a story yet - if the behavior does become problematic we can revisit, but unless we end up hitting a cap on the domain names we can set in Gandi, it is harmless.

PR #191, opened 2017-03-29
Remove XQueue version pinning.

smarnach:
This was only required for the Eucalyptus release to include [this commit](https://github.com/open-craft/xqueue/commit/7fbb72461b2f9c88f787ddc5ca51dda855735dd4).  Since this is included in the upstream Ficus release, we don't need to pin this anymore.


itsjeyd:
@smarnach This looks good to me. We'll test it when deploying new app servers for Cloudera :+1:

PR #194, opened 2017-04-07
Allow setting additional monitoring emails per instance (OC-2496)

bradenmacdonald:
This PR introduces a new per-instance setting, `additional_monitoring_emails`, which can be used to define extra emails (in addition to those in `settings.ADMINS`) who should receive email alerts from New Relic for that instance. The setting can be set via the shell or the django admin, and it appears in both the API and the UI.

**Test Instructions**:

1. Log in to New Relic, select the OpenCraft.Hosting account, go to Account Settings, API Keys, and create/note your "Admin user's API key" (we do *not* need the "REST API key" even though we are using a REST API).
1. In your OC IM devstack `.env` file, configure the following settings (substitute the real values):
    ```
    NEWRELIC_ADMIN_USER_API_KEY='your_admin_api_key_here'
    ADMINS = '[["Your Name", "your_email@opencraft.com"]]'
    ```
1. On your OC IM devstack, have at least one instance that is configured but offline. It does not need to have any associated online VMs, so there's no need to go through the provisioning process.
1. Go the the django admin, find that instance, and set its "Additional monitoring emails" to "your_email+somethingelse@opencraft.com"
1. In the UI, ensure that additional email is shown on the instance's "Settings" tab (you'll need to force your browser to ignore its cache of angular HTML templates)
1. On your devstack, run `make shell`
1. In the shell, find the instance by its ID, e.g. `inst = InstanceReference.objects.get(pk=50).instance`
1. Enable monitoring with `inst.enable_monitoring()`
1. Within a few minutes, you should get three emails from New Relic indicating that the instance's three URLs are offline. Verify that each email was sent to both your_email@opencraft.com (from `ADMINS`) and your_email+somethingelse@opencraft.com (from "Additional monitoring emails")
1. Run `inst.disable_monitoring()` to delete the monitors

**Other Notes**: I didn't have time during this task, but at some point we should upgrade to the New Relic Synthetics API v3 instead of v1, which is deprecated. The parts of the API that we use appear to be unchanged, but more thorough testing would be required to say for sure.

**FYI**: API documentation: https://docs.newrelic.com/docs/synthetics/new-relic-synthetics/synthetics-api/synthetics-rest-api-version-1


instance/models/mixins/openedx_monitoring.py
```diff
@@ -42,28 +42,34 @@ def enable_monitoring(self):
...
-        for monitor in self.new_relic_availability_monitors.exclude(
-                pk__in=already_enabled_ids):
-            monitor.delete()
+        self.logger.info('Checking New Relic Synthetics monitors')
+
+        already_monitored_urls = set()
+
+        for monitor in self.new_relic_availability_monitors.all():
+            url = newrelic.get_synthetics_monitor(monitor.pk)['uri']
+            if url in self._urls_to_monitor:
```
smarnach: The set `_urls_to_monitor` is a property that gets evaluated in every call (and all its entries also come from evaluating properties, which depend on further properties).  In this case it does not really matter, but in general I would prefer to store this to a local variabl once and reuse it.  Keeping a bit of hygiene in this regard helps avoiding lots of performance problems in the long run.

instance/models/mixins/openedx_monitoring.py
```diff
@@ -42,28 +42,34 @@ def enable_monitoring(self):
...
+        # We add emails here but never remove them - that must be done manually (or the monitor deleted)
+        # in order to reduce the chance of bugs or misconfigurations accidentally supressing monitors.
+        emails_to_monitor = set([email for name, email in settings.ADMINS] + self.additional_monitoring_emails)
+        if emails_to_monitor:
+            for monitor in self.new_relic_availability_monitors.all():
+                emails_current = set(newrelic.get_synthetics_notification_emails(monitor.id))
+                emails_to_add = list(emails_to_monitor - emails_current)
+                if emails_to_add:
+                    self.logger.info('Adding email(s) to monitor %s: %s', monitor.id, ', '.join(emails_to_add))
+                    newrelic.add_synthetics_email_alerts(monitor.id, emails_to_add)
```
smarnach: This code won't ever remove email addresses from monitors.  Is this intentional, e.g. to allow configuring additional email addresses via the NewRelic web interface?

I'd tend to let OCIM "own" the monitor configuration, and allow it to remove email addresses again if we remove them from the OCIM config.

instance/models/openedx_appserver.py
```diff
@@ -176,6 +176,16 @@ class Meta:
             "behind a firewall. (In the django admin, separate group names with a comma.)"
         )
     )
+    additional_monitoring_emails = ArrayField(
```
smarnach: At some point we should consolidate how we represent list of strings in our models.  It's a bit all over the place at the moment.  For the list LMS users, we use a many-to-many relation (which is probably fine), for the github admin users we use a `JSONField`, and for this field we use an `ArrayField`.

instance/tests/test_newrelic.py
```diff
@@ -42,47 +42,28 @@ class NewRelicTestCase(TestCase):
     @responses.activate
     def test_get_synthetics_monitors(self):
```
smarnach: Since you renamed the function, it makes sense to rename this test function as well (i.e. remove the trailing "s").

smarnach:
:+1:

- [x] I tested this: I followed the testing instructions, looked at the API calls and the resulting database objects.
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~
- [x] Includes documentation: The newly introduced configuration setting is documented in its help string.

instance/models/mixins/openedx_monitoring.py
bradenmacdonald: Good point - I'll do that.

instance/models/mixins/openedx_monitoring.py
bradenmacdonald: Yes, I intentionally did not implement any way to automatically remove email addresses from monitors - see the comment block a few lines above (62-64). And yes, in part, that would prevent the unwanted removal of emails that were added via the web interface, even though that's not an encouraged way to configure monitors.

I think it will be pretty rare that we'll remove an email address from a monitor, but keep the monitor online. We can revisit this in the future if we find that that would be helpful, but for now I believe this is a reasonable and safe way forward.

instance/models/openedx_appserver.py
bradenmacdonald: IMHO a Foreign Key / m2m makes sense whenever we're referencing any data that's in the database and an `ArrayField` makes the most sense in other cases for simple string lists (or other lists, since it's a fully-typed list field so you can do lists of integers etc. and the typing is enforced by the DB and converted correctly to python - a big advantage over JSONField). `ArrayField` was probably not available in Django at the time the GitHub admin users list was added.

When we upgrade Postgres we should convert all the JSONFields to official django postgres `jsonb` fields and/or `ArrayField` as appropriate.

instance/tests/test_newrelic.py
bradenmacdonald: Yes, I simply missed that trailing `s`. Thanks.

PR #197, opened 2017-04-20
Update integration tests to use Ficus

haikuginger:
This pull request updates our integration tests to use Ficus, rather than Eucalyptus, which no longer works due to missing dependencies.

### DIFF
--- a/playbooks/collect_activity/collect_activity.yml
+++ b/playbooks/collect_activity/collect_activity.yml
@@ -15,13 +15,15 @@
       copy:
         src: stats.py
         dest: /tmp/collect_activity.py
-      become_user: edxapp
+      become: yes
+      become_user: www-data
 
     - name: Run stats script
       command: >
         /edx/bin/python.edxapp /tmp/collect_activity.py {{ inventory_hostname }} {{ remote_output_filename }}
       register: stats_results
-      become_user: edxapp
+      become: yes
+      become_user: www-data
 
     # For debugging purposes:
     #   stdout contains the computed stats
--- a/playbooks/collect_activity/collect_activity.yml
+++ b/playbooks/collect_activity/collect_activity.yml
@@ -15,13 +15,13 @@
       copy:
         src: stats.py
         dest: /tmp/collect_activity.py
-      become: yes
+      become_user: edxapp
 
     - name: Run stats script
       command: >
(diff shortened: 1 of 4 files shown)


haikuginger:
@smarnach, tests are passing!

playbooks/collect_activity/collect_activity.yml
```diff
@@ -15,13 +15,13 @@
       copy:
         src: stats.py
         dest: /tmp/collect_activity.py
-      become_user: www-data
+      become: yes
```
smarnach: So you are running the collect_activity script as root now?  Why is this necessary?

@bdero Would you have any concerns about this change?

haikuginger: It looks like www-data doesn't have even read access for any of the requisite files anymore, and I figured that, since this is just a test, it was a better choice to just use a user that'll always have the necessary privileges than to spend time figuring out who has those privileges now, to be further updated as needed.

bdero: @smarnach Yes, we should definitely try to keep these scripts running with the same user as the LMS/CMS uses at runtime.

`collect_activity.py` initializes Django in the LMS/CMS environment, and my main concern is that doing this as root may cause subtle issues down the line that are difficult to track down if, for example, the Django setup process caches anything to disk (which I suspect does happen, including with python bytecode caching, etc.), or if edX decides to add anything to the setup process that could result in files being written.

haikuginger: All right; let me see if this works when run as `edxapp`. :)

haikuginger:
After some digging, located this thread:

https://github.com/ansible/ansible/issues/14635

It looks like `become_user` does not automatically set `become: yes`. So, the fact that testing this command worked previously was likely a total fluke; our default SSH user used to have access to the directories in question, but no longer does. Adding `become: yes` to the existing `become_user: www-data` should fix it.

playbooks/collect_activity/collect_activity.yml
smarnach: @haikuginger Re "since this is just a test": it isn't just a test.  The playbook is used for production instances; otherwise I wouldn't be that worried.

smarnach:
@haikuginger Thanks for figuring this out!  The changes look good to me, :+1:.  I'm merging and deploying this now, so I can rebase #198 and get its build passing.

bdero:
@haikuginger Ah, thanks for tracking that down!

PR #198, opened 2017-04-21
Store public IP addresses in the database

smarnach:
This PR contains several vaguely related changes:

1. Store the public IP address on the `OpenStackServer` model once it becomes known.  The public IP address of a server doesn't ever change, so we can just as well cache it.
1. Detect if a server was terminated in `OpenStackServer.update_status()`.  This should make Ocim handle VMs that are shut down via Horizon correctly.
1. Move the lock protecting load balancer reconfigurations one level up, so it also covers collecting the configuration.  So far I assumed that there is no need to protect that part, but since collecting the config can take a significant amount of time, it's better to include it in the lock.  With IP addresses being cached in the database, the time to collect the config will be significantly reduced, but moving the lock one level out does not hurt anyway.

The common goal of these changes is to make LB reconfigurations more robust.  Caching the public IP will also speed up the web interface.


instance/models/server.py
```diff
@@ -384,7 +395,7 @@ def start(self,
         else:
             self.openstack_id = os_server.id
             self.logger.info('Server got assigned OpenStack id %s', self.openstack_id)
-            # Persist OpenStack ID
+            # Persist OpenStack ID ans public IP
```
smarnach: This change is wrong anyway and should be reverted.  (I first tried to store the public IP immediately after VM creation, but it's not yet available at that time.)

instance/models/server.py
```diff
@@ -321,38 +325,45 @@ def vm_not_yet_requested(self):
...
+            except novaclient.exceptions.NotFound:
+                # This exception is raised before the server is created, and after it has been
+                # terminated.  Because of the first "if", we can't get her in Pending state, so the
+                # server must have been terminated.
+                self.logger.debug('Server does not exist anymore: %s', self)
+                self._status_to_terminated()
+            except (requests.RequestException, novaclient.exceptions.ClientException):
+                self.logger.debug('Could not reach the OpenStack API')
+                if self.status != Status.Unknown:
+                    self._status_to_unknown()
```
smarnach: This is the existing behaviour.  The status is updated everytime it's shown in the UI, since the server serializer calls `update_status()`.  And after provisioning the status doesn't do much anyway, apart from being shown in the UI.

I'm not completely sure this behaviour is correct in all cases, but it's what we used so far, and I don't know how to improve it.  I'm also not aware of any major problems it has caused.

smarnach:
@bdero I added the integration test you asked for.  I had to add a separate test, since the branch only gets triggered for servers that get terminated via a different mechanism than the `OpenStackServer.terminate()` method.

instance/models/load_balancer.py
```diff
@@ -198,19 +198,24 @@ def run_playbook(self, ansible_vars):
...
+        )
         if returncode != 0:
             self.logger.error("Playbook to reconfigure load-balancing server %s failed.", self)
             raise ReconfigurationFailed
 
+    def _configuration_lock(self):
+        """
+        A Redis lock to protect reconfigurations of this load balancer instance.
+        """
+        return cache.lock("load_balancer_reconfigure:{}".format(self.domain), timeout=900)
```
haikuginger: Is 15 minutes universally long enough? Should this timeout value be configurable?

smarnach: @haikuginger This timeout was already existing in the code.  Currently, reconfigurations take about two to three minutes, but with this change applied, we should be down to less than one minute, and we plan to further reduce the time, so I never expect this to become an issue.

haikuginger: Sounds good!

smarnach:
@bdero Jesse took a quick look before the meeting today, but did not have time to test this.  Could you please test once you are back?  I won't be able to do any work this week, but I can reply to questions.

bdero:
@smarnach Alrighty, deployed this to OCIM stage and I'm happy to say it worked quite well! :+1:

- [x] I tested this: I confirmed that the IP address is getting getting cached from Nova by checking the _ip_field for a fresh appserver VMs. I restarted OCIM during provisioning to lock the appserver state to "Configuring VM" and observed that the VM Status was "Ready (Booted and ready)". I then terminated the appserver in horizon, refreshed the page on OCIM, and verified thatthe VM Status changes to "Terminated (Stopped forever)". I also confirmed that appserver data fetches are around twice as fast after viewed for the first time.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ N/A.
- [x] Includes documentation. These changes alter underlying implementation for performance, not outward facing features. The docstrings are good!

PR #202, opened 2017-05-04
Add features to support ecommerce and course discovery

haikuginger:
In conjunction with open-craft/configuration#18, this pull request enables support for deploying an integrated ecommerce and course discovery environment.

Testing instructions:

Using this branch, create a new instance using the `opencraft-release/ficus.3` edx-platform branch and the configuration branch in open-craft/configuration#18. Include the following extra vars:

```yaml
DISCOVERY_ELASTICSEARCH_URL: "http://149.202.166.156:9200/"
SANDBOX_ENABLE_DISCOVERY: yes
SANDBOX_ENABLE_ECOMMERCE: yes
COMMON_HOSTNAME: "custom-local-hostname-here"
DISCOVERY_VERSION: "c96c77bbf80650b3d3cc238335cc5205e24b775f"
```

Currently, we don't support Elasticsearch as a database backend, so we have to pass that in manually. Additionally, discovery and ecommerce still aren't first-class citizens, so they must be enabled via extra var. We need to set the hostname, other than the FQDN, that the appserver refers to itself as so that we can properly go out to the load balancer-terminated SSL connection. And finally, newer versions of course discovery don't work properly with OIDC login.

Once the spawn is complete, you'll need to take the following steps to finish setup (these are one-time actions that are stored in the database):

1. Update the staff user so that a first name is associated with it (some routines look for a user profile).
2. Update the database-backed OAuth2 clients for course discovery and ecommerce so that they're associated with the staff user.
3. In the sandbox ecommerce env, configure the Enterprise Partner, Site, and SiteConfiguration [per the instructions on this page](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/install_ecommerce.html#configure-a-site-partner-and-site-configuration).
4. In the sandbox discovery env, configure a partner [as described here](https://open-edx-course-catalog.readthedocs.io/en/latest/getting_started.html#configure-partners).
5. If using a load balancer that does not fetch real certs by default, ssh into that load balancer and use a set of commands that looks something like this to get real certificates for your site:
    ```bash
    sudo letsencrypt certonly --email ops@opencraft.com --authenticator standalone --standalone-supported-challenges http-01 --http-01-port 8080 --force-renewal -d haikuginger-ecom2.plebia.net -d studio-haikuginger-ecom2.plebia.net -d ecommerce-haikuginger-ecom2.plebia.net -d discovery-haikuginger-ecom2.plebia.net
    sudo service haproxy restart
    ```
6. You should now be able to hit the `/login` endpoint on both discovery and ecommerce and have the OIDC flow log you in once you've authenticated with the LMS.
7. In the sandbox discovery env, run the following command and verify that it finishes without incident:
    ```bash
    ./manage.py refresh_course_metadata
    ```


instance/models/openedx_instance.py
```diff
@@ -172,20 +205,38 @@ def get_active_appservers(self):
...
+        """
+        Regex that matches either the internal or external URL for the site.
+
+        This is meant exclusively for filling in the server_name regex in nginx configs.
+        """
+        domain_attrs = ['external_{}_domain', 'internal_{}_domain']
+        domains = [getattr(self, attr.format(site_name), None) for attr in domain_attrs]
+        domains = [domain for domain in domains if domain is not None]
+        choices = '|'.join(map(re.escape, domains))  # pylint: disable=bad-builtin
+        return '~^({})$'.format(choices)
```
e-kolpakov: @haikuginger I believe this `~` does not belong to the regex itself, but to nginx - is there a way to add it somewhere later, so that this method actually returns a working regex.

instance/models/openedx_instance.py
```diff
@@ -111,10 +115,7 @@ def domain(self):
 
         Returns external domain if present, otherwise falls back to internal domain.
         """
-        if self.external_lms_domain:
-            return self.external_lms_domain
-        else:
-            return self.internal_lms_domain
+        return self.get_domain('lms')
```
e-kolpakov: @haikuginger Might be good to extract these domain labels into constants (or even enumeration)

smarnach: @haikuginger Why?  It would just increase the level of indirection, and I can't see what we would gain.

instance/models/openedx_instance.py
smarnach: @e-kolpakov This is just copied from the existing code, and I introduced it originally.  In nginx, `~` marks a regular expression.  This is why I called the function `..._nginx_regex`.  We could add the `~` in the template instead of here, if you prefer.  I don't feel strongly either way.

instance/templates/instance/ansible/vars.yml
```diff
@@ -181,11 +181,33 @@ POSTFIX_QUEUE_SENDER_CANONICAL_MAPS: '{{ smtp_relay_settings.source_address }}
 {% endif %}
 {% endwith %}
 
+
+
+# EDXAPP_CMS_NGINX_PORT: 80
```
smarnach: I think this should be uncommented rather than removed.  Why did you comment it out?

instance/models/openedx_instance.py
haikuginger: @smarnach - sorry, are you asking me "why include this code" or asking @e-kolpakov "why move them to a constant"?

My reasoning: We're adding a couple more domains that can be either internal or external, and I wanted to DRY up the specific implementation of doing that.

An alternative: move to a mixin with a `__getattr__` magic method, and a blessed list of names such that if accessing the attribute `blessed_domain`, it does the `external_blessed_domain`/`internal_blessed_domain` fetching automatically.

instance/models/openedx_instance.py
```diff
@@ -124,17 +125,40 @@ def domain_slug(self):
         prefix = ('edxins-' + slugify(self.domain))[:20]
         return "{prefix}-{num}".format(prefix=prefix, num=self.id)
 
+    def get_domain(self, site):
+        """
+        Returns the external domain for the given site if present; otherwise, falls back to the internal domain.
+        """
+        internal_attr = 'internal_{}_domain'.format(site)
+        external_attr = 'external_{}_domain'.format(site)
+        return getattr(self, external_attr, None) or getattr(self, internal_attr)
```
smarnach: It's a bit of a hack that the field names now define the structure of the data.  In general, it's not a good idea to put data in variable names.

An alternative would be to have a field `domains`, which stores a JSON dictionary mapping site names to lists `["external.domain", "internal.domain"]`, or a similar construct.  However, since the version of Postgres we are using does not support JSONB fields yet, it's probably best to just leave it as you implemented it for now.

instance/models/openedx_instance.py
smarnach: @haikuginger I meant to ping Eugeny, not you.  I'm asking what the additional variables would buy us.  I think the problem here is that the structure of the data is encoded in the field names (see my other comment), but this would not be addressed by Eugeny's suggestion.

smarnach:
@haikuginger I've got a few questions about this.

1. Do our standard VMs have enough memory to run both discovery and ecommerce in addition to the usual stuff?  Even normal instances can be somewhat short on memory, so I expect it to be tight with all the additional stuff.
1. If we ever want to enable ecommerce and discovery by default, we need to automate the manual steps that are steps in your testing instructions.
1. Why is it necessary to request production certificates?  The staging certificates are generally perfectly fine for testing purposes.

haikuginger:
1. They do run; performance isn't great, though. Not sure how much of that is due to SSL negotiation with the dev LB, though. Latency seems to be more of an issue than throughput.

2. Yes, it's true. This is a first step so that we can get this online for Pearson.

3. Because Discovery and ecommerce communicate via validated HTTPS, they reject invalid SSL certificates.

instance/templates/instance/ansible/vars.yml
haikuginger: Ah, this is cruft; I had copied it from above (where it isn't commented) and was using it as a mental template to build the vars for discovery and ecommerce. Will remove.

instance/models/openedx_instance.py
```diff
@@ -172,20 +205,38 @@ def get_active_appservers(self):
...
 
+    def domain_nginx_regex(self, site_name):
+        """
+        Regex that matches either the internal or external URL for the site.
+
+        This is meant exclusively for filling in the server_name regex in nginx configs.
+        """
+        domain_attrs = ['external_{}_domain', 'internal_{}_domain']
+        domains = [getattr(self, attr.format(site_name), None) for attr in domain_attrs]
+        domains = [domain for domain in domains if domain is not None]
```
haikuginger: Will do.

instance/models/openedx_instance.py
e-kolpakov: @smarnach I don't feel strongly too, but moving the `~` to the template have slight benefit of return value of this function being an actual regex - so it can be used to validate something, or in tests, or something like that.

instance/models/openedx_instance.py
e-kolpakov: @smarnach It buys us protection from typos. :) Obviously, my suggestion wasn't to address the problem you highlighted, because this comment was posted before you highlighted it :)

Also, you might misunderstood what I meant - I only suggested to extract those magic strings "lms", "studio", "ecommerce" etc. into constants - that's it.

instance/models/openedx_instance.py
e-kolpakov: @smarnach It's not `let` though, it just a plain old loop :)

smarnach: @e-kolpakov That's why I said "kind of".  I don't think Python needs any additional syntax for this.

instance/models/openedx_instance.py
smarnach: @e-kolpakov As far as I can tell, the strings only occurs once in the code, so moving this single occurrence to a different place does not seem to help.  Moreover, we already have protection from typos, since line 134 will fail if we use a string that does not have a corresponding attribute name.

instance/models/openedx_instance.py
smarnach: @e-kolpakov I can see that.  @haikuginger Can you move the tilde to the template, please?

smarnach:
@haikuginger Could you please expand on "Discovery and ecommerce communicate via validated HTTPS"?  In what way to the communicate?

Do they make API requests that are routed via the LB?  If so, that seems wrong to me.  They can just use plain HTTP directly to the application port, even circumventing the local nginx.

Or do they redirect the user to one of the other domains?  If so, the right solution would be to cajole your browser into trusting Let's Encrypt staging certificates (which is a relatively safe thing to do anyway.)

haikuginger:
@smarnach, they use the hostname settings both for API requests and for redirects via OIDC. Unfortunately, we can't separate the two concerns without making major codebase changes - so we have to route API calls out to the LB so that they can have the same scheme as redirects. Hence the `COMMON_HOSTNAME` override - we can't have the appserver believing it can contact itself at itself, because port 443 isn't open (and even if it were, it wouldn't have a valid cert). We have to force it to do a full DNS lookup and contact itself through the LB.

smarnach:
@haikuginger Thanks for the explanation.  I can understand the problem now.  Fixing the upstream code to allow to separate the API URL from the redirect URL would be the optimal solution, but I understand we don't have the resources to do that now.

A simple short-term fix would be to simply install a self-signed certificate on the Pearson instance and make it trust its own certificate.  We might need to open port 443 in the security group rules.  However, this is unrelated to this PR – it's just a different configuration for the instance.

haikuginger:
@smarnach, we'd also need to install hostfile entries for the ecom and discovery domains, correct?

smarnach:
@haikuginger My understanding was that it's just discovery and ecommerce accessing the LMS API.  If it's also the other way around, then yes.  But as noted above, this is unrelated to this PR.

haikuginger:
@smarnach, back to you - suggested changes have been made, and unit tests are passing.

pylintrc
```diff
@@ -321,13 +321,13 @@ max-statements=50
 max-parents=7
 
 # Maximum number of attributes for a class (see R0902).
-max-attributes=7
+max-attributes=8
```
smarnach: Xavier does not want us to change these limits, nor to add pylint overrides to the code base.  I'm not sure what to do, since to me the code looks good.

Using the single structured `domains` attribute I suggested above would avoid hitting this limit, but I don't think it's a good idea to use it before we can use proper JSONB fields.

CC @antoviaque

smarnach:
:+1: once it is rebased, squashed and has a green build.  I added a comment about changing the pylint limits; I'm fine with that, but Xavier probably isn't, so please discuss this point with him while I'm away.

It would also be nice to have slightly more detailed instructions how to set this up on a production instance.  If we ever want to do this again for a production instance, or if we want to automate the process, detailed instructions will be very handy.  It's not really obvious how to get this working, after all.  Kudos for figuring it out!

- [x] I tested this: I followed the testing instructions and set up [this instance on stage](https://stage.console.opencraft.com/instance/1240/).  It was a bit of a rough ride, since I experienced various issues on the way, but none really related to this PR, except maybe that the testing instructions were a bit _too_ concise in some places. :)
- [x] I read through the code

haikuginger:
Squashed and rebased from 279750b.

instance/models/openedx_instance.py
antoviaque: The parts of the OpenEdXInstance class dedicated to support e-commerce could be put on a separate mixin? And/or splitting out the generic parts of the domains management?

haikuginger: To avoid unnecessary complication, I'm going to split the domain-fetching logic out onto a separate mixin.

smarnach: @haikuginger You will hit another pylint limit when moving this to a new mixin.  The maximum number of base classes is set to 7 in our pylint config, and we've already reached this limit.

haikuginger:
@antoviaque, I've refactored domain name handling into a new mixin.

As @smarnach commented above, it looks like this'll result in a new Pylint error, since we limit the number of base classes. Should I update `pylintrc` to allow this, or do you have a different workaround in mind?

antoviaque:
@haikuginger Would grouping the mixins by categories in an intermediary class work, and make sense from the code perspective?

Btw, another point is documentation - it's missing a README update for the new conf variables, and it might be worth also adding the installation instructions from this PR's description there?

haikuginger:
Rebased from a89c270.

haikuginger:
@antoviaque, it looks like we have a green build; the anticipated linting issue did not emerge. I've also added documentation. Are we set to merge this and the configuration PR?

haikuginger:
@antoviaque, ping^

antoviaque:
@haikuginger Sorry that I missed your ping last week - yes, this is good on my side. :+1:

haikuginger:
Rebased from bd07240.

PR #203, opened 2017-05-04
Mention archive() method in README.

mtyaka:
Explain that in production `archive()` should be used instead of `destroy()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,13 +637,22 @@ for appserver in instance.appserver_set.all():
     appserver.terminate_vm()
 ```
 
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
:+1:

- [ ] ~~I tested this:~~ Did not test yet, see ping on Mattermost
- [x] I read through the code
- [x] Includes documentation: is documentation

bradenmacdonald:
We seem to be hitting some limit on our integration test OpenStack account causing the build to fail due to not being able to obtain fresh VMs.

PR #207, opened 2017-05-08
Add info about bulk-updating instances to README.

itsjeyd:
Part of the scope of [OC-2608](https://tasks.opencraft.com/browse/OC-2608).

The example script was used successfully to bulk-upgrade instances from Ficus.1 to Ficus.3.

**TODO** (as part of future tasks involving instance upgrades):

* Turn the script for upgrading instances into a management command.
    * Rework step 3 and 4 so that they will block until the low priority huey queue is empty.
    * Parameterize the management command on:
        * The main tag name to use (`ficus3-redeployment` in the script). We can derive success and failure tags from that name automatically.
        * The source release to use for selecting instances (`open-release/ficus.1` in the script).
        * The target release to upgrade to (`open-release/ficus.3` in the script).
    * To avoid the possibility of being left with some instances that failed to upgrade after the command finishes, modify step 4 to iterate until the set of instances currently tagged with `failure_tag` is empty.

**Reviewers**

- [x] @bradenmacdonald

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ flavor:
 ### OpenStack Security Groups
 
 Every VM used to host Open edX will automatically be added to an OpenStack
-network security group, which is provides a firewall that limits what 
+network security group, which is provides a firewall that limits what
 ports/services on the VM are exposed to the Internet. The security group will
 automatically be created and managed by OpenCraft IM.
 
@@ -653,6 +653,74 @@ instance.delete()
 ```
 
 **Do not use delete() in production!**
+
+Bulk-updating sandboxes
+-----------------------
+
+OpenCraft IM provides a tagging mechanism that you can use to update instances in bulk.
+The following script shows how to update multiple instances at once from the Django shell:
+
+```python
+# Author: Brandon DeRosier <brandon@opencraft.com>
+
+# 1. Set up tags
+tag, _ = InstanceTag.objects.get_or_create(name='ficus3-redeployment')
+success_tag, _ = InstanceTag.objects.get_or_create(name=tag.name + "-succeeded")
+failure_tag, _ = InstanceTag.objects.get_or_create(name=tag.name + "-failed")
+
+# 2. Find all the affected instances (ficus.1) and tag them
+upgrade_instances = OpenEdXInstance.objects.filter(
+    openedx_release='open-release/ficus.1',
+    edx_platform_commit='opencraft-release/ficus.1',
+    configuration_version='opencraft-release/ficus.1',
+    ref_set__is_archived=False,
+)
+for instance in upgrade_instances:
+    instance.tags.add(tag)
+    instance.openedx_release='open-release/ficus.3'
+    instance.edx_platform_commit='opencraft-release/ficus.3'
+    instance.configuration_version='opencraft-release/ficus.3'
+    instance.save()
+
+# 3. Spawn the new appservers
+for instance in tag.openedxinstance_set.iterator():
+    spawn_appserver(instance.ref.pk, success_tag=success_tag, failure_tag=failure_tag)
+
+# 4. Try again on any failed instances (run multiple times if necessary)
+for instance in failure_tag.openedxinstance_set.iterator():
+    spawn_appserver(instance.ref.pk, success_tag=success_tag, failure_tag=failure_tag)
+
+# 5. Activate only the latest appserver for all upgraded instances
+for instance in success_tag.openedxinstance_set.iterator():
+    print('Swapping active instance for: ' + instance.domain)
+    # Deactivate currently active servers
+    for appserver in instance.appserver_set.filter(_is_active=True):
+        print(' - Deactivating {} ({})'.format(appserver.name, appserver.id))
+        appserver.is_active = False
+        appserver.save()
+    # Activate the latest server
+    latest_appserver = instance.appserver_set.latest('created')
+    print(' - Activating {} ({})'.format(latest_appserver.name, latest_appserver.id))
+    latest_appserver.is_active = True
+    latest_appserver.save()
+
+# 6. Reconfigure the load balancer once
+lb = LoadBalancingServer.objects.get(accepts_new_backends=True)
+lb.reconfigure()
+
+# 7. Update the monitoring for each of the instances after the load balancer is reconfigured
+for instance in success_tag.openedxinstance_set.iterator():
+    instance.set_active_vm_dns_records()
+    instance.enable_monitoring()
+
+# 8. Remove tags
+all_tags = [tag, success_tag, failure_tag]
+untag = success_tag.openedxinstance_set.all()
+
+for inst in untag:
+    inst.tags.remove(*all_tags)
+    inst.save()
+```
 
 manage.py
 ---------


README.md
```diff
@@ -654,6 +654,74 @@ instance.delete()
...
 **Do not use delete() in production!**
 
+Bulk-updating sandboxes
+-----------------------
+
+OpenCraft IM provides a tagging mechanism that you can use to update instances in bulk.
+The following script shows how to update multiple instances at once from the Django shell:
+
+```python
+# Author: Brandon DeRosier <bdero@opencraft.com>
```
antoviaque: @itsjeyd @bradenmacdonald Can this go to a management command script?

bradenmacdonald: @antoviaque I would prefer *not* to make this a management command. There are often slightly different things that we need to do with each mass re-deployment, and I think its useful to be able to customize the behavior each time.

bradenmacdonald:
I'm not going to test this again, but 👍  from me once you fix Brandon's email.

README.md
itsjeyd: @bdero Sorry :) Fixed in a55b886512568fbef181cce9b181ad659c5ba9dd.

itsjeyd:
Squashed from a55b886512568fbef181cce9b181ad659c5ba9dd.

itsjeyd:
@bradenmacdonald Thanks for the review! Will merge once the build is green.

README.md
antoviaque: Alright, can we list these things precisely, so that i can become a todo list for the next iterations? And use the opportunity/iteration to discuss the approach, as part of the current review? I'd like to keep this work as part of the deployment tasks, rather than something separate. It would also be useful to have your feedback from this iteration - what did take you time? What was missing in terms of automation?

> I would prefer not to make this a management command. There are often slightly different things that we need to do with each mass re-deployment, and I think its useful to be able to customize the behavior each time.

If there are things that are different each time, can we cleanly separate them from the workflow of rebuilding all the instances?

bradenmacdonald: @antoviaque 

> If there are things that are different each time, can we cleanly separate them from the workflow of rebuilding all the instances?

Somewhat. For example, in this case, we could separate the tagging of instances and the changing of settings (steps 1, 2, and 8) from the process of spawning and activating the new appservers (steps 3-7). So steps 3-7 could be put into a management command (or, better, an API endpoint which is fairly easy to do and also gives us [an autogenerated UI](https://console.opencraft.com/api/v1/)).

There are cases where that approach won't work, though: for example, if an upgrade includes a migration that's not backwards compatible. Then the script needs to deploy a new instance without migrations, take the current instance offline (display a maintenance page), run the migrations on the new instance, and point the LB to the new instance, which is a completely different process than we're seeing here. So even if we put steps 3-7 into a command/API, there is a need to document how to do these things via a flexible script.

antoviaque: > There are cases where that approach won't work, though: for example, if an upgrade includes a migration that's not backwards compatible. Then the script needs to deploy a new instance without migrations, take the current instance offline (display a maintenance page), run the migrations on the new instance, and point the LB to the new instance, which is a completely different process than we're seeing here. So even if we put steps 3-7 into a command/API, there is a need to document how to do these things via a flexible script.

@bradenmacdonald These also sound like steps that can be automated - and actually, talking about migrations, separating the DB migrations from the VM updates is a good practice that we will need to implement to have a reliable CI workflow. 

Generally, rather than relying on updating a script manually every time, I'd rather look at what needs to be changed from one time to the next in details. I would be surprised if we couldn't find a way to automate everything -- at least on he Ocim part. Ie there could still be work on each release on the git repos Ocim points to, but not to Ocim itself, which needs to be on a more generic level of abstraction, and not require changes on each release of the applications it supports.

bradenmacdonald: @antoviaque Ok, well for our next iteration, I suggest we put steps 3-7 into an API within the codebase and to think about how to make them a bit more generic to handle cases like migrations.

For the rest, just because we have to type a few commands into the shell doesn't mean it's not "automated" - it's the same situation as provisioning a new instance right now, which requires typing some commands into a shell. The shell is the UI, and 99% of the work is still automated. Same here - because of the complexity of the types of upgrades, migrations, changes, deployments, etc. that we may need to do, it makes sense to me to keep the shell as the UI for such things, even though 99% of the work is automated within the codebase. Later on we can create another UI that's actually HTML-based if we see enough common patterns that fit that workflow.

antoviaque: @bradenmacdonald That can be a useful intermediary step, but the goal is to use the concrete use cases to push the formalization, and get to a continuous integration workflow. The end goal is to not require UI at all for updates - that once we are done, pressing "merge" on a PR on github is all that is required to get the production instance to update.

bradenmacdonald: @antoviaque How do you envision that working for things like instance upgrades (e.g. ficus.2 to ficus.3), settings changes (e.g. enable CREATOR_GROUP for all instances), resource sharding (e.g. moving 1/2 of our instances to a new RabbitMQ server), etc.? There's no pull request involved for those things so I don't see it as "continuous integration".

antoviaque: @bradenmacdonald 

> How do you envision that working for things like instance upgrades (e.g. ficus.2 to ficus.3)

The instance points toward a branch called "production", which gets updated from the content of the ficus.2 branch to the ficus.3 branch. A PR could be used to merge it there.

> settings changes (e.g. enable CREATOR_GROUP for all instances)

The settings changes would be listed in the PR, like for our sandboxes currently.

> resource sharding (e.g. moving 1/2 of our instances to a new RabbitMQ server)

That uses configuration settings, no?

itsjeyd:
Rebased from a17613688d0496646d44c9340ed74d7cf3044cb9.

README.md
itsjeyd: @antoviaque Responding to your request from [OC-2608](https://tasks.opencraft.com/browse/OC-2608):

> There are often slightly different things that we need to do with each mass re-deployment, and I think its useful to be able to customize the behavior each time.

> Alright, can we list these things precisely, so that i can become a todo list for the next iterations?

@bdero's [comment](https://github.com/open-craft/opencraft/pull/207#discussion_r115268569) covers the some of the immediate steps that we need to take in order to turn the script into a management command:

> For making this reference into a management command, step 4 and 5 would need to be reworked so that they will block until the low priority huey queue is empty; I'm not sure how to do this but I'm sure it can be done.

... except that I think it should be step 3 and 4 that should be blocking; step 5 does not involve any asynchronous processing.

Beyond that, we should parameterize the management command on:

* The main tag name to use (`ficus3-redeployment` in the script). We can derive success and failure tags from that name automatically.
* The source release to use for selecting instances (`open-release/ficus.1` in the script).
* The target release to upgrade to (`open-release/ficus.3` in the script).

Lastly, to avoid the possibility of being left with some instances that failed to upgrade after the command finishes, we could modify step 4 to iterate until the set of instances currently tagged with `failure_tag` is empty.

CC @bradenmacdonald @bdero @smarnach

UmanShahzad:
@itsjeyd @antoviaque Is this PR required given the instance redeployment command now?

UmanShahzad:
@antoviaque Yep, that's https://github.com/open-craft/opencraft/blob/master/instance/management/commands/instance_redeploy.py.

@itsjeyd Any objection on your part?

itsjeyd:
@UmanShahzad Nope, no objections. I'll go ahead and close this.

PR #209, opened 2017-05-16
Update integration tests to use ficus.3.

mtyaka:
The `opencraft-release/ficus.3` branch includes a fix for the mathplotlib bug which prevents instances from building successfully.

/cc @haikuginger


bradenmacdonald:
👍  if the build passes.

(testing, doc, a11y n/a here)

haikuginger:
👍 as the build passed, and that's what we were looking for. This fixes all the locations I'm aware of where we have release-specific code.

PR #211, opened 2017-05-26
Remove dj-static

snowcrshd:
Since we will start serving static files through NGINX, we can drop dj-static.
I'm still working on this, just sending the PR upfront in case there is anything that should be considered too.

This change is tied to open-craft/ansible-opencraft#7

When we are good to go, I believe we should merge `ansible-opencraft` PR
and then update this PR to make `deploy/requirements.yml` point to the correct version of `ansible-opencraft`.

### DIFF
--- a/deploy/requirements.yml
+++ b/deploy/requirements.yml
@@ -21,24 +21,24 @@
 
 - name: backup-to-tarsnap
   src: https://github.com/open-craft/ansible-backup-to-tarsnap
-  version: 3d2f6f982ab02112ede9e39e1553b106e5079b87
+  version: origin/master
 
 - name: common-server
   src: https://github.com/open-craft/ansible-common-server
-  version: 8423f76a12f5ade7bd008efffedfee12f9506259
+  version: origin/master
 
 - name: configure-apt
   src: https://github.com/open-craft/ansible-configure-apt
-  version: 97877cadc7927dab043bff70a1e621af3b277b5f
+  version: origin/master
 
 - name: sanity-checker
   src: https://github.com/open-craft/ansible-sanity-checker/
-  version: 5387389cbb640e125595fd57e5ab787504a9537c
+  version: origin/master
 
 - name: forward-server-mail
   src: https://github.com/open-craft/ansible-forward-server-mail
-  version: c252ef531802fecace6c0173e647e0a0f13df256
+  version: origin/master
 
 - name: opencraft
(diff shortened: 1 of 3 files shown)


smarnach:
@snowcrshd For our other Ansible repositories, we've moved to using branch names in requirements.yml instead of commit hashes, at least for the roles we control ourselves.  Could you change this for the roles owned by the open-craft org as part of this PR, please?  Just change the version to `origin/master` instead of the commit hash, and we will always follow the latest version.

snowcrshd:
@smarnach sure thing. Just updated the PR with a new commit.

deploy/requirements.yml
```diff
@@ -17,28 +17,28 @@
 # Can be cleared when this is closed: https://github.com/pmbauer/ansible-tarsnap/pull/3
 - name: pmbauer.tarsnap
   src: https://github.com/open-craft/ansible-tarsnap
-  version: d96b1e69ae6b930e0841d7b025a3070c408a23f4
+  version: origin/master
```
smarnach: We can't use master for this one.  It's a fork of a third-party role, and the master branch is the upstream master at the time of forking, and not the branch containing our modification.

Sorry, I didn't think of this one when I said "all roles owned by the open-craft org".

snowcrshd: No problem! Just reverted that change.

smarnach:
:+1: Looks good to me once [the last minor comment](https://github.com/open-craft/opencraft/pull/211#discussion_r120012480) is addressed and the commits are squashed.  I will trigger a CircleCI build and merge then.

- [x] I tested this: (describe what you tested)
- [x] I read through the code
- [x] Includes documentation

snowcrshd:
@smarnach PR is corrected now. I'm not changing the tarsnap playbook version anymore.

smarnach:
@snowcrshd Thanks!  I currently can't figure out how to trigger the CircleCI build for this PR.  I'll wait whether someone can tell me in Mattermost until tomorrow.

bradenmacdonald:
@smarnach Push @snowcrshd 's branch to the `opencraft` fork, and it should build.

@snowcrshd Next time we can just give you push permissions for the `opencraft` fork and you won't have that issue.

snowcrshd:
@bradenmacdonald thanks! That would be very useful 😄

bradenmacdonald:
@snowcrshd Done!

PR #212, opened 2017-06-14
Add ecommerce URL variables

haikuginger:
This PR adds additional variables that were missed in the first pass of #202.

Testing: On the `courses.pearsonx.com` instance (where these variables had to be set manually), verify that when selecting the edX Demo Course and the Verified course mode, you're successfully redirected to the ecommerce stack.


antoviaque:
I'll defer to @smarnach on this review - no objection on my side as long as he is fine with it.

smarnach:
@haikuginger Looks good to me. :+1:

I've tested this as described, and it works fine.  I only noticed that currently the live PearsonX instance uses the Paypal sandbox, but apparently the demo course is the only one currently offering a paid option anyway.  We should remove the paid option for the demo course when switching to an actual payment processor.

PR #214, opened 2017-07-10
Improve Gandi domain error message

bdero:
This small PR improved the domain matching error message to include the actual domain name. This makes it easy to see specifically which domains are not in the record list whenever this exception is thrown.

CC @smarnach


instance/tests/test_gandi.py
```diff
@@ -88,8 +88,12 @@ def test_split_domain_name(self):
         self.assertEqual(self.api.split_domain_name('sub.domain.test.com'), ('sub.domain', 'test.com'))
         self.assertEqual(self.api.split_domain_name('sub.domain.opencraft.co.uk'), ('sub.domain', 'opencraft.co.uk'))
         self.assertEqual(self.api.split_domain_name('example.com'), ('@', 'example.com'))
-        with self.assertRaises(ValueError):
+        with self.assertRaises(ValueError) as error:
             self.api.split_domain_name('sub.domain.unknown.com')
+        self.assertEqual(
+            error.exception.message,
```
smarnach: This should be `str(error.exception)` instead.

bdero:
@smarnach Thanks, I've amended that change

PR #217, opened 2017-08-09
Don't shut down instances more than once

itsjeyd:
... when cleaning up obsolete instances.

This is to avoid unnecessarily creating and activating new versions of the Gandi zone file that we are using for opencraft.hosting.

Cf. [OC-2954](https://tasks.opencraft.com/browse/OC-2954)

**Reviewers**

- [x] @pomegranited


itsjeyd:
@pomegranited @e-kolpakov Thanks for following up, this LGTM 👍 Build is green, so I'll flip the switch now.

PR #221, opened 2017-08-17
Optional rabbitmq on dev vagrant VM

carlio:
Adding instructions and helper provisioning script to include a rabbitmq server on the vagrant dev box.

This is useful if you do not have access to an external rabbitmq server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,11 @@ edit its security group rules to only allow access to VMs in the
 * `DEFAULT_INSTANCE_RABBITMQ_URL`: The RabbitMQ AMQPS URI to be used by instances. E.g.,
   `amqps://rabbitmq.example.com:5671`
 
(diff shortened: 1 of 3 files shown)


UmanShahzad:
@bradenmacdonald Any reason we wouldn't want this?

bradenmacdonald:
@UmanShahzad  I actually don't know why we _would_ want this. RabbitMQ must be reachable by the VMs that we spawn (e.g. on OVH), and a local RabbitMQ within vagrant will not be. Our private .env file template contains the setting for a shared RabbitMQ server that we can all use, so we don't need this ourselves. See the long (collapsed) comment thread on our internal documentation PR 196.

UmanShahzad:
@bradenmacdonald Ah, yeah I read that earlier, makes sense :+1:

PR #222, opened 2017-08-17
Move instance log data to new API endpoint

carlio:
This pull request references OC-2877. It removes log entry information from the `instance/:id` endpoint and moves it into a new `instance/:id/logs` endpoint instead. 

This means that log data will not be loaded by default when viewing an instance detail, thus saving unnecessary data being streamed to the frontend. When a user clicks on the "Logs" tab, the data is loaded at that point from the new endpoint.

**To Test:**

* Set up an OCIM and create some instances
* Verify that all tabs except Logs work as normal
* Click on Logs tab with a browser inspector open to verify that a new network request is made, and that the log data is loaded and displayed in the Logs tab (and a "Loading logs" placeholder message is there until log data arrives).
* See also the API end point for verification (`/api/v1/instance/:id/logs/`)
* Check that from `/api/v1/instance/:id/` you can follow a link to `/api/v1/instance/:id/logs/`

### DIFF
--- a/instance/api/instance.py
+++ b/instance/api/instance.py
@@ -29,8 +29,11 @@ from rest_framework.response import Response
 
 from instance.models.instance import InstanceReference
 from instance.models.openedx_appserver import OpenEdXAppServer
-from instance.serializers.instance import (InstanceReferenceBasicSerializer, InstanceReferenceDetailedSerializer,
-                                           InstanceLogSerializer)
+from instance.serializers.instance import (
+    InstanceReferenceBasicSerializer,
+    InstanceReferenceDetailedSerializer,
+    InstanceLogSerializer
+)
 
 
 # Views - API #################################################################
--- a/instance/api/instance.py
+++ b/instance/api/instance.py
@@ -24,11 +24,13 @@ Instance API
 
 from django.db.models import Prefetch
 from rest_framework import viewsets
+from rest_framework.decorators import detail_route
 from rest_framework.response import Response
 
 from instance.models.instance import InstanceReference
 from instance.models.openedx_appserver import OpenEdXAppServer
-from instance.serializers.instance import InstanceReferenceBasicSerializer, InstanceReferenceDetailedSerializer
+from instance.serializers.instance import (InstanceReferenceBasicSerializer, InstanceReferenceDetailedSerializer,
+                                           InstanceLogSerializer)
 
 
 # Views - API #################################################################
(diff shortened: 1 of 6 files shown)


clemente:
@bradenmacdonald later today I will be improving the other PR (#223) so that if you want you can check both at the same time. But this one (#222) is finished and you can review it and merge it now if you prefer.

clemente:
@bradenmacdonald By the way, CircleCI tests don't pass for unrelated reasons (DNS); how can I trigger a rebuild?

bradenmacdonald:
@clemente  You just click the "Rebuild" button in the top right of CircleCI (if you're logged in):
screen shot 2017-09-25 at 10 34 55 am

clemente:
@bradenmacdonald I don't have a CircleCI account, therefore the rebuild button is disabled („you need write permissions to trigger builds“).

instance/api/instance.py
```diff
@@ -24,11 +24,13 @@
...
 from django.db.models import Prefetch
 from rest_framework import viewsets
+from rest_framework.decorators import detail_route
 from rest_framework.response import Response
 
 from instance.models.instance import InstanceReference
 from instance.models.openedx_appserver import OpenEdXAppServer
-from instance.serializers.instance import InstanceReferenceBasicSerializer, InstanceReferenceDetailedSerializer
+from instance.serializers.instance import (InstanceReferenceBasicSerializer, InstanceReferenceDetailedSerializer,
+                                           InstanceLogSerializer)
```
bradenmacdonald: Nit: once imports get longer than 1 line, I much prefer this format:

```python
from instance.serializers.instance import (
    InstanceReferenceBasicSerializer,
    InstanceReferenceDetailedSerializer,
    InstanceLogSerializer,
)
```

instance/serializers/instance.py
```diff
@@ -110,3 +110,15 @@ class InstanceReferenceDetailedSerializer(InstanceReferenceBasicSerializer):
...
+class InstanceLogSerializer(serializers.ModelSerializer):
+    """
+    Provide the log entries for an Instance
+    """
+    log_entries = LogEntrySerializer(many=True, read_only=True)
+    log_error_entries = LogEntrySerializer(many=True, read_only=True)
+
+    class Meta:
+        model = InstanceReference
+        fields = ('log_entries', 'log_error_entries',)
```
bradenmacdonald: `Instance` does not have a `log_error_entries` field, so that isn't doing anything - please remove it from this serializer.

clemente:
@bradenmacdonald  I did these two changes. Waiting for CircleCI…

instance/serializers/instance.py
```diff
@@ -121,4 +121,4 @@ class InstanceLogSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = InstanceReference
-        fields = ('log_entries', 'log_error_entries',)
+        fields = ('log_entries', )
```
bradenmacdonald: @clemente Thanks, but you should also remove this above on line 120

clemente:
CircleCI is also failing with:
> requests.exceptions.ConnectionError: HTTPSConnectionPool(host='dc7f8e33.integration.plebia.net', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x2b11918ab978>: Failed to establish a new connection: [Errno -2] Name or service not known',))

Restarting…

PR #223, opened 2017-08-17
Load no more than 5 appservers per instance and add button to load all

carlio:
This PR references OC-2877. It removes all AppServer data apart from the most recent 5 from the basic `instance/:id` end point and moves it into a new `instance/:id/app_servers` end point which lists all.

This means that the app server list on an instant detail page will only show the first 5, until the link is clicked to "load entire list". This helps reduce the amount of data streamed to the client and reduces data streamed from the DB.

~This PR is a series of additional commits after #222 and contains everything in #222 as well. It was intended as helpful "extra" functionality on top of the main goal of issue OC-2877.~ This feature (limit to 5 servers) is one of the optimizations mentioned in OC-2877. The other one is avoiding loading logs, and is handled by #222. Both improvements are independent of each other (they share no commits).

This is ready to be merged.

You should try:
- when there are many app servers, only the last 5 are displayed
- a button „show all [NUMBER]“ appears before the first app server, and shows how many will be loaded after clicking it
- the button disappears when clicked
- the button doesn't appear when there are <5
- the API includes a link from instance/10/ to instance/10/app_servers/
- it feels faster than before, because only 5 are loaded

The UI for "Load entire App Server List" is simple (just a text "show all 11") and could use a better design like an icon.


clemente:
I rebased the code but onto `master` (fixing some conflicts), to make it clear that this improvement is independent from the one in #222. We can discuss it, test it or apply it indepently. So now the commits from #222 are not in this branch, they are in #222 only.
I don't know if it was the right thing to do, because when merging there will be conflicts again (but simple ones).

clemente:
@bradenmacdonald This is ready for review.

bradenmacdonald:
> I rebased the code but onto master (fixing some conflicts), to make it clear that this improvement is independent from the one in #222. 

Sounds good - thanks!

instance/serializers/instance.py
```diff
@@ -110,3 +114,21 @@ class InstanceReferenceDetailedSerializer(InstanceReferenceBasicSerializer):
...
+    """
+    class Meta:
+        model = InstanceReference
+        fields = ('app_servers',)
+
+    def to_representation(self, obj):
+        output = super().to_representation(obj)
+        request = self.context['request']
+        output['app_servers'] = [
+            AppServerBasicSerializer(appserver, context={'request': request}).data
```
bradenmacdonald: I think you should just say `context=self.context` instead of declaring a local `request` variable and passing it in explicitly like this.

instance/tests/api/test_instance.py
```diff
@@ -156,3 +163,19 @@ def test_get_log_entries(self):
...
+        self.api_client.login(username='user3', password='pass')
+        instance = OpenEdXInstanceFactory(name="Test!")
+        instance.logger.info("info")
+        instance.logger.error("error")
+
+        response = self.api_client.get('/api/v1/instance/{pk}/app_servers/'.format(pk=instance.ref.pk))
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        self.assertTrue('app_servers' in response.data)
```
bradenmacdonald: Can we add a test or assertion that actually checks the format of the data returned by this API?

instance/serializers/openedx_instance.py
```diff
@@ -150,7 +150,7 @@ def to_representation(self, obj):
         """
         output = super().to_representation(obj)
         output['appservers'] = [
-            AppServerBasicSerializer(appserver, context=self.context).data for appserver in obj.appserver_set.all()
+            AppServerBasicSerializer(appserver, context=self.context).data for appserver in obj.appserver_set.all()[:5]
```
UmanShahzad: I feel like this number should be configurable.

instance/serializers/instance.py
clemente: Done

instance/tests/api/test_instance.py
clemente: Done. I also improved other tests, one of them was wrong because it didn't create appservers in the instance. Now I create e.g. 7 appservers and check that the API returns less (e.g. 5).

instance/serializers/openedx_instance.py
clemente: Ok, I made this into a variable, and we can set it to 5, 3, or any number.

UmanShahzad: Awesome, thanks!

UmanShahzad:
@clemente If the build issues related to MongoDB are fixed, go ahead and squash & push.

PR #226, opened 2017-09-05
Allow for longer PR branch names and mitigate GitHub rate limit issues

haikuginger:
This pull request has two main goals: allow for automatic sandbox creation for longer branch names (255 characters up from the current 50 character limit) and also mitigates issues due to GitHub rate limiting.

To accomplish the second goal, we check the GitHub ratelimit status endpoint before making each call, and fail early if we don't have the resources available. We also deduplicate the requests that search for open pull requests by checking for PRs from multiple usernames at once.

**Dependencies**: None

**Testing instructions**:

1. Create a branch on your Ocim devstack's watched repo that has an excessively long name (50 characters or longer)
2. In the shell for your Ocim devstack, run the following:
    ```python
    >>> from pr_watch.tasks import watch_pr
    >>> watch_pr()
    ```
3. Verify that a new PR object is created without any issues.
4. Also in your Ocim devstack shell, run the following:
    ```python
    >>> from pr_watch.tasks import watch_pr
    >>> for i in range(100):
            watch_pr()
    ```
5. Watch the output, specifically eyeing the ratelimit HTTP responses, and note that once the "remaining" calls for search resources reach 0, you start seeing warnings that look like the following:
    ```
    Rate limit exceeded when requesting a 'search' resource at https://api.github.com/search/issues?sort=created&q=is:open is:pr author:opencraft-fake-account repo:random-fake-organization/edx-platform
    ```

1. This pull request shifts from calling the GitHub API once for every user on a team to calling it for all users on a team simultaneously by passing multiple `author:username` parameters.
2. We currently don't do anything when a ratelimited request fails; this is because every action we take (i.e., spawning new sandboxes) is logged in the database, so if a request fails, then we can simply pick up where we were at the next iteration.


pr_watch/github.py
```diff
@@ -143,14 +145,21 @@ def get_pr_list_from_username(user_name, fork_name):
...
+    q = 'is:open is:pr {authors} repo:{repo}'.format(authors=authors, repo=fork_name)
     r_pr_list = get_object_from_url('https://api.github.com/search/issues?sort=created&q={}'.format(q))
-    logger.debug('List of PRs received for user %s: %s', user_name, r_pr_list)
 
     pr_list = []
     for pr_dict in r_pr_list['items']:
-        pr = get_pr_by_number(fork_name, pr_dict['number'])
-        pr_list.append(pr)
+        logger.debug('Received PR for user %s: %s', pr_dict['user']['login'], pr_dict)
+        pr_list.append(get_pr_by_number(fork_name, pr_dict['number']))
```
smarnach: It took me a while to understand why we need to re-request each PR, since we already got information about each PR in the search results.  The information we get from the search result is not detailed enough, though, so we have to query each individual PR.

pr_watch/github.py
```diff
@@ -183,6 +192,19 @@ def parse_date(date):
...
+    """
+    Check the ratelimit to ensure that we'll be able to successfully make a request.
+    """
+    query_type = 'search' if 'search' in url.split('/') else 'core'
+    response = get_object_from_url('https://api.github.com/rate_limit', rate_limited=False)
+    resource = response['resources'].get(query_type, {})
+    if not resource.get('remaining'):
+        warn_string = 'Rate limit exceeded when requesting a \'{}\' resource at {}'.format(query_type, url)
+        logger.warning(warn_string)
+        raise RateLimitExceeded(warn_string)
```
smarnach: I don't see how this check is useful.  We basically just replace one traceback in the logs by a different one, and instead of trying to make the actual request we make an API request to query the rate limit, so we don't gain anything.  We actually increase the total number of API requests, and we have additional code that can break, so I think we should just try the request and let Github check the rate limit for us.

pr_watch/models.py
```diff
@@ -103,8 +103,8 @@ class WatchedPullRequest(models.Model):
     # TODO: Remove parameters from 'update_instance_from_pr'; make it fetch PR details from the
     # api (including the head commit sha hash, which does not require a separate API call as
     # is currently used.)
-    branch_name = models.CharField(max_length=50, default='master')
-    ref_type = models.CharField(max_length=50, default='heads')
+    branch_name = models.CharField(max_length=255, default='master')
```
smarnach: 255 characters ought to be enough for anybody. :-)

pr_watch/models.py
```diff
@@ -103,8 +103,8 @@ class WatchedPullRequest(models.Model):
...
+    ref_type = models.CharField(max_length=255, default='heads')
```
smarnach: This has only three possible values: `"repository"`, `"branch"` or `"tag"`, so there is no real need to reserve more space in the DB for this.

smarnach:
With the change to do only a single search request for each invocation of the PR watcher, I don't expect the rate limit to be a problem again;  if it is anyway, for the record here is what we can do:

* Run the PR watcher less frequently, like once every five minutes.  Given that deploying a sandbox take 1.5 hours anyway, we don't gain much by running it every minute.
* Use the [Events API|https://developer.github.com/v3/activity/events/] instead.  It's actually meant to be polled, and supports adding an `ETag` header to only retrieve new events.  If there are no new events, the API call does not count against the rate limit.

pr_watch/github.py
```diff
@@ -143,14 +145,21 @@ def get_pr_list_from_username(user_name, fork_name):
...
-    q = 'is:open is:pr author:{author} repo:{repo}'.format(author=user_name, repo=fork_name)
+    return get_pr_list_from_usernames([user_name], fork_name)
+
+
+def get_pr_list_from_usernames(user_names, fork_name):
+    """
+    Retrieve the current active PRs for a given set of users
+    """
+    authors = ' '.join('author:{author}'.format(author=author) for author in user_names)
+    q = 'is:open is:pr {authors} repo:{repo}'.format(authors=authors, repo=fork_name)
```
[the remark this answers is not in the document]
haikuginger: Yeah, I double-checked it just to see if it was an option, and was glad to see that it was. Should save us a massive number of queries.

pr_watch/github.py
haikuginger: @smarnach, I think you're right. At first I was concerned, because GitHub actually just returns a 403 response code when we hit the limit, rather than a much-more-helpful 429, but we can check a header to see if we're out of requests, and use that to make sure we behave properly.

pr_watch/models.py
haikuginger: Ah, good call.

haikuginger:
Rebased from 07effa5.

pr_watch/github.py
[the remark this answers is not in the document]
smarnach: Great find!  I hope they won't "fix" this behaviour. :)

pr_watch/tasks.py
```diff
@@ -46,10 +50,11 @@ def watch_pr():
     organization on the watched repository
     """
     team_username_list = get_username_list_from_team(settings.WATCH_ORGANIZATION)
```
smarnach: Theoretically, we could get rate-limited for this request as well, so you could move it inside the try/except block.

I know the limit for this kind of query is much higher than for search queries, but we now mostly have non-search queries.

haikuginger:
Final comments and squashed from 66aebd2.

PR #228, opened 2017-09-12
Add ansible var for custom certificate templates

cgopalan:
As part of [OC-3011](https://tasks.opencraft.com/browse/OC-3011) and [OC-3032](https://tasks.opencraft.com/browse/OC-3032) a variable needs to be set to enable custom templates for certificates. This PR just sets the variable in the ansible vars so that its set in all OCIM instances by default.

**JIRA tickets**: Needed by [OC-3011](https://tasks.opencraft.com/browse/OC-3011) and [OC-3032](https://tasks.opencraft.com/browse/OC-3032)

**Merge deadline**: None

**Testing instructions**:

1. Spin up an OCIM instance.
2. Verify that the `CUSTOM_CERTIFICATE_TEMPLATES_ENABLED` exists and is set to `true`

**Reviewers**
- [ ] @haikuginger

PR #229, opened 2017-09-20
Enable sysadmin dashboard

clemente:


cgopalan:
@clemente seems like there are build failures here.

pomegranited:
@clemente CC @cgopalan  I just merged a commit that fixes the timeout issues you're seeing with CircleCI.

Could you please rebase this PR on the latest master and try again?

clemente:
@cgopalan @pomegranited I rebased and pushed to trigger the rebuild. The build failures were due to Python3 code in node_modules, I think this was fixed, so this time it should build correctly.

cgopalan:
@clemente how do I test this before merging to see if a new instance spawned has the new variable set? @haikuginger I remember you did this for me on https://github.com/open-craft/opencraft/pull/228 Is this something I would have access to? Or does it have to be tested by a core member?

clemente:
@cgopalan I don't really know the fast way; probably it can be done from a local installation of OCIM.

However… are we suspicious that this variable won't be set in some cases? If all the other variables are being set correctly and all variables are treated equally (this is my supposition), can't we be certain that the deployed instance will have the variable?

haikuginger:
@cgopalan, if you have a local copy of Ocim running in a Vagrant box (there are instructions in this repo), you can create a new instance and spawn an appserver, then terminate the process just so you can look at the variables it creates.

cgopalan:
@clemente I am not sure if the variable will be set or not. I would assume that it would. Since a similar change from me was actually tested before merge by @haikuginger, I was wondering if I could do the same.
@haikuginger if you have a test instance running (or anybody else who has), can you test this? I am not sure that I have the bandwidth right now to start setting up a local instance and fiddling with it. However, if you think this is not necessary, I am good with that too.

haikuginger:
Okay, this is not a formal review, but I can confirm that the variable was added to the `EDXAPP_FEATURES` var as expected.

cgopalan:
@haikuginger thanks!

PR #230, opened 2017-09-22
Enable debug toolbar in DEBUG at all pages including admin and api

clemente:
There's already code to add django-debug-toolbar when `DEBUG==True`, but it's not working (it doesn't show).
- I had to add the setting `SHOW_TOOLBAR_CALLBACK` to force always showing it.
- Then it failed because the URL isn't added, so I added it to urls.py too. Tip from https://github.com/django-debug-toolbar/django-debug-toolbar/issues/529

It's disabled by default, and you can enable it by changing one setting as explained in settings.py

### DIFF
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -49,6 +49,8 @@ ALLOWED_HOSTS = env.json('ALLOWED_HOSTS', default=[])
 
 DEBUG = env.bool('DEBUG', default=False)
 
+ENABLE_DEBUG_TOOLBAR = env.bool('ENABLE_DEBUG_TOOLBAR', default=False)
+
 
 # Auth ########################################################################
 
@@ -108,9 +110,14 @@ MIDDLEWARE = (
     'django.middleware.security.SecurityMiddleware',
 )
 
-if DEBUG:
+if DEBUG and ENABLE_DEBUG_TOOLBAR:
     INSTALLED_APPS += ('debug_toolbar',)
     MIDDLEWARE += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
+    # Enable in all pages
+    DEBUG_TOOLBAR_CONFIG = {
+        'SHOW_TOOLBAR_CALLBACK': lambda request: True,
+    }
+
 
 ROOT_URLCONF = 'opencraft.urls'
 
--- a/opencraft/urls.py
+++ b/opencraft/urls.py
@@ -22,6 +22,7 @@ Global URL Patterns
 
 # Imports #####################################################################
 
+from django.conf import settings
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.views.generic.base import RedirectView
@@ -41,3 +42,12 @@ urlpatterns = [
     url(r'^favicon\.ico$', RedirectView.as_view(url='/static/img/favicon/favicon.ico', permanent=False)),
     url(r'^$', views.IndexView.as_view(), name='index'),
 ]
+
+if settings.DEBUG and settings.ENABLE_DEBUG_TOOLBAR:
+    # Enable debug toolbar URLs
+    import debug_toolbar # pylint: disable=wrong-import-position,wrong-import-order
+    # "debug" URL pattern must be before "site" URL pattern.
+    # See https://github.com/jazzband/django-debug-toolbar/issues/529
+    urlpatterns = [
+        url(r'^__debug__/', include(debug_toolbar.urls)),
+    ] + urlpatterns


opencraft/urls.py
```diff
@@ -41,3 +41,9 @@
     url(r'^favicon\.ico$', RedirectView.as_view(url='/static/img/favicon/favicon.ico', permanent=False)),
     url(r'^$', views.IndexView.as_view(), name='index'),
 ]
+
+if settings.DEBUG:
+    import debug_toolbar
+    urlpatterns = [
```
UmanShahzad: Just use `+=`?

clemente: Ok, I tested and it works the same. I squashed the commits and force-pushed to keep it at 1 commit.

opencraft/urls.py
```diff
@@ -22,10 +22,14 @@
 
 # Imports #####################################################################
 
+from django.conf import settings
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.views.generic.base import RedirectView
 
+if settings.DEBUG:
+    import debug_toolbar
```
smarnach: This causes a linter error, because of the following import.  I suggest moving this inside the `if` statement at the bottom of the file (as I did when I [temporarily fixed this a while ago](https://github.com/open-craft/opencraft/commit/6fa502f4a60f520fceb008206bba124c9d22d55b)).

clemente: I put it at the top to avoid a pylint error. When I had it at the bottom of the file, pylint complained:
{code}
************* Module opencraft.urls
C: 48, 4: Import "import debug_toolbar" should be placed at the top of the module (wrong-import-position)
C: 48, 4: external import "import debug_toolbar" comes before "import opencraft.views as views" (wrong-import-order)
{code}
If I make pep8 pass, then pylint complains, and viceversa.
I now put the import at the end as you suggest and see we'll see whether pylint still complains in remote, and repushed.

UmanShahzad: If it doesn't fix, you could temporarily go with `# pylint: disable=wrong-import-position,wrong-import-order`.

opencraft/urls.py
clemente: I had to revert the `+=` because tests failed. It seems it's important the debug URLs come _before_ the other URLs (https://github.com/jazzband/django-debug-toolbar/issues/529#issuecomment-314851121).

clemente:
I addressed the comments and after the tests finish (2 to 3 h) you can review it.

bradenmacdonald:
@clemente Did you find the toolbar was useful for debugging? I personally usually find it more a nuisance than a help.

smarnach:
@bradenmacdonald For finding out the databse queries caused by an API call, the debug toolbar can be really useful.  I agree, though, that it should be disabled by default.

clemente:
I added a commit to disable it by default, and some code to toggle it easily with only 1 setting (instead of commenting INSTALLED_APPS and urls).

smarnach:
@clemente The debug toolbar has a significant and very noticeable performance impact.  When it is disabled, I'd rather the middleware wouldn't slow everything down.  So maybe add a new configuration setting, like `ENABLE_DEBUG_TOOLBAR`, which is configured from the environment, and replace `if DEBUG:` by

    if DEBUG and ENABLE_DEBUG_TOOLBAR:

opencraft/settings.py
```diff
@@ -49,6 +49,8 @@
 
 DEBUG = env.bool('DEBUG', default=False)
 
+ENABLE_DEBUG_TOOLBAR = False
```
smarnach: This should be

    ENABLE_DEBUG_TOOLBAR = env.bool('ENABLE_DEBUG_TOOLBAR', default=False)

so we can change it via the `.env` or the environment if needed.

clemente:
> Can you please also add something to the .env file in our internal documentation? e.g.

@bradenmacdonald https://github.com/open-craft/documentation/pull/198

bradenmacdonald:
Thanks @clemente . Please rebase and merge this.

clemente:
I'm getting this error in CircleCI tests, even after restarting:

> requests.exceptions.ConnectionError: HTTPSConnectionPool(host='8e1bd224.integration.plebia.net', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x2b7775269438>: Failed to establish a new connection: [Errno -2] Name or service not known',))

bradenmacdonald:
@clemente Yeah, I see there haven't been any successful builds in the last 20 hours. Seems to be that DNS issue and `Can't take a write lock while out of disk space`. I may be able to help you with this tomorrow, but in the meantime since the issue with CI seems unrelated to your current task, @UmanShahzad or @bdero would you be able to investigate and help unblock @clemente ?

UmanShahzad:
@bradenmacdonald @clemente I'll try and investigate this quite soon.

UmanShahzad:
@bradenmacdonald @clemente It looks like the DB is legitimately out of space, as `db.collection.stats()` being run while connected to our integration test's Mongo instance returns the same error. I can technically select one of the integration test DBs and drop them -- is that safe to do here?

clemente:
@UmanShahzad which DB, and is there something I can do in the code to reduce the disk space requirements? I'll let @bradenmacdonald answer about dropping other DBs.

UmanShahzad:
@clemente I'm referring to the MongoDB instance used for integration tests for OpenCraft (requires access). Not sure if this is related to your code at all, but maybe just that a lot of CircleCI builds have been happening in a relatively short amount of time.

clemente:
@UmanShahzad The recent builds I did include changes in which I call Ocim's `instance.tests.models.factories.openedx_appserver.make_test_appserver` many times (e.g. 10). This runs very fast in local but I hope it's not spawning real instances in CircleCI.

smarnach:
@UmanShahzad What Mongo instance are we using for the integration tests?  Is it some mlab sandbox, or one of our own servers?  In any case, if you can see collections unambiguously belonging to the integration tests, you can delete them all.  The worst that can happen is that some integration test run fails, which is currently happening anyway.

UmanShahzad:
@smarnach Yeah, they are very clearly integration tests because I think the whole VM is meant just for them. I'll post more details on the ticket for reference, though.

UmanShahzad:
@clemente I've deleted the old integration test DBs that were clogging the integration test server's MongoDB, and restarted your build.

clemente:
@UmanShahzad  Thanks! I will be restarting builds and merging.  The build in this ticket is still failed, it wasn't restarted; I'll restart it manually later (now the queue is still full).

PR #234, opened 2017-09-29
[OC-2042] Constrain subdomains to be RFC 1034-conformant.

UmanShahzad:
# Problem

As it stands, a beta tester can register a subdomain that isn't RFC 1034-conformant, or that conflicts with another without raising any errors by changing letter case.

# Solution

This PR introduces a regex based validation scheme to ensure RFC 1034 conformance. It also cleans the subdomain to be lower case.

# Testing

1. Open up a local Ocim and go to `/registration/`.
1. Try subdomains you'd expect to fail under RFC 1034, and ensure that they do.
1. Try subdomains you'd expect to pass under RFC 1034, and ensure that they do.
1. Register the instance with a subdomain in all caps.
1. Check the instance in the admin panel and ensure that the subdomain is actually in lower case.

# Remaining work

The remaining work must be done outside of this PR, and that is:

1. Convert any existing non-conformant subdomains to meet RFC 1034 standards. This will require contacting the instance owner. This includes making sure all subdomains are lower case.
1. We may need to add some subdomains to the `SUBDOMAIN_BLACKLIST` setting in our production Ocim instance.
1. There are some extra details of remaining work on the internal ticket.


registration/models.py
```diff
@@ -76,8 +76,9 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
                 'The subdomain name must at least have 3 characters.',
             ),
             validators.RegexValidator(
-                r'^[\w.-]+$',
-                "Please include only letters, numbers, '_', '-' and '.'",
+                r'^[a-zA-Z0-9](?:[a-zA-Z0-9\-]{1,63}[a-zA-Z0-9])$',
```
smarnach: If I remember correctly, the maximum length of a domain name is 63 characters, but this regex matches names of up to 65 characters in length.

I suggest enforcing the length via the `max_length` property of the field rather than the regex.  This would allow us to simplify the regex to 

    r'^[a-zA-Z0-9]([a-zA-Z0-9\-]+[a-zA-Z0-9])?$'

This would also make the `MinLengthValidator` redundant, but I'd keep it for clarity (and in case we want to increase the minimum length).

UmanShahzad: Ah, that sounds good. I'll address this when we pick the ticket back up.

UmanShahzad:
Closing in favor of #257.

PR #235, opened 2017-10-05
Integrate with simple_theme: add design fields to instance registration form, send e-mail when values change, include ansible variables

clemente:
This adds extra fields to Ocim's instance registration form to ask about 4 colors, favicon and logo.
Files are uploaded to SWIFT and values are saved in DB.
This is done for https://tasks.opencraft.com/browse/OC-3082

In addition, when it detects changes to the saved values, it sends an e-mail to the address configured in the .env. This was part of a future task, but it's already implemented. It sends e-mails only when fields change, not on creation. Note that you can't change the instance name.

In addition, it implements https://tasks.opencraft.com/browse/OC-3259 : it changes the ansible variables to be able to use simple-theme (https://github.com/open-craft/configuration/tree/clemente/ansible-theme), so that the provisioning will use them and the instance will have the desired colors and logo/favicon. 

It includes an integration test to check that the spawned instance uses the provided files and colors.

# Testing 1: form fields and e-mail
1. Add to your .env settings the settings from https://github.com/open-craft/documentation/pull/203 (use your e-mail in VARIABLES_NOTIFICATION_EMAIL, and use all the values which are in comments there; you can use my SWIFT test container too)
2. Migrate the DB (there are minor migrations)
3. Start Ocim, Go to http://localhost:5000/registration/
4. Check that there are 6 design fields, that they show a default color (not only black), and that they're using your browser's color chooser
5. Play with the form, e.g. send it without choosing and image and see it fail
6. Send the form with all data and your favourite colors
7. Notice that no extra message about design fields should be shown, only the ones about application status
8. Notice that no e-mail was sent to us about design changes
9. Go back to the form
10. Check that it has your favourite colors and both the images, and they are served from an OVH address
11. Update colors and images, and save
12. Check that the new settings are saved, and that on top of the form there's a message saying we will review the changes
13. Check (in console) that an e-mail was sent with the information of the new colors

# Testing 2: provisioning
1. Create a SWIFT container and assign it to MEDIAFILES_SWIFT_CONTAINER_NAME in .env
2. Follow the comment above the field `logo` in `registration/models.py` in order to upload the default logo and favicon to your SWIFT container
3. Get yourself a working Ocim that can deploy (good MySQL servers, versions, …) or adjust values later from admin if it fails.
4. In your .env (or later in the admin if you prefer) choose configuration repository `https://github.com/open-craft/configuration/` and configuration commit `clemente/ansible-theme`, so that simple_theme be used. That branch is near master. You can use ginkgo.1 for the other branches.
5. Create a new user through the registration form, and note which colors you choose. Don't upload a logo but do upload your icon (or the other way around) so that one of them stays with the default value (OpenCraft's)
6. Validate the e-mail of that user, and let the provisioning start. In the admin check that the configuration version points to the simple-theme branch, and that everything else is correct (e.g. ginkgo.1), if not, modify the instance values and create a new appserver.
7. Check in the instance settings page that the `configuration_extra_settings` contains SIMPLETHEME_… variables with the colors you choose.
8. Wait 2 hours. Check the log, the first time the simple_theme role will be skipped (because of: celery workers), the 2nd one it will collect files (including downloading them from SWIFT)
9. Ocim will finish; check the appserver through the authenticated link, you should see the colors and logo/favicon (if you let the default value, you should see OpenCraft's logo/favicon).
10. Change the values in the registration form as a user. Notice (from the instances list admin) that the ansible variables in the instance (configuration_extra_settings) have been updated.

# Testing 3
1. The automated integration test should run in CircleCI (2h). The code already contains (in circle.yml) some references to a SWIFT container used to upload files during the test.


antoviaque:
@clemente Can you ping me for a product review on this, once the code is up on a sandbox/stage server somewhere, and before merging? Thanks!

registration/forms.py
```diff
@@ -355,6 +385,41 @@ def update_user(self, application, commit=True):
                 application.user.profile.save()
                 application.save()
 
+        if self.restart_fields_changed():
+            messages.add_message(self.request, messages.INFO,
+                                 "Thank you for submitting these changes - we will rebuild your instance to "
+                                 "apply them, and email you to confirm once it is up to date.")
```
bdero: I noticed that the "Thank you for applying for the OpenCraft beta. We will contact you shortly" message still shows up when submitting changes as well. The reason for this is that this "Thank you" message shows up whenever the [application is pending](https://github.com/open-craft/opencraft/blob/d891795e7bf7a568222331859403e9584dbf22c3/registration/templates/registration/registration.html#L27-L28), so I don't feel it's that weird.

registration/forms.py
```diff
@@ -355,6 +385,41 @@ def update_user(self, application, commit=True):
...
+                       ).format(
+                           instance_id=application.instance_id,
+                           domain=application.domain,
+                           main_color=application.main_color,
+                           link_color=application.link_color,
+                           bg_color_1=application.bg_color_1,
+                           bg_color_2=application.bg_color_2,
+                           logo=application.logo.url,
+                           favicon=application.favicon.url,
+                       )
```
bdero: @clemente Could you place this email template into a separate file, and then render it using the django template engine instead?

registration/forms.py
```diff
@@ -355,6 +385,41 @@ def update_user(self, application, commit=True):
...
+                        "registration form.\n"
+                        "\n"
+                        "The new values are:\n"
+                        "- main color: {main_color}\n"
+                        "- link color: {link_color}\n"
+                        "- primary background color: {bg_color_1}\n"
+                        "- secondary background color: {bg_color_2}\n"
+                        "- logo: {logo}\n"
+                        "- favicon: {favicon}\n"
+                       ).format(
```
bdero: @antoviaque, could you take a look at the content of this email to see if it'll be sufficient for the betatest workflow needs?

registration/forms.py
clemente: Yes, this was a conscious decision, one message is for the changes, the other for the status. Both apply. To clarify, to could prepend „Application status: “ to the 2nd one. But I don't think it's needed.

registration/forms.py
clemente: Note that I updated it (I removed "[OpenCraft]: " from the subject).
Check also the sending address (now: `settings.DEFAULT_FROM_EMAIL`).

UmanShahzad:
@clemente This isn't on stage anymore, because it's out of date -- can you rebase it? I'll update stage again then.

clemente:
@UmanShahzad rebased

UmanShahzad:
@clemente Thanks :) Unfortunately stage seems a bit busy lately so...

clemente:
Corrected test and re-rebased.

registration/forms.py
antoviaque: @clemente Yes, that seem fine. Maybe just remove that exclamation mark - from both actually. It looks like something went wrong, when it's just informative.

registration/forms.py
antoviaque: @bdero @clemente That looks good yes. When we receive this on help@, the only action the  @aliciaopencraft17 would need to do then would be to start a new VM and activate it, right? Could you write down exactly these steps, so she can keep this in one of her recurring tasks?

aliciaopencraft17: @clemente Daniel, I created https://tasks.opencraft.com/browse/ADMIN-58 for this. Thank you in advance for writing down these steps - please keep in mind to "dumb" down the language and explain acronyms for me to try and understand it from a non-dev mind :) I might still have questions but hopefully this will reduce them. Thank you!
cc @antoviaque

antoviaque: @aliciaopencraft17 Thank you!

clemente: @antoviaque @aliciaopencraft17 Yes it will be like that, Alicia will only need to click the „Launch new appserver“ button. I added instructions to ADMIN-58, including manual checks that might be necessary if we need to debug things.

registration/forms.py
clemente: I removed the exclamation marks.

registration/forms.py
antoviaque: @clemente Thank you!

clemente:
Rebased.

static/scss/components/_components.scss
```diff
@@ -14,6 +14,8 @@
         position: relative;
         color: currentColor;
 
+        /*
+        // Disabled exclamation icon because it looks like something went wrong, but in fact it's informative. Maybe look for better icon
```
mtyaka: Please remove the code rather than commenting it out if we don't need it.

registration/forms.py
```diff
@@ -355,6 +391,32 @@ def update_user(self, application, commit=True):
...
 
+        if self.restart_fields_changed():
+            messages.add_message(self.request, messages.INFO,
+                                 "Thank you for submitting these changes - we will rebuild your instance to "
+                                 "apply them, and email you to confirm once it is up to date.")
+
+            # Recreate the ansible variables in the instance
+            instance = application.instance
+            if instance:
+                instance.configuration_extra_settings = get_design_fields_as_yaml(application)
```
mtyaka: Simple theme configuration variables should not be stored in `configuration_extra_settings`. The `configuration_extra_settings` field should only be used for settings which are entered manually and are not automated by Ocim.

Add a new field `configuration_theme_settings` instead. You can check how this is done for storage settings, which are stored in `configuration_storage_settings` and most related code is contained in the [openedx_storage.py mixin](https://github.com/open-craft/opencraft/blob/0afe1e19eeb1c6fc2b338294ba1c1cbddd64b365/instance/models/mixins/openedx_storage.py). The value [get set whenever creating a new appserver](https://github.com/open-craft/opencraft/blob/0afe1e19eeb1c6fc2b338294ba1c1cbddd64b365/instance/models/openedx_instance.py#L228). If you grep the code for `configuration_storage_settings`, you will see how it all fits together. Please follow the same approach for `configuration_theme_settings`.

registration/provision.py
```diff
@@ -65,11 +83,15 @@ def _provision_instance(sender, **kwargs):
...
         return
+
+    design_fields_yaml = get_design_fields_as_yaml(application)
+
     with transaction.atomic():
         application.instance = production_instance_factory(
             sub_domain=application.subdomain,
             name=application.instance_name,
             email=application.public_contact_email,
+            configuration_extra_settings=design_fields_yaml,
```
mtyaka: As noted in another comment, these settings should not be stored in `configuration_extra_settings`. Also, I don't think they need to be set at this stage at all - they can be set when spawning a new app server (on the new appserver itself).

registration/forms.py
```diff
@@ -25,13 +25,18 @@
...
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.models import User
+from django.core.mail import send_mail
 from django.db import transaction
 from django.utils.text import capfirst
+from django.template.loader import get_template
 from djng.forms import NgDeclarativeFieldsMetaclass, NgFormValidationMixin, NgModelForm, NgModelFormMixin
 
 from registration.models import BetaTestApplication
+#from registration.provision import get_design_fields_as_yaml
```
bdero: Please remove this commented code.

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -0,0 +1,130 @@
...
+                     'value': '#ffff11', },
+                    {'variable': 'button-color',
+                     'value': '#001122', },
+                    {'variable': 'action-primary-bg',
+                     'value': '#001122', },
+                    {'variable': 'action-secondary-bg',
+                     'value': '#001122', },
+                ],
+                'SIMPLETHEME_STATIC_FILES_URLS': [
+                    {'url': 'static/img/png/opencraft_logo_small.png',
```
mtyaka: These don't look like URLs, is that expected?

clemente: I changed that in the other branch: https://github.com/open-craft/opencraft/blob/0eeb150c248ee65f7654c00ff1481115a73d224d/instance/tests/models/test_openedx_theme_mixins.py#L110-L114
I had to [modify](https://github.com/open-craft/opencraft/blob/0eeb150c248ee65f7654c00ff1481115a73d224d/circle.yml#L11) CircleCI to do SWIFT uploads because otherwise simple_theme can't get the logo/favicon. But the local tests in my computer don't upload to SWIFT, I think because they're using `.env.test`, and probably it should be like that (fast, mocked); there's no simple_theme in local tests and therefore noone needs to download from remote URLs. Now I allow both values: non-SWIFT (that is, non-URLs), and SWIFT (with valid URLs). I welcome ideas for cleaner approaches.

clemente:
To make it easier to merge, I squashed everything into two commits, one for integration, another for the test. This is now the main branch, to one to merge, it includes everything.
I have copies of the old branch (unsquashed) at https://github.com/open-craft/opencraft/tree/clemente/ui-choose-colors-backupbeforesquash , and the other one with the integration test (https://github.com/open-craft/opencraft/tree/clemente/ui-choose-colors-integration-test, unsquashed at https://github.com/open-craft/opencraft/tree/clemente/ui-choose-colors-integration-test-backupbeforesquash) isn't needed anymore.

instance/tests/integration/factories/instance.py
```diff
@@ -64,3 +64,16 @@ def create(cls, *args, **kwargs):
...
+
+
+class OpenEdXInstanceFactoryWithSimpleTheme(OpenEdXInstanceFactory):
+    """
+    Factory for OpenEdXInstance including simple_theme. It requires different branches
+    because simple_theme wasn't merged upstream yet.
+    """
+    # This branch is based on 'integration-ginkgo' and has simple_theme role on top
+    configuration_source_repo_url = 'https://github.com/open-craft/configuration.git'
+    configuration_version = 'integration-ginkgo-with-simpletheme'
```
mtyaka: @clemente The simpletheme role looks pretty good to me, so we can just merge it into `integration-ginkgo` and remove this temporary `OpenEdXInstanceFactoryWithSimpleTheme` class.

Can you please open a PR against `integration-ginkgo` branch to add simpletheme role? I'll approve it and we can merge, then remove this class.

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -0,0 +1,138 @@
...
+                     'value': '#001122', },
+                ],
+                'EDXAPP_DEFAULT_SITE_THEME': 'simple-theme',
+                # for SIMPLETHEME_STATIC_FILES_URLS, see below
+            }
+            for ansible_var, value in expected_settings.items():
+                self.assertEqual(value, parsed_vars[ansible_var])
+
+            # Testing of files is a bit special because if we use SWIFT, we will see URLs, but in local tests we have
+            # SWIFT disabled so it's ok to see file names. We allow for both.
```
mtyaka: Is this comment still correct? It looks like we are always expecting URLs now.

instance/tests/integration/factories/instance.py
clemente: PR: https://github.com/open-craft/configuration/pull/26
I'll squash changes again after doing this change.

instance/tests/models/test_openedx_theme_mixins.py
clemente: The problem is in tests run locally with .env.test, because there's no SWIFT upload there (and probably there shouldn't be? They're supposed to be local). I think that if I accept only URLs, working non-SWIFT will fail. We could mock the SWIFT storage method but it would be code only for tests run in local. I thought of adding an `if`: „if using SWIFT storage, then check this; if not check this other thing“, would that be clearer?

mtyaka: I might be missing something, but in [this line](https://github.com/open-craft/opencraft/pull/235/files/af68a8c23ffb9384340c4a4d8c33a9884485ea3f#diff-4d7dfc4152cfcf430d6487dd0f2ff228R115), it looks like we are asserting that `logo['url']` is in fact a URL pointing to OVH storage? Does the test still work locally?

I actually think it's totally fine to require SWIFT to be configured even when developing locally. We currently require OVH access to be able to spawn VMs anyway. Unless you disagree, I think it's fine to assume SWIFT storage is set up even when running in the devstack, and just remove the comment.

clemente: It doesn't accept only URLs. In the regexp there's a `?` to make the left URL path optional, so it allows both.
So then I'll remove the regexp thing and accept only URLs, ok? I'll add some comment to .env.test

mtyaka: Ah, sorry I see... I knew I was missing something.

If I understand correctly, when running locally and you don't have swift set up `url` entries will be paths to local files. Does the simpletheme role task that's using `get_url` work if you pass it paths to local files?

clemente: No it doesn't work, it crashed when I tried (ansible's `get_url` expects a URL). The correct way to include local files in the theme would be through the other variable `SIMPLETHEME_STATIC_FILES_DIR`

mtyaka: Ok, in that case we just assume SWIFT is always enabled in dev environment.

clemente:
@mtyaka I did those 2 changes and nothing more (nothing in .env.test) and squashed (sorry, I know it makes review harder, but then the merge will be faster).
I had a long night with headache and I need to go to sleep (I'm working at nights lately); later in the evening I'll handle any other comment (or test failure), or in the weekend if it's ok.

clemente:
The last test said „Success“ but it actually failed. I fixed whitespace and repushed.

clemente:
@mtyaka Ok, now the tests passed! Any other task before it can be merged? When can we merge it to Ocim production and test it there?

mtyaka:
By the way, the latest code is deployed on staging.

UmanShahzad:
(note to also rebase for conflicts)

UmanShahzad:
@mtyaka I think the SWIFT container in GRA3 won't be used, because this update is only for edxapp VMs; we're leaving SWIFT on GRA1 for now. Or did I misunderstand/miss something?

clemente:
@mtyaka Actually the goal was just to remove an asterisk. I made these fields „not required“ but then Django added the whole business of clearing them, which broke UI and tests.
I think clearing a logo and favicon isn't a feature we need to support (with tests and extra code and UI). Someone who has already chosen a logo could want to *change* it to another logo, but I bet no client has ever requested to remove their logo once they gave one. So I'll try to revert these migrations and do another way to just remove the asterisk in those 6 fields.

clemente:
Rebased on master; fixes still to do.
@mtyaka @UmanShahzad Then it's ok to use GRA1 SWIFT for client logos/icons? I didn't find the container there, so I can create one in GRA1 (`simpletheme_files`) and update the variables in CircleCI as needed.

mtyaka:
@UmanShahzad 

>I think the SWIFT container in GRA3 won't be used, because this update is only for edxapp VMs; we're leaving SWIFT on GRA1 for now. Or did I misunderstand/miss something?

This container is different from individual SWIFT containers assigned to each instance - it's a global container for all theme related files uploaded in Ocim. It used to be created in the same OpenStack account as edxapp VMs, but we have a separate account for edxapp SWIFT containers, because those credentials are shared between all VMs.

We could potentially keep using the old container in GRA1 region where our old edxapp VMs are located, but I think it will be less confusing if we just move it to the new GRA3 region/account.

UmanShahzad:
@clemente Yeah the situation is that SWIFT containers for integration tests with CircleCI are on GRA3, because those are always new instances. But stage SWIFT is on GRA1. Please only update CircleCI if you need to make the integration tests use GRA1 SWIFT containers for some reason.

UmanShahzad:
@mtyaka Ah, ok, that makes sense!

mtyaka:
@clemente 

> So I'll try to revert these migrations and do another way to just remove the asterisk in those 6 fields.

That sounds good 👍

> Then it's ok to use GRA1 SWIFT for client logos/icons?

I think it would make more sense to use the new container in GRA3 rather than the old one in GRA1 (see above).

clemente:
I did the asterisk fix and I'll be fixing the tests which fail; I'm curious to see what happens with a failing test that requests F,G,E servers to be redeployed in a certain order.
```
FAIL: test_redeployment_failure (instance.tests.management.test_instance_redeploy.InstanceRedeployTestCase)
[…]
AssertionError: sequence not as expected:
```
One day it changed from FFGGEE to GGEEFF. I'll see the expected order and adapt the test to match the results.

static/scss/components/_forms.scss
```diff
@@ -172,6 +172,23 @@ form .input
...
+
+	// The following fields (colors and logo/favicon) are still required, but we 
+	// prefer not to show the "*" because the user isn't required to change them
+	// (the defaults are good and it's not even possible not to submit a color)
+    &--required[for="id_header_bg_color"],
+	&--required[for="id_footer_bg_color"],
+	&--required[for="main_color"],
+	&--required[for="link_color"],
+	&--required[for="id_logo"],
+	&--required[for="id_favicon"]
```
mtyaka: Nit: when you squash, can you please change the indentation so that all lines are indented the same amount as the first line?

mtyaka:
Thanks @clemente! Deployed to stage. All that remains is to squash the commits, then we can merge and deploy to production.

It looks like the order in the test you had to change is not entirely deterministic, so we'll probably have to rethink how we test that in the future (or make the ordering deterministic), but that's out of scope of this task.

mtyaka:
:+1:

- [x] I tested this: I spawned several appservers with theming enabled, with different combinations of theme variables.
- [x] I read through the code
- [x] Includes documentation - inline documentation on the form

clemente:
@mtyaka I fixed the indentation and squashed and repushed right now. After tests pass, can you merge and deploy it to production? Thanks

clemente:
@mtyaka now that the django-wiki and the nltk bugs are fixed, I restarted the tests to see if they pass; if yes then it can be installed in production.

PR #236, opened 2017-10-11
Protect references to Ansible variables from the Django template engine.

smarnach:
The Ansible variable templates get rendered by two different template engines with similar syntax.  First the whole file is rendered by the Django template engine before being passed on to Ansible.  Ansible then renders the values with the Jinja2 template engine.  The syntax for variable references in both template engines is identical, so all variable references that are meant to be rendered by Jinja2 are already rendered by Django.  Django ignores unknown variables and replaces them by empty strings.

This fix protects all variables that should actually be rendered by Jinja2 with verbatim blocks.

### DIFF
--- a/instance/templates/instance/ansible/swift.yml
+++ b/instance/templates/instance/ansible/swift.yml
@@ -4,8 +4,6 @@ EDXAPP_SETTINGS: 'openstack'
 XQUEUE_SETTINGS: 'openstack_settings'
 
 VHOST_NAME: 'openstack'
-# COMMON_VHOST_ROLE_NAME is deprecated; keeping it around for compatibility
-COMMON_VHOST_ROLE_NAME: '{{ VHOST_NAME }}'
 
 EDXAPP_DEFAULT_FILE_STORAGE: 'swift.storage.SwiftStorage'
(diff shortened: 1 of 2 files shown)


smarnach:
@bdero I've deployed this to stage and started to provision and [instance](https://stage.console.opencraft.com/instance/1400/) for testing.

bdero:
:+1:

- [x] I tested this: I verified that the new instance spawned without errors, and that all of the text surrounded by verbatim tags contain only vars that should be rendered by Ansible.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ N/A.
- [ ] ~Includes documentation~ Bug fix.

smarnach:
For the record, I don't think this is the optimal solution.  Having two different template engines with similar syntax rendering the same strings calls for errors.  We can't get rid of the second one, Jinja2, so we should get rid of the first one, the Django one.  There isn't really a good reason to use a template engine here anyway.  Rewriting these templates as simple Python functions returning a dictionary and taking the context variables as parameters seems the better approach in all regards.  The dictionaries can then easily be rendered as YAML.

I'm merging this PR now anyway, since we need the fix in production.  I'll create a task for the later cleanup.

PR #237, opened 2017-10-14
Use .only() to reduce the amount of information read from DB in the instances list API

clemente:
This avoids loading all unneeded columns of OpenEdXAppServer models.

Before using `only`:
image

After using `only`:
image

### DIFF
--- a/instance/api/instance.py
+++ b/instance/api/instance.py
@@ -72,16 +72,22 @@ class InstanceViewSet(viewsets.ReadOnlyModelViewSet):
         List all instances. No App server list is returned in the list view, only the newest app server information.
 
         """
+        # Don't load all columns, because some of them have very big data
+        appservers_few_columns = OpenEdXAppServer.objects.only('_is_active', '_status', 'id', 'name', 'owner_id',
+                                                               'created', 'modified')
         queryset = self.queryset.prefetch_related(
             # Use prefetching to make the number of database queries required to
             # generate this list O(1).
-            # Note that prefetching all app server information is still required, as the "newest" is not decideable
+            # Note that prefetching all app servers information is still required, as the "newest" is not decideable
             # at this point. This will cause more data than necessary to be streamed from the DB, but removing this
             # prefetch without first selecting only the "newest" here results in O(n).
-            Prefetch('instance__ref_set__openedxappserver_set'),
             Prefetch(
                 'instance__ref_set__openedxappserver_set',
-                queryset=OpenEdXAppServer.objects.filter(_is_active=True),
+                queryset=appservers_few_columns,
+            ),
+            Prefetch(
+                'instance__ref_set__openedxappserver_set',
+                queryset=appservers_few_columns.filter(_is_active=True),
                 to_attr='_cached_active_appservers'
             ),
         )


instance/api/instance.py
```diff
@@ -72,16 +72,22 @@ def list(self, request):
...
 
         """
+        # Don't load all columns, because some of them have very big data
+        appservers_few_columns = OpenEdXAppServer.objects.only('_is_active', '_status', 'id', 'name', 'owner_id',
+                                                               'created', 'modified')
         queryset = self.queryset.prefetch_related(
             # Use prefetching to make the number of database queries required to
             # generate this list O(1).
-            # Note that prefetching all app server information is still required, as the "newest" is not decideable
+            # Note that prefetching all app servers information is still required, as the "newest" is not decideable
```
UmanShahzad: The older comment was actually grammatically correct. :p

instance/api/instance.py
```diff
@@ -72,16 +72,22 @@ def list(self, request):
...
             # at this point. This will cause more data than necessary to be streamed from the DB, but removing this
             # prefetch without first selecting only the "newest" here results in O(n).
-            Prefetch('instance__ref_set__openedxappserver_set'),
             Prefetch(
                 'instance__ref_set__openedxappserver_set',
-                queryset=OpenEdXAppServer.objects.filter(_is_active=True),
+                queryset=appservers_few_columns,
+            ),
```
UmanShahzad: I think the `Prefetch` below (minus caching) will become redundant because this one already contains all app servers where `_is_active == True`.

Let me know if I misunderstood something.

instance/api/instance.py
clemente: It's not about grammar, it's about the word „all“. Before the patch, it was really loading all „server information“. Now, it's not loading all information, just a part of it. But it's still true that it's loading information for all app servers…
I can change it to something else.

UmanShahzad: Nevermind, it's not that important :)

UmanShahzad:
I tested this locally and can confirm that the SQL query is much more slim now. Will put it on stage to test there.

clemente:
About your other comment (deleted?):
I know, the first one contains all of them (active & non-active), so Django still wants to do a 2nd query to find which ones are active and which not. I tried to remove the top Prefetch, and also the bottom Prefetch, and in both cases Django ends up doing too many small queries (so: no prefetch).
The comment about the Prefetch also speaks about this.
However, I expect there must be a Django solution to remember 1 column (`is_active`) and not go to DB again, but I don't know which one.

UmanShahzad:
@clemente Yeah I removed it, after testing locally with different combos and realizing it just doesn't work unless we use 2 prefetches.

clemente:
Do I need a rebase? github says „This branch is out-of-date with the base branch“ but I don't see how; my local git says that it's up to date when I do rebase.

UmanShahzad:
@clemente Most likely you haven't pulled the latest `master`.

smarnach:
I've deployed this to production, since I needed to deploy anyway.

UmanShahzad:
@smarnach Thank you!

PR #239, opened 2017-10-16
To deal with memory issues, default to less edxapp workers.

UmanShahzad:
We've noted some memory issues that can be solved by reducing the number of LMS/CMS workers.

**Testing instructions**:

1. See stage for a new instance (TBD).
1. See its configuration setup and make sure you see:

```yaml
EDXAPP_WORKERS:
  lms: 3
  cms: 2
```

### DIFF
--- a/instance/templates/instance/ansible/vars.yml
+++ b/instance/templates/instance/ansible/vars.yml
@@ -173,7 +173,12 @@ COMMON_USER_INFO:{% for github_username in github_username_list %}
 {% endif %}
 {% endwith %}
 
-# Workers
+# Gunicorn workers
+EDXAPP_WORKERS:
+  lms: 3
+  cms: 2
+
+# Celery workers
 EDXAPP_WORKER_DEFAULT_STOPWAITSECS: 1200
 
 # Monitoring


instance/templates/instance/ansible/vars.yml
```diff
@@ -174,6 +174,9 @@ COMMON_USER_INFO:{% for github_username in github_username_list %}
 {% endwith %}
 
 # Workers
+EDXAPP_WORKERS:
+  lms: 3
+  cms: 2
```
smarnach: The naming of the Ansible variables is terrible. `EDXAPP_WORKER_DEFAULT_STOPWAITSECS` is about the Celery workers, while the one you added is about the Gunicorn workers.  I propse to introduce two different sections, with comments that make clear which workers are meant.

(As an aside, the variable to set the number of Celery workers is called `edxapp_workers`, so the same, just in lower case.)

UmanShahzad: That makes me facepalm so hard. Will change it.

UmanShahzad: 2af57223a3fb15d500b22c32453ed361fea43c0b (squashed and pushed, to prevent multiple CircleCI builds)

PR #240, opened 2017-10-17
[OC-3157,2517] Handle potentially wasteful load balancer reconfigurations internally

UmanShahzad:
This PR introduces some logic internal to the load balancer API that makes it so that the load balancer is not reconfigured multiple times wastefully/redundantly whenever a call to reconfigure/deconfigure the load balancer is made.

In the process it also updates the mass redeployment command, to take into account this update to the load balancer. Previously, we were trying to be quite careful with not reconfiguring the load balancer for each instance in a batch. Now, we can ignore this, and focus on mass redeployment-related logic only, and let the LB itself handle preventing wasteful reconfigurations.

This also introduces a small task that runs every minute -- it makes sure that any load balancer that requires reconfiguration (marked `dirty` for when potentially redundant/wasteful calls are made to the LB model API for reconfiguration), gets reconfigured, but no others.

**Testing Instructions**:

I think the best way to test the mass redeployment section of this PR would be to just run the mass redeployment command on stage. There are currently 4 instances there. What we should do is run the following command:

```bash
make manage "instance_redeploy --tag oc-3157-2517 --num-attempts 2 --batch-size 2" >& ~/oc-3157-2517.log
```

For one, we're obviously looking to see if the mass redeployment command still works correctly given these changes, and we'll find that out inevitably.

More specifically, we want to know the following:

1. Are new appservers that provision successfully marked active?

We can figure this out by just looking at the GUI and seeing whether the UI shows the latest appserver as being active.

2. Are old appservers deactivated?

We can figure this out by just looking at the GUI and seeing whether the UI shows all appservers besides the latest as being inactive.

---

For the LB portion of these changes, we *could* try and see the LB configuration version after the mass redeployment command succeeds, but I'm not sure exactly what numbers to look for and I think it depends a lot on whether 2 appservers both complete provisioning at roughly the same time (within ~10 minutes of each other).

Instead, we can try this:

1. Spawn new appservers for 2 different instances.
1. Take note of the LB deployed configuration version & the normal configuration version.
1. When provisioning completes for both of them, manually activate 1.
1. Confirm quickly that configuration version > deployed configuration version.
1. Manually activate the other appserver.
1. Confirm that configuration version is now 1 greater than before, and deployed configuration version hasn't changed yet.
1. Wait about 10-15 minutes and come back to see that deployed configuration version is 1 less than the normal configuration version.
1. Wait about 10-15 minutes and come back to see that the values are still the same.

I think the above steps may be overkill, I'm not really sure. Let me know if any of these steps can be squashed / made simpler.


instance/tasks.py
```diff
@@ -81,13 +83,17 @@ def spawn_appserver(
...
     """
     Mark an AppServer as active or inactive.
     """
     logger.info('%s AppServer: ID=%s', "Activating" if active else "Deactivating", appserver_id)
     appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
     appserver.make_active(active)
+    if deactivate_old:
+        for appserver_to_deactivate in appserver.instance.appserver_set.filter(_is_active=True):
+            logger.info(' - Deactivating %s [%s]', appserver, appserver.id)
+            appserver_to_deactivate.mark_active(active=False)
```
smarnach: I think this will trigger a LB reconfiguration twice.  The LB reconfiguration is a rather slow process (taking minutes), so it would be worth avoiding that.  I like Jill's suggestion to add a flag to `make_active()` to disable the LB reconfiguration.

instance/management/commands/instance_redeploy.py
```diff
@@ -276,6 +297,37 @@ def _do_mysql_commands(self, instance):
...
+            for appserver in instance.appserver_set.filter(_is_active=True):
+                LOG.info(' - Deactivating %s [%s]', appserver, appserver.id)
+                appserver.is_active = False
+                appserver.save()
+
+            # Activate the latest server
+            LOG.info(' - Activating %s [%s]', latest_appserver, latest_appserver.id)
+            latest_appserver.is_active = True
+            latest_appserver.instance.enable_monitoring()
+            latest_appserver.instance.set_active_vm_dns_records()
```
smarnach: We could avoid having to call these function explicitly here by adding a flag to `make_active()` to disable the LB reconfiguration.  We basically want `make_active()` to do everything except for the LB reconfiguration, since that is the slowest bit.

instance/tasks.py
UmanShahzad: Yes, you're right. The flag suggestion makes sense, and I'm going to go down that route. Thanks!

instance/management/commands/instance_redeploy.py
UmanShahzad: Yep, saw the comment below. Makes sense -- will add a flag.

instance/tasks.py
antoviaque: Could we instead make `make_active()` handle concurrent requests for reconfiguration gracefully? That would allow to avoid exposing its internals to functions that need to call it but not know what it does, like this, along with the mental overload that comes with it.

What about having the LB reconfiguration call turned into just flagging the LB config as modified, and needing a reconfiguration - and the actual reconfig could be done in a separate worker, either immediately via a task call if none is working, or once the previous reconfig is done if there is already one in progress?

smarnach: @antoviaque We have a ticket for doing exactly what you said in the backog, but we just don't have the time right now to implement this.  There are loads of things we want to improve in Ocim, and I don't feel the LB configurations are the highest priority right now.  If you feel it actually _is_ the top priority, we can start working on it, but that may delay our other plans.

My approach would be to keep it simple for now.  Adding the parameter Jill suggested now will simplify the code immediately, and we can further simplify it when we get to that ticket in the backlog.

antoviaque: @smarnach  It's not the top priority as a task, but if we make a change on this, I'd rather take the clean approach now, rather than adding a new layer of complexity, which we will need to extricate when we would get to it later. It's exactly in this type of situation that we need to think about leaving the code cleaner after us, rather than dirtier.

And I assumed the change was relatively small - this is basically moving some of the code to another function/task, and gating the execution with a flag? 

For reference, this is https://tasks.opencraft.com/browse/OC-2517

UmanShahzad: I don't mind pulling in OC-3157 and OC-2517 next sprint and making this PR basically solve both of them in one go, as they touch quite related parts/functionality. I don't intend on actually having this merged by the end of this week as part of OC-3242, because that wasn't actually planned (a small iteration was, but what I ended up doing, plus if we add OC-2517's work, wouldn't be) Does that sound good?

smarnach: @antoviaque Fair enough.  I don't really see the problem with the approach we planned, but it's not that important.  Implementing OC-2517 will work as well, so let's do that.  It won't resolve the double reconfiguration I mentioned above, but that's not a big problem – it will eventually disappear once we implement app server sets.

antoviaque: @UmanShahzad @smarnach Sounds good - thanks to both of you!

instance/models/load_balancer.py
```diff
@@ -84,24 +83,74 @@ def generate_fragment_name(length):
...
-    ssh_username = models.CharField(max_length=32)
-    # Whether new backends can be assigned to this load-balancing server
-    accepts_new_backends = models.BooleanField(default=True)
-    # A random postfix appended to the haproxy configuration file names to avoid clashes between
-    # multiple instance managers (or multiple concurrently running integration tests) sharing the
-    # same load balancer.
+
+    ssh_username = models.CharField(
+        max_length=32,
+        help_text='The username sed to SSH into the server.'
```
UmanShahzad: Note to self: "The username *used*..."

UmanShahzad:
Updated the title/description to reflect new scope of this PR.

instance/tasks.py
```diff
@@ -81,13 +88,20 @@ def spawn_appserver(
...
 
 @db_task()
-def appserver_make_active(appserver_id, active=True):
+def make_appserver_active(appserver_id, active=True, deactivate_others=False):
     """
     Mark an AppServer as active or inactive.
     """
-    logger.info('%s AppServer: ID=%s', "Activating" if active else "Deactivating", appserver_id)
+    # Disallow a request to deactivate this appserver *and* all others.
+    deactivate_others = deactivate_others if active else False
```
UmanShahzad: Note to self: this can be removed in favor of `if active and deactivate_others` below.

instance/tests/test_tasks.py
```diff
@@ -107,6 +122,56 @@ def test_one_attempt_default(self):
...
+        """
+        By default, we activate the appserver.
+        """
+        tasks.make_appserver_active(self.appserver.id, active=active)
+        mock_make_active.assert_called_once_with(active=active)
+
+    @patch('instance.models.openedx_appserver.OpenEdXAppServer.make_active')
+    def test_deactivate_others(self, mock_make_active):
+        """
+        When activating the appserver, optionally deactivating others.
```
UmanShahzad: Note to self: "optionally *deactivate* others".

instance/tests/test_tasks.py
```diff
@@ -107,6 +122,56 @@ def test_one_attempt_default(self):
...
+        """
+        for dummy in range(5):
+            appserver = make_test_appserver(self.appserver.instance)
+            appserver.is_active = True
+            appserver.save()
+        tasks.make_appserver_active(self.appserver.id, active=True, deactivate_others=True)
+        mock_make_active.assert_has_calls(
+            # Calls to make the appserver active.
+            [call(active=True)] +
+            # Calls to make other appservers deactivated.
```
UmanShahzad: Note to self: "Calls to *deactivate other appservers*".

instance/tests/test_tasks.py
```diff
@@ -219,3 +284,21 @@ def test_clean_up_task(self, mock_shut_down_sandboxes, mock_terminate_appservers
...
+
+    @patch('instance.models.load_balancer.LoadBalancingServer.reconfigure')
+    def test_reconfigure_dirty_load_balancers(self, mock_reconfigure):
+        """
+        `reconfigure_dirty_load_balancers` calls `reconfigure` on all dirty load balancers, and no others.
+        """
+        for dummy in range(5):
+            LoadBalancingServerFactory(configuration_state=LoadBalancingServer.DIRTY)
+        LoadBalancingServerFactory(configuration_state=LoadBalancingServer.CLEAN)
+        LoadBalancingServerFactory(configuration_state=LoadBalancingServer.IN_PROGRESS)
```
UmanShahzad: Note to self: add a `IN_PROGRESS_AND_DIRTY` version, too.

UmanShahzad:
Note to whoever reviews this: the changes to the instance redeployment command only make sense in the context of the LB changes, so the LB changes should be looked at first.

instance/models/load_balancer.py
```diff
@@ -111,6 +145,55 @@ def __init__(self, *args, **kwargs):
...
+        Flag an internal method as one that modifies the state of load balancer configuration,
+        to allow managing configuration state outside and around the method call.
+        """
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            """
+            Used to wrap the function that modified load balancer configuration state.
+            """
+            try:
+                # Prevent race conditions by updating state only if the LB is actually clean or dirty.
```
UmanShahzad: For the curious, this is [optimistic locking](https://en.wikipedia.org/wiki/Optimistic_concurrency_control), so we can avoid [pessimistic locking](https://en.wikipedia.org/wiki/Concurrency_control#Categories), which I found a lot harder to implement, and harder to understand/read in the code.

instance/models/load_balancer.py
```diff
@@ -111,6 +145,55 @@ def __init__(self, *args, **kwargs):
...
+                updated = LoadBalancingServer.objects.filter(
+                    pk=self.pk,
+                    configuration_state__in=[self.CLEAN, self.DIRTY]
+                ).update(
+                    configuration_state=self.IN_PROGRESS
+                )
+
+                # A race condition or just a normal queuing up of LB reconfiguration requests is handled.
```
UmanShahzad: Note by 'queueing up' I'm referring to a bulk queue -- the periodic task will of course "handle" these "queued up" requests in a batch through a single reconfiguration.

instance/tasks.py
```diff
@@ -127,3 +141,15 @@ def clean_up():
...
+@db_periodic_task(crontab())
+def reconfigure_dirty_load_balancers():
+    """
+    Any load balancers that are dirty need to be reconfigured.
+
+    This task runs every minute.
+    """
+    logger.info('Reconfiguring all dirty load balancers')
+    for load_balancer in LoadBalancingServer.objects.filter(configuration_state=LoadBalancingServer.DIRTY):
+        load_balancer.reconfigure()
```
UmanShahzad: I think that if the program crashed or was control-c'd out of while a reconfiguration was going on, we'd be stuck in that state. Maybe we should just exclude `CLEAN` states instead? That wouldn't have an effect on the logic, and it would save us from these cases, but there would then guaranteed be at least 1 call to the DB per minute from this task (the one to perform the 'update'. We could maybe prevent that by first checking to see if the state is `IN_PROGRESS` or `IN_PROGRESS_AND_DIRTY` and kick the thread out early to prevent that DB call.

instance/models/load_balancer.py
```diff
@@ -111,6 +145,55 @@ def __init__(self, *args, **kwargs):
...
+            except ReconfigurationFailed:
+                # If the active reconfiguration failed for any reason, the LB is dirty and ready
+                # for another round of reconfiguration, either by request or periodic task.
+                self.configuration_state = self.DIRTY
+            else:
+                self.configuration_state = (
+                    self.DIRTY
+                    if self.configuration_state == self.IN_PROGRESS_AND_DIRTY
+                    else self.CLEAN
+                )
```
smarnach: `self.configuration_state` may be out of date here.  At the very least, we need to do `self.refresh_from_db()` before doing that check, but we still would have a race condition.  This needs to become an atomic update, like the one above, but I'm not sure that can be expressed in Django without using a database transaction (i.e. the `atomic` context manager).

instance/tasks.py
smarnach: If an LB is stuck in `IN_PROGRESS` state, calling the `reconfigure()` method won't help, since it won't do anything except for marking the LB as in progress and dirty, but we would still be stuck in that state; see my comment above.

instance/ansible.py
```diff
@@ -42,6 +42,19 @@
...
 # Functions ###################################################################
 
+def load_yaml(string):
+    """
+    Return the dict parsed from the given yaml string.
+
+    If the string begins with @, then read from the file there named.
+    """
+    if string.startswith('@'):
+        content = open(string[1:])
```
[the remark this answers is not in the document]
UmanShahzad: Yep, this was from before, and I actually hadn't even read the implementation of it at the time, until now, and I agree with your suggestions (I'll use the context manager). I don't think it matters that much if we have 2 return statements, as long as it does what it should.

instance/management/commands/instance_redeploy.py
```diff
@@ -91,6 +73,13 @@ def add_arguments(self, parser):
                  ' Note that archived instances are automatically excluded. Omit to re-spawn all un-archived instances.'
         )
         parser.add_argument(
+            '--exclude',
+            type=load_yaml,
+            default='{}',
+            help='YAML containg the OpenEdXInstance exclusion queryset used to exclude instances for redeployment. '
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

instance/models/load_balancer.py
```diff
@@ -49,7 +48,7 @@ def _create_default(self):
         Create the default load balancing server configured in the Django settings, if any.
         """
         if settings.DEFAULT_LOAD_BALANCING_SERVER:
-            ssh_username, unused, domain = settings.DEFAULT_LOAD_BALANCING_SERVER.partition("@")
+            ssh_username, _, domain = settings.DEFAULT_LOAD_BALANCING_SERVER.partition("@")
```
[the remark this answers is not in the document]
UmanShahzad: Yeah, I get your opinion on this, and I'm personally neutral; `_` is just something I got used to from working on edx-enterprise, but leaving it `unused` sounds good to me!

instance/models/load_balancer.py
```diff
@@ -111,6 +145,55 @@ def __init__(self, *args, **kwargs):
...
+                LoadBalancingServer.objects.filter(
+                    pk=self.pk,
+                    configuration_state=self.IN_PROGRESS
+                ).update(
+                    configuration_state=self.IN_PROGRESS_AND_DIRTY
+                )
+            except ReconfigurationFailed:
+                # If the active reconfiguration failed for any reason, the LB is dirty and ready
+                # for another round of reconfiguration, either by request or periodic task.
+                self.configuration_state = self.DIRTY
```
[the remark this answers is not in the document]
UmanShahzad: This is a good point. I think we can try and cover more random cases by just trying and using a higher level exception class. That's not the best idea though. If we move it to `finally`, then the next thread that comes in would be able to mark this as being in-progress again.

What we could do to replace the function of the timeout, is just use a timeout context manager, which we can real quickly build, or just use [stopit](https://pypi.python.org/pypi/stopit). It would turn the method call into:

```python
with stopit.ThreadingTimeout(15) as reconfiguration:
    try:
        method(self, *args, **kwargs)
    except ReconfigurationFailed:
        # This particular exception probably doesn't make sense to raise here -- it's what `stopit` expects to change the state, though.
        raise TimeoutException
if reconfiguration.state != reconfiguration.EXECUTED:
    self.configuration_state = self.DIRTY
```

This technique would handle the network timeout case, but not necessarily for exceptions that aren't `ReconfigurationFailed`, so that's one detail to work out.

I know I'm using `self.configuration_state` here. We can just replace that with `update`, and call `save()` at the end, in `finally`, to guarantee the signals. Or whatever approach works best.

instance/models/load_balancer.py
```diff
@@ -111,6 +145,55 @@ def __init__(self, *args, **kwargs):
...
+            Used to wrap the function that modified load balancer configuration state.
+            """
+            try:
+                # Prevent race conditions by updating state only if the LB is actually clean or dirty.
+                updated = LoadBalancingServer.objects.filter(
+                    pk=self.pk,
+                    configuration_state__in=[self.CLEAN, self.DIRTY]
+                ).update(
+                    configuration_state=self.IN_PROGRESS
+                )
```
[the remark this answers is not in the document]
UmanShahzad: It does come with a single command, yes, otherwise this would be totally useless for the purpose of locking.

Yeah, I think we can refresh from the DB if this does succeed (and there wouldn't be a race condition there because only a single thread will be able to succeed at a time here). Or we can just replace `self.configuration_state` with querysets.

instance/models/load_balancer.py
UmanShahzad: See above -- we can maybe just use a queryset.

instance/tasks.py
UmanShahzad: Yeah, not sure what I was thinking when I commented that. :)

instance/models/load_balancer.py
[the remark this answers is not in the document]
UmanShahzad: Clarification on `update`:

From https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.update --

> Using update() also prevents a race condition wherein something might change in your database in the short period of time between loading the object and calling save().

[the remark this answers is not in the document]
UmanShahzad: I also tested locally and got:

```sql
UPDATE "instance_loadbalancingserver"
SET "configuration_state" = 'in_progress'
WHERE ("instance_loadbalancingserver"."configuration_state" IN ('dirty',
                                                                'clean')
       AND "instance_loadbalancingserver"."id" = 2)

```

UmanShahzad:
@smarnach Thanks for the review, really appreciate it! I've replied, but won't go about implementing anything until we agree on how we want to continue. This is indeed tricky, but I think with some creative approaches we can make it work.

instance/models/load_balancer.py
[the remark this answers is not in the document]
smarnach: @UmanShahzad Yeah, I realize that this is kind of moot, since I'm not going to stop this pattern from being used, but I just needed to take the opportunity to rant about it.

instance/models/load_balancer.py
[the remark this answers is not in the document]
smarnach: I also believed that it results in a single `UPDATE`, but I wanted to know for sure.  Thanks!

instance/models/load_balancer.py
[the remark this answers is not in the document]
smarnach: I believe the only way we can handle the dropped network connection case is by storing a timestamp somewhere – either in Redis or in th DB.  If the network dies, we can't access the DB or Redis anymore, since the network is dead.  So anything we do with `stopit` or context managers or whatever won't be able to save us, since we won't be able to change anything in the DB anyway.

This case actually happens, and the result would be that the LB would get stuck in "in progress" state.  This is a particularly bad failure mode, since it's not really obvious for the user.  You click "Activate" in the UI and your app server gets shown as "active".  You may notice that the feature you meant to deploy does not seem to work yet, but it won't be obvious to you that the UI is lying, and that your app server actually isn't active.  So while this case is easy to fix manually if it happens, it may lead to difficult to debug problems, so I'd prefer if we get this right.

So I really believe we need to use a lock that expires after some time, since this is the only way to avoid that error case.  I think the suggestion I made on Jira using the timestamp and the flag actually works, but it is a bit fiddly.  I will comment on Jira with a suggestion that is easier to understand and implement.

instance/models/load_balancer.py
smarnach: The problem here is the if condition.  You need two updates to express this in the same way as above:

    LoadBalancingServer.objects.filter(
        configuration_state=self.IN_PROGRESS
    ).update(
        configuration_state=self.CLEAN
    )
    LoadBalancingServer.objects.filter(
        configuration_state=self.IN_PROGRESS_AND_DIRTY
    ).update(
        configuration_state=self.DIRTY
    )

Because it's two updates, they are not atomic anymore.  If you put them in the order given above, it would actually work, but if you swap them, you would have a race condition again.

Anyway, we might not even stick with this solution, so we don't need to further discuss it.

UmanShahzad:
@smarnach I've addressed all outstanding reviews and implemented the new algorithm for this that you posted on Jira. I'm not 100% sure if I implemented it as you had in mind, so do let me know.

instance/models/load_balancer.py
```diff
@@ -111,6 +148,28 @@ def __init__(self, *args, **kwargs):
...
+        to allow managing configuration state outside and around the method call.
+        """
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            """
+            Used to wrap the function that modified load balancer configuration state.
+            """
+            with transaction.atomic():
+                self.configuration_version += 1
+                self.save()
```
smarnach: I'm not sure this ensures atomicity of the update, since `self` will already be loaded from the DB before the transaction even starts.  To be honest, I'm not even sure this _needs_ to be atomic, since if two processes update this field in parallel, we don't really care whether the value gets increased by one or by two – both would mark the LB config as dirty.

So this is kind of up to you.  If you think this needs to be atomic, make sure it is, e.g. by using an F-expression like

    from django.db.models import F
    self.configuration_version = F("configuration_version") + 1

(Not sure I got the details right.)  If you think it's fine if this isn't atomic, just remove the transaction context manager.

instance/models/load_balancer.py
```diff
@@ -111,6 +148,28 @@ def __init__(self, *args, **kwargs):
...
+
+            with self._configuration_lock(blocking_timeout=1):
+                candidate_configuration_version = self.configuration_version
+                method(self, *args, **kwargs)  # pylint: disable=not-callable
+                self.deployed_configuration_version = candidate_configuration_version
+
+            self.save()
```
smarnach: I think you should save while still holding the lock.

instance/models/load_balancer.py
```diff
@@ -210,22 +264,28 @@ def reconfigure(self, triggering_instance_id=None):
...
         Delete the LoadBalancingServer from the database.
         """
         self.deconfigure()
         super().delete(*args, **kwargs)
+
+    def _configuration_lock(self, **kwargs):
+        """
+        A Redis lock to protect reconfigurations of this load balancer instance.
+        """
+        return cache.lock("load_balancer_reconfigure:{}".format(self.domain), timeout=900, **kwargs)
```
smarnach: Should we make the timeout configurable?  We'll probably never need to change it, so I think it's fine.

instance/models/load_balancer.py
```diff
@@ -210,22 +264,28 @@ def reconfigure(self, triggering_instance_id=None):
...
         """
+        self.logger.info("Deconfiguring load-balancing server %s", self.domain)
         fragment_name = settings.LOAD_BALANCER_FRAGMENT_NAME_PREFIX + self.fragment_name_postfix
-        with self._configuration_lock():
-            self.run_playbook(
-                "FRAGMENT_NAME: {fragment_name}\nREMOVE_FRAGMENT: True".format(fragment_name=fragment_name)
-            )
+        self.run_playbook(
+            "FRAGMENT_NAME: {fragment_name}\nREMOVE_FRAGMENT: True".format(fragment_name=fragment_name)
+        )
```
smarnach: I didn't realize this in the first version, but adding the same decorator here doesn't really do what we want.  We want to remove the fragment, not mark the LB for reconfiguration.  If we want to remove the fragment, we should just block until we are allowed to, so basically do exactly what we did before.

instance/models/load_balancer.py
UmanShahzad: We can do that, sure. I don't like raw numbers in a codebase too much, so I'll add the var.

instance/models/load_balancer.py
UmanShahzad: Ah, OK!

instance/models/load_balancer.py
UmanShahzad: Oh woops, good point. I had out of the lock's block because I thought I was going to be adding some extra logic, but I ended up not doing that.

instance/models/load_balancer.py
UmanShahzad: I tried some things to see if I could keep it atomic, including with F-expressions, but it was just unnecessarily complicating things, and I'm thinking the same as you now -- that atomicity here doesn't seem to be needed. Only one thread will grab the lock, and actually, we want that thread to have the version provided in-memory through `self.configuration_version`, when it originally updated it, so removing the context manager and the `candidate_configuration_version` below and just using `self.configuration_version` should suffice.

instance/models/load_balancer.py
UmanShahzad: P.S. I'm going to do this near the end of the PR's lifetime, to give myself time to do higher-priority things.

instance/models/load_balancer.py
smarnach: Please do not remove `candidate_configuration_version`.  Relying on `self.configuration_version` seems fragile, since we would rely on the value not being updated from the DB.  So if any of the methods involved in bulding the configuration calls `self.refresh_from_db()`, we would have a race condition again, so it seems more robust to keep the local variable.

UmanShahzad: Ah, yeah, good point!

UmanShahzad:
Rebased on `master` from 6811e3d8b368a629c69fa6a9560ac5081e92e24e

instance/models/load_balancer.py
```diff
@@ -210,13 +263,13 @@ def reconfigure(self, triggering_instance_id=None):
         reconfiguration of the load balancer.
         """
         self.logger.info("Reconfiguring load-balancing server %s", self.domain)
-        with self._configuration_lock():
-            self.run_playbook(self.get_ansible_vars(triggering_instance_id))
+        self.run_playbook(self.get_ansible_vars(triggering_instance_id))
 
     def deconfigure(self):
```
bradenmacdonald: Shouldn't this method also be using `@modifies_configuration` ?

instance/models/load_balancer.py
```diff
@@ -111,6 +148,27 @@ def __init__(self, *args, **kwargs):
...
+        """
+        Flag an internal method as one that modifies the state of load balancer configuration,
+        to allow managing configuration state outside and around the method call.
+        """
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            """
+            Used to wrap the function that modified load balancer configuration state.
+            """
+            self.configuration_version += 1
```
bradenmacdonald: This increment is subject to a race condition - does that matter?

If necessary it can be changed to use an [F expression](https://docs.djangoproject.com/en/1.11/ref/models/expressions/#f-expressions):

```python
from django.db.models import F
...
self.configuration_version = F('configuration_version') + 1
self.save()
self.refresh_from_db()  # Required to read the result, and to avoid future .save() calls from incrementing the value again
```

instance/tests/management/test_instance_redeploy.py
```diff
@@ -247,12 +244,8 @@ def _spawn_appserver_success(
             # Verify the logs
             captured_logs.check(*expected_logs)
 
-        # Ensure that the load balancer was reconfigured only once per batch of successful appservers
-        self.assertEqual(mock_reconfigure.call_count, 2)
```
bradenmacdonald: Is there no way to test the LB re-configuration anymore here?

instance/tests/test_tasks.py
```diff
@@ -75,8 +77,20 @@ def test_mark_active_on_success(self, provisioning_succeeds):
...
+        """
+        If `mark_active_on_success` and `deactivate_old_appservers` are both passed in as `True`,
+        the spawn appserver task will mark the newly provisioned AppServer as active, and deactivate
+        old appservers, if provisioning succeeded.
+        """
+        self.mock_spawn_appserver.return_value = 10 if provisioning_succeeds else None
+        instance = OpenEdXInstanceFactory()
+        tasks.spawn_appserver(instance.ref.pk, mark_active_on_success=True, deactivate_old_appservers=True)
+        self.assertEqual(self.mock_spawn_appserver.call_count, 1)
+        self.assertEqual(self.mock_make_appserver_active.call_count, 1 if provisioning_succeeds else 0)
```
bradenmacdonald: This test doesn't seem to actually check if the old appservers got deactivated? Or even just if `deactivate_others=True` got passed to `make_appserver_active`?

(I know `test_deactivate_others` does, but it's a more narrow unit test)

instance/models/load_balancer.py
```diff
@@ -91,17 +99,46 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
...
+    )
+
+    configuration_version = models.PositiveIntegerField(
+        default=1,
+        help_text=(
+            'The current version of configuration for this load balancer. '
+            'By default, a version of 0 means that the load balancer is clean, '
+            'i.e. it does not currently require a reconfiguration. '
+            'The version value is essentially equivalent to the number of requests made '
+            'to reconfigure the load balancer since it last was in progress.'
```
bradenmacdonald: This docstring seems out of date.

It says "By default, a version of 0 ", but the version starts at 1 by default and never gets reset to zero.

It also says "The version value is essentially equivalent to the number of requests made to reconfigure the load balancer since it last was in progress." but that doesn't seem to be the case anymore, since the latest code seems to only ever increment the version number. So it would be more accurate to say "The version value is the total number of requests ever made to reconfigure the load balancer".

instance/models/load_balancer.py
smarnach: @bradenmacdonald We've discussed this, and came to the conclusion that it doesn't matter.  If tow processes increase the version number concurrently, we might only increase by one instead of two, but that would still mark the LB as dirty, and only one of the processes will be able to grab the lock to actually reconfigure it.  We should add a comment here, though, to explain why this is fine.

instance/models/load_balancer.py
smarnach: It did originally, but that doesn't really do what we want.  It try to grab the lock, and if it can't it would mark the LB as requiring _reconfiguration_, while it actually require _deconfiguration_.  This method is only called when an LB is deleted, and in this case the only reasonable behaviour is to block until we can grab the lock, and then deconfigure the LB.  Deferring the deconfiguration is not helpful.

instance/management/commands/instance_redeploy.py
```diff
@@ -285,77 +297,54 @@ def _do_redeployment(self):
...
-            # 1. Activate successful instances
-            reconfigure_load_balancer = False
+            # 1. Log successful instances
             for instance in self.success_tag.openedxinstance_set.iterator():
                 LOG.info("SUCCESS: %s [%s]", instance, instance.id)
                 instance.tags.remove(self.ongoing_tag)
                 instance.tags.remove(self.success_tag)
 
-                # Make sure that the latest appserver is running
+                # 1.1 Log completed instances
```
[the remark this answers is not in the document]
UmanShahzad: I actually think the `completed` tag may now be irrelevant. I've delegated the logic that makes the appservers active into `spawn_appserver`, which also tags instances as failed/successful depending on whether they were actually successfully provisioned. We were marking things with an *extra* tag, namely `complete`, for a reason I'm quite uncertain of... I guess it may just have marked a way for us to see that, yes, we successfully made it past the logic that makes the appserver active. But now we don't need that because I modified `spawn_appserver` to make the appserver active upon success, and deactivate old ones.

So the work here should be to fix the potentially misleading docstrings, and maybe explain a little more what's happening.

instance/management/commands/instance_redeploy.py
```diff
@@ -248,20 +256,24 @@ def _redeployment_complete(self):
 
         Also logs a redeployment status report from the counts taken.
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

instance/management/commands/instance_redeploy.py
```diff
@@ -151,6 +140,24 @@ def handle(self, *args, **options):
         else:
             LOG.info("** Redeployment canceled **")
 
+    def get_statistics(self, pending=False, ongoing=False, failed=False, successful=False, completed=False,
+                       all_statistics=False):
+        """
+        Get the number of pending, ongoing, failed, successful, and/or completed instance redeployments.
```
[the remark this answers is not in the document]
UmanShahzad: Yes, sure. But since I'll delete `completed` (see comment below), I'll just explain each of them.

instance/models/load_balancer.py
UmanShahzad: Yeah, it's definitely out of date, sorry! I had tried a different technique before this one but forgot to update the help text.

instance/models/load_balancer.py
UmanShahzad: Will explain this, yep.

instance/tests/management/test_instance_redeploy.py
UmanShahzad: Since we're now delegating the LB re-configuration logic to `make_active`, which is itself now being called only in `spawn_appserver`, which we're mocking above, we can't really test this, no. But tests for `spawn_appserver` elsewhere should give us confidence that `make_active` is called, and tests for `make_active` should give us confidence that the LB is reconfigured.

instance/tests/test_tasks.py
UmanShahzad: I've changed it to test that the right arguments are passed in. :+1:

UmanShahzad:
@bradenmacdonald Thanks for the review! I've addressed it.

I think the best way to test the mass redeployment section of this PR would be to just run the mass redeployment command on stage. There are currently 4 instances there. What we should do is run the following command:

```bash
make manage "instance_redeploy --tag oc-3157-2517 --num-attempts 2 --batch-size 2" >& ~/oc-3157-2517.log
```

For one, we're obviously looking to see if the mass redeployment command still works correctly given these changes, and we'll find that out inevitably.

More specifically, we want to know the following:

1. Are new appservers that provision successfully marked active?

We can figure this out by just looking at the GUI and seeing whether the UI shows the latest appserver as being active.

2. Are old appservers deactivated?

We can figure this out by just looking at the GUI and seeing whether the UI shows all appservers besides the latest as being inactive.

---

For the LB portion of these changes, we *could* try and see the LB configuration version after the mass redeployment command succeeds, but I'm not sure exactly what numbers to look for and I think it depends a lot on whether 2 appservers both complete provisioning at roughly the same time (within ~10 minutes of each other).

Instead, we can try this:

1. Spawn new appservers for 2 different instances.
1. Take note of the LB deployed configuration version & the normal configuration version.
1. When provisioning completes for both of them, manually activate 1.
1. Confirm quickly that configuration version > deployed configuration version.
1. Manually activate the other appserver.
1. Confirm that configuration version is now 1 greater than before, and deployed configuration version hasn't changed yet.
1. Wait about 10-15 minutes and come back to see that deployed configuration version is 1 less than the normal configuration version.
1. Wait about 10-15 minutes and come back to see that the values are still the same.

I think the above steps may be overkill, I'm not really sure. Let me know if any of these steps can be squashed / made simpler.

instance/models/load_balancer.py
```diff
@@ -111,6 +145,31 @@ def __init__(self, *args, **kwargs):
...
+        Flag an internal method as one that modifies the state of load balancer configuration,
+        to allow managing configuration state outside and around the method call.
+        """
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            """
+            Used to wrap the function that modified load balancer configuration state.
+            """
+            # This is subject to a race condition, but it doesn't matter:
+            # if two processes increased the version concurrently, the LB would be the marked dirty
```
UmanShahzad: note to self: "would be marked dirty"

instance/models/load_balancer.py
bradenmacdonald: Nit: Ok, in that case, maybe the logic from `modifies_configuration` should be moved inside of `reconfigure(...)`? It would make the code easier to follow.

instance/management/commands/instance_redeploy.py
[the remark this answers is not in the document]
bradenmacdonald: Ah, good - I was actually thinking of suggesting removing the `completed` tag, so I'm glad you went with that option.

instance/models/load_balancer.py
UmanShahzad: Yep, I initially had it as a decorator to allow me to use it on both `reconfigure` and `deconfigure`, but now it makes more sense to be directly in `reconfigure`.

UmanShahzad:
@bradenmacdonald Thanks! Mass redeployment is now in progress. If redeployment works fine, I'll then give the LB thing a test right after.

UmanShahzad:
The migrations got stuck running on prod for some reason.

instance/models/load_balancer.py
```diff
@@ -196,27 +230,33 @@ def run_playbook(self, ansible_vars):
...
-        with self._configuration_lock():
+        # This is subject to a race condition, but it doesn't matter:
+        # if two processes increased the version concurrently, the LB would be marked dirty
+        # either way, and only one could successfully grab the lock, while the other would get
+        # kicked out after the blocking timeout.
+        self.configuration_version += 1
+        self.save()
+
+        # Memorize the configuration version, in case new threads change it.
+        with self._configuration_lock(blocking_timeout=1):
```
smarnach: Snap, we missed a major problem here:  This will try to grab the lock, but only block for 1 second in case the lock is held by another process.  However, it will enter the `with` block _regardless of whether it managed to grab the lock_, since there is no conditional code here at all.  I just ran into this issue, and it caused the Cloudera instance (and possibly other instances) to get into a weird state (fixed by reconfiguring the LB again).  I think we need to roll back until we have found a solution here.

PR #241, opened 2017-10-25
[OC-3036] Fixed double quotes issue in OCIM's JSON fields

macornwell:
This pull request fixes a double quotes json bug that occurs when adding a new application instance on OCIM. This fixes an issue with deserializing the django_extensions library's JSONField where invalid json would be created.

**JIRA tickets**: OC-3036

**Dependencies**: None

**Merge deadline**: "None" 

**Testing instructions**:

1. Open up a running version of OCIM with this pull request. 
2. Go to the django admin.
3. Attempt to add a new "Open edX Instances"
4. Scrolling down "Openstack server flavor" and "Openstack server base image" contain valid json where all string values are encased in double quotes rather than single quotes.
5. Put in valid values for all required fields.
6. Save. There should be no exceptions that occur.

1. In the long run it is probably better to use Django's own JSONField, but that requires upgrading postgresql's version. For now this works.

**Reviewers**
- [ ] mtyaka

### DIFF
--- a/instance/tests/test_widgets.py
+++ b/instance/tests/test_widgets.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Tests Django Widgets
+"""
+import re
+
+from instance.tests.base import TestCase
+from instance.widgets import JSONWidget
+
+
+class JSONWidgetTestCase(TestCase):
+    """
+    Tests the JSONWidget class.
+    """
+
+    def setUp(self):
+        self.html_front = re.compile(r'^.+?\>')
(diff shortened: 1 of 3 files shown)


mtyaka:
Thanks @macornwell, this is looking good! There are two more things I'd like to see before we merge. Can you please:

1. Add a unit test for the widget. It can be very simple, just checking that when you render an instance of the widget with the value set to a `dict`, it renders it with double quotes, and if you set the value to an already encoded JSON string, it doesn't escape the double quotes when rendered. Example widget tests can be found [in django's test folder](https://github.com/django/django/blob/ec5f7cb92e315846cb020b909b0ab7e0489aad05/tests/forms_tests/widget_tests/test_textarea.py).
2. Add a comment explaining that the widget will most likely be no longer required when we upgrade postgresql and start using django's built-in `JSONField`.

macornwell:
@mtyaka Unit tests added, and I added a comment in the wiget's class. (I am assuming that's where you meant). I also have it in the PR description.
Anything else you think this might need?

instance/tests/test_widgets.py
```diff
@@ -0,0 +1,67 @@
...
+        self.html_back = re.compile(r'\<.+?\>$')
+
+    def _strip_html(self, html_text):
+        """
+        Strips the front and end off an HTML string.
+        """
+        front_match = self.html_front.search(html_text).group()
+        back_match = self.html_back.search(html_text).group()
+        cleaned = html_text.replace(front_match, '')
+        cleaned = cleaned.replace(back_match, '')
```
mtyaka: Nit: could this be simplified as:

```python
r = re.compile(r'^.+?\>(.*)\<.+?\>$')
return r.match(html_text).group(1)
```

mtyaka:
@macornwell Not a big deal, but next time please remember to squash the commits before merging.

macornwell:
@mtyaka  it wasn't squashed? Is there some other way than selecting from the drop down that says "Squash and Merge"?

mtyaka:
@macornwell Ah sorry, it was squashed on merge. I was only looking at the commits in the PR. When working on edx repositories, we typically need to squash manually, but I guess automatic squash on merge is fine for our own repos.

macornwell:
@mtyaka got it. I'll make sure to do that next time!

PR #242, opened 2017-10-26
Convert ansible templates to python dictionary format

kaizoku:
This PR converts the ansible variable templates to python dictionaries, since previously they were rendered with django templates. This was problematic as both django and Jinja2 templates use the same variable expansion syntax, and both template languages were chained.

**JIRA tickets**: Implements OC-3214

**Testing instructions**:

1. Run unit tests with `make test`
2. Provision new instance and see that templates are rendered properly

**Reviewers**
- [ ] (smarnach)


instance/utils.py
```diff
@@ -133,3 +138,31 @@ def sufficient_time_passed(earlier_date, later_date, expected_days_since):
...
+
+class DictTemplate(object):
+    """
+    Represents a dictionary based template for ansible configuration.
+    """
+
+    def __init__(self, path):
+        spec = importlib.util.spec_from_file_location(os.path.basename(path), path)
+        self.module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(self.module)
```
antoviaque: Why go through the trouble of importing manually based on a file name? Wouldn't it be simpler (and flexible) to pass the right function to get_template?

smarnach:
@kaizoku Thanks for the changes!  This version looks much more like what I had in mind when I wrote the ticket.

I think this can still be further simplified.  Much of the structure of the old code was forced by the settings being Django templates: You had to put all of them in one folder, since all Django templates need to live under the `templates` folder and can't be put alongside the code the belong to.  You had to create a context dictionary, since that's the only way of passing in values to a template.

Both of these don't make sense anymore with simple Python functions.  It's much more natural to have the functions alongside the related code, and there isn't really a point in first putting parameters in a `context` dictionary just to build another dictionary from them.  This was only necessary when we actually had templates.

I'm sorry that I did not make this intention clearer in the ticket description.  I'll push an example commit of what I have in mind to this PR, so you can take a look.  Don't hesitate to start a discussion if you disagree with the approach in my commit.   It's not the law – it's just what I had in mind when I wrote the ticket, and I still think it's the simplest approach.

smarnach:
@kaizoku Apparently I made the tests fail since I did not remove the now unused import, and I also did not remove the imports in `ansiblevars` for the modules I deleted, but my commit is only meant to explain the idea.

kaizoku:
OK, I definitely see what you mean now. I guess I was a bit afraid of changing too much around in a codebase I don't know well, but I can see what you're looking for and how it's simpler now.

I've updated the code based on your example, dropped the old commits, and pushed here.
Thanks for the extra explanation and example code.

instance/models/openedx_appserver.py
```diff
@@ -295,40 +294,317 @@ def get_playbooks(self):
             playbooks.append(self.lms_user_creation_playbook())
         return playbooks
 
+    def _get_configuration_variables(self):
```
smarnach: You could consider moving this particular function to a separate module again.  This file is already rather hard to navigate, and this is the biggest settings block of all.

instance/models/openedx_appserver.py
```diff
@@ -295,40 +294,317 @@ def get_playbooks(self):
...
+            "ECOMMERCE_CYBERSOURCE_SOAP_API_URL": 'https://set-me-please',
+            # * PayPal
+            "ECOMMERCE_PAYPAL_MODE": 'SET-ME-PLEASE',
+            "ECOMMERCE_PAYPAL_CLIENT_ID": 'SET-ME-PLEASE',
+            "ECOMMERCE_PAYPAL_CLIENT_SECRET": 'SET-ME-PLEASE',
+            "ECOMMERCE_PAYPAL_RECEIPT_URL": '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/receipt/',
+            "ECOMMERCE_PAYPAL_CANCEL_URL": '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/cancel/',
+            "ECOMMERCE_PAYPAL_ERROR_URL": '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/error/',
+
+            # Comprehensive theming
```
smarnach: Can you please remove the commented out code?  It's unclear why it is there, and we have git to recover it if we need it.

instance/models/openedx_appserver.py
kaizoku: I've moved this into a mixin which is imported and subclassed by OpenEdxAppServer. Is this a suitable loading strategy?

instance/models/openedx_appserver.py
kaizoku: Good point, I just kept everything from the original file. These should be removed now.

smarnach:
I tested this on stage, and it seems to work fine.  However, I found the old formatting of the templates more readable, so could you please add back `default_flow_style=False` to all `yaml.dump()` calls?  I wouldn't bother factoring this out into a separate function.  Apart from this and the comment about renaming the class and the file, everything looks good to me.

smarnach:
@kaizoku Thanks for addressing all the comments!  This looks good to me now, :+1:.

- [x] I tested this: I deployed this to stage and manually tried launching an instance and looked at the configuration.
- [x] I read through the code

PR #245, opened 2017-11-02
Fix locking around the LB configuration version updates.

smarnach:
This PR makes the version updates for LB reconfigurations atomic.

The most important change is to refresh the LB object from the database after the LB reconfiguration.  The current version of the code still has the old version of `configuration_version` stored on `self`, so if `configuration_version` was increased by a different process while the reconfiguration was running this increase will simply be overwritten by the final `self.save()` call.

### DIFF
--- a/instance/models/load_balancer.py
+++ b/instance/models/load_balancer.py
@@ -231,26 +231,32 @@ class LoadBalancingServer(ValidateModelMixin, TimeStampedModel):
             self.logger.error("Playbook to reconfigure load-balancing server %s failed.", self)
             raise ReconfigurationFailed
 
-    def reconfigure(self, triggering_instance_id=None):
+    def reconfigure(self, triggering_instance_id=None, mark_dirty=True):
         """
         Regenerate the configuration fragments on the load-balancing server.
 
         The triggering_instance_id indicates the id of the instance reference that initiated the
         reconfiguration of the load balancer.
+
+        The mark_dirty flag indicates whether the LB configuration should be marked as dirty.  If
+        this method is called because the configuration changed, the flag should be set to True (the
+        default).  If this method is called because the LB was marked dirty earlier, the flag
+        should be set to False.
         """
-        # We need to use an F expression here.  The problem is not other processes trying to
-        # increase this counter concurrently – that wouldn't matter, since we don't care whether
-        # we increase this counter by one or by two, since both marks the LB as dirty.  However, if
-        # another process is making a completely unrelated change to the LB object we might lose
-        # the increment altogether.
-        LoadBalancingServer.objects.filter(pk=self.pk).update(
-            configuration_version=models.F("configuration_version") + 1
-        )
-        self.refresh_from_db()
+        if mark_dirty:
+            # We need to use an F expression here.  The problem is not other processes trying to
+            # increase this counter concurrently – that wouldn't matter, since we don't care whether
+            # we increase this counter by one or by two, since both marks the LB as dirty.  However, if
+            # another process is making a completely unrelated change to the LB object we might lose
+            # the increment altogether.
+            LoadBalancingServer.objects.filter(pk=self.pk).update(
+                configuration_version=models.F("configuration_version") + 1
+            )
 
         try:
             with self._configuration_lock(blocking=False):
                 # Memorize the configuration version, in case new threads change it.
+                self.refresh_from_db()
                 candidate_configuration_version = self.configuration_version
                 self.logger.info("Reconfiguring load-balancing server %s", self.domain)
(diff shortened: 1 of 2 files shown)


instance/models/load_balancer.py
```diff
@@ -238,21 +238,30 @@ def reconfigure(self, triggering_instance_id=None):
...
                 self.logger.info("Reconfiguring load-balancing server %s", self.domain)
                 self.run_playbook(self.get_ansible_vars(triggering_instance_id))
-                self.deployed_configuration_version = candidate_configuration_version
-                self.save()
+
+                with transaction.atomic():
+                    # Reload to pick up the potentially changed configuration_version
+                    self.refresh_from_db()
+                    self.deployed_configuration_version = candidate_configuration_version
+                    self.save()
```
bradenmacdonald: Can you just replace this whole `with` statement with another `LoadBalancingServer.objects.filter(pk=self.pk).update(deployed_configuration_version= candidate_configuration_version)` to be sure you're not changing other fields, while keeping the code simple?

Also, I think there may still be a race condition whereby another process might set a new version, and then this process could set a lower version.

Process A:
* Increment `configuration_version` to 2 and `refresh_from_db()`

Process B:
* Increment `configuration_version` to 3 and `refresh_from_db()`
* Obtain `_configuration_lock`; at this point, `candidate_configuration_version` is 3
* In atomic transaction, set `deployed_configuration_version` to 3

Process A:
* Obtain `_configuration_lock`; at this point, `candidate_configuration_version` is 2
* In atomic transaction, set `deployed_configuration_version` to 2

Perhaps you should move the first `refresh_from_db()` within the lock and/or change the `with transaction.atomic()... save()` expression to:

```
LoadBalancingServer.objects.filter(pk=self.pk, deployed_configuration_version__lt=candidate_configuration_version).update(deployed_configuration_version=candidate_configuration_version)
```

UmanShahzad: I think in a race condition situation, Process A wouldn't actually be able to acquire the lock in that scenario because we aren't blocking - it'd just get kicked out of the try/except block as soon as it tried to grab it because Process B already did. Or did I misunderstand something?

smarnach: I can change this to an `update()` call if you prefer, though I don't think the current code could change any other fields.  We refresh from the DB after starting the transaction, so the DB row will be locked.

And I don't think what you describe is a race condition.  First, I don't think it can happen, since we don't block when trying to grab the lock, so this would only happen when process A does not get a time slice for a full LB reconfiguration, which is for a minute or so.  And even if it does happen, the result would still be correct – we just would trigger a redundant reconfiguration, but I don't think it's even possible to avoid _any_ redundant reconfiguration while at the same time never missing one (in fact I think it's quite easy to prove that this is impossible).

So I can move the `refresh_from_db()` inside the `with` statement, but I think it's clearer as it is, and it does not make much of a difference.

smarnach: @UmanShahzad It's actually theroretically possible, since processes can be interrupted by the OS at any time, and aren't any guarantees at all for how long they will be interrupted.  In practice, however, and interruption of a minute (or even just several seconds) is rather unlikely unless the system is in a very unhealthy state (thrashing, ooming, lots of proccesses waiting for CPU, etc).

smarnach: @bradenmacdonald Sounds all reasonable, I made that change.  the downside is that we now have to wait at least two hours again before we can merge this.  I'll test on stage in the meantime.

smarnach:
Side note: We can't use `F()` expressions to increase the version in the usual form, i.e.

    self.configuration_version = F("configuration_version") + 1
    self.save()

The reason why this does not work for us is that [we call `full_clean()` before saving the model](https://github.com/open-craft/opencraft/blob/481c6a8eb9dced501be1be710fb66eb3fee2abc9/instance/models/utils.py#L99), and the `F()` expression won't pass validation as an integer.  (I have severe doubts it's a good idea to always call `full_clean()`, but that's a different discussion.)

instance/tasks.py
```diff
@@ -153,4 +153,4 @@ def reconfigure_dirty_load_balancers():
     for load_balancer in LoadBalancingServer.objects.filter(
             configuration_version__gt=F('deployed_configuration_version')
     ):
-        load_balancer.reconfigure()
+        load_balancer.reconfigure(mark_dirty=False)
```
UmanShahzad: Nice, thanks!

smarnach:
There is still a problem with the tests.  I haven't fully understood why the test is failing, but I believe it's a problem with the test rather than with the code.  I'm going to bed now, so this will have to wait until tomorrow.

PR #247, opened 2017-11-09
Don't set COMMON_HOSTNAME to instance domain.

mtyaka:
`COMMON_HOSTNAME` is mapped to `127.0.0.1` on edxapp app servers, making the instance inaccessible from the app server itself.

This patch uses a different value for `COMMON_HOSTNAME`, derived from the domain and the appserver name. The value itself probably  doesn't matter much.

**Testing instructions**:

Spawn a new appserver, SSH into it, and verify that `/etc/hostname` is set to a value derived from the instance domain and the app server name, instead of the instance domain.


UmanShahzad:
@mtyaka Am spinning up https://stage.console.opencraft.com/instance/1460/edx-appserver/364/ with the change cherry-picked to #235 which is also on stage (it has like 3-4 migrations, so I figured it'd be easier to cherry-pick the single commit than try and revert things).

mtyaka:
Rebased from ffec30403e17a9744ed380107705a98b8b0ba38c.

mtyaka:
Thanks @UmanShahzad! I rebased and will merge once the build succeeds.

PR #250, opened 2017-11-17
Fix the "Launch new app server" button

clemente:
This fixes a bug where the button was left disabled.

# Testing instructions

1. I suggest you apply this patch to disable the spawning of appservers and to make testing faster 
```
modified   instance/models/openedx_instance.py
@@ -172,7 +172,7 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
 
         Returns the ID of the new AppServer on success or None on failure.
         """
-        if not self.load_balancing_server:
+        if False and not self.load_balancing_server:
             self.load_balancing_server = LoadBalancingServer.objects.select_random()
             self.save()
             self.reconfigure_load_balancer()
@@ -180,10 +180,11 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
         # We unconditionally set the DNS records here, though this would only be strictly needed
         # when the first AppServer is spawned.  However, there is no easy way to tell whether the
         # DNS records have already been successfully set, and it doesn't hurt to alway do it.
-        self.set_dns_records()
+        if False:
+            self.set_dns_records()
 
         # Provision external databases:
-        if not self.use_ephemeral_databases:
+        if False and not self.use_ephemeral_databases:
             # TODO: Use db row-level locking to ensure we don't get any race conditions when creating these DBs.
             # Use select_for_update(nowait=True) to lock this object's row, then do these steps, then refresh_from_db
             self.logger.info('Provisioning MySQL database...')
@@ -197,8 +198,8 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
 
         app_server = self._create_owned_appserver()
 
-        if app_server.provision():
-            self.logger.info('Provisioned new app server, %s', app_server.name)
+        if True or app_server.provision():
+            self.logger.info('Provisioned new app server [BUENO, NO…), %s', app_server.name)
             self.successfully_provisioned = True
             self.save()
             return app_server.pk

```
2. Go to an instance with 0 appservers
3. Click the the „Launch new appserver“, wait, reclick (it should be reclickable)
4. See it advance in the special cases: 0→1, 4→5, 5→6, and after 5.  (5 is special because the list is limited to 5).

### DIFF
--- a/instance/static/js/src/instance.js
+++ b/instance/static/js/src/instance.js
@@ -142,7 +142,7 @@ app.controller("Details", ['$scope', '$state', '$stateParams', 'OpenCraftAPI',
             $scope.is_spawning_appserver = false;
             $scope.is_updating_from_pr = false;
             $scope.instance_active_tabs = {};
-            $scope.old_appserver_count = 0;
+            $scope.old_appserver_count = 0; // Remembers number of servers to detect when a new one appears
 
             $scope.instanceLogs = false;
             $scope.isFetchingLogs = false;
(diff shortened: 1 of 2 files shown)


UmanShahzad:
I put this on stage, am testing now.

PR #251, opened 2017-11-18
Don't watch forks marked as disabled

clemente:
Small fix to honour the `WatchedFork.enabled` attribute.

### DIFF
--- a/pr_watch/tasks.py
+++ b/pr_watch/tasks.py
@@ -49,7 +49,7 @@ def watch_pr():
     organization on the watched repository
     """
     try:
-        for watched_fork in WatchedFork.objects.all():
+        for watched_fork in WatchedFork.objects.filter(enabled=True):
             team_username_list = get_username_list_from_team(watched_fork.organization)
             for pr in get_pr_list_from_usernames(team_username_list, watched_fork.fork):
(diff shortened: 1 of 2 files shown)


pr_watch/tasks.py
```diff
@@ -49,7 +49,7 @@ def watch_pr():
     organization on the watched repository
     """
     try:
-        for watched_fork in WatchedFork.objects.all():
+        for watched_fork in WatchedFork.objects.filter(enabled=True):
```
pomegranited: Thanks for spotting this @clemente !

Just needs a test added to https://github.com/open-craft/opencraft/blob/master/pr_watch/tests/test_tasks.py

pomegranited:
@clemente :+1: 

- [x] I tested this by using the test instructions from https://github.com/open-craft/opencraft/pull/246, and verifying that an de-activating the `WatchedFork` causes the `WatchedPullRequest` to not be created.
  Also ensured that the unit tests run without error.
- [x] I read through the code
- [x] Includes documentation - N/A (field is straightforward)

UmanShahzad:
@clemente Can you squash/rebase? We can stop the current build.

clemente:
Squashed and rebased.

PR #252, opened 2017-11-28
Catch `EndpointNotFound` error for getting server status update.

UmanShahzad:
While testing the new region in OVH, we're having to use a new project with new creds that Ocim uses for both spawning new appservers and getting appserver information for currently running appservers. Since the currently running appservers are in the old region and project that require a different set of credentials, we get errors when attempting to access their detail view. This is a tiny patch to catch that error and log it to handle the case gracefully, until we improve Ocim with the ability to handle multi-OpenStack-project deployments.

*Testing instructions:*

1. On stage, don't be on this branch, and make sure the `.env` file contains the new OpenStack credentials for all `OPENSTACK_X` variables (i.e. user/pass/tenant/region).
2. See an instance that is setup to use the new region (i.e. 'Theme Academy'). Click on the older active instance and notice you get an error message (top right, or for a traceback see logs) when clicking on it to see the detail view.
3. With this patch, see that clicking on it does not raise an error.

### DIFF
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -359,8 +359,10 @@ class OpenStackServer(Server):
                 # server must have been terminated.
                 self.logger.debug('Server does not exist anymore: %s', self)
                 self._status_to_terminated()
-            except (requests.RequestException, novaclient.exceptions.ClientException):
-                self.logger.debug('Could not reach the OpenStack API')
+            except (requests.RequestException,
+                    novaclient.exceptions.ClientException,
+                    novaclient.exceptions.EndpointNotFound) as exc:
+                self.logger.debug('Could not reach the OpenStack API due to %s', exc)
                 if self.status != Status.Unknown:
                     self._status_to_unknown()
             else:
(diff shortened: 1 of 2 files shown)


UmanShahzad:
@itsjeyd Let me know if the last commit is confusing; with the new creds we have now, that's going to be required if we want CircleCI to use them.

UmanShahzad:
@itsjeyd Probably safer to discuss that detail in the JIRA thread; will do that now.

UmanShahzad:
@itsjeyd I made some minor additions (more places to handle the possible exception) and added the comment :)

itsjeyd:
@UmanShahzad Thanks! You [read my mind](https://tasks.opencraft.com/browse/OC-3378?focusedCommentId=53588&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-53588) :) It would be nice to give an example, though, as you did when explaining your reasoning on the ticket.

0ede8de5f89e98bc1614b27b85a02840a29fce37 looks good!

UmanShahzad:
@itsjeyd Alrighty, added an example.

UmanShahzad:
Getting failed builds because we reached the quota again on the new Horizon account :p I'll contact OVH to raise it.

UmanShahzad:
Squashed from 64f7e1c6cb8f1bf154c511535ed16c5d7dc7e1c2

PR #253, opened 2017-12-01
As a temporary measure, ignore action for endpoint errors.

UmanShahzad:
If we move status to 'Unknown' for an instance (which we're doing as part of moving to GRA3 without multi-region support in Ocim), it will deconfigure it from the LB, when it really shouldn't.


UmanShahzad:
Darn, bad time to drop below coverage requirements.

PR #256, opened 2017-12-20
Add setting to avoid memory leak, and add comment about number of workers

clemente:
Steps taken to alleviate memory-related performance issues and errors occurring since the Ginkgo upgrade.

Note that 80000 is an arbitrary number that should restart workers every few days. With 3 LMS this gives 240k requests, +10% static 260k, this means that only after 260k requests will a worker restart, this can be 6 days if 40k requests/day, 12 days if 20k requests/day, etc.

An alternative solution would be to do this change in the `configuration` branch instead of Ocim.

## Testing
- deploy an appserver in an instance that has already appservers 
- check that the new setting has been correctly applied to the new appserver (but not changed in the old ones)

## Reviewers
- [ ] @kaizoku

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -206,7 +206,11 @@ class OpenEdXConfigMixin(models.Model):
             },
 
             # Restart workers regularly to work around a memory leak
-            "EDXAPP_LMS_MAX_REQ": 80000,
+            # Tailor the max_requests number to suit the average request load on the server.
+            # e.g if the instance receives an average of around 42000 requests per day,
+            # with 10% for static assets, restarting after 20000 requests means
+            # restarting each of the 4 LMS workers about every 2-3 days.
+            "EDXAPP_LMS_MAX_REQ": 20000,
 
             # Celery workers
             "EDXAPP_WORKER_DEFAULT_STOPWAITSECS": 1200,
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -198,10 +198,15 @@ class OpenEdXConfigMixin(models.Model):
             },
 
             # Gunicorn workers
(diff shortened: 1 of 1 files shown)


kaizoku:
I've tested the functions locally, but cannot deploy from my test instance.
@clemente Do you have openstack setup locally to test deployments, or is there a test OCIM instance we can use to test this?

clemente:
@kaizoku We don't need a local openstack. It's possible to deploy from the local Ocim through the „launch new appserver“ button, by putting the username and password from `opencraft-im.env` into your `.env`; you need to adjust many other variables too (upload SSH key, etc.). In the end the instance will appear in Ocim stage. You can ask me for variables in Mattermost or ticket too.

clemente:
@kaizoku However you might consider a shortcut: check that Ocim works well enough to set up the gunicorn variables in the right places. You don't need to wait 2 h for Open edX to deploy correctly; you could ssh to the instance and check the variables there.

clemente:
@kaizoku Could you comment on the 80000 number? I fear it could be too high and therefore not be used. E.g. with 1000 requests per day and 4 workers, that's 250 requests/worker_day.  80000/250=320, each worker will restart after 320 days maximum. Maybe it makes sense for a server without much traffic, but we could lower the value too.

kaizoku:
@clemente 80000 does seem high for less active servers.
Even some of the other heavily used servers are only set to 20000, a quarter of that.
Even with only 3 workers, the max requests must be very low to risk multiple workers restarting at the same time.
Would 20,000, or even 10,000 be too low?

kaizoku:
@clemente I tested deploying a new appserver, all the configuration is set appropriately.
We just need to tailor that request limit to a good default, and this should be good to merge.

clemente:
@kaizoku Thanks, I merged it with 20000, and I included a comment which was in the other branches.

PR #257, opened 2017-12-26
Ensure only valid subdomain names can be used.

xitij2000:
Currently there is very little validation of subdomain names while registering a beta instance. This means that:
- It is possible to pick subdomain names that are simply invalid (for instance they include an underscore). 
- It is possible to override someone else's subdomain by using different the same name with a different case. 
- It is possible to use dots in the subdomain names, which can have unexpected effects if one user picks a subdomain under another user. 
- It is possible to pick a subdomain name that is longer than the allowed 63 characters.

This pull request includes changes that ensure that it is no longer possible to register domain names that cause the above issues. 

**Merge deadline**: None

**Testing instructions**:

1. Set up Ocim using the patches in this pull request. 
2. Visit /registration/ 
3. Try to register an instance that is too long, uses capital letters, a dot or other invalid characters like an underscore. 
4. The registration should fail and the registration form should show the appropriate error message.

Existing instances that don't pass the above validation will need manual intervention. 

**Reviewers**
- [ ] @pomegranited


registration/migrations/0006_auto_20171225_0938.py
```diff
@@ -0,0 +1,77 @@
...
+from __future__ import unicode_literals
+
+from django.core import validators
+from django.db import migrations, models
+from django.db.models import Q, Func, F
+
+import registration.models
+
+
+def migrate_to_lowercase_domains(apps, schema_editor):
```
pomegranited: Could you please split this migration into two parts, one which adds the subdomain validation to the model, and a separate data migration for fixing the existing internal domains?

Then, if you submit the data change as a separate PR, then we can merge the schema change and deploy it, and at least stop any new invalid domains from being created until we can sort out the manual changes required.

registration/models.py
```diff
@@ -79,7 +79,7 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
 
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
     subdomain = models.CharField(
-        max_length=255,
+        max_length=63,
```
pomegranited: Why 63?  (I haven't read the RFC).

Have you checked to see if any of the existing instances' subdomains exceed 63 characters?

Since this change will actually cause the database field to be truncated for all instances (be they archived or not), it's probably best to avoid changing the actual `max_length` and instead just add another validator.

registration/models.py
```diff
@@ -91,8 +91,9 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
                 'The subdomain name must at least have 3 characters.',
             ),
             validators.RegexValidator(
-                r'^[\w.-]+$',
-                "Please include only letters, numbers, '_', '-' and '.'",
+                r'^[a-zA-Z0-9]([a-zA-Z0-9\-]+[a-zA-Z0-9])?$',
```
pomegranited: We don't need to allow uppercase letters here, if we're just going to lowercase them later anyway.

registration/tests/test_views.py
```diff
@@ -224,14 +224,37 @@ def test_valid_application(self):
...
+        self.form_data['subdomain'] = good_subdomain
+        self._assert_registration_succeeds(self.form_data)
+
+    @data(
+        '?hogwarts', '?hogwarts?', 'hogwarts?',
+        '-hogwarts', '-hogwarts-', 'hogwarts-',
+        '_hogwarts', 'hog_warts', 'hogwarts_',
+        'hog.warts',
+    )
+    def test_invalid_subdomain(self, bad_subdomain):
```
pomegranited: Please add tests for uppercase subdomains too.

registration/tests/test_views.py
```diff
@@ -224,14 +224,37 @@ def test_valid_application(self):
...
+    def test_invalid_long_subdomain(self):
+        """
+        Subdomains not RFC-1034 conformant throw an error.
+        """
+        self.form_data['subdomain'] = 'hogwarts-school-of-witchcraft-and-wizardry-the-un-official-website'
         self._assert_registration_fails(self.form_data, expected_errors={
-            'subdomain': ["Please include only letters, numbers, '_', '-' "
-                          "and '.'"],
+            'subdomain': ['Ensure this value has at most 63 characters (it has 66).'],
         })
```
pomegranited: Please also add a tests that illustrates what happens when uniqueness is violated (including case insensitivity).

registration/models.py
UmanShahzad: I suppose this is a UX thing: do we want to tell the user they have to lowercase their chars? (This also means the error message has to change).

pomegranited: Yep, I think it's ok to tell the user to lowercase their characters.

We also don't want to have to deal with uniqueness checks that fail *after* we lowercase their domains.

registration/models.py
xitij2000: I did do a check of all the instances. I exported all instances using the API and used a Python script to validate the maximum length and the regex. No subdomains were too long, but a couple failed validation. Those were attached to the ticket. 

I will change this to a validator though, as that will also give us better control over the error message when validation fails.

registration/models.py
xitij2000: Perhaps we can accept upper case letters but force the domain name to lower case via JavaScript on blur?

UmanShahzad: That would work but is extra work for not that much gain I think :P

registration/tests/test_views.py
xitij2000: The ``test_existing_subdomain`` already tests for violating uniqueness, and I just realised that with the regex change, it won't be possible to add a subdomain in upper case since it'll already be rejected for violating the regex.

xitij2000:
Sorry about the WIP. I forgot to remove that once I knew I was done. 
I have rebased and pushed code that addresses your comments other than a test for checking case-insensitive uniqueness, which (I think) won't be an issue anymore since subdomains are forced to be lower-case.

registration/migrations/0006_auto_20180105_1631.py
```diff
@@ -0,0 +1,50 @@
...
+                        'The subdomain name must at least have 3 characters.'
+                    ),
+                    django.core.validators.MaxLengthValidator(
+                        63,
+                        'The subdomain name can have at most have 63 '
+                        'characters.'
+                    ),
+                    django.core.validators.RegexValidator(
+                        '^[a-z0-9]([a-z0-9\\-]+[a-z0-9])?$',
+                        'Please include only lower-case letters, numbers, '
```
pomegranited: nit: since your regex handles the "minimum 3 characters" requirement, I don't ever see the `'The subdomain name must at least have 3 characters.'` error.

I think it's best to keep the `MinLengthValidator` as-is, just in case someone comes along later and changes our regex.  But maybe just update the message to `"Please include at least 3 lower-case..."`

pomegranited:
Closed in favour of https://github.com/open-craft/opencraft/pull/263.

PR #258, opened 2017-12-29
Use `cached_db` session engine to persist sessions

xitij2000:
Switch to using ``django.contrib.sessions.backends.cached_db``  for sessions so redeploying an Instance doesn't lead to expired sessions. 

**Merge deadline**: ASAP

**Testing instructions**:
1. Deploy an instance using an Ocim with this path
2. Sessions running on that AppServer for that instance should persist as a new AppServer is launched, or if the running AppServer is restarted.

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -161,7 +161,8 @@ class OpenEdXConfigMixin(models.Model):
 
             # Available as ENV_TOKENS in the django setting files.
             "EDXAPP_ENV_EXTRA": {
-                "LANGUAGE_CODE": 'en'
+                "LANGUAGE_CODE": 'en',
+                "SESSION_ENGINE": 'django.contrib.sessions.backends.cached_db',
             },
 
             # Features


pomegranited:
@xitij2000 I rebased your PR on `master` just to unblock the merge here.

pomegranited:
@xitij2000 @tomaszgy CC @smarnach  I haven't tested this PR directly, but I've tested its effects on the [oc2096.opencraft.hosting](https://oc2096.opencraft.hosting) test instance, and it maintains sessions as desired.  👍 

Here's what I did:
1. Activated 2 appservers for the instance, Appserver 4 and Appserver 5.
   Waited for the load balancer to reconfigure.
1. Logged into [oc2096.opencraft.hosting](https://oc2096.opencraft.hosting) as the `honor` user
1. Watched the logs on both appservers, and determined that the load balancer was routing me to Appserver 4.
1. Deactivated Appserver 4, and waited for the LB.
1. Refreshed the page on [oc2096.opencraft.hosting](https://oc2096.opencraft.hosting), and noted that there's no disruption to my session, and I'm still logged in as the `honor` user.

PR #259, opened 2017-12-31
WIP: New application to periodically build edx/edx-platform, and notification system when deployment fails

clemente:
Prototype for https://tasks.opencraft.com/browse/OC-2167 ([discovery document](https://docs.google.com/document/d/1kZUGgFp-rl-YVzRBOp8QaK20mDS4vB418HN0dX0725w/edit?ts=5a46a3d2#)).

This adds two features:
- a task that regularly (every two hours) builds `edx/edx-platform` (`master` branch) at its current status, using `edx/configuration`.
- a notification system for appserver deployment failures, it will send e-mails to DevOps OpenCraft when OVH or the ansible playbook fails. This applies to all appservers. In the case of the servers built for the previous task (edx CI), it will also send e-mails to edX devops.

WIP because there are still FIXMEs for things that need to be decided or improved..

**JIRA tickets**: None
**Discussions**: None
**Dependencies**: Please merge https://github.com/open-craft/documentation/pull/240 too
**Screenshots**: None
**Sandbox URL**: None
**Merge deadline**: None

**Testing instructions**:

1. Define variables in your `.env`
```
INFRASTRUCTURE_DEPLOYMENT_PROBLEMS_EMAIL = "ops@example.com"
OPENEDX_DEPLOYMENT_PROBLEMS_EMAIL = "edx-devops@example.com"
```
2. `honcho run python3 ./manage.py shell_plus`
3. `from periodic_builds.tasks import deploy_edx_edxplatform; deploy_edx_edxplatform()` to test that the task runs. But from now on, you can also use `instance=OpenEdXInstance.objects.get(internal_lms_domain__startswith='master')` and then just `spawn_appserver(instance.ref.pk, mark_active_on_success=False, num_attempts=2)` each time you test (no need to reload `shell_plus`, but Ocim itself needs to be reloaded)
4. The first time, an `OpenEdXInstance` will be created and a server spawned (check it in Django's admin). Watch the logs from console or from Ocim's admin to check that it's deploying
5. Repeat the past steps in many more situations, including different types of failures. Details follow:
6. To simulate an OVH failure, go to `instance/models/mixins/load_balanced.py`, at `set_dns_records` function, and add a `raise Exception("DNS on strike today")`. This should cause an infrastructure error and en e-mail should be sent to OpenCraft only, with a stack trace
7. To simulate an ansible playbook failure, go to `instance/models/openedx_instance.py`, at `spawn_appserver` and a `return None` very early. This should cause an openedx error, with e-mail sent to OpenCraft and edX
8. Verify that the server deploys correctly. WIP. This won't happen yet in the prototype due to a missing "openstack" role; see the discovery document or https://github.com/edx/configuration/pull/3522
9. Delete all test servers from OVH

**Reviewers**
- [ ] TBD, not yet

**Author concerns**: None
**Settings**: None

### DIFF
--- a/instance/emails.py
+++ b/instance/emails.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""
+Send notification emails when certain deployment events happen.
+"""
+
+import traceback
+
+from functools import wraps
+from django.conf import settings
+from django.core.mail import send_mail
+from django.template.loader import get_template
+
+def send_emails_on_deployment_failure(method):
+    """
+    Decorator around spawn_appserver that catches all the emited exceptions and notifies by email
+    when it detects that a server deployment failed because of infrastructure problems.
+    It doesn't change spawn_server's behaviour (it will still raise exceptions on error).
+    It is done as a decorator to separate the e-mail code from the server-spawning code.
+
+    Different instances might have different e-mail requirements. E.g. an instance that builds the 'master'
+    branch of edx/edx-platform will send an e-mail to edX devops in case of failure.
+    In a future version, we could detect the 'owner' of the instance (e.g. the person who created the PR) and notify
+    them by e-mail when a deployment finished with or without error.
+    """
+    @wraps(method)
+    def wrapper(self, *args, **kwds): #pylint: disable=missing-docstring
+
+        try:
+            result = method(self, *args, **kwds)
+        except Exception as e:
+            # This means that the deployment failed due to infrastructure problems (which raise exceptions), e.g.
+            # OVH down, or an incorrect MySQL/MongoDB login/password, etc.
+            # Notify OpenCraft
+            if settings.INFRASTRUCTURE_DEPLOYMENT_PROBLEMS_EMAIL:
+                stacktrace = traceback.format_exc()
+                context = {
+                    'instance': self,
+                    'exception': e,
+                    'exception_type': e.__class__.__name__,
+                    'stacktrace': stacktrace,
+                }
+                subject = get_template('instance/emails/infrastructure_deployment_error_subject.txt').render(context)
+                subject = ''.join(subject.splitlines())
+                text = get_template('instance/emails/infrastructure_deployment_error_body.txt').render(context)
+                sender = settings.DEFAULT_FROM_EMAIL
+                dest = [settings.INFRASTRUCTURE_DEPLOYMENT_PROBLEMS_EMAIL]
+                send_mail(subject, text, sender, dest)
+
+            raise
+
+        if not result:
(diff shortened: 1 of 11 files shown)


periodic_builds/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    It uses the master branch from https://github.com/edx/edx-platform, https://github.com/edx/configuration and
+    downloads all dependencies specified in the requirements files.
+    It acts as a "continuous integration" system that uses Ocim's servers, including OVH and OpenStack, instead of
+    the other technologies used in edX's CI system.
+    The built version doesn't depend on any PRs (see WatchedFork and pr_watch for that).
+    Because a deployment can take up to 2 hours, don't run this task more often than that, or else you'll have appserver overflow.
+    """
+
+    # Find suitable MongoDB server to avoid choosing one randomly
+    # FIXME probably delete this part, we should not hardcode values. Check whether we need to tell which server to use vs. when it's randomly selected (the default) and decide whether to use this code or not
```
smarnach: The random server selection only selects among the servers that are flagged to accept new clients.  In production this is currently just a single one, and we will probably need to adapt the server selection strategy at some point.  However, this is completely out of scope for this work here – just let the code select a server, as it does for any other instance we deploy.

periodic_builds/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    # try:
+    #     mongodb_server = MongoDBServer.objects.get(name="OCIM Default")
+    # except MongoDBServer.DoesNotExist:
+    #     mongodb_server = select_random_mongodb_server()
+
+
+    instance, created = OpenEdXInstance.objects.get_or_create(
+        internal_lms_domain=generate_internal_lms_domain('master'),
```
smarnach: You should be able to simply pass in `subdomain='master'` or similar.

periodic_builds/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+        configuration_source_repo_url='https://github.com/edx/configuration',
+        configuration_version='master',
+        edx_platform_commit='master',
+        openedx_release='master',
+        deploy_simpletheme=True, # FIXME add extra configuration variables that actually change some color
+        #mongodb_server=mongodb_server.pk, # FIXME remove, see above
+    )
+    if created:
+        # Name is set separately because it's stored in InstanceReference
+        instance.name = 'Integration - Open edX master periodic build'
```
smarnach: This function contains various hard-coded values.  I'm not really sure whether it makes sense to make them configurable at the current point – probably not, since I don't see how that would be useful.

periodic_builds/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+        instance.save()
+
+    # There is a wrapper around spawn_appserver that will check the build result and will send e-mails,
+    # see in send_emails_on_deployment_failure
+    spawn_appserver(instance.ref.pk, mark_active_on_success=False, num_attempts=2)
```
smarnach: Note that this can easily take longer than two hours.  I would remove `num_attempts` – there will be a new attempt in two hours anyway.  (This still wouldn't guarantee we don't have overlap between deployments, but this isn't a problem in itself.)

instance/emails.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
```
smarnach: We already get emails when app server builds fail.  How does this new implementation relate to the old one?

periodic_builds/tasks.py
clemente: Thanks for explaining it; I didn't find it after a first look. I removed it.

periodic_builds/tasks.py
clemente: Added as FIXME.

periodic_builds/tasks.py
clemente: Removed.

instance/emails.py
clemente: I didn't know about those e-mails (I'm not in the devops list). Probably some of those e-mails are not needed then. But there's the option of providing more details, or do a better detection of whom to e-mail. E.g. e-mail the person who clicked the blue button. Or find the github username of the PR owner and e-mail their e-mail. Or e-mail also in case of success („build finished“).

clemente:
I rebased on master (latest code was >1 year ago) and edited some FIXMEs. Didn't test it yet.

clemente:
A better version of this code was built in SE-1544 and merged.

PR #260, opened 2018-01-03
Use Ansible to provision Vagrant box instead of bootstrap script

xitij2000:
With this update Vagrant will use Ansible for provisioning the developer environment. 

**Dependencies**: 
Needs open-craft/ansible-opencraft#10 to work as expected.

**Merge deadline**: None

**Testing instructions**:
Provision developer Vagrant box as normal by running ``vagrant up`` and the provisioning procedure should now use Ansible instead of ``bin/bootstrap.sh``.

**Reviewers**
- [ ] TBD


kaizoku:
@xitij2000 Should we also remove the bin/bootstrap script as it's not used anymore?

Also I like the docker addition, but it may be better to split it into another pull request so we can have 2 seperate entries in the git log.

Otherwise this looks good. I deployed locally and it worked for me.

deploy/opencraft.yml
```diff
@@ -14,6 +14,10 @@
       UNATTENED_UPGRADES_ERRORS_RELAY_TO: "{{ inventory_hostname }}"
     - name: forward-server-mail
       FORWARD_MAIL_MYDOMAIN: "{{ inventory_hostname }}"
+      when: vagrant_mode is undefined or not vagrant_mode
     - name: geerlingguy.nginx
       nginx_remove_default_vhost: true
+      when: vagrant_mode is undefined or not vagrant_mode
     - opencraft
+    - name: pmbauer.tarsnap
+      when: vagrant_mode is not undefined or vagrant_mode
```
UmanShahzad: We're running tarsnap for vagrant?

xitij2000: I did this because of this comment https://github.com/open-craft/opencraft/blob/3666dd84d8d986167b8231b8f02f46ac332ec2aa/deploy/install_tarsnap.yml#L1-L4
I think I misinterpreted it. Reading it again it does seem like it should be possible to optionally install tarsnap on Vagrant.

xitij2000:
Sure, will split out the Docker bits and remove the bootstrap script

Vagrantfile
```diff
@@ -20,10 +20,16 @@ Vagrant.configure(2) do |config|
...
   config.ssh.forward_x11 = true
 
-  config.vm.provision 'shell',
-                      path: 'bin/bootstrap',
-                      privileged: false,
-                      keep_color: true
+  config.vm.provision "ansible" do |ansible|
+    ansible.playbook = "deploy/opencraft.yml"
+    ansible.raw_arguments = ["--extra-vars", "@deploy/vagrant_vars.yml"]
+    if FileTest.exists?("deploy/private.yml")
```
UmanShahzad: Sorry I'm not so good with Ruby but could we loop through these as a list of `['deploy/private.yml', 'private.yml']`?

deploy/opencraft.yml
UmanShahzad: What I meant was, *should* we be? I don't think we had tarsnap running on vagrant dev environments before, and I don't think we even need it for Ocim itself because it's stateless AFAIK.

UmanShahzad: @xitij2000 I think the confusion is that, I think you meant to put `when: vagrant_mode is undefined or not vagrant_mode` here.

Vagrantfile
xitij2000: I'm not overly familiar with Ruby either, so tried to keep it simple, but I think looping over it makes sense, will actually be simpler.

deploy/opencraft.yml
xitij2000: What I understood from that comment is that keeping the tarsnap Ansible script separate allows us to install tarsnap on Vagrant for testing, so we shouldn't restrict tarsnap based on whether the deployment is via Vagrant or not. 
If ``tarsnap_tarsnapper_conf`` is not configured, then, in any case, the tarsnap scripts and chron config won't be set up. If I add ``when: vagrant_mode is undefined or not vagrant_mode`` then it will be hard to set up tarsnap for testing inside vagrant (using this play at least).

UmanShahzad: @xitij2000 Ah, I see what you mean. In that case :+1:

Vagrantfile
UmanShahzad: Your solution looks fine to me, as long it runs :P :+1:

xitij2000: :-) I did a basic test. i.e. Just created those two files, and saw them pop up on the command line for the Ansible process' command line params. If the file is correctly being passed to Ansible, after that it's all up to Ansible anyway.

kaizoku: My bit of ruby experience tells me this loop is textbook ruby. ;]

kaizoku:
:+1:

- [✓] I tested this: (deployed a new vagrant devstack using this branch)
- [✓] I read through the code
- [✓] Includes documentation

PR #262, opened 2018-01-04
Fix an edge case in get_pr_list_from_usernames.

mtyaka:
When invoked with an empty username list, the function would construct the Github API query as `is:open is:pr  repo:edx/edx-platform`, which returns all PRs from the requested fork, instead of no PRs.

**Testing instructions**:

1. Without this patch, start the django shell. Invoke `pr_watch.github.get_pr_list_from_usernames([], 'edx/edx-platform')` and observe it fetches a page of *all* PRs from the Github API.
1. Install the patch & restart the shell.
1. Invoke the function again, verify it returns an empty list.

**Reviewers**:
- @pomegranited

### DIFF
--- a/pr_watch/github.py
+++ b/pr_watch/github.py
@@ -152,6 +152,9 @@ def get_pr_list_from_usernames(user_names, fork_name):
     """
     Retrieve the current active PRs for a given set of users
     """
+    if not user_names:
+        return []
+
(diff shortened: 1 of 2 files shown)


mtyaka:
Rebase from 1fe061d97032aaeb115067839d6c76bf8e1b89c7.

PR #263, opened 2018-01-05
Validate domain names, and fix invalid domain names

xitij2000:
Currently there is very little validation of subdomain names while registering a beta instance. This means that:
- It is possible to pick subdomain names that are simply invalid (for instance they include an underscore). 
- It is possible to override someone else's subdomain by using different the same name with a different case. 
- It is possible to use dots in the subdomain names, which can have unexpected effects if one user picks a subdomain under another user. 
- It is possible to pick a subdomain name that is longer than the allowed 63 characters.

This pull request includes changes that ensure that it is no longer possible to register domain names that cause the above issues. It also includes a data migration that will convert fix existing domain names so that they pass validation.

**Merge deadline**: None

**Testing instructions**:
1. Take note any invalid domain names in Ocim.
2. Set up Ocim using the patches in this pull request. 
3. Check previously invalid domain names; they should now be fixed to pass validation.
4. Visit /registration/ 
5. Try to register an instance that is too long, uses capital letters, a dot or other invalid characters like an underscore. 
6. The registration should fail and the registration form should show the appropriate error message.

**Reviewers**
- [ ] @pomegranited


registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('registration', '0006_auto_20180105_1631'),
+    ]
+
+    operations = [
+        migrations.RunPython(fix_invalid_domain_names),
```
pomegranited: Have to provide a `reverse_code` function, even if in this case, it's a no-op.
```python
def migrate_noop(apps, schema_editor):
   # Cannot reverse the data migration, so just no-op
   pass
...
      migrations.RunPython(fix_invalid_domain_names, migrate_noop)
```

registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -0,0 +1,58 @@
...
+        function='btrim',
+    )
+    remove_trailing_hyphens = Func(
+        trim_hyphens,
+        Value(r'[-]+\.'),
+        Value('.'),
+        Value('g'),
+        function='regexp_replace'
+    )
+    return Lower(remove_trailing_hyphens)
```
pomegranited: Good that you're not stripping out `.`s here -- that would break `internal_lms_domain` 👍

registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -0,0 +1,58 @@
...
+        Value('_'),
+        Value('-'),
+        function='replace'
+    )
+    trim_hyphens = Func(
+        remove_underscores,
+        Value('-'),
+        function='btrim',
+    )
+    remove_trailing_hyphens = Func(
```
pomegranited: note: this won't strip trailing hyphens out of any `subdomain`s which created  `internal_lms_domain`, but that's probably ok.

registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -0,0 +1,58 @@
...
+        trim_hyphens,
+        Value(r'[-]+\.'),
+        Value('.'),
```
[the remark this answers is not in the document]
ciuin: No, actually. I've only ever seen it when managing DNS, but I'm pretty sure applications are always supposed to honor it, according to the RFCs. It makes the implicit root domain explicit. https://en.wikipedia.org/wiki/Fully_qualified_domain_name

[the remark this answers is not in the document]
bradenmacdonald: No I think it searches for "-." or "----." etc. and replaces it with ".", preserving the existing dot?  e.g. bad-.example.com becomes bad.example.com, not badexample.com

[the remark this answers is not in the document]
pomegranited: Oh ah, gotcha.  Thanks for the clarification!

registration/migrations/0007_fix_invalid_domain_names.py
xitij2000: I think the ``remove_trailing_hyphens`` should do that.

registration/migrations/0007_fix_invalid_domain_names.py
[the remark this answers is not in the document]
xitij2000: Yes. I added this because the old regex would allow trailing hyphens or underscores, and the previous step would convert a trailing underscore to a hyphen. 

There is still a chance for collision where two users have subdomains like ``_bad_subdomain`` and ``bad_subdomain`` which would both resolve to ``bad-subdomain``, but I the main Ocim doesn't have any of those. 

Should I add code to handle those cases if we / others have additional Ocim deployments?

registration/migrations/0007_fix_invalid_domain_names.py
xitij2000: Yes, that is why I left it out. A  dot in internal_lms_domain is valid anyway. I think the registration subdomain case will be taken care of when we change the subdomain manually in the CMS based on feedback (if any).

registration/migrations/0007_fix_invalid_domain_names.py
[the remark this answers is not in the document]
pomegranited: @xitij2000 AFAIK, no one else runs Ocim yet.  We have a client who is considering it for their production deployments, but their domains would likely be very clean, so I wouldn't worry about handling collisions here.

registration/migrations/0007_fix_invalid_domain_names.py
pomegranited: Yep 👍

registration/migrations/0007_fix_invalid_domain_names.py
pomegranited: Yep, so it will!  My regex fu must be rusty :)

pomegranited:
Note: we'll wait to merge and deploy until the existing invalid beta instances have been given notice of the change.

PR #265, opened 2018-01-10
Log more information on the failure of swift container download.

tomaszgy:
This PR introduces enhanced logging of errors on the failure of swift container download.

**JIRA tickets**: This PR is a first step in investigation for the issue in [OC-3611](https://tasks.opencraft.com/browse/OC-3611). Another PR might be needed, once the source of the issue is found.

**Dependencies**: None

**Screenshots**: None

**Sandbox URL**: None

**Merge deadline**: If possible, very soon so that further investigation of issue from [OC-3611](https://tasks.opencraft.com/browse/OC-3611) can be made.

**Testing instructions**:

1. Run the existing tests.

1. It would be great to test it with the connection to OpenStack configured - I haven't done it yet as I don't have one.

**Reviewers**
- [x] @kaizoku 

CC: @smarnach 

Squashed and rebased from de9479af5a83b9a9df4e866412e7d1876a1e0bc1 onto master.

Signed-off-by: Tomasz Gargas <tomasz@opencraft.com>

### DIFF
--- a/instance/openstack_utils.py
+++ b/instance/openstack_utils.py
@@ -39,7 +39,8 @@ logger = logging.getLogger(__name__)
 
 # Data objects ################################################################
 
-FailedContainer = namedtuple('FailedContainer', ['name', 'number_of_failures'])
+FailedContainer = namedtuple(
+    'FailedContainer', ['name', 'number_of_failures', 'extra_information'])
 StatContainer = namedtuple('StatContainer', ['read_acl', 'write_acl', 'bytes'])
 # A simplified version of openstack.network.v2.security_group_rule.SecurityGroupRule
 # This version removes fields like 'id', 'security_group_id', 'updated_at' so
@@ -219,7 +220,7 @@ def download_swift_account(download_target, **kwargs):
              containers backed up successfully.
     """
 
-    errors = defaultdict(lambda: 0)
+    errors = defaultdict(lambda: {'failures_count': 0, 'extra_information': []})
 
     with swift_service(**kwargs) as service:
         downloader = service.download(options={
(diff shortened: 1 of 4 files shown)


instance/openstack_utils.py
```diff
@@ -234,11 +235,17 @@ def download_swift_account(download_target, **kwargs):
                 if file_download_result['response_dict']['status'] == 304:
                     continue
 
-                errors[file_download_result['container']] += 1
+                errors[file_download_result['container']]['failures_count'] += 1
```
kaizoku: nit: With the extra index here, the nested index that was already there is a bit confusing. Perhaps we could add a variable like so
```
container_name = file_download_result['container']
errors[container_name]['failures_count'] += 1
errors[container_name]['extra_information'].append(
...
```

tomaszgy: Good point!

instance/openstack_utils.py
```diff
@@ -39,7 +39,8 @@
 
 # Data objects ################################################################
 
-FailedContainer = namedtuple('FailedContainer', ['name', 'number_of_failures'])
+FailedContainer = namedtuple(
+    'FailedContainer', ['name', 'number_of_failures', 'extra_information'])
```
kaizoku: Looks like this change broke some of the tests in instance/tests/test_openstack_utils.py.
Might this affect other uses of FailedContainer? If not we can change those tests too.

tomaszgy: @kaizoku Thanks for pointing that out!

`FailedContainer` is used only in `openstack_utils.download_swift_account` function which itself is used only in `backup_swift.tasks.do_backup_swift` so I've just updated the failing tests.

tomaszgy:
@kaizoku Thanks for the review! All the tests pass now. If you have a second, have a look at this again and if it's fine please give +1 so I could deploy it :)

kaizoku:
:+1:

- [✓] I tested this: (checked out code, ran tests locally)
- [✓] I read through the code
- [n/a] Includes documentation

tomaszgy:
Thanks @kaizoku. Just rebased and squashed. If the CircleCI goes OK, I'll merge it and deploy it.

PR #266, opened 2018-01-11
Fix LB stickiness by inserting a dedicated cookie.

bdero:
Also use /heartbeat for health checking; the current version only verifies the
server accepts TCP connections.

I'm going to try and expedite this fix by @smarnach on [OC-3169](https://tasks.opencraft.com/browse/OC-3169) through this week. See the ticket for details. In a nutshell, our method of populating the sticky map (hashing by the session cookie) broke for Ginkgo, which causes requests from the same client/session to be uniformly distributed between all servers in the backend. Of course, this causes all kinds of problems for any instance with more than one appserver, including:
* Static asset 404s if, for any reason, static assets happen to not be hashed the same during collection between different appservers (we saw this happen on an ecommerce deployment recently).
* Issues starting sessions when not using the persistent session backend (we also saw this happen).
* Issues with any highly ephemeral state stored between requests in memcached.

### DIFF
--- a/instance/templates/instance/haproxy/openedx.conf
+++ b/instance/templates/instance/haproxy/openedx.conf
@@ -1,10 +1,10 @@
 {% autoescape off %}
     # Backend configuration for {{ domain }}
-    stick-table type string len 128 size 20k expire 12h
-    stick on req.cook(sessionid)
+    cookie openedx-backend insert postonly indirect
     acl has-authorization req.hdr(Authorization) -m found
     http-request set-header Authorization 'Basic {{ http_auth_info_base64 }}' unless has-authorization
+    option httpchk /heartbeat
     {% for server in appservers %}
-    server {{ server.name }} {{ server.ip_address }}:80 check
+    server {{ server.name }} {{ server.ip_address }}:80 cookie {{ server.name }} check
     {% endfor %}
 {% endautoescape %}


bradenmacdonald:
Nice, and sorely needed. Thanks for this, @bdero!

bdero:
Build failed, rebuilding once to verify it's not a flake.

instance/templates/instance/haproxy/openedx.conf
```diff
@@ -1,10 +1,10 @@
 {% autoescape off %}
     # Backend configuration for {{ domain }}
-    stick-table type string len 128 size 20k expire 12h
-    stick on req.cook(sessionid)
+    cookie openedx-backend insert postonly indirect
     acl has-authorization req.hdr(Authorization) -m found
     http-request set-header Authorization 'Basic {{ http_auth_info_base64 }}' unless has-authorization
+    option httpchk GET /heartbeat HTTP/1.1
```
smarnach: I apparently got this wrong.  For HTTP version to 1.1, the Host header is mandatory, but we don't provide it, so all health checks return a 400 and all backends are disabled.  The easiest solution is to use the default of HTTP version 1.0 instead, so this line can simply become

    option httpchk /heartbeat

I already tested this by manually editing the configuration of your sandbox on the LB, and it worked fine.

instance/templates/instance/haproxy/openedx.conf
```diff
@@ -1,10 +1,10 @@
 {% autoescape off %}
     # Backend configuration for {{ domain }}
-    stick-table type string len 128 size 20k expire 12h
-    stick on req.cook(sessionid)
+    cookie openedx-backend insert postonly indirect
```
smarnach: This only sets the cookie on POST requests, since I expected it only to be relevant for logged in users.  You had some issues with the static assets, so maybe we need this for all users.  We can change `postonly` to `nocache` to set the cookies for anonymous users as well.

smarnach:
@bdero I fixed the bug I found and deployed the fix to stage.  I'd still like to hear your opinion on the question I asked.

I also tested what happens if the cookie is set to an appserver that's no longer available, but the behaviour seems to be fine: The cookie will keep the invalid value until the next POST request, and while the cookie is invalid the user is routed to a random backend.

smarnach:
I've actually done enough testing on my end.  If you are happy with `postonly` and have tested this on your end, feel free to merge and deploy to production.

instance/templates/instance/haproxy/openedx.conf
bdero: I think we should make it sticky for all users. Do you see any potential downsides to it?

If such a mismatched static asset problem happens again, we definitely don't want to discover that kind of failure the same way we did this time, and I would rather we just take the extra hit of clients possibly downloading the static assets multiple times without us noticing for a while in the worst case.

bdero:
@smarnach Thanks! What was the bug you found exactly? Was it the healthcheck that was causing the problem? I noticed in both the integration test and when I deployed to stage, I was getting 503s but had to redirect my attention elsewhere again.

One more question:
What do you expect an appserver failure scenario will look like exactly? If we're running 2 appservers for an instance, and one of them is marked unhealthy/taken out, is there a way we'll know it happened other than SSHing into the load balancer and tailing the logs? If not, we'll have to be really careful to figure out the state of HAProxy before deciding to take appservers out (without having any indication of which active appservers are healthy on Ocim, one of us could easily wind up taking out the one _good_ active appserver without realizing it).
It would be ideal if we get notified specifically when HAProxy decides to mark something unhealthy, since I don't think the NewRelic synthetics checks will catch this if HAProxy detects dead appservers and takes them out of the pool quickly.

smarnach:
@bdero Sorry, I completely missed your reply.

The bug was the broken health check, which caused HAProxy to disable all backend servers.  Everything else seems to be working as intended.

Your concerns that we don't really notice when one of two app servers is failing is completely valid.  Our support for multiple app servers is still pretty limited, but it might at least help to deal with load spikes, and slightly reduce the likelihood of downtimes.

We might be moving in a completely different direction with this anyway – there is an emerging consensus on [OC-3406](https://tasks.opencraft.com/browse/OC-3406) about moving to some Kubernetes-based solution, which would solve this problem. Kubernetes  deals with all this more or less automatically, by evicting all pods from nodes that have become unreachable and restarting containers when they fail liveness probes.

bdero:
@smarnach I think a kubernetes solution with containerized edX services would be great for many, many reasons, so +1 to that. For this solution, I agree we should roll it out for now, but it's a bit of a dragon that we should make others aware of (for when confusion/emergencies arise).

@bradenmacdonald Could you take a look at this conversation, so you understand this pitfall as well?

smarnach:
@bdero Thanks for your reply.

I'm also not really sure what to do when only one backend server is active.  Disabling the only backend server because the health check is failing does not seem very useful – the only thing it can help with is to show a nicer error page.

In my previous job, we used a special exception:  When _all_ backends reported unhealthy, we treated them all as healthy instead.  In can't be worse sending requests to unhealthy backends than not sending them anywhere at all.  I can't see how configure a similar rule with HAProxy.

I'm waiting for @bradenmacdonald to comment here before merging this.  I still think this PR is an improvement over the status quo.

bradenmacdonald:
@smarnach @bdero Thanks for the ping. I don't have a strong opinion on the healthcheck here. If haproxy doesn't have an option to ignore healthchecks when all servers are unhealthy, then I guess there's not much difference whether we're showing an error at the haproxy level or at the appserver level. Just make sure that errors triggered by the /heartbeat endpoint will still show in the appserver logs so we can debug issues as easily.

In any case, let's get this fix merged ASAP, with or without that healthcheck change.

UmanShahzad:
@smarnach Randomly chipping in, not exactly sure where this is, but was reading about a similar topic: would client IP hashing help here (ip's hash mapping to 1 appserver)?

smarnach:
@UmanShahzad Argh, this is still open!  Everytime I get here, I need to rebase, and then I forget to merge it two hours later.

We are using cookie now to store the backend, which is fine.  The problem we were discussing is whether the health checks do more harm than good – all our instances have only one app server at the moment, so the LB does not learn anything useful by health checking that app server, since removing it from the balancing set if it reports unhealthy is not helpful.

UmanShahzad:
@smarnach Ah, OK, makes sense. And if you want, I can merge it for you in 2 hours.

UmanShahzad:
Oops, pressing 'Update branch' didn't do what I expected.

UmanShahzad:
@smarnach Yep it merges. I got some pushback when doing that before though, so I haven't done that.

smarnach:
@UmanShahzad Yes, we are supposed to rebase before merging.

PR #267, opened 2018-01-12
Make instance list scrollable.

UmanShahzad:
Since we have so many instances now, it got annoying to try and scroll all the way down while losing sight of the appserver details, which I needed to see in some cases while scrolling.

To fix this, we make the instance column scrollable:

selection_001

selection_002

### DIFF
--- a/instance/static/scss/instance.scss
+++ b/instance/static/scss/instance.scss
@@ -1,3 +1,8 @@
+nav.top-bar {
+    position: fixed;
+    width: 100%;
+}
+
 .loading-msg {
     position: fixed;
     top: 3px;
@@ -21,6 +26,23 @@ ul.side-nav li.active {
 .content.instance-app {
     .row {
         max-width: none;
+    }
+
+    .scrollable {
+        overflow: auto;
+        height: calc(100vh - #{$header-height});
+    }
+
+    .scrollable::-webkit-scrollbar {
+        width: 1em;
+    }
+
+    .scrollable::-webkit-scrollbar-thumb {
+        background-color: lightgray;
+    }
+
+    .scrollable::-webkit-scrollbar-track {
+        -webkit-box-shadow: inset 0 0 2px rgba(0, 0, 0, 0.3);
     }
 
     .side-instance-list {
--- a/instance/static/scss/instance.scss
+++ b/instance/static/scss/instance.scss
@@ -1,6 +1,17 @@
+// Variables
+
+$header-height: 2.8125em;
+
+// SCSS
+
+header ~ main {
+    padding-top: $header-height;
+}
+
 nav.top-bar {
     position: fixed;
     width: 100%;
+    z-index: 1;
 }
 
 .loading-msg {
@@ -42,7 +53,7 @@ ul.side-nav li.active {
     }
 
     .scrollable::-webkit-scrollbar-track {
-        -webkit-box-shadow: inset 0 0 2px rgba(0, 0, 0, 0.3);
+        -webkit-box-shadow: inset 0 0 2px rgba(0, 0, 0, 0.2);
     }
 
     .side-instance-list {
--- a/requirements.txt
+++ b/requirements.txt
@@ -88,7 +88,7 @@ positional==1.1.1
 prettytable==0.7.2
 prompt-toolkit==1.0.9
 prospector==0.12.4
-psycopg2==2.6.2
+psycopg2==2.7.3.2
 ptyprocess==0.5.1
 pycodestyle==2.0.0
 pydocstyle==1.0.0
--- a/instance/static/html/instance/index.html
+++ b/instance/static/html/instance/index.html
@@ -8,7 +8,7 @@
 
 <div class="row">
   <!-- Instances List -->
-  <div class="large-3 columns">
+  <div class="large-3 columns scrollable">
     <ul class="side-nav side-instance-list">
       <li ng-repeat="instance in instanceList"
           ng-class="{active: instance.id == state.params.instanceId}"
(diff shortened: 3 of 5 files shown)


UmanShahzad:
Unfortunately: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar... but it works in my Chrome, although I have not tested it on Firefox, and it might just not work there. Still, I think it doesn't have a bad effect.

UmanShahzad:
@antoviaque Pinging in case this is something you'd like to have added.

bradenmacdonald:
@UmanShahzad Yes. Please. Thanks so much :)

I won't have time to review today but I'd love to see this merged ASAP. I could probably review Tuesday if nobody beats me to it.

UmanShahzad:
@bradenmacdonald Awesome, thanks!

I think we have a separate problem to solve though that is now affecting us: https://github.com/psycopg/psycopg2/issues/594

UmanShahzad:
@bradenmacdonald This is on stage now for easy verification.

instance/static/scss/instance.scss
```diff
@@ -1,3 +1,19 @@
+// Variables
+
+$header-height: 2.8125em;
```
UmanShahzad: I thought this could get picked up easily through the SCSS in `static/` by placing it in the dedicated global variables file, but the global SCSS files aren't actually *global*: they're overridden in the "custom CSS" block with this SCSS file. Adding them to it doesn't seem to help (or maybe I did it incorrectly).

Trying to then instead `@import` them directly in the SCSS file worked beautifully on stage, but then tests were failing because they couldn't find the import, since it wasn't built in the tests.

I finally gave up and just placed it right here. We should really consider either moving all SCSS/JS files to a single place, or ensure that *global* files really are *global*.

antoviaque:
@UmanShahzad Nice! :D Thanks for doing this - looks good from the screenshots.

UmanShahzad:
@antoviaque You can check it out live on stage, too. :)

instance/static/scss/instance.scss
```diff
@@ -23,6 +39,23 @@ ul.side-nav li.active {
         max-width: none;
     }
 
+    .scrollable {
+        overflow: auto;
+        height: calc(100vh - #{$header-height});
+    }
+
+    .scrollable::-webkit-scrollbar {
```
e-kolpakov: @UmanShahzad I assume you know what you're doing by using vendor-specific pseudo-elements, especially those that don't have analogs with other popular browser vendors (i.e. firefox and ie):

https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

Does it make sense to completely remove them or replicate the same effect using "some creativity", like [this SO answer](https://stackoverflow.com/a/30464993/882918) suggests?

e-kolpakov: @UmanShahzad I'd like to point out it also affected other scrollables, at least for me - note right scrollbar is also styled the same way:

image

While on prod it uses native scrollbar:

image

Unless it is intentional there might be a problem of (re)using too generic css class.

requirements.txt
```diff
@@ -88,7 +88,7 @@ positional==1.1.1
 prettytable==0.7.2
 prompt-toolkit==1.0.9
 prospector==0.12.4
-psycopg2==2.6.2
+psycopg2==2.7.3.2
```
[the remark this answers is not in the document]
UmanShahzad: I had to do this: https://github.com/open-craft/opencraft/pull/267#issuecomment-357379331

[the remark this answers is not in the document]
e-kolpakov: @UmanShahzad Maybe split it into separate PR and address independently?

instance/static/scss/instance.scss
UmanShahzad: Yeah I knew about this being vendor-specific, as you can [see in my comment above](https://github.com/open-craft/opencraft/pull/267#issuecomment-357342018). But it's just styling so I didn't bother with alternatives. I saw that SO answer but if you see the actual CSS suggested, that is very ugly, fragile code. If you insist, I can just delete this styling, but I thought it's better to have for at least some, than not having it at all.

UmanShahzad: Yes the 2nd scrollbar is intentional: that div itself becomes scrollable. See what happens when you make it responsive -- the actual browser scrollbar is not affected.

requirements.txt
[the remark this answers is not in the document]
UmanShahzad: @e-kolpakov Builds were failing without it. That would be a lot of overhead to make a new PR and also to wait 2+ hours for that build, no?

instance/static/scss/instance.scss
e-kolpakov: @UmanShahzad That's the question of taste of curse, so do what you see fit best, but is there particular reason you wanted a custom scrollbar in the first place (as opposed to just let browser use native one?)

UmanShahzad: No good reason, I just wanted to spice it up ;)

requirements.txt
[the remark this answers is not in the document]
e-kolpakov: @UmanShahzad ok (but the fact that long-running test suite prevents us from doing so is a problem on its own).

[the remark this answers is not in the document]
UmanShahzad: @e-kolpakov Very much agreed...

instance/static/scss/instance.scss
e-kolpakov: @UmanShahzad I'd slightly prefer going with native control just to reduce the amount of code (and non-standard solutions), but since I'm not the author of this PR I'll leave it up to your good judgement.

UmanShahzad: Good point, but since I already invested the effort to get that to work, and it doesn't negatively effect any users of non-Chrome browsers, I will opt to leave this here. Oh and then I can hit merge without waiting again :D

PR #268, opened 2018-01-22
Update ecommerce database configuration

xitij2000:
The Ansible variables for ecommerce database configuration have changed such that Ocim can no longer deploy an instance based on the standard edx/configuration repo. 

This change adds variables to support the new variables for configuring ecommerce. 

**Sandboxes**:

`edx:master`:
* LMS: https://ocpr268.sandbox.stage.opencraft.hosting/
* Studio: https://studio-ocpr268.sandbox.stage.opencraft.hosting/
* Ecommerce: https://ecommerce-ocpr268.sandbox.stage.opencraft.hosting/

`open-craft:opencraft-release/ginkgo.1`
* LMS: https://ocpr268-ginkgo.sandbox.stage.opencraft.hosting/
* Studio: https://studio-ocpr268-ginkgo.sandbox.stage.opencraft.hosting/
* Ecommerce: https://ecommerce-ocpr268-ginkgo.sandbox.stage.opencraft.hosting/

**Merge deadline**: None

**Testing instructions**:
1. Set up an Ocim instance based on this pull request. 
2. Create an instance based on edx/configuration:master and edx/edx-platform:master
3. Launch an AppServer for that instance
4. If all else works, the ecommerce configuration step should not fail during database migrations.

**Reviewers**
- [ ] @pomegranited

### DIFF
--- a/instance/models/mixins/openedx_database.py
+++ b/instance/models/mixins/openedx_database.py
@@ -234,14 +234,20 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
             "EDXAPP_MYSQL_PORT": self.mysql_server.port,
 
             # ecommerce
-            "ECOMMERCE_DEFAULT_DB_NAME": self._get_mysql_database_name("ecommerce"),
+            "ECOMMERCE_DATABASE_NAME": self._get_mysql_database_name("ecommerce"),
+            "ECOMMERCE_DATABASE_USER": self._get_mysql_user_name("ecommerce"),
+            "ECOMMERCE_DATABASE_PASSWORD": self._get_mysql_pass_from_dbname("ecommerce"),
+            "ECOMMERCE_DATABASE_HOST": self.mysql_server.hostname,
+
+            # Old ecommerce database settings kept around for compatibility with Ginkgo.
+            "ECOMMERCE_DEFAULT_DB_NAME": "{{ ECOMMERCE_DATABASE_NAME }}",
             "ECOMMERCE_DATABASES": {
                 "default": {
                     "ENGINE": 'django.db.backends.mysql',
-                    "NAME": self._get_mysql_database_name("ecommerce"),
-                    "USER": self._get_mysql_user_name("ecommerce"),
-                    "PASSWORD": self._get_mysql_pass_from_dbname("ecommerce"),
-                    "HOST": self.mysql_server.hostname,
+                    "NAME": "{{ ECOMMERCE_DATABASE_NAME }}",
+                    "USER": "{{ ECOMMERCE_DATABASE_USER }}",
+                    "PASSWORD": "{{ ECOMMERCE_DATABASE_PASSWORD }}",
+                    "HOST": "{{ ECOMMERCE_DATABASE_HOST }}",
                     "PORT": self.mysql_server.port,
                     "ATOMIC_REQUESTS": True,
                     "CONN_MAX_AGE": 60
--- a/instance/tests/models/test_openedx_database_mixins.py
+++ b/instance/tests/models/test_openedx_database_mixins.py
@@ -321,12 +321,24 @@ class MySQLInstanceTestCase(TestCase):
                 database={"name": "edx_notes_api", "user": "notes"},
                 include_port=False
             ),
-            "ECOMMERCE_": make_nested_group_info(
-                ["DEFAULT_DB_NAME", "DATABASES"],
-                [{"name": "ecommerce", "user": "ecommerce", "additional_settings": {
-                    "ATOMIC_REQUESTS": True,
-                    "CONN_MAX_AGE": 60,
-                }}]
+            "ECOMMERCE_": {
+                "vars": ["DATABASES"],
+                "values": [{
+                    "default": {
+                        "ENGINE": 'django.db.backends.mysql',
+                        "NAME": "{{ ECOMMERCE_DATABASE_NAME }}",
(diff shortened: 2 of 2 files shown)


pomegranited:
@xitij2000  FYI I rebased your PR so I can test it on stage.

instance/models/mixins/openedx_database.py
```diff
@@ -235,6 +235,10 @@ def _get_mysql_settings(self):
 
             # ecommerce
             "ECOMMERCE_DEFAULT_DB_NAME": self._get_mysql_database_name("ecommerce"),
```
pomegranited: Could you please add a comment here explaining that this variable has been replaced by `ECOMMERCE_DATABASE_NAME`, but that we're keeping this one around to maintain compatibility with Ginkgo?

instance/models/mixins/openedx_database.py
```diff
@@ -235,6 +235,10 @@ def _get_mysql_settings(self):
...
+            "ECOMMERCE_DATABASE_NAME": self._get_mysql_database_name("ecommerce"),
+            "ECOMMERCE_DATABASE_USER": self._get_mysql_user_name("ecommerce"),
+            "ECOMMERCE_DATABASE_PASSWORD": self._get_mysql_pass_from_dbname("ecommerce"),
+            "ECOMMERCE_DATABASE_HOST": self.mysql_server.hostname,
             "ECOMMERCE_DATABASES": {
```
pomegranited: Similar comment required here..  We keep this stanza for backwards compatibility with Ginkgo.

So to ease confusion, could you also update the `ECOMMERCE_DATABASES["NAME"]`, `["USER"]`, `["PASSWORD"]`, and `["HOST"]` variables below to use the new `ECOMMERCE_DATABASE_NAME`, `_USER`, ... variables you added above?

That way, we can still override individual `ECOMMERCE_DATABASE_NAME` variables in our Ocim instance configuration, and things will continue to work as expected, with both Ginkgo and post-ginkgo releases.

pomegranited: Raises a sticky question for Ocim, and its future maintainability, testing, and support when it starts being used by organisations outside of OpenCraft: what versions of the {{edx-platform}} and {{edx:configuration}} repos will we support deployment for?  And how will we ensure that these versions remain compatible with future changes?  CC @antoviaque @smarnach

instance/models/mixins/openedx_database.py
xitij2000: I've moved all old style settings together and put a comment explaining why they are there.

instance/models/mixins/openedx_database.py
xitij2000: I've also updated the old-style settings to use the new style settings allowing them to be overridden together.

smarnach: @pomegranited In the short term, I suggest we simply keep doing what we did so far:  Keep backwards compatible with the last release, and remove the old stuff when it's not needed anymore.  In the long term, we plan to make Ocim a lot more flexible.  We want to support instance types other than Open edX, and targets other than OpenStack.  At that point, I don't think we will keep the specifics of individual instance types in the core code base – they will be stored externally, e.g. in a different repository or in the database.  We will then be able to have multiple templates for different releases of Open edX, and whenever we want to support a new release, we will add a new template.

This all hasn't been planned or discussed in detail, but that's my idea of where we are going in the long term.

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -321,12 +321,24 @@ def make_nested_group_info(var_names, databases):
...
+            "ECOMMERCE_": {
+                "vars": ["DATABASES"],
+                "values": [{
+                    "default": {
+                        "ENGINE": 'django.db.backends.mysql',
+                        "NAME": "{{ ECOMMERCE_DATABASE_NAME }}",
+                        "USER": "{{ ECOMMERCE_DATABASE_USER }}",
+                        "PASSWORD": "{{ ECOMMERCE_DATABASE_PASSWORD }}",
+                        "HOST": "{{ ECOMMERCE_DATABASE_HOST }}",
+                        "PORT": expected_port,
```
xitij2000: All I wanted to test here was the port value, but I couldn't see a way around doing this other than adding a one-off test for port.

pomegranited: Oh yeah.. weird that they didn't add `ECOMMERCE_DATABASE_PORT` upstream too?

Fine to do it this way -- since it also verifies the presence of the `ECOMMERCE_DATABASES` stanza.

xitij2000: The new sandboxes based on this change seem to have worked. 
I'll just rebase / squash and update this pull request.

pomegranited:
:+1: Thanks @xitij2000 ! Ready to merge whenever you're ready.  I can help deploy to production if needed.

- [x] I tested this on stage.console by:
  * deploying Ginkgo test servers with [ecommerce disabled (default)|https://stage.console.opencraft.com/instance/1840/edx-appserver/510/] and [ecommerce enabled|https://stage.console.opencraft.com/instance/1840/edx-appserver/504/], and ensured that the service was disabled/enabled as expected.
  * deploying master test servers with [ecommerce disabled (default)|https://stage.console.opencraft.com/instance/1830/edx-appserver/508/] and [ecommerce enabled|https://stage.console.opencraft.com/instance/1830/edx-appserver/509/], and ensured that the service was disabled/enabled as expected.
- [x] I read through the code
- [x] Includes documentation -- comments explaining why duplicate settings remain in config.

PR #269, opened 2018-01-25
Update readme to add missing Ansible provision step

xitij2000:
In #260 we switched from the bootstrap script to Ansible for provisioning Vagrant. This relies on the Ansible roles being present in the ``deploy/roles`` directory, however that is not the case by default. 

This pull request updates the readme to include instructions to install the Ansible roles for deployment and provisioning Vagrant.

**Merge deadline**: None

**Reviewers**
- [ ] (TBD)


smarnach:
@xitij2000 Thanks, this looks good to me!

PR #271, opened 2018-01-31
Revert #258 use `cached_db` session engine to persist sessions

xitij2000:
The setting change might have had unforeseen consequences and is being reverted until its impact can be better analysed.


mtyaka:
Thanks @xitij2000! 👍

PR #274, opened 2018-02-03
Adds XQUEUE_SESSION_ENGINE.

pomegranited:
Uses a cache backend to store xqueue sessions, instead of the Django default of `django.contrib.sessions.backends.db`.

OpenCraft PRs of relevance:
https://github.com/open-craft/xqueue/pull/1
https://github.com/open-craft/configuration/pull/33

Proof of concept:

**Sandbox URLs**
* LMS: https://jinja-fix.stage.opencraft.hosting/
* Studio: https://studio-jinja-fix.stage.opencraft.hosting/

[Appserver 14](https://stage.console.opencraft.com/instance/1400/edx-appserver/529/)


instance/models/mixins/openedx_database.py
```diff
@@ -377,6 +377,14 @@ def _get_rabbitmq_settings(self):
             "XQUEUE_RABBITMQ_HOSTNAME": self.rabbitmq_server.instance_host,
             "XQUEUE_RABBITMQ_PORT": self.rabbitmq_server.instance_port,
             "XQUEUE_RABBITMQ_TLS": True,
+            "XQUEUE_SESSION_ENGINE": "django.contrib.sessions.backends.cache",
+            "XQUEUE_CACHES": {
+                "default": {
+                    "BACKEND": "django.core.cache.backends.memcached.MemcachedCache",
+                    "KEY_PREFIX": "xqueue",
+                    "LOCATION": "localhost:11211",
```
smarnach: Can we maybe default to `"{{ EDXAPP_MEMCACHE }}"` here, to explicitly use the same memcached as edxapp?  Not sure which version is less confusing, but I feel it could be a bit error prone to hardcode the location of memcached here.

UmanShahzad: Yep, that's fine with me. Note that upstream uses different vars for each role requiring memcache, i.e. `INSIGHTS_MEMCACHE`, `DISCOVERY_MEMCACHE`, and `CREDENTIALS_MEMCACHE`, so hopefully this is not confusing to have here.

instance/models/mixins/openedx_config.py
```diff
@@ -135,12 +135,15 @@ def _get_configuration_variables(self):
...
+            # Temporary: We want a configurable `XQUEUE_SESSION_ENGINE` ASAP but it's not in upstream Ginkgo.
+            "xqueue_source_repo": "https://github.com/open-craft/xqueue",
 
             # Pin down dependencies to specific (known to be compatible) commits.
             "edx_platform_version": self.edx_platform_commit,
             "configuration_version": self.configuration_version,
             "forum_version": self.openedx_release,
-            "xqueue_version": self.openedx_release,
+            # Temporary: We want a configurable `XQUEUE_SESSION_ENGINE` ASAP but it's not in upstream Ginkgo.
+            "xqueue_version": 'opencraft-release/ginkgo.1',
```
smarnach: @tomaszgy Can you please take a note on the Hawthorn epic to remove this after the Hawthorn upgrade?  Otherwise we will forget about this, and it might cause issues.

UmanShahzad:
@smarnach Thanks! I'm not going to rebase and merge until #277 is merged, since that seems to be higher priority. Actually, we should probably get rid of this block on the merge for when a rebase is required, because waiting 2 hours each time like this is very inefficient.

smarnach:
@UmanShahzad Yeah, having mandatory tests that take two hours to run seems a bit too much; we've actually discussed this many times, and I hope this will finally get somewhat better once we split out the builder – then at least changes that don't affect instance building won't have to run the slow tests.  (Wouldn't help for this particular change, though.)

instance/models/mixins/openedx_config.py
tomaszgy: @smarnach Sure, will do!

PR #275, opened 2018-02-04
Disables the certificates service by default on new instances

pomegranited:
Will need to be manually added to existing (i.e. unmaintained) before
reprovisioning in order to take effect.

cf https://github.com/open-craft/configuration/pull/35

**JIRA issues**: OC-3839

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -129,6 +129,8 @@ def production_instance_factory(**kwargs):
         # Don't create default users on production instances
         "DEMO_CREATE_STAFF_USER": False,
         "demo_test_users": [],
+        # Disable certificates process to reduce load on RabbitMQ and MySQL
+        "SANDBOX_ENABLE_CERTIFICATES": False,
     }
     configuration_extra_settings = kwargs.pop("configuration_extra_settings", "")
     configuration_extra_settings = yaml.load(configuration_extra_settings) if configuration_extra_settings else {}
--- a/instance/tests/test_factories.py
+++ b/instance/tests/test_factories.py
@@ -40,7 +40,8 @@ class FactoriesTestCase(TestCase):
     Test cases for functions in the factories module
     """
 
-    CONFIGURATION_EXTRA_SETTINGS = "{'demo_test_users': [], 'DEMO_CREATE_STAFF_USER': False}"
+    CONFIGURATION_EXTRA_SETTINGS = "{'demo_test_users': [], 'DEMO_CREATE_STAFF_USER': False, "
+                                   "'SANDBOX_ENABLE_CERTIFICATES': False}"
     SANDBOX_DEFAULTS = {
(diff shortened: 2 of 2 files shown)


pomegranited:
:+1:

- [x] I tested this by
  * deploying to stage.console
  * creating a new instance via the shell
     ```python
     In [1]: from instance.factories import production_instance_factory
     In [2]: ocpr275 = production_instance_factory(name="OC PR#275 disable certificates", sub_domain="ocpr275.sandbox")
     ```
  * checking the [instance](https://stage.console.opencraft.com/instance/1900/) to ensure that `SANDBOX_ENABLE_CERTIFICATES: false` is set as expected:
    ```python
    In [3]: ocpr275.configuration_extra_settings
    Out[3]: 'DEMO_CREATE_STAFF_USER: false\nSANDBOX_ENABLE_CERTIFICATES: false\ndemo_test_users: []\n'
    ```
- [x] I read through the code
- [x] Includes documentation - code comments

instance/factories.py
```diff
@@ -129,6 +129,8 @@ def production_instance_factory(**kwargs):
         # Don't create default users on production instances
         "DEMO_CREATE_STAFF_USER": False,
         "demo_test_users": [],
+        # Disable certificates process to reduce load on RabbitMQ and MySQL
+        "SANDBOX_ENABLE_CERTIFICATES": False,
```
smarnach: We should also include this for sandboxes.  Sandboxes almost never need the certs service, and if we do we can still override the setting via the PR description or via `configuration_extra_settings`.

instance/tests/test_factories.py
```diff
@@ -40,7 +40,8 @@ class FactoriesTestCase(TestCase):
     Test cases for functions in the factories module
     """
 
-    CONFIGURATION_EXTRA_SETTINGS = "{'demo_test_users': [], 'DEMO_CREATE_STAFF_USER': False}"
+    CONFIGURATION_EXTRA_SETTINGS = "{'demo_test_users': [], 'DEMO_CREATE_STAFF_USER': False, "
+    "'SANDBOX_ENABLE_CERTIFICATES': False}"
```
smarnach: This string is on a line of its own, so it doesn't seem to be doing anything.

instance/factories.py
UmanShahzad: Question: where would one make this a default for sandboxes, if they wanted to? I'm looking around but it's not obvious to me.

UmanShahzad:
Closing this, since I'm moving forward with OC-3840 and OC-3857 which together will allow us to control performance better for the certs/xqueue combo.

pomegranited:
Thank you @UmanShahzad !  Don't think we need this branch anymore, so is it ok to delete?

UmanShahzad:
@pomegranited Yep, safe to delete now!

PR #276, opened 2018-02-05
Use a configured certificate xqueue poll frequency.

UmanShahzad:
See OC-3857 for the details, and the incident reports on why.


UmanShahzad:
@clemente We have to wait on #274 before continuing here unfortunately, because an instance is being provisioned for testing and I can't restart stage. We shouldn't merge this until we test it, of course.

PR #277, opened 2018-02-07
Completely disable health checks in the haproxy config.

smarnach:
The health checks have proven problematic, since they will completely disable a backend even if we only have intermittent connection failures to MongoDB.  Since the health cheks are only useful in a multi-appserver setting we completely disable them for now, and plan to implement a fix for enabling them only for multi-appserver instances next week.


smarnach:
@tomaszgy @xitij2000 Can you please review this PR?  This needs to be tested by deploying it to Ocim stage and running an LB reconfiguration, which can be done by running this from the Django shell:

    lb = LoadBalancingServer.objects.get(accepts_new_backends=True)  # There should be only one.
    lb.reconfigure()

tomaszgy:
👍

PR #278, opened 2018-02-09
Set up a cron job to clear sessions

clemente:
This PR sets variables which will be used by the task introduced at https://github.com/open-craft/configuration/pull/38

Untested in local because my VirtualBox refuses to work. I'll fix tests here later.

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -4,6 +4,8 @@ Instance app model mixins - OpenEdX Instance Configuration
 
 # Imports #####################################################################
 
+from random import randint
+
 from django.db import models
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


instance/models/mixins/openedx_config.py
```diff
@@ -65,6 +67,12 @@ def _get_configuration_variables(self):
             # We cannot do this on OC IM for security reasons (we don't want different *instances* to share cookies).
             "EDXAPP_SESSION_COOKIE_DOMAIN": '',
 
+            # Run a command to delete expired sessions once a day. The time is random and different in each server
+            # to avoid the case when all servers connect to the database at exactly the same time
+            "EDXAPP_CLEARSESSIONS_CRON_ENABLED": True,
+            "EDXAPP_CLEARSESSIONS_CRON_HOURS": randint(0, 23),
```
smarnach: I had to double check that this method is only called once for each app server, and the resulting configuration is stored in `configuration_settings`, so it's fine that it isn't consistent across multipe calls.  Maybe add a comment expliaining this here?

clemente: Ok, I added the comment.

clemente:
Squashed and rebased. We must wait again for tests to finish.

PR #280, opened 2018-02-13
Update activity stats to use OpenStack settings

xitij2000:
The current stats script for collecting data from an AppServer is throwing errors when running against the current ``ginkgo.1``-based AppServers. This pull request updates the activity stats collection script to use openstack settings instead of AWS settings. 

**Testing instructions**:

1. Setup Ocim to this pull request. 
2. Run the [activity_csv](/instance/management/commands/activity_csv.py) command to collect usage activity of instances.
3. The command should complete successfully without errors and report the full statistics for all valid instances

I tested this script on a master instance as well, and this change fixes it for those cases as well.

**Reviewers**
- [ ] @e-kolpakov

### DIFF
--- a/playbooks/collect_activity/stats.py
+++ b/playbooks/collect_activity/stats.py
@@ -17,7 +17,7 @@ LOG_PATH = '/edx/var/log/nginx'
 
 if __name__ == '__main__':
     # Set up the environment so that edX can be initialized as the LMS with the correct settings.
-    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'lms.envs.aws')
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'lms.envs.openstack')
     os.environ.setdefault('SERVICE_VARIANT', 'lms')
     django.setup()
 


itsjeyd:
@xitij2000 This should be safe to test on OCIM stage. Could you deploy it there (or let me know if you need help with that step)? Running `activity_csv` against the set of instances deployed from stage would be more efficient than running it locally (because that would involve setting up a realistic number of instances and generating traffic on them), and give results that are easy to compare/verify.

xitij2000:
@itsjeyd Sure. Will test this change on stage.

itsjeyd:
@xitij2000 Great, let me know when you're done deploying so I can test on stage as well.

xitij2000:
@itsjeyd The activity collection has finished on stage, you can see the result in ``/var/www/activity.csv``.

itsjeyd:
@xitij2000 OK, I'll have a look.

xitij2000:
Closing this since it's no longer required, useful or valid.

PR #281, opened 2018-02-15
Enable LB health checks if more than one app server is active.

smarnach:
We have disabled the load balancer health checks in #277 since they proved to cause more harm than good for single-VM instances.  We still need them for multi-VM instances, though – otherwise the load balancer would keep sending traffic to a server that is down, so this PR re-enables them if more than one app server is active.

For testing, activate two app servers for an instance and verify that health checks in the generated config are only configured for the instance with two active app servers.  Log into the instance with two active app servers and check the stickiness cookie to learn what app server you are routed to.  Then stop the LMS on that app server, and verify that you can still successfully access the instance.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -137,6 +137,7 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
             domain=self.domain,
             http_auth_info_base64=self.http_auth_info_base64(),
             appservers=appserver_vars,
+            health_check=len(appserver_vars) > 1,
         ))
         backend_map = [(domain, backend_name) for domain in self.get_load_balanced_domains()]
         backend_conf = [(backend_name, config)]
--- a/instance/templates/instance/haproxy/openedx.conf
+++ b/instance/templates/instance/haproxy/openedx.conf
@@ -5,6 +5,6 @@
     http-request set-header Authorization 'Basic {{ http_auth_info_base64 }}' unless has-authorization
     option httpchk /heartbeat
     {% for server in appservers %}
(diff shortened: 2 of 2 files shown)


tomaszgy:
@smarnach 

:+1: (conditionally, see comments)

- [x] I tested this:
1. I followed the instructions from this PR's description, using the instance from [OC-3880](https://tasks.opencraft.com/browse/OC-3880?focusedCommentId=62395&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-62395).
2. I noticed that after stopping the LMS on the app server to which I'd been routed I could still access the instance. I needed to login again, though.

- [x] I read through the code

~~- [ ] I checked for accessibility issues~~
~~- [x] Includes documentation~~

**Comments:**
1. I see the CI build has failed. Could you please have a look?
2. I needed to log in again after I stopped the LMS (see point 2 in testing). If that is how it's supposed to be, then it's fine, if not, could you please check it? :)

Thanks!

smarnach:
@tomaszgy Thanks for your review!  Re your comments:

1. The build fails because of a bug with the Gandi API that is currently blocking all Ocim PRs.  See [this ticket](https://tasks.opencraft.com/browse/OC-3949) for details.
2. Yes, it's currently expected that you need to log in again.  Since we reverted the patch to store sessions in the database, they are now exclusively stored in the memcached running on the app server again, i.e. each app server is only seeing its own sessions.  If you are failed over to the other app server, your session is lost and you need to log in again. This problem is independent of this change.  It does not affect our paid instances (since we are still using the `cached_db` backend for them) and we are discussing possible solutions for the other instances.

tomaszgy:
@smarnach Thanks for the explanations! Please feel free to merge it :)

PR #284, opened 2018-03-04
New servers should decline new clients by default.

UmanShahzad:
We want to decline new clients for shared services by default. This prevents accidents where we create a new server on Ocim prod which isn't meant to accept new clients just yet, which ends up accepting clients on re-/new provisions of client instances.

Test instructions:

1. Make a new shared server object of some sort on Stage (i.e. MySQL/MongoDB/RabbitMQ/Load Balancer).
2. Ensure it is not accepting new clients/backends by default.


UmanShahzad:
@smarnach No need to review this right now, but I just brought a PR up in case we get to the ticket, OC-3765.

smarnach:
@UmanShahzad Thanks!  If I end up having a bit of time at some point I'll try to fit in the review.

You will probably also need to change the places where the default instances are created based on the Django settings, e.g. in `LoadBalancingServerManager._create_default()`, since these default instances need to accept new clients/backends.  (The default instance creation code is needed for the integration tests, and to make setting up the dev environment more convenient.)

UmanShahzad:
@smarnach Sure, anytime works, not urgent!

Thanks for catching that about `_create_default` -- I clearly still need to find my way around this codebase.

smarnach:
@UmanShahzad The code looks good to me, and I don't really see a need for manual testing with this PR.  The automated tests make sure that the migration works and that the default objects are still accepting clients.  The only thing I could test manually is that the defaults have actually changed, but that seems pointless.  So let's just merge it. :+1:

UmanShahzad:
@smarnach Sounds good :+1: Will squash and merge.

smarnach:
@UmanShahzad Thanks for deploying! I've just tested creating a new LB on the production server, and the new defaults work fine (as expected).

PR #285, opened 2018-03-09
Registration form text on self-hosting page - remove beta

akovari:
This PR updates the registration form.

**Screenshots**: screenshot-2018-3-9 register for beta - opencraft -

**Merge deadline**: None

**Testing instructions**:

1. Run OCIM
2. Go to /registration

**Reviewers**
- [ ] clemente


registration/forms.py
```diff
@@ -101,7 +101,6 @@ class Meta:
         widgets = {
             'instance_name': TextInput,
             'public_contact_email': EmailInput,
-            'project_description': Textarea,
```
clemente: Unless it breaks anything, it may be good to keep the widget configuration untouched (`Textarea`) and then just not show the form field.

akovari:
@antoviaque sure, will do that

registration/tests/test_views.py
```diff
@@ -532,7 +527,7 @@ def test_modifying_design_fields_sends_email(self):
         # Modifying most fields shouldn't send an e-mail
         with self.assertTemplateNotUsed('registration/fields_changed_email.txt'):
             modified.update({
-                'project_description': 'Learn',
+                'status': 'Pending',
```
clemente: This test simulates that the user manually modified some of the fields by typing (when returning to the form after having submitted it). „status“ isn't one of those visible fields, it's something internal. So better replace `project_description` by a visible field from https://github.com/open-craft/opencraft/blob/adam%2Fregistration-updates/registration/forms.py#L112, I'd suggest `full_name` because it's also text, but you can also use the checkbox `subscribe_to_updates` or maybe another.

clemente:
@akovari By the way, I just saw Gabriel suggested a phrase after „Sorry, your application has been rejected.“. If you can please add it too; see document.

antoviaque:
@akovari Thank you!

clemente:
@akovari thanks, merged. I'll continue with the deployment.

PR #287, opened 2018-03-16
Migrate config to CircleCI 2

rocioar:
Migrating to CircleCI version 2.

JIRA Ticket: OC-4053

Testing instructions:

Check that tests are passing.

Notes to reviewer:

- Had to upgrade to python version 3.5.4, since 3.5.3 is not in the list of docker images: https://github.com/CircleCI-Public/circleci-dockerfiles/tree/master/python/images.
- Using browsers image, so that it comes with firefox already installed
- Had to add hosts to mysql commands and also update links from localhost to 127.0.0.1, so that mysql connects using tcp (required when using docker images for dbs)
- Added cache for python requirements

Reviewers:

[Sven]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ install_system_dependencies: apt_get_update ## Install system-level dependencies
 	sudo -E apt-get install -y `grep -v '^#' debian_packages.lst | tr -d '\r'`
 
 create_db: ## Create blanket DBs, i.e. `opencraft`.
-	createdb --encoding utf-8 --template template0 opencraft || \
+	createdb --host 127.0.0.1 --encoding utf-8 --template template0 opencraft || \
 	    echo "Could not create database 'opencraft' - it probably already exists"
 
 .PHONY: static
@@ -100,7 +100,7 @@ test.quality: clean ## Run quality tests.
 	prospector --profile opencraft --uses django
 
 test.unit: clean static_external ## Run all unit tests.
-	honcho -e .env.test run coverage run --source='.' --omit='*/tests/*' ./manage.py test --noinput
+	honcho -e .env.test run coverage run --source='.' --omit='*/tests/*,venv/*' ./manage.py test --noinput
 	coverage html
 	@echo "\nCoverage HTML report at file://`pwd`/build/coverage/index.html\n"
 	@coverage report --fail-under $(COVERAGE_THRESHOLD) || (echo "\nERROR: Coverage is below $(COVERAGE_THRESHOLD)%\n" && exit 2)
--- a/circle.yml
+++ b/circle.yml
@@ -1,35 +1,81 @@
-machine:
-  python:
-    version: 3.5.3
-  services:
-    - redis
-  environment:
-    # Don't add sensitive information (e.g. passwords) here but in https://circleci.com/gh/open-craft/opencraft/edit#env-vars
-    DEBUG: 'true'
-    DEFAULT_FORK: 'open-craft/edx-platform'
-    INSTANCE_EPHEMERAL_DATABASES: 'false'
-    LOAD_BALANCER_FRAGMENT_NAME_PREFIX: 'integration-'
-    TEST_RUNNER: 'opencraft.tests.utils.CircleCIParallelTestRunner'
-dependencies:
-  pre:
-    - pip install --upgrade pip
-    - pip install --upgrade virtualenv
-    # Update python to latest 2.6.* release from the deadsnakes PPA.
-    # Ubuntu 14.04 which we currently use on CircleCI comes with 2.7.6 preinstalled,
-    # which does not support SNI. When CircleCI starts providing Ubuntu 16.04 images,
-    # this won't be needed anymore.
-    - sudo add-apt-repository ppa:fkrull/deadsnakes-python2.7 -y
-    - sudo apt-get update; sudo apt-get install python2.7 python2.7-dev
-  override:
-    - pip install -r requirements.txt
-database:
-  override:
-    - make create_db
-test:
-  override:
-    - bin/run-circleci-tests:
-        timeout: 2800
-        parallel: true
-  post:
-    - make test.integration_cleanup:
-        timeout: 1200
+version: 2
+jobs:
+  build:
+    docker:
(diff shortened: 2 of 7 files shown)


circle.yml
```diff
@@ -1,35 +1,56 @@
...
+      - image: circleci/python:3.5.4
+        environment:
+          DEBUG: 'true'
+          DEFAULT_FORK: 'open-craft/edx-platform'
+          INSTANCE_EPHEMERAL_DATABASES: 'false'
+          LOAD_BALANCER_FRAGMENT_NAME_PREFIX: 'integration-'
+          TEST_RUNNER: 'opencraft.tests.utils.CircleCIParallelTestRunner'
+      - image: redis
+      - image: mongo
+      - image: mysql
```
rocioar: @smarnach the PR isn't ready yet, but would like to hear your thoughts on this. So basically this is using docker for redis, mongo and mysql, is this ok? or would you prefer for them to be installed with apt?

smarnach: I don't really have a preference here – it seems easiest to use Docker containers to me.  Apparently we used the services provided by CircleCI in the past as well, so that seems fine.

smarnach:
@rocioar You can consider disabling the integration tests to get the other tests running first.  The integration tests are really slow, so by disabling them you may be able to iterate faster.  Once everything else is passing, you can re-enable the integration tests.

Makefile
```diff
@@ -46,9 +46,9 @@ help: ## Display this help message.
 all: run.dev
 
 clean: ## Remove all temporary files.
-	find -name '*.pyc' -delete
-	find -name '*~' -delete
-	find -name '__pycache__' -type d -delete
+	sudo find -name '*.pyc' -delete
+	sudo find -name '*~' -delete
+	sudo find -name '__pycache__' -type d -delete
```
smarnach: Do you know why we need `sudo` here?  I would prefer to make do without it, since it is inconvenient when used locally rather than as part of the tests.

Makefile
```diff
@@ -63,7 +63,7 @@ install_system_dependencies: apt_get_update ## Install system-level dependencies
 	sudo -E apt-get install -y `grep -v '^#' debian_packages.lst | tr -d '\r'`
 
 create_db: ## Create blanket DBs, i.e. `opencraft`.
-	createdb --encoding utf-8 --template template0 opencraft || \
+	createdb --host 127.0.0.1 --encoding utf-8 --template template0 opencraft || \
```
smarnach: This seems to be telling `createdb` to connect to the Postgres server via TCP rather than Unix domain sockets, right?  Will this still work when creating the devstack, i.e. is Postgres on Ubuntu configured to listen on localhost and allow connections without authentication?  I suspect not, but you can test it by trying to build a devstack from this branch.

circle.yml
```diff
@@ -1,35 +1,74 @@
...
+        environment:
+          MYSQL_ALLOW_EMPTY_PASSWORD: True
+      - image: circleci/postgres:10.1-alpine
+        environment:
+          PG_HOST: 127.0.0.1
+          POSTGRES_USER: circleci
+          POSTGRES_DB: circle_test
+    parallelism: 2
```
smarnach: Is this the number of parallel containers used?  We are currently using four containers, and should continue to do so to keep the build times somewhat reasonable.

circle.yml
```diff
@@ -1,35 +1,74 @@
...
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install --upgrade pip
+            pip install --upgrade virtualenv
+            sudo apt-get update; sudo apt-get install python2.7-dev
+            sudo apt-get install mysql-client
+            pip install -r requirements.txt
+            sudo apt-get install postgresql-client
+      - save_cache:
+          key: deps1-{{ .Branch }}-{{ checksum "requirements.txt" }}
```
smarnach: Why do we need to include the branch name here?  I believe caching based on the checksum of `requirements.txt` alone should be fine.

circle.yml
```diff
@@ -1,35 +1,74 @@
...
+          POSTGRES_USER: circleci
+          POSTGRES_DB: circle_test
+    parallelism: 2
+    steps:
+      - checkout
+      - restore_cache:
+          key: deps1-{{ .Branch }}-{{ checksum "requirements.txt" }}
+      - run:
+          name: Install Dockerize
+          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
```
smarnach: This is only used for waiting for the ports to be open, right?  It may be easier to use `netcat` instead, which I believe is already installed even in minimal Ubuntu images.  (I'm fine with what you have – just suggesting an alternative solution you can consider.)

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -74,7 +74,8 @@ def _assert_privileges(self, database):
             # Pass password using MYSQL_PWD environment variable rather than the --password
             # parameter so that mysql command doesn't print a security warning.
             env = {'MYSQL_PWD': password}
-            mysql_cmd = "mysql -u {user} -e 'SHOW TABLES' {db_name}".format(user=user, db_name=database_name)
+            mysql_cmd = "mysql -h 127.0.0.1 -u {user} -e 'SHOW TABLES' {db_name}".format(user=user,
```
smarnach: Same comment as above – will this work on Vagrant?

rocioar: Tested on vagrant and works.

circle.yml
rocioar: Agree 👍

Makefile
rocioar: Tested it on vagrant and works.

circle.yml
rocioar: Agree, much better 👍

rocioar:
@smarnach this is ready for review again. After approval I will rebase.

PR #290, opened 2018-03-26
Create a simple documentation for new developers

clemente:
This adds a missing chapter to Ocim's README, with a screenshot and a very fast HOWTO on how to create instances, appservers, how to SSH them and debug them.
It is made for people who will know and use Ocim only through web, this includes new developers (who didn't work on Ocim's code yet), some clients, and non-technical workers.

It addresses the list of TODOs found at https://tasks.opencraft.com/browse/OC-3115

Later this documentation will be linked to from our onboarding course, to say that at least you should be familiar with this chapter.


README.md
```diff
@@ -10,6 +10,50 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+---------------------------------------------------------
+
+This section focuses on how to use the web interface, as opposed to how to install the code,
+debug it or develop it (see next sections for that).
+
+Ocim is a web interface to manage a list of Open edX servers. With some clicks you can create
+new servers, check their status and their configuration.
+It looks like this:
+
+![Ocim's main screen](documentation/ocim_main_screen.png)
```
bradenmacdonald: Please make sure the screenshot doesn't contain any client names - this one does.

README.md
```diff
@@ -10,6 +10,50 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+
+To create a new instance, you use Django's admin and you need to fill in the domain name,
+the prefixed domain names (for Studio, e-commerce, etc.), the edx-platform/configuration branches to use,
+and extra ansible variables to pass to Open edX's playbook (if any).
+The instance settings are used for new deployments only
+(changing the instance settings doesn't retroactively redeploy appservers).
+
+How do we use Ocim at OpenCraft, as of March 2018:
+we have two Ocim installations:
+[stage](https://stage.console.opencraft.com/) and [production](https://stage.console.opencraft.com/).
```
bradenmacdonald: Please don't publicly publish the URLs of stage sites.

README.md
clemente: @bradenmacdonald ok, I edited it, the IPs are also fake now. The URL too.
I changed the document too and deleted your comment, which showed the URL in a diff.

bradenmacdonald: Thanks @clemente !

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+so you can always SSH to an appserver's IP, *even if Open edX's deployment failed*, and then debug it.
+You can use your GitHub username and key.
+
+To create a new instance, you use Django's admin and you need to fill in the domain name,
+the prefixed domain names (for Studio, e-commerce, etc.), the edx-platform/configuration branches to use,
+and extra ansible variables to pass to Open edX's playbook (if any).
+The instance settings are used for new deployments only
+(changing the instance settings doesn't retroactively redeploy appservers).
+
+How do we use Ocim at OpenCraft, as of April 2018:
```
UmanShahzad: I don't know if this belongs in the public documentation or not. It may be helpful to others in the community wanting to use Ocim, but it doesn't have to be specific about how we use it, otherwise it may give away some undesirable information and/or have to be updated in case we use it differently later (evidenced by the fact that we had to enter a date).

But I do get the point here, and it's definitely good to have -- so maybe instead just re-word it to be more generic, i.e. "Some use cases of Ocim:", going on to explain what special things one can do by adopting Ocim. Basically, "sell it" to them in this paragraph. This paragraph may also want to go up a bit higher, in the main header as its own section, by updating/replacing what's currently there.

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+Failed servers are automatically cleaned up after some days.
+An important feature is that Ocim *grants SSH access to members of our organization* (OpenCraft)
+so you can always SSH to an appserver's IP, *even if Open edX's deployment failed*, and then debug it.
+You can use your GitHub username and key.
+
+To create a new instance, you use Django's admin and you need to fill in the domain name,
+the prefixed domain names (for Studio, e-commerce, etc.), the edx-platform/configuration branches to use,
+and extra ansible variables to pass to Open edX's playbook (if any).
+The instance settings are used for new deployments only
+(changing the instance settings doesn't retroactively redeploy appservers).
```
UmanShahzad: Good explanation!

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+We can create a new appserver through the **Launch new AppServer** button;
+it will automatically get the current configuration from the instance settings
+and use it for this server.
+After 1 to 2 hours, it will finish and then you need to **activate** the new one
+and **deactivate** the old one, to make the domain name direct to the new one.
+Normally we want just 1 active appserver per instance.
+Before activating a server, there's the option to test it through a password-protected link.
+
+Sometimes Open edX playbook fails, and then you need to read the log,
+which is shown in real-time in the web.
```
UmanShahzad: *in the web **console**.*

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+You can fix the settings and then spawn another server.
+Failed servers are automatically cleaned up after some days.
```
UmanShahzad: *Failed **and old inactive servers**...**after some configurable amount of days**.*

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+An important feature is that Ocim *grants SSH access to members of our organization* (OpenCraft)
```
UmanShahzad: See comment about making certain parts generic -- we should make this generic rather than specific to OpenCraft. Maybe mention that it gives SSH access to members of a configurable GitHub organization.

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+It looks like this:
+
+![Ocim's main screen](documentation/ocim_main_screen.png)
+
+At the left there's a list of instances, and each instance has many appservers.
+We can create a new appserver through the **Launch new AppServer** button;
+it will automatically get the current configuration from the instance settings
+and use it for this server.
+After 1 to 2 hours, it will finish and then you need to **activate** the new one
+and **deactivate** the old one, to make the domain name direct to the new one.
```
UmanShahzad: Nit: we can mention the load balancer here, i.e. "to make the load balancer update its configuration so that the domain name of the instance directs to the new appserver".

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+Normally we want just 1 active appserver per instance.
```
UmanShahzad: Should mention why, either explicitly or implicitly.

Explicitly: ", because having two or more could be overkill in terms of resource requirements in some cases."
Implicitly: ", but two or more active at once may be required in some high resource-utilization cases."

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+Before activating a server, there's the option to test it through a password-protected link.
```
UmanShahzad: *through a **basic auth-protected link in the "Authenticated Link" section -- the username & password are embedded in the link**.*

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
 GitHub pull requests.
 
 
+What does OpenCraft Instance Manager do and how to use it
```
UmanShahzad: Should be "What Ocim does and how to use it", or just "Using the Ocim Web Console" if we move out the "what" part as a generic section, as suggested in an earlier comment.

(We can use the acronym 'Ocim' safely now because we mentioned it above).

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+---------------------------------------------------------
+
+This section focuses on how to use the web interface, as opposed to how to install the code,
```
UmanShahzad: *how to install, **debug, or develop Ocim; see the following sections for that**.*

README.md
```diff
@@ -10,6 +10,49 @@ intended for testing new features, and can deploy sandboxes automatically from
...
+debug it or develop it (see next sections for that).
+
+Ocim is a web interface to manage a list of Open edX servers. With some clicks you can create
```
UmanShahzad: I'd say Ocim is more of a continuous delivery tool -- the web console is just a small part of it.

README.md
clemente: I tried to write the things that I was missing when I started: Ocim's URL (I didn't realize at first that „Console“ and „Ocim“ were the same), the stage Ocim, ephemeral DBs, …
But ok, if it can't go here, then better leave it for the onboarding course, or mentoring.
The use cases are already explained in the introduction, and the note about ephemeral DBs is too specific, so better I remove all this.

README.md
clemente: Well, I wouldn't define it as CI either… it's not rather „continuous“ yet. I left the definition at the top of the README untouched, it's better; and here I'll just say that it includes an interface.

clemente:
@UmanShahzad I applied your changes, thanks. Please review.
You can merge (by squashing) if ready.

clemente:
Tests have been failing with `instance.models.load_balancer.ReconfigurationFailed`. E.g. `instance.tests.integration.integration_instance.InstanceIntegrationTestCase`. I'll keep re-running them. I also rebased. When it's ready I can squash. the new commits.

README.md
```diff
@@ -10,6 +10,46 @@ intended for testing new features, and can deploy sandboxes automatically from
 GitHub pull requests.
 
 
+Using the Ocim web console
```
UmanShahzad: I have just one last comment about this: should this whole section be moved down _after_ the install part? Whenever I look at a repo for some software, I care about 2 things right away: what is it, and how do I install it?

The "what is it" part is answered above this section, and the "how do I install it" part is below. This part seems to get in the way and throws too many intricacies about Ocim to the reader before they even know how to get started.

clemente: I want the document clearly divided in two parts: the user part and the developer part. It could be a different page if you want.
I put the user part first because everyone needs it in the beginning, and because too often I have been lost in OpenCraft's documentation because it has too many sections.
If a random developer browsing GitHub doesn't know Ocim yet, he still has to know what it does and if he likes it or not, before installing it. This „user“ part should be the one that sells it. We could say that the screenshot is enough for that and we could move the screenshot to before this header so that the usage section is smaller.
Or we can add some short links at the top to directly go to the „Install“ header (or to all key sections: install, settings, run, debug, tests).

UmanShahzad: @clemente Yeah I get your point.

I had the same alternative idea: the image is a good "seller" here, so we can just move that up. I'd be good with that approach. A table of contents would also be really helpful, if you have time for it. We can delay implementing a ToC part until later in the week if you don't have time right now, it's not as urgent.

clemente: Ok, I moved the image and added a very simple TOC.

UmanShahzad:
@clemente Thanks! I've "completed" the table of contents to include all h2 sections and pushed that commit here. As soon as tests pass we can merge.

PR #292, opened 2018-03-29
Safer load balancer reconfigurations

afzaledx:
Added a called to update_status() if an active server was found without IP address. If the IP is still missing after the call, allow the load balancer configuration to fail.

**JIRA tickets**: OC-4205

**Testing instructions**:

1. Spawn an appserver. but before making it active, terminate the OpenStack VM server. If we then try to make the appserver active. It should not be allowed to be made active. The VM server should be in the Ready state before making the appserver active.

2. When reconfiguring the loadbalancer, the VM server should have a public IP address, otherwise it will throw an exception. 

**Reviewers**
- [ ] @UmanShahzad

### DIFF
--- a/instance/api/openedx_appserver.py
+++ b/instance/api/openedx_appserver.py
@@ -96,11 +96,11 @@ class OpenEdXAppServerViewSet(viewsets.ReadOnlyModelViewSet):
         Add this AppServer to the list of active app server for the instance.
         """
         app_server = self.get_object()
-        if not app_server.status.is_healthy_state:
+        if not app_server.status.is_healthy_state or not make_appserver_active(app_server.pk):
             return Response(
                 {"error": "Cannot make an unhealthy app server active."}, status=status.HTTP_400_BAD_REQUEST
             )
-        make_appserver_active(app_server.pk)
+
         return Response({'status': 'App server activation initiated.'})
 
     @detail_route(methods=['post'])
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -40,6 +40,7 @@ from instance.models.mixins.openedx_storage import OpenEdXStorageMixin
 from instance.models.mixins.openedx_theme import OpenEdXThemeMixin
 from instance.models.mixins.secret_keys import SecretKeyInstanceMixin
 from instance.models.openedx_appserver import OpenEdXAppConfiguration
+from instance.models.utils import WrongStateException
 from instance.utils import sufficient_time_passed
 
 
@@ -116,13 +117,17 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
         appserver_vars = []
         for appserver in active_appservers:
             server_name = "appserver-{}".format(appserver.pk)
-            ip_address = appserver.server.public_ip
-            if ip_address:
-                appserver_vars.append(dict(ip_address=ip_address, name=server_name))
-            else:
-                appserver.logger.error(
-                    "Active appserver does not have a public IP address. This should not happen."
+            if not appserver.server.public_ip:
+                self.logger.error(
+                    "Active appserver does not have a public IP address. This should not happen. "
+                    "Updating internal status to double check."
                 )
+                appserver.server.update_status()
+                if not appserver.server.public_ip:
+                    raise WrongStateException("Public IP still not available for active appserver. "
+                                              "Canceling reconfiguration process.")
+
+            appserver_vars.append(dict(ip_address=appserver.server.public_ip, name=server_name))
 
         if len(appserver_vars) == 0:
             self.logger.error(
@@ -182,7 +187,7 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
 
         # We unconditionally set the DNS records here, though this would only be strictly needed
         # when the first AppServer is spawned.  However, there is no easy way to tell whether the
-        # DNS records have already been successfully set, and it doesn't hurt to alway do it.
+        # DNS records have already been successfully set, and it doesn't hurt to always do it.
         self.set_dns_records()
 
         # Provision external databases:
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -70,6 +70,10 @@ class ServerState(ResourceState):
     # until the target server has reached one of these statuses.
     vm_available = False
 
+    # We can only reconfigure the load balancer in case the server is in Status.Ready.
+    # This field is to be set just in the Ready state and False in other states.
+    is_healthy_state = False
+
 
 class Status(ResourceState.Enum):
     """
@@ -94,6 +98,7 @@ class Status(ResourceState.Enum):
         is_steady_state = True
         accepts_ssh_commands = True
         vm_available = True
+        is_healthy_state = True
 
     class Terminated(ServerState):
         """ Stopped forever """
--- a/instance/serializers/server.py
+++ b/instance/serializers/server.py
@@ -50,8 +50,6 @@ class OpenStackServerSerializer(serializers.ModelSerializer):
         )
 
     def to_representation(self, obj):
-        # Make sure we have an up-to-date representation of the server
-        obj.update_status()
         output = super().to_representation(obj)
         # Convert the state values from objects to strings:
         output['status'] = obj.status.state_id
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -73,14 +73,12 @@ def spawn_appserver(
                 instance.tags.remove(failure_tag)
             if success_tag:
                 instance.tags.add(success_tag)
-            if mark_active_on_success:
-                # If the AppServer provisioned successfully, make it active.
-                make_appserver_active(
+            if mark_active_on_success and make_appserver_active(
                     appserver_id,
                     active=mark_active_on_success,
                     deactivate_others=deactivate_old_appservers
-                )
-            break
+            ):
+                break
     else:
         if failure_tag:
             instance.tags.add(failure_tag)
@@ -92,8 +90,19 @@ def spawn_appserver(
 def make_appserver_active(appserver_id, active=True, deactivate_others=False):
     """
     Mark an AppServer as active or inactive.
+
+    :returns
+        True if the appserver was made active,
+        False if the corresponding server was not healthy and thus the appserver was not made active.
     """
     appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
+    if not appserver.server.status.is_healthy_state:
+        appserver.server.update_status()
+        if not appserver.server.status.is_healthy_state:
+            logger.info('Not %s %s: ID=%s since the server is not Ready.', "Activating" if active else "Deactivating",
+                        appserver, appserver_id)
+            return False
+
     logger.info('%s %s: ID=%s', "Activating" if active else "Deactivating", appserver, appserver_id)
     appserver.make_active(active=active)
     if active and deactivate_others:
@@ -101,6 +110,8 @@ def make_appserver_active(appserver_id, active=True, deactivate_others=False):
         for appserver_to_deactivate in other_appservers:
             logger.info('Deactivating %s [%s]', appserver_to_deactivate, appserver_to_deactivate.id)
             appserver_to_deactivate.make_active(active=False)
+
+    return True
 
 
 @db_task()
--- a/instance/tests/models/factories/openedx_appserver.py
+++ b/instance/tests/models/factories/openedx_appserver.py
@@ -28,7 +28,7 @@ from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFact
 # Functions ###################################################################
 
 
-def make_test_appserver(instance=None):
+def make_test_appserver(instance=None, server=None):
     """
     Factory method to create an OpenEdXAppServer (and OpenStackServer).
     """
@@ -37,4 +37,10 @@ def make_test_appserver(instance=None):
     if not instance.load_balancing_server:
         instance.load_balancing_server = LoadBalancingServer.objects.select_random()
         instance.save()
-    return instance._create_owned_appserver()
+    appserver = instance._create_owned_appserver()
+
+    if server:
+        appserver.server = server
+        appserver.save()
+
+    return appserver
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -24,7 +24,7 @@ OpenEdXInstance model - Tests
 
 from datetime import timedelta
 import re
-from unittest.mock import patch, Mock
+from unittest.mock import patch, Mock, PropertyMock
 
 import ddt
 from django.conf import settings
@@ -41,6 +41,7 @@ from instance.models.load_balancer import LoadBalancingServer
 from instance.models.openedx_appserver import OpenEdXAppServer
 from instance.models.openedx_instance import OpenEdXInstance, OpenEdXAppConfiguration
 from instance.models.server import OpenStackServer, Server, Status as ServerStatus
+from instance.models.utils import WrongStateException
 from instance.tests.base import TestCase
 from instance.tests.models.factories.openedx_appserver import make_test_appserver
 from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
@@ -433,7 +434,6 @@ class OpenEdXInstanceTestCase(TestCase):
             "ecommerce-test.load_balancer.example.com",
             "discovery-test.load_balancer.example.com",
         ]
-
         # Test configuration for preliminary page
         backend_map, config = instance.get_load_balancer_configuration()
         self._check_load_balancer_configuration(
@@ -444,20 +444,23 @@ class OpenEdXInstanceTestCase(TestCase):
         appserver_id = instance.spawn_appserver()
         appserver = instance.appserver_set.get(pk=appserver_id)
         appserver.make_active()
-        backend_map, config = instance.get_load_balancer_configuration()
+        with patch('instance.models.server.OpenStackServer.public_ip', new_callable=PropertyMock) as mock_public_ip:
+            mock_public_ip.side_effect = [None, "1.1.1.1", "1.1.1.1", "1.1.1.1"]
+            backend_map, config = instance.get_load_balancer_configuration()
         self._check_load_balancer_configuration(
             backend_map, config, domain_names, appserver.server.public_ip,
         )
 
         # Test configuration in case an active appserver doesn't have a public IP address anymore.
         # This might happen if the OpenStack server dies or gets modified from the outside, but it
-        # is not expected to happen under normal circumstances.  We deconfigure the backend and log
-        # an error in this case.
+        # is not expected to happen under normal circumstances.  In case the public IP address is not there,
+        # We log an Error, and call update_status(). In case the IP address is still not there, we stop further
+        # activity and raise an Exception.
         with patch('instance.openstack_utils.get_server_public_address', return_value=None), \
                 self.assertLogs("instance.models.instance", "ERROR"):
             appserver.server._public_ip = None
             appserver.server.save()
-            self.assertEqual(instance.get_load_balancer_configuration(), ([], []))
+            self.assertRaises(WrongStateException, instance.get_load_balancer_configuration)
 
     def test_get_load_balancer_config_ext_domains(self):
         """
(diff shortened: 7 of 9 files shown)


instance/models/openedx_instance.py
```diff
@@ -117,6 +117,10 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
         for appserver in active_appservers:
             server_name = "appserver-{}".format(appserver.pk)
             ip_address = appserver.server.public_ip
+            if not ip_address:
+                # if the ip_address is not there, try to update_status() before failing.
```
UmanShahzad: This comment is a code smell: there is little that we gain from it which the code doesn't already make clear right after.

Also, by *fail*, we actually literally mean to make this function fail, i.e. cancel the load balancer reconfiguration process (This probably sounds confusing, so see https://github.com/open-craft/ansible-load-balancer#architecture-notes if it may be helpful), which probably entails throwing an exception or something to prevent Ocim from actually reconfiguring the load balancer in this state.

UmanShahzad: (BTW, for the comment itself, we can probably still have a comment but just rewrite it to explain in some more detail the motive behind this -- up to you).

instance/models/openedx_instance.py
```diff
@@ -117,12 +117,19 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+                appserver.server.update_status()
+                ip_address = appserver.server.public_ip
             if ip_address:
                 appserver_vars.append(dict(ip_address=ip_address, name=server_name))
             else:
                 appserver.logger.error(
                     "Active appserver does not have a public IP address. This should not happen."
+                    "Deconfiguring the load balancer backend."
                 )
+                return [], []
```
afzaledx: @UmanShahzad , this is pretty similar to what's happening on line 139. Raising an Exception would mess with the entire call stack.

instance/models/openedx_instance.py
```diff
@@ -117,12 +117,21 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
             if ip_address:
                 appserver_vars.append(dict(ip_address=ip_address, name=server_name))
             else:
-                appserver.logger.error(
-                    "Active appserver does not have a public IP address. This should not happen."
+                self.logger.error(
+                    "Active appserver still does not have a public IP address after calling update_status()."
+                    "Deconfiguring the load balancer backend."
                 )
+                return [], []
```
UmanShahzad: This does nothing different than what it already would've, actually, for the case we're trying to prevent. Notice that `appserver_vars` will be empty if we don't return here if there is an issue retrieving the public IP, and it'll just go down to line 136, and do the same thing.

As mentioned earlier, we want to kill the call stack, because this case "should not happen" and if it does, it's too dangerous to proceed and potentially accidentally deconfigure an active and healthy backend.

instance/models/openedx_instance.py
```diff
@@ -117,12 +117,21 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
         for appserver in active_appservers:
             server_name = "appserver-{}".format(appserver.pk)
             ip_address = appserver.server.public_ip
+            if not ip_address:
+                appserver.logger.error(
+                    "Active appserver does not have a public IP address. This should not happen."
+                    "Calling update_status()"
+                )
+                appserver.server.update_status()
+                ip_address = appserver.server.public_ip
```
UmanShahzad: :+1: Looks good.

If we introduce raising an exception instead of returning `[], []` in the case that the IP is still not available, that exception should probably go right here in this conditional block. Then later below we can erase the extra `if ip_address` check.

instance/models/openedx_instance.py
```diff
@@ -117,12 +117,21 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
         for appserver in active_appservers:
             server_name = "appserver-{}".format(appserver.pk)
             ip_address = appserver.server.public_ip
+            if not ip_address:
+                appserver.logger.error(
+                    "Active appserver does not have a public IP address. This should not happen."
+                    "Calling update_status()"
```
UmanShahzad: We should probably not mention the function name explicitly, in case it has to change; then it'll have to change here, too, and additionally it's not as descriptive.

Better to just describe the process, i.e. "Updating internal status to double check.", or something along those lines.

instance/serializers/server.py
```diff
@@ -50,8 +50,6 @@ class Meta:
         )
 
     def to_representation(self, obj):
-        # Make sure we have an up-to-date representation of the server
-        obj.update_status()
```
UmanShahzad: :+1: looks good.

instance/tasks.py
```diff
@@ -75,12 +76,15 @@ def spawn_appserver(
...
                 # If the AppServer provisioned successfully, make it active.
-                make_appserver_active(
-                    appserver_id,
-                    active=mark_active_on_success,
-                    deactivate_others=deactivate_old_appservers
-                )
-            break
+                appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
+
+                if appserver.server.status == OpenStackServer.Status.Ready:
```
UmanShahzad: This is the wrong level at which to check healthiness. It is *one* place, but we want to cover all possible places where activating an appserver takes place, and this is only one of them.

So we need to go to a lower level, in particular in the `make_appserver_active` function itself. It would be good to also log a note there for when the OpenStack server status is not correct for whatever reason.

UmanShahzad: Also, I think we should import `Server` rather than `OpenStackServer` to get `.Status.Ready`, since it lets it be more generic for the future (i.e. an `AwsServer` would also be covered).

instance/api/openedx_appserver.py
```diff
@@ -100,7 +100,12 @@ def make_active(self, request, pk):
             return Response(
                 {"error": "Cannot make an unhealthy app server active."}, status=status.HTTP_400_BAD_REQUEST
             )
-        make_appserver_active(app_server.pk)
+        successfully_activated = make_appserver_active(app_server.pk)
```
UmanShahzad: Similar to my comment earlier, we can move this into the previous `if` check to decrease repeat code, i.e.

```python
if not app_server.status.is_healthy_state or not make_appserver_active(app_server.pk):
         return Response(
             {"error": "Cannot make an unhealthy app server active."}, status=status.HTTP_400_BAD_REQUEST
         )
```

This'll make it active if the appserver is healthy and move on if successful, otherwise fail with the same error.

instance/models/openedx_instance.py
```diff
@@ -117,12 +118,17 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
-                appserver_vars.append(dict(ip_address=ip_address, name=server_name))
-            else:
+            if not ip_address:
                 appserver.logger.error(
                     "Active appserver does not have a public IP address. This should not happen."
+                    "Updating internal status to double check."
                 )
+                appserver.server.update_status()
+                ip_address = appserver.server.public_ip
+                if not ip_address:
```
UmanShahzad: Nit: since we're not reusing `ip_address` now, can we just do `if not appserver.server.public_ip`?

instance/models/openedx_instance.py
```diff
@@ -117,12 +118,17 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+                    raise WrongStateException("Active appserver does not have a public IP address.")
```
UmanShahzad: Nit: can we improve this a bit to tell a story? :D I.e., "Public IP still not available for active appserver. Canceling reconfiguration process."

instance/models/openedx_instance.py
```diff
@@ -117,12 +118,17 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
         for appserver in active_appservers:
             server_name = "appserver-{}".format(appserver.pk)
             ip_address = appserver.server.public_ip
-            if ip_address:
-                appserver_vars.append(dict(ip_address=ip_address, name=server_name))
-            else:
+            if not ip_address:
                 appserver.logger.error(
                     "Active appserver does not have a public IP address. This should not happen."
+                    "Updating internal status to double check."
```
UmanShahzad: We either need a space at the end of the previous line, or in front of this one. Otherwise they'll be concatenated awkwardly.

instance/tasks.py
```diff
@@ -93,15 +95,22 @@ def make_appserver_active(appserver_id, active=True, deactivate_others=False):
...
+    if appserver.server.status == Server.Status.Ready:
+        appserver.make_active(active=active)
+        success = True
+        if active and deactivate_others:
+            other_appservers = appserver.instance.appserver_set.filter(_is_active=True).exclude(pk=appserver_id)
+            for appserver_to_deactivate in other_appservers:
+                logger.info('Deactivating %s [%s]', appserver_to_deactivate, appserver_to_deactivate.id)
+                appserver_to_deactivate.make_active(active=False)
+    else:
+        success = False
```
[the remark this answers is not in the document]
afzaledx: @UmanShahzad, what if `active` was passed in as `False`, i.e. when deactivating the app server? It does not seem appropriate to not deactivate when the appserver.server is not ready.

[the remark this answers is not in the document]
UmanShahzad: With the code as you currently have it, this will still happen -- deactivating the appserver only happens under the condition of the server status being ready.

The only case I can think of right now where the VM is not 'Ready' but is still active is in the 'Unknown' state, and in that state we don't want to deactivate anyway right, because it's dangerous?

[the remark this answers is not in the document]
afzaledx: @UmanShahzad , I have changed the logic around a bit. can you take a look at it again?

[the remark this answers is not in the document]
UmanShahzad: @afzaledx I took a look, and it will do as you said, but is it what we want? See my last paragraph above.

> The only case I can think of right now where the VM is not 'Ready' but is still active is in the 'Unknown' state, and in that state we don't want to deactivate anyway right, because it's dangerous?

instance/tasks.py
```diff
@@ -92,16 +91,28 @@ def spawn_appserver(
...
     appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
-    logger.info('%s %s: ID=%s', "Activating" if active else "Deactivating", appserver, appserver_id)
-    appserver.make_active(active=active)
-    if active and deactivate_others:
-        other_appservers = appserver.instance.appserver_set.filter(_is_active=True).exclude(pk=appserver_id)
-        for appserver_to_deactivate in other_appservers:
-            logger.info('Deactivating %s [%s]', appserver_to_deactivate, appserver_to_deactivate.id)
-            appserver_to_deactivate.make_active(active=False)
+    vmserver_ready = appserver.server.status == Server.Status.Ready
+
```
afzaledx: @UmanShahzad , this is the new code. you are looking at this ? I am a bit confused. We do want to deactivate without checking the vm server, right?

instance/tasks.py
```diff
@@ -92,16 +91,28 @@ def spawn_appserver(
 def make_appserver_active(appserver_id, active=True, deactivate_others=False):
     """
     Mark an AppServer as active or inactive.
+
+    :returns True if successful, False otherwise.
```
UmanShahzad: I would clarify this comment a bit more to be explicit about what successful really means in this function's context.

instance/tasks.py
```diff
@@ -92,16 +91,28 @@ def spawn_appserver(
...
     """
     appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
-    logger.info('%s %s: ID=%s', "Activating" if active else "Deactivating", appserver, appserver_id)
-    appserver.make_active(active=active)
-    if active and deactivate_others:
-        other_appservers = appserver.instance.appserver_set.filter(_is_active=True).exclude(pk=appserver_id)
-        for appserver_to_deactivate in other_appservers:
-            logger.info('Deactivating %s [%s]', appserver_to_deactivate, appserver_to_deactivate.id)
-            appserver_to_deactivate.make_active(active=False)
+    vmserver_ready = appserver.server.status == Server.Status.Ready
```
UmanShahzad: A nit that came up: maybe we should move this check to be a property of `appserver.server`? I.e. `appserver.server.is_healthy_state` or something.

instance/tasks.py
```diff
@@ -92,16 +91,28 @@ def spawn_appserver(
...
+        if vmserver_ready:
+            logger.info('Activating %s: ID=%s', appserver, appserver_id)
+            appserver.make_active(active=True)
+            if deactivate_others:
+                other_appservers = appserver.instance.appserver_set.filter(_is_active=True).exclude(pk=appserver_id)
+                for appserver_to_deactivate in other_appservers:
+                    logger.info('Deactivating %s [%s]', appserver_to_deactivate, appserver_to_deactivate.id)
+                    appserver_to_deactivate.make_active(active=False)
+        else:
+            logger.info('Did not Activate %s: ID=%s. The server was not Ready.', appserver, appserver_id)
```
UmanShahzad: @afzaledx Btw, one more note for when you change this, I _think_ we should do something similar here as above where since the VM is unhealthy, we re-check the status with appserver.server.update_status, and check again if it is healthy, in case the status was 'Unknown' and we need to just refresh it with another call to the Nova API.

instance/tasks.py
```diff
@@ -75,12 +76,13 @@ def spawn_appserver(
                 instance.tags.add(success_tag)
```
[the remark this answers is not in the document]
afzaledx: @UmanShahzad , Still trying to figure out how to do that in a proper way, since tags is a 1 to many field, not a simple list.

instance/models/server.py
```diff
@@ -70,6 +70,10 @@ class ServerState(ResourceState):
     # until the target server has reached one of these statuses.
     vm_available = False
 
+    # We want to reconfigure the load balancer in case the server is in Status.Ready.
+    # This field be set just in the Ready state and False in other states.
```
UmanShahzad: "This field *is to* be set...".

instance/models/server.py
```diff
@@ -70,6 +70,10 @@ class ServerState(ResourceState):
     # until the target server has reached one of these statuses.
     vm_available = False
 
+    # We want to reconfigure the load balancer in case the server is in Status.Ready.
```
UmanShahzad: "We *can only* reconfigure the load balancer...".

instance/tests/test_tasks.py
```diff
@@ -146,13 +172,29 @@ def test_make_appserver_active(self, active, mock_make_active):
...
+                mock_make_active.assert_called_once_with(active=True)
+
     @patch('instance.models.openedx_appserver.OpenEdXAppServer.make_active')
     def test_deactivate_others(self, mock_make_active):
         """
         When activating the appserver, optionally deactivate others.
         """
         for dummy in range(5):
             appserver = make_test_appserver(self.appserver.instance)
+            appserver.server = ReadyOpenStackServerFactory()
```
UmanShahzad: I noticed we are adding this in **lot** of different places. This is bad: it's code repetition everywhere. Please move this to a common place, like `make_test_appserver`, so it's not done like this. You can add it as a keyword arg that'll default to the ready OpenStack server, or whichever way you think is best.

instance/tests/models/test_openedx_instance.py
```diff
@@ -457,7 +459,7 @@ def test_get_load_balancer_configuration(self, mocks):
                 self.assertLogs("instance.models.instance", "ERROR"):
             appserver.server._public_ip = None
             appserver.server.save()
-            self.assertEqual(instance.get_load_balancer_configuration(), ([], []))
+            self.assertRaises(WrongStateException, instance.get_load_balancer_configuration)
```
UmanShahzad: Please change the comment above this `with` block to reflect the new reality of what happens.

instance/tests/models/test_openedx_instance.py
```diff
@@ -733,9 +735,10 @@ def test_archive_no_active_appserver(self, mock_reconfigure, mock_disable_monito
         ])
         self.assertTrue(instance.ref.is_archived)
 
+    @patch('instance.models.server.OpenStackServer.public_ip')
```
UmanShahzad: I don't understand immediately -- why is this needed now?

instance/tests/test_tasks.py
```diff
@@ -72,29 +73,53 @@ def test_mark_active_on_success(self, provisioning_succeeds):
         Test that we when mark_active_on_success=True, the spawn_appserver task will mark the
```
UmanShahzad: I know you didn't comment this, but since we're here can you remove that extraneous 'we'? Thanks :P

instance/tests/test_tasks.py
```diff
@@ -146,13 +172,29 @@ def test_make_appserver_active(self, active, mock_make_active):
         tasks.make_appserver_active(self.appserver.id, active=active)
         mock_make_active.assert_called_once_with(active=active)
 
+    @ddt.data([False, True], [False, False])
+    @patch('instance.models.server.OpenStackServer.update_status')
+    @patch('instance.models.openedx_appserver.OpenEdXAppServer.make_active')
+    def test_make_appserver_active_server_unhealthy(self, health, mock_make_active, mock_update_status):
```
UmanShahzad: Can we call this `test_make_appserver_active_but_server_unhealthy`?

instance/tests/test_tasks.py
```diff
@@ -146,13 +172,29 @@ def test_make_appserver_active(self, active, mock_make_active):
...
+        """
+        By default, we activate the appserver.
+        """
+        with patch('instance.models.server.OpenStackServer.Status.Ready.is_healthy_state',
+                   new_callable=PropertyMock) as mock_is_healthy_state:
+            mock_is_healthy_state.side_effect = health
+            tasks.make_appserver_active(self.appserver.id, active=True)
+            mock_update_status.assert_called_once_with()
+            if not all(not data for data in health):
```
UmanShahzad: This is hard to read. It's a `not` operation applied on a group of `AND`'d `not`s, so it expands to be a group of just `or`s. So this can be simplified to `if any(data for data in health)`. (I'd rename `data` to `healthy`, btw).

(For an example, see this:)
```
~(~a and ~b and ~c)
= a or b or c
```

And in the load-balancer reconfiguration context, if we see that the server is healthy _any_ time, it's good to be made active.

instance/tests/models/test_openedx_instance.py
afzaledx: the call to public_ip is an external call. On line 772, there's a call to
        # Set single app server active - this will be the reference date
        with freeze_time(reference_date):
            active_appserver.make_active()

so that will check the public_ip. Need to patch the public_ip call to avoid failures due to the external API call. Basically whereever the make_active() funciton is called, we need to patch the public_ip because we have added it as a precondition before making the appserver active.

instance/tests/test_tasks.py
afzaledx: Thanks for that. Really helpful.

instance/tests/models/test_openedx_instance.py
afzaledx: Well, I removed the mock and it passed anyway, so removing it now.

instance/tests/models/factories/openedx_appserver.py
```diff
@@ -37,4 +37,10 @@ def make_test_appserver(instance=None):
     if not instance.load_balancing_server:
         instance.load_balancing_server = LoadBalancingServer.objects.select_random()
         instance.save()
-    return instance._create_owned_appserver()
+    appserver = instance._create_owned_appserver()
+
+    if server:
```
UmanShahzad: This doesn't fix the problem of repeat code I mentioned -- we're still just repeating the creation of `ReadyOpenStackServerFactory` everywhere.

What we should do instead is, `appserver.server = ReadyOpenStackServerFactory() if server is None else server`.

instance/tests/models/test_openedx_instance.py
UmanShahzad: :+1:

instance/tests/test_tasks.py
```diff
@@ -146,6 +170,22 @@ def test_make_appserver_active(self, active, mock_make_active):
         tasks.make_appserver_active(self.appserver.id, active=active)
         mock_make_active.assert_called_once_with(active=active)
 
+    @ddt.data([False, True], [False, False])
+    @patch('instance.models.server.OpenStackServer.update_status')
+    @patch('instance.models.openedx_appserver.OpenEdXAppServer.make_active')
+    # pylint: disable=invalid-name
```
UmanShahzad: Ah snap, I didn't know this would introduce a pylint violation :) In that case, sorry for my old comment -- it's probably better to leave this as you originally had it.

instance/tests/models/factories/openedx_appserver.py
afzaledx: Well, I changed the default to Ready state, and 20 tests started failing, because the default server being created is in the Pending state, and that's why we have to overwrite it with the Ready state server. I think it's suitable to use it they what I have it now because it's very similar to how `instance` is being used.

UmanShahzad: Ah, OK, I did not know about that default. Sorry, my recommendation was not very well informed here. But the addition to the make_test_appserver function is still pretty nice, thanks!

instance/tests/models/test_openedx_instance.py
afzaledx: actually it didn't pass. had to add it back. Sorry about the confusion.

UmanShahzad:
@afzaledx Can you squash commits here and get these tests to pass? Then we should be good for merge!

PR #293, opened 2018-03-29
Enhance admin BetaTestApplication and OpenEdXInstance admin pages

xitij2000:
This PR makes some basic improvements to the admin pages for Ocim. 

It adds the instance name, email, status and days since activation for beta registrations to make it easy to get all that information in one place. It also adds support for filtering by beta registrations by status and searching registrations by name, email and domain

The admin page for OpenEdxAdmin now also displays whether the instance is successfully provisioned, and allows searching by internal_lms_domain.

**Reviewers**
- [ ] @clemente


clemente:
@xitij2000 Thanks, the search features will help.

Some missing steps:
- pylint says `Too many public methods`, and just because of this we need to refactor… It's true that the are many methods, and many are internal helpers (the ones starting with underscore). One idea: you could move all those to a class (it can even be a TestCase without tests) and make other classes that inherit from it; these new classe can be themed too, e.g. one about DNS, another about archiving, etc.
- the `InstanceReference` stacked-inline admin is also visible when *adding* a new instance from admin, and if I fill the instance data (top form [here](http://localhost:5000/admin/instance/openedxinstance/add/)) *and write a name for my instance* (bottom form) I get a `django.db.utils.IntegrityError: django.db.utils.IntegrityError: duplicate key value violates unique constraint "instance_instancereference_instance_type_id_f6d25849_uniq". DETAIL:  Key (instance_type_id, instance_id)=(3, 16) already exists.`. Fails at [save](https://github.com/open-craft/opencraft/blob/kshitij%2Fupdate_admin_pages/instance/models/instance.py#L93). It works well if I don't change the name (then the name becomes `Instance`). It shouldn't fail.
- and by the way, the „Is archived“ checkbox, which is now in a stacked inline admin, doesn't really archive an instance; it's just an indicator of whether the instance was archived. Changing it from shell and saving is intantaneous, and doesn't change remote servers. This isn't a change in behaviour, because it does the same as it did before, so it's not a problem. Only the commit message is wrong, but if you squash it won't be seen. You may add some comment if you want to clarify is_archived.

Tested and works:
- at first I was concerned with the speed, but these are not integration tests, so they should run fast
- I verified that instance search works
- Betatest search works
- „first activated“ works: I faked some activation data of an betatest instance's appservers, and the date was shown correctly

clemente:
:+1:

  - [X] I tested this: create new server and edit the name (through new inline form). Search betatest. Search instances. First_activated tested by creating fake servers.
  - [X] I read through the code
  - [X] I checked for accessibility issues: none
  - [X] Includes documentation: help_text expanded

Can you squash and proceed with deploy? We'll need to make CircleCi pass first, I don't know what's happening lately. Retrying usually fixes it.

clemente:
:+1: about the new tests. I traced them and checked that they verify the important scenarios (only APIs, only pending OpenStack servers).

This seems to need a rebase.
Squash not needed because it's two different features, but as you wish.

UmanShahzad:
@xitij2000 Yay, it passed!

xitij2000:
@UmanShahzad  :-) finally

PR #294, opened 2018-04-06
WIP: Updates the docs for how to set up ecommerce and discovery

pomegranited:
Setting up ecommerce/discovery sandbox for a client, and so this PR fixes some formatting issues and broken links in the existing documentation, and updates it to provide some more information.

However, even though this was just a README change, the CircleCI tests weren't passing.  So subsequent commits were dedicated to fixing the incompatibilities between the latest firefox and selenium, and adding a geckodriver dependency.

**Reviewer**
- [ ] @clemente

### DIFF
--- a/README.md
+++ b/README.md
@@ -882,39 +882,66 @@ or if individual instances receive a large amount of traffic.
 ### Configuring ecommerce and Course Discovery
 
 By default, instances are not provisioned with either ecommerce or the Course Discovery
-service. However, support is available to manually enable those services. To do so, the
-following additional variables must be est on the instance:
+service. However, support is available to manually enable those services. To do so:
 
-```yaml
-DISCOVERY_ELASTICSEARCH_URL: "http://149.202.166.156:9200/"
-SANDBOX_ENABLE_DISCOVERY: yes
-SANDBOX_ENABLE_ECOMMERCE: yes
-DISCOVERY_VERSION: "c96c77bbf80650b3d3cc238335cc5205e24b775f"
-```
+* Ensure that the [`refresh_course_metadata` cron task](https://github.com/open-craft/configuration/blob/c1e576eabefea82d21d7785810126e39752cd14e/playbooks/roles/discovery/tasks/main.yml)
+  is added to the ansible discovery tasks.
+* Set these extra variables on the instance:
+  ```yaml
+  SANDBOX_ENABLE_DISCOVERY: yes
+  SANDBOX_ENABLE_ECOMMERCE: yes
+  DISCOVERY_VERSION: 'open-release/ginkgo.1'
+  nginx_discovery_gunicorn_hosts:
+      - "127.0.0.1"
+  COMMON_HOSTNAME: "your-site-name"
+  ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
+      your-partner-code:
+        paypal:
+          ...
+  ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -886,35 +886,59 @@ service. However, support is available to manually enable those services. To do
...
 ```
 
-Currently, we don't support Elasticsearch as a database backend, so we have to pass
-that in manually. Additionally, discovery and ecommerce still aren't first-class
-citizens, so they must be enabled via extra var. We need to set the hostname, other
-than the FQDN, that the appserver refers to itself as so that we can properly go out
-to the load balancer-terminated SSL connection. And finally, newer versions of course
-discovery don't work properly with OIDC login yet.
+We need to set the hostname to something other than the FQDN, that the appserver
+refers to itself as so that requests can properly through the load
```
clemente: This was in the original, but if you want to change it…:
„as so that“ → „so that“
„that the appserver refers to itself“ ← I don't understand the „refers to itself“, maybe add punctuation?
„can properly through“ ← missing verb

README.md
```diff
@@ -886,35 +886,59 @@ service. However, support is available to manually enable those services. To do
...
+   [`create_or_update_partner`](https://github.com/edx/course-discovery/blob/master/course_discovery/apps/core/management/commands/create_or_update_partner.py).
+   Use the same partner code as what you used for ecommerce.
+1. [Configure LMS to use ecommerce](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/ecommerce/install_ecommerce.html#switch-from-shoppingcart-to-e-commerce)
+
+Test your configuration:
+
+1. Verify the OIDC login for both services:
+    https://ecommerce-<your-instance>.opencraft.hosting/login
+    https://discovery-<your-instance>.opencraft.hosting/login
+1. Verify that the discovery cronjob runs without errors:
```
clemente: Does this need to be manually set up as a cron job too? I didn't find where in configuration the cron job is added

clemente:
:+1:

I verified that the instructions are consistent with what was required to set up the sandbox. The new links are more helpful, and the documentation is more complete.

Because I didn't actually follow the instructions and built a my own server, it would be better if @kaizoku could also take a quick look to make sure that the new instructions make sense to set up any other e-commerce.

README.md
pomegranited: Thanks for spotting those typos -- I'd glossed over them!  Addressed by https://github.com/open-craft/opencraft/pull/294/commits/efab883dbcd784a4950ee15737be7a8b9e79e363#diff-04c6e90faac2675aa89e2176d2eec7d8R904

README.md
pomegranited: True..  It's in our `opencraft-release/ginkgo.1` playbook, but unfortunately hasn't been contributed (or wasn't accepted?) upstream.  I've added a note about it https://github.com/open-craft/opencraft/pull/294/commits/efab883dbcd784a4950ee15737be7a8b9e79e363#diff-04c6e90faac2675aa89e2176d2eec7d8R887

clemente: It's also not in the [opencraft-release/ginkgo.1-cloudera branch](https://github.com/open-craft/configuration/blob/opencraft-release/ginkgo.1-cloudera/playbooks/roles/discovery/tasks/main.yml).

README.md
clemente: Looks good, thanks.

PR #296, opened 2018-04-13
Add option to skip activation of instance in redeployment.

UmanShahzad:
Adds an option to skip the activation of an appserver after successful instance redeployment. Useful for pre-provisioning instances en masse.

Testing instructions:

Go on stage, checkout this branch, restart, run the mass redeployment command for a single server (use a tag that selects a single server somehow) with this flag set, and notice the resulting server is not activated, whereas it normally would have been.

### DIFF
--- a/instance/management/commands/instance_redeploy.py
+++ b/instance/management/commands/instance_redeploy.py
@@ -99,6 +99,11 @@ class Command(BaseCommand):
             help='Pass --force to start redeployment without confirming first.'
         )
         parser.add_argument(
+            '--no-activate',
+            action='store_true',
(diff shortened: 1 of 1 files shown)


xitij2000:
> (use a tag that selects a single server somehow)
You can do this with a yaml file that just selects by id. For instance:
``yaml
---
id: 123
# Multiple ids
id__in:
- 123
- 456
``

UmanShahzad:
@xitij2000 Thanks!

kaizoku:
:+1:

- [✓] I tested this: (deployed on stage, provisioned a new appserver and checked that it's not active)
- [✓] I read through the code
- [✓] Includes documentation

PR #297, opened 2018-04-18
Support for Mongo replica set

rocioar:
Add support for Mongo replica set on OpenEdxInstances

Notes for reviewer:

- This PR adds settings on .env to setup the default replica set for new instances
- Creates one new model for the Replica set
- Adds settings needed for edx to work with the replica

About testing:

I've tested the app, by creating some dummy settings and trying to create an instance. I ensured that the default replica set is created using the default settings in .env.
Since we don't have a staging replica set, testing will happen on production, according to Sven.


rocioar:
There is an issue with migrations, it's caused by the order in which they are running, I will review this and fix it tomorrow, but should be easy to do it.

rocioar:
As it is right now, its not creating the default mongo db server, setup using: DEFAULT_INSTANCE_MONGO_URL setting. We could change the code so that it creates both the default replica set, and the default_instance_mongo

instance/migrations/0101_auto_20180418_1601.py
```diff
@@ -0,0 +1,50 @@
...
+# Generated by Django 1.10.7 on 2018-04-18 22:21
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+import django_extensions.db.fields
+import instance.models.mixins.database
+
+
+class Migration(migrations.Migration):
```
e-kolpakov: @rocioar Might be good to give this migration a sensible name (i.e. `add_mongodb_replia_set_model`)

instance/migrations/0078_field_defaults.py
```diff
@@ -85,7 +85,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='openedxinstance',
             name='mongodb_server',
-            field=models.ForeignKey(blank=True, default=instance.models.mixins.database.select_random_mongodb_server, null=True, on_delete=django.db.models.deletion.PROTECT, to='instance.MongoDBServer'),
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, to='instance.MongoDBServer'),
```
e-kolpakov: @rocioar I think modifying old migrations is not a good idea. Maybe you could preserve the `select_random_mongodb_server` method where it was? Or maybe just move it to some "legacy" package.

Alternative approach (probably better, but more work) is to squash the migrations. Looks like there are over a hundred mmigrations now, so I bet we'll never need anything below 0090. I think Django provides some means for doing so.

@pomegranited @bradenmacdonald Might be good to know your thoughts about squashing migrations.

instance/migrations/0102_auto_20180419_1648.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2018-04-19 16:48
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+import instance.models.mixins.database
+
+
+class Migration(migrations.Migration):
```
e-kolpakov: @rocioar Sensible file name as well.

instance/migrations/0102_auto_20180419_1648.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('instance', '0101_auto_20180418_1601'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='openedxinstance',
+            name='mongodb_server',
+            field=models.ForeignKey(blank=True, default=instance.models.mixins.database.select_random_mongodb_server, null=True, on_delete=django.db.models.deletion.PROTECT, to='instance.MongoDBServer'),
```
e-kolpakov: @rocioar Looks like the difference between this and 0100 is this default - can't it be done in 0100?

Another question - should this field remain nullable at the end? It's a geniune question, not a suggestion to change something :)

instance/models/database_server.py
```diff
@@ -47,10 +47,44 @@
...
     Custom manager for the DatabaseServer model.
     """
+    def create_database_server(self, host=None, user=None, password=None, port=None,
+                               primary=None, replica_set=None):
+        """
+        Create MongoDB server
+        """
+        logger.info("Creating DatabaseServer %s", host)
+        extra_args = {}
+        if self.model._meta.model_name == 'mongodbserver':
```
e-kolpakov: @rocioar This is essentially a type check, which is in general discouraged in object-oriented programming.

Maybe it should be a method with overrides in descendent classes? Something like

```
# this class
def extra_args(self, arg1, arg2, arg3, ...):
    return {}

# here
extra_args = self.extra_args(**necessary_parameters

# in MongoDBServerManager (create if not exists)
def extra_args(self, host, primary, replica_set):
    return dict(...)
```

instance/models/database_server.py
```diff
@@ -198,13 +213,112 @@ def default_port(self):
...
+            'hosts': self._get_setting('HOSTS'),
+            'port': self._get_setting('PORT')
+        }
+
+    def _create_default(self):
+        """
+        Create the default database replica set configured in the Django settings, if any.
+        """
+        replica_settings = self.get_replica_set_settings()
+        for setting in replica_settings:
```
e-kolpakov: @rocioar suggestion: this code gets the job done, but a bit unelegant. Do you think something like this ould work better:

```
optiona_fields = set("port")
required_fields = set(replica_settings.keys()) - optional_fields
all_required_present = all(replica_settings[field] is not None for field in required_fields)

if not all_required_present:
    logger.error(...)

File: instance/models/database_server.py [hunk 7]
```diff
@@ -198,13 +213,112 @@ def default_port(self):
...
+            if setting != 'port' and replica_settings[setting] is None:
+                logger.error(
+                    "Error creating the default servers for the replica set, please ensure that"
+                    " all needed settings are configured."
+                )
+                return
```
e-kolpakov: @rocioar Should it just silently fail (except error message) or throw an exception and let caller decide what to do?

instance/models/database_server.py
```diff
@@ -198,13 +213,112 @@ def default_port(self):
...
+
+        replica_set_hosts = replica_settings['hosts'].split(',')
```
e-kolpakov: @rocioar nit: might be good to trim the splitted values - people tend to put spaces after commas :)

I meann, might be good to support both:

* mongo1,mogno2,mongo3
* mongo1, mongo2, mongo3

instance/models/database_server.py
```diff
@@ -198,13 +213,112 @@ def default_port(self):
...
+        """
+        Select a replica set for a new instance.
+        The current implementation selects one of the replica sets with servers that accept new clients at random.
+        If no database server accepts new clients, DoesNotExist is raised.
+        """
+        self._create_default()
+
+        # The set of servers might change between retrieving the server count and retrieving the random server,
+        # so we make this atomic.
+        with transaction.atomic():
```
e-kolpakov: @rocioar I understand this was migrated from previous version, but can't we simplify it to not require atomic transaction?

Looks like something like this might just do the same:

1. Fetch all distinct mongoDB servers **replica sets**
2. Count them
3. Do the "count > 0" check
4. Return random replica set.

The difference is that we fetch replica sets right away, and than just pick one random object from "memory", while current implementation fetches objects, than picks one random from them, than goes to DB to fetch a field.

instance/models/mixins/database.py
```diff
@@ -265,12 +287,26 @@ def mongo_database_names(self):
         """
         return NotImplementedError
 
+    def _get_main_database_url(self):
+        """
+        Returns main database url from replica set, or url from signle server
+        """
+        #if self.mongodb_replica_set:
```
e-kolpakov: @rocioar nit: please remove commented code

instance/tests/models/test_database_server.py
```diff
@@ -223,6 +223,50 @@ def test_set_field_defaults(self):
...
+    DEFAULT_MONGO_REPLICA_SET_PASSWORD="test",
+    DEFAULT_MONGO_REPLICA_SET_PRIMARY="test.opencraft.hosting",
+    DEFAULT_MONGO_REPLICA_SET_HOSTS="test.opencraft.hosting,test1.opencraft.hosting,test2.opencraft.hosting"
+)
+class MongoDBReplicaSetManagerTest(TestCase):
+    """
+    Test cases for MongoDBReplicaSetManager.
+    """
+
+    def test__create_default_replica_set(self):
```
e-kolpakov: @rocioar nit: double underscore: test__

instance/tests/models/test_database_server.py
```diff
@@ -223,6 +223,50 @@ def test_set_field_defaults(self):
...
+    @override_settings(
+        DEFAULT_MONGO_REPLICA_SET_USER=None
+    )
+    def test_select_random_fails(self):
+        """
+        Test that `select_random` returns DoesNotExist when one of the settings is not configured
+        """
+        with self.assertRaises(MongoDBReplicaSet.DoesNotExist):
+            MongoDBReplicaSet.objects.select_random()
+        self.assertLogs("instance.models.database_server" "ERROR")
```
e-kolpakov: @rocioar Missing comma between "ERROR" and "instance.models.database_server", or am I missing something? :)

pomegranited:
Thanks for your code review @e-kolpakov -- I'm doing a pass too, testing with my devstack and its default single-mongo server configuration, and then will do a careful check using the production mongo replicaset.

instance/models/mixins/openedx_database.py
```diff
@@ -351,20 +352,44 @@ def _get_mongo_settings(self):
         """
         Return dictionary of mongodb settings
         """
-        return {
+        extra_settings = {}
+        if self.mongodb_replica_set:
+            mongodb_servers = MongoDBServer.objects.filter(
+                mongodb_replica_set=self.mongodb_replica_set
+            ).values_list('hostname', flat=True)
```
pomegranited: `values_list` still returns a QuerySet object, which doesn't serialise.  To turn it into the list of hostnames that `EDXAPP_MONGO_HOSTS` expects, you need to wrap it in `list()`.

instance/models/mixins/openedx_database.py
```diff
@@ -351,20 +352,44 @@ def _get_mongo_settings(self):
...
+                "EDXAPP_LMS_SPLIT_DOC_STORE_READ_PREFERENCE": "{{ EDXAPP_MONGO_LMS_READ_PREFERENCE }}"
+            }
+        elif self.mongodb_server:
+            mongodb_servers = [self.mongodb_server.hostname]
+        settings = {
             "EDXAPP_MONGO_USER": self.mongo_user,
             "EDXAPP_MONGO_PASSWORD": self.mongo_pass,
-            "EDXAPP_MONGO_HOSTS": [self.mongodb_server.hostname],
+            "EDXAPP_MONGO_HOSTS": mongodb_servers,
             "EDXAPP_MONGO_PORT": self.mongodb_server.port,
```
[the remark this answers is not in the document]
pomegranited: e.g. 
```diff
diff --git a/instance/models/mixins/openedx_database.py b/instance/models/mixins/openedx_database.py
index 877a0cc..4155314 100644
--- a/instance/models/mixins/openedx_database.py
+++ b/instance/models/mixins/openedx_database.py
@@ -355,8 +355,10 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
         extra_settings = {}
         if self.mongodb_replica_set:
             mongodb_servers = MongoDBServer.objects.filter(
-                mongodb_replica_set=self.mongodb_replica_set
-            ).values_list('hostname', flat=True)
+                replica_set=self.mongodb_replica_set
+            )
+            mongodb_servers_list = list(mongodb_servers.all().values_list('hostname', flat=True))
+            primary_mongodb_server = mongodb_servers.filter(primary=True)[0]
             extra_settings = {
                 "EDXAPP_MONGO_REPLICA_SET": self.mongodb_replica_set.name,
                 # Used only if EDXAPP_MONGO_REPLICA_SET is provided.
@@ -366,19 +368,20 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
                 "EDXAPP_LMS_SPLIT_DOC_STORE_READ_PREFERENCE": "{{ EDXAPP_MONGO_LMS_READ_PREFERENCE }}"
             }
         elif self.mongodb_server:
-            mongodb_servers = [self.mongodb_server.hostname]
+            primary_mongodb_server = self.mongodb_server
+            mongodb_servers_list = [self.mongodb_server.hostname]
         settings = {
             "EDXAPP_MONGO_USER": self.mongo_user,
             "EDXAPP_MONGO_PASSWORD": self.mongo_pass,
-            "EDXAPP_MONGO_HOSTS": mongodb_servers,
-            "EDXAPP_MONGO_PORT": self.mongodb_server.port,
+            "EDXAPP_MONGO_HOSTS": mongodb_servers_list,
+            "EDXAPP_MONGO_PORT": primary_mongodb_server.port,
             "EDXAPP_MONGO_DB_NAME": self.mongo_database_name,
 
             "FORUM_MONGO_USER": self.mongo_user,
             "FORUM_MONGO_PASSWORD": self.mongo_pass,
             # search for all servers in MongoDBServer(replica_set=...)
-            "FORUM_MONGO_HOSTS": mongodb_servers,
-            "FORUM_MONGO_PORT": self.mongodb_server.port,
+            "FORUM_MONGO_HOSTS": mongodb_servers_list,
+            "FORUM_MONGO_PORT": primary_mongodb_server.port,
             "FORUM_MONGO_DATABASE": self.forum_database_name,
             "FORUM_REBUILD_INDEX": True,
             "FORUM_MONGO_URL": (
```

instance/tests/models/test_database_server.py
rocioar: wow, how did this pass the pylint checks?

instance/tests/models/test_database_server.py
rocioar: this is because _create_default is a private method, and starts with an underscore, the first underscore on the name of this tests it's just to divide the words, and the other one is part of the name of the function this is testing.

instance/models/database_server.py
rocioar: @e-kolpakov This function is also checking that the MongoDB severs are accepting clients, do you think it would be a good idea to avoid that check?

Also to avoid the extra query I can add a select_related

instance/models/database_server.py
rocioar: @e-kolpakov to be honest I rather use for and ifs than using all(), I think it's much more easier to read. But I will make a small change to see if looks better for you.

instance/migrations/0078_field_defaults.py
bradenmacdonald: @e-kolpakov @rocioar Squashing migrations sounds like a good idea, yep.

instance/migrations/0101_add_mongodb_replia_set_model.py
```diff
@@ -10,6 +10,8 @@
 
 class Migration(migrations.Migration):
 
+    replaces = [('instance', '0101_auto_20180418_1601'), ('instance', '0102_auto_20180419_1648'), ('instance', '0103_auto_20180420_1600')]
```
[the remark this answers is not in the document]
pomegranited: Good point -- the CircleCI tests will tell us once the quality tests succeed.

rocioar:
@pomegranited @e-kolpakov I've updated the code addressing your comments.

Checking now on the issues @pomegranited is having with migrations.

rocioar:
@pomegranited I've merged your PR and fixed the tests

instance/migrations/0102_auto_20180419_1648.py
rocioar: Updated the filename

instance/models/database_server.py
rocioar: fixed :)

instance/migrations/0102_auto_20180419_1648.py
rocioar: I updated it so that it's all done on the same migration file.

About this field being nullable, yes, I think so, it's going to be null when a replica set is setup instead of a single server.

instance/migrations/0102_add_mongo_replica_set.py
```diff
@@ -0,0 +1,51 @@
...
+        ),
+        migrations.AddField(
+            model_name='mongodbserver',
+            name='primary',
+            field=models.BooleanField(default=False, help_text='Indicates if the server is the primary server on a Replica Set, only applies when replica_set is set.'),
+        ),
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='mongodb_replica_set',
+            field=models.ForeignKey(blank=True, default=instance.models.mixins.database.select_random_mongodb_replica_set, null=True, on_delete=django.db.models.deletion.PROTECT, to='instance.MongoDBReplicaSet'),
```
pomegranited: This is causing an issue for existing instances which have a non-replicaset mongodb selected.

For example, our Cloudera instance points to its own separate, single-node mongodb server.  If we enable replicasets and run this migration, then the Cloudera instance would be updated to point to use the new default replicaset, instead of its own mongodb server, and new appservers will fail.

We can deal with this using the following procedure:
* Leave `DEFAULT_INSTANCE_MONGO_URL` set
* Deploy this change
* Run migrations
* Unset `DEFAULT_INSTANCE_MONGO_URL` and provide the replicaset alternative settings
* Start Ocim

So any *new* instances will use the replicaset by default, but existing ones won't change. To use the replicaset for existing instances, we'll need to:
* Create a new instance to trigger `select_random_mongodb_replica_set` to create the new `MongoDBReplicaSet`
* Update each instance we want to use replicaset to provide `mongodb_replica_set`.

@rocioar Does this sound ok?  Can you think of a better way?

rocioar: @pomegranited this sounds ok Jill, good catch actually! I will make sure that this is done this way when deploying

instance/migrations/0078_field_defaults.py
rocioar: Squashed migrations from 0090 down, I kept the files, so that deployment can go well, and once deployment is done, we can remove them :)

rocioar:
@pomegranited thanks for the great review, I've squashed the commits, squashed migrations, and fixed the nit. It should be good to merge once tests are green :)

instance/models/database_server.py
```diff
@@ -47,11 +47,37 @@
...
+    def extra_args(self, host, primary, replica_set):  # pylint: disable=no-self-use
+        """
+        Return extra attributes needed for database creation
+        """
+        return {}
+
+    def create_database_server(self, host=None, user=None, password=None, port=None,
+                               primary=None, replica_set=None):
+        """
+        Create MongoDB server
```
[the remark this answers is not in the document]
rocioar: you are right @UmanShahzad will fix :)

PR #299, opened 2018-04-22
Run node-exporter role after opencraft-role.

smarnach:
The node-exporter role needs the SSL certificates already in place, so we can only run it after the opencraft role.


UmanShahzad:
Looks good! Just waiting on a comment on JIRA before approval.

PR #300, opened 2018-04-24
Run an extra custom playbook on all appservers.

UmanShahzad:
A default custom playbook will run on all new Open edX appservers that installs some common services like Consul, Prometheus Node Exporters, Filebeat, etc..

To test this:

1. Go on stage and set proper stage settings for Prometheus Node Exporters & Consul according to the new settings that were added in this PR.
2. Redeploy Ocim stage.
3. Deploy a new appserver.
4. Check that its configuration settings show the new variables, plus the consul service definition.
5. Check that the custom playbook is the one being run.
6. When the playbook finishes running, check on the server that the node exporter, consul and filebeat are all installed successfully.


instance/models/mixins/openedx_config.py
```diff
@@ -311,3 +314,41 @@ def _get_configuration_variables(self):
             })
 
         return template
+
```
smarnach: You also need `_get_prometheus_variables()` here with these settings:
```
NODE_EXPORTER_NGINX_CONFIG_PATH: /edx/app/nginx/sites-available/node-exporter
NODE_EXPORTER_HTPASSWD_PATH: /edx/app/nginx/nginx.htpasswd
NODE_EXPORTER_USE_SSL: false
```

UmanShahzad:
@smarnach Tests finally passed here.

instance/models/mixins/openedx_config.py
```diff
@@ -311,3 +314,50 @@ def _get_configuration_variables(self):
...
+    def _get_consul_variables(self):
+        """Get all Consul-related ansible variables specific to Open edX appservers."""
+        return {
+            **super()._get_consul_variables(),
+            'consul_service_config': {
+                'service': [
+                    {
+                        "id": "edxapp-lms",
+                        "name": "edxapp",
+                        "port": 8000,
```
smarnach: I don't think it's useful to advertise the internal ports the services are running on.  The service catalogue is meant for service discovery by other servers, e.g. the load balancer, but no other server will send traffic to the internal port.  We should simply advertise edxapp on port 80 instead, without distinguishing LMS and CMS.  They are both part of the same HTTP service, and the difference between the two can only be expressed in terms of HTTP notions (e.g. the Host header), while the service catalogue is only about IP address and port.

instance/models/mixins/openedx_config.py
```diff
@@ -311,3 +314,50 @@ def _get_configuration_variables(self):
...
+                        "tags": ["edxapp", "lms"],
+                        "enable_tag_override": True,
+                        "checks": [
```
smarnach: It's fine to remove the checks for now – using the heartbeat URL from the load balancer should be good enough.

instance/models/mixins/openedx_config.py
UmanShahzad: :+1: I was not entirely sure what to do either here so just chose these ports, thanks for clarifying.

instance/models/mixins/openedx_config.py
UmanShahzad: :+1: makes it easier!

UmanShahzad:
@smarnach OK I'll have to dig into the integration tests to make that work. My main concern is making Consul actually "leave" the cluster when the integration tests are finished. I can't remember the default behavior off the top of my head, but I think it needs to explicitly leave to let all other nodes know through gossip that it's not something accidental.

UmanShahzad:
@smarnach Since we're still using the old `requirements.yml` way of doing things for Ocim deployments, and we don't have separate repos anymore for those roles we listed there, how do you want to consolidate the Ocim playbook with the others? Should we just move it in there as we've done with the others so far? My concern is that this will break the Vagrantfile ansible provisioner, so we'd have to pull the playbooks repo no matter what, or provide ready-to-go Vagrant boxes.

smarnach:
@UmanShahzad I think I will go with option 1.  I'll work on this now, since we kind of need this to proceed.

opencraft/settings.py
```diff
@@ -596,3 +600,22 @@
...
+# Filebeat ####################################################################
+
+# The Logstash host to forward logs to.
+FILEBEAT_LOGSTASH_HOSTS = env.list('FILEBEAT_LOGSTASH_HOSTS', default=[])
+
+# The URL to retrieve the public CA certificate from.
+FILEBEAT_CA_CERT_URL = env('FILEBEAT_CA_CERT_URL', default='')
+
+# Common fields for all Filebeat prospectors.
+FILEBEAT_COMMON_PROSPECTOR_FIELDS = env.dict('FILEBEAT_COMMON_PROSPECTOR_FIELDS', default={})
```
smarnach: `env.dict` is weird.  The format is a [comma-separated list of `key=value` pairs](http://django-environ.readthedocs.io/en/latest/#supported-types), without support for quoting special characters.  I think we should use `env.json()` here, and make sure the Ansible code wriites out proper JSON.

opencraft/settings.py
```diff
@@ -596,3 +600,22 @@
...
 AWS_S3_BUCKET_PREFIX = env('S3_BUCKET_PREFIX', default='ocim')
 AWS_IAM_USER_PREFIX = env('IAM_USER_PREFIX', default='ocim')
+
+# Consul ######################################################################
+
+# The encryption key used to gossip in a Consul cluster.
+CONSUL_ENCRYPT = env('CONSUL_ENCRYPT', default='')
+
+# The list of server agents in the Consul cluster.
+CONSUL_SERVERS = env.list('CONSUL_SERVERS', default=[])
```
smarnach: Same here – we need to make sure the Ansible code emits a comma-separated list instead of the Python representation we are currently writing.  Alternatively, we can consider using JSON instead.

instance/models/mixins/openedx_config.py
```diff
@@ -311,3 +314,48 @@ def _get_configuration_variables(self):
...
+
+    def _get_consul_variables(self):
+        """Get all Consul-related ansible variables specific to Open edX appservers."""
+        return {
+            **super()._get_consul_variables(),
+            'consul_service_config': {
+                'service': [
+                    {
+                        "id": self.server_hostname,
+                        "name": "edxapp",
```
smarnach: I noticed that we are using the same service name for all Open edX instances, which I don't think we should.  I think all appservers of an Open edX instance should use the same service name, but different Open edX instances provide different services.  For this reason, I propose to change this name to `"edxapp-{}".format(self.id)"`.

UmanShahzad:
@smarnach This is ready for another (hopefully final) review.

UmanShahzad:
@smarnach No worries, I've addressed the TODOs and made the change you requested.

smarnach:
@UmanShahzad I think that the settings should to be persisted in the database as well, just as we do for `configuration_settings` and `lms_user_settings`.  Otherwise, this looks great, and it's working great as well. :)

UmanShahzad:
@smarnach Sounds good, I'll do that.

UmanShahzad:
Tests passed, so now just squashing one last time.

UmanShahzad:
Had to rebase because the latest master had another migration at number 102.

pomegranited:
@UmanShahzad  CC @smarnach  I'm getting a failure on a [new appserver deployment](https://console.opencraft.com/instance/5270/edx-appserver/6968/) due to this change.  Is it easy to fix with ssh config maybe?

```
INFO 2018-05-03 22:42:20+0930 INFO Running playbook "playbooks/appserver.yml" from "https://github.com/open-craft/ansible-playbooks.git"
...
INFO 2018-05-03 22:42:31+0930 INFO PLAY [Set up common services on appservers] **********************************************************************************************************************************************************************************************************************
2018-05-03 22:42:31+0930 INFO TASK [Gathering Facts] *******************************************************************************************************************************************************************************************************************************************
2018-05-03 22:42:32+0930 INFO fatal: [149.202.172.137]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\r\n@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @\r\n@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\r\nIT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!\r\nSomeone could be eavesdropping on you right now (man-in-the-middle attack)!\r\nIt is also possible that a host key has just been changed.\r\nThe fingerprint for the ECDSA key sent by the remote host is\nSHA256:oHtPoGBbrrh5lq6deLIpO6uYp9meGGirG1MeudnsLPM.\r\nPlease contact your system administrator.\r\nAdd correct host key in /var/www/.ssh/known_hosts to get rid of this message.\r\nOffending ECDSA key in /var/www/.ssh/known_hosts:1998\r\n  remove with:\r\n  ssh-keygen -f \"/var/www/.ssh/known_hosts\" -R 149.202.172.137\r\nECDSA host key for 149.202.172.137 has changed and you have requested strict checking.\r\nHost key verification failed.\r\n", "unreachable": true}
```

smarnach:
@pomegranited Hmm, I'm not really sure what's going on there.  Why should the host key of the target host change between running the edx_sandbox playbook and our playbook?  And we certainly haven't seen this problem in our tests.  I will take a look to see whether I can identify the problem.  For what it's worth, I don't think Ocim can perform meaningful host key checks without significant additional work so for now we should probably just disable them, at least as a quick workaround.

smarnach:
@pomegranited I may have a vague idea what is causing this.  For now I should have fixed the problem by simply disabling host key checking altogether.

pomegranited:
Thanks @smarnach !

smarnach:
@pomegranited @UmanShahzad I finally figured this out.  The [Ansible configuration in edx/configuration disables SSH host key checking](https://github.com/edx/configuration/blob/837308626ff278a39303f8b987313d7e38c36ef3/playbooks/ansible.cfg#L8).  The first time a playbook runs for a given IP, SSH will accept the host key unconditionally and add it to `~/.ssh/known_hosts`.  The `ansible.cfg` in ansible-playbooks, by contrast, does not disable host key checking.  Most of the time this isn't a problem, since by the time the playbook runs the host key is already in `known_hosts`.  If we by chance get assigned an IP that we already had for an earlier app server, we will have a different host key in `known_hosts`.  Ansible will ignore the discrepancy for the upstream playbook, since host key checking is disabled, but SSH won't add the key to `known_hosts`, because there already is a different key for the same IP.  In this case, our playbook will fail.  This also explains why we didn't see this problem in our tests – we just didn't get an IP we already had in `known_hosts`.

I think the best solution for now is to set `ANSIBLE_HOST_KEY_CHECKING=false` when running Ocim from Ansible – I've filed PR #308, though I didn't have time to test it.

UmanShahzad:
@smarnach @pomegranited Thanks for figuring this out guys :)

PR #301, opened 2018-04-26
WIP: Link to instance page from admin

clemente:
This adds a „View on site“ button to the openedx_instance admin page, to be able to move from Django's admin to Ocim's admin.

view-on-site

Tested in local.

This PR was created to test CircleCI's cleanup process. **WIP** because I might still use it to trigger some builds.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -26,6 +26,7 @@ from django.db import models, transaction
 from django.db.backends.utils import truncate_name
 from django.db.models import F
 from django.template import loader
+from django.urls import reverse
 from django.utils import timezone
 
 from instance import gandi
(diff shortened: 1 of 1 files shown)


UmanShahzad:
@clemente Cool :) I wish there was a way to go backwards though, to the admin. But that might be way out of the scope here.

clemente:
@UmanShahzad for that other part we'd need to decide where to put the link without cluttering the design.

UmanShahzad:
@clemente Right now the instance name above the box with the LMS/CMS links is just plain text, so maybe a clutter-free way to add this would be to make that an anchor to the corresponding admin. Not sure if that feels right from a UX perspective, but seems like a good start.

clemente:
@UmanShahzad Sven had a simpler idea: in the „Settings“ tab, after the words „Current configuration“, just add an „(edit)“ link to Django.
Unfortunately, this new change is out of scope because this PR doesn't even belong to a ticket (I confirmed it with Xavier some time ago). So I guess we can just merge this one, and I'll try to sneak in the other change in other future Ocim PR. I rebased. When/if the tests pass, could you please merge it? Oh, I see you approved it, so I' can merge it tomorrow too.

UmanShahzad:
@clemente I didn't formally approve nor test this just yet actually, but I think it's a fine feature with a small footprint. If you can get tests passing and put it on stage for testing and ping me I'd be up for double-checking.

clemente:
@UmanShahzad ok but time can't be logged to any task.
It's best to finish these tasks as part of other tickets.

swalladge:
@clemente this has been merged in #478

PR #304, opened 2018-04-27
Remove deploy/ subdirectory in favour of ansible-playbooks repo.

smarnach:
The playbooks are now in ansible-playbooks, just like everything else.  See https://github.com/open-craft/ansible-playbooks/pull/26 for details.

I updated the instructions for setting up the Vagrant devstack in `README.md` and the Vagrantfile.  I moved the `deploy/README.md` file to the ansible-playbooks repository.


README.md
```diff
@@ -86,14 +86,15 @@ environment:
...
-Once you have these tools installed, you will need to download the Ansible
-roles used to build the Vagrant instance. Run the following in the ``deploy``
-folder in directory where you have cloned Ocim:
+Once you have these tools installed, you will need to download the [Ansible
+playbooks](https://github.com/openc-craft/ansible-playbooks) used to build the
+Vagrant instance.  If you haven't checked it out yet, you can clone it into the
+`deploy/` subdirectory using this command at the root of the repository:
 
-    ansible-galaxy install -r requirements.yml
+    git clone https://github.com/openc-craft/ansible-playbooks/deploy
```
UmanShahzad: `openc-craft` -> `open-craft`

UmanShahzad: I think you meant to replace the `/` between `ansible-playbooks` and `deploy` with a space.

Vagrantfile
```diff
@@ -21,19 +21,19 @@ Vagrant.configure(2) do |config|
...
+    end
   end
 
   config.vm.provider :virtualbox do |vb|
     vb.memory = 1024
     # Allow DNS to work for Ubuntu host
     # http://askubuntu.com/questions/238040/how-do-i-fix-name-service-for-vagrant-client
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
+    vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+    vb.customize ["modifyvm", :id, "--nictype1", "virtio"]
```
UmanShahzad: Yay!

README.md
UmanShahzad: Also, because we're running the playbook from the host, we actually need the correct Ansible 2.5 install through `deploy/requirements.txt`, so that should be another step.

UmanShahzad:
@smarnach Just one tiny note: Ocim uses [netaddr==0.7.18](https://github.com/open-craft/opencraft/blob/master/requirements.txt#L70) while ansible-playbooks uses [netaddr==0.7.19](https://github.com/open-craft/ansible-playbooks/blob/master/requirements.txt#L2). Can we upgrade Ocim's?

README.md
smarnach: The installation of Ansible is mentioned above, as it hass been before my change.  This actually has always been somewhat out of order, so I fixed it now.

smarnach:
@UmanShahzad Thanks for the review! I'm not sure why we should upgrade `netaddr`, since Ocim's dependencies are installed inside the Vagrant VM, while Ansible is installed in some virtual environment on the host machine.  Anyway, since I didn't see anything problematic in the [changelog](https://github.com/drkjam/netaddr/compare/netaddr-0.7.18...netaddr-0.7.19#diff-d3bb3391c79904494c60ee2ac2f33070), I've upgraded it anyway.

UmanShahzad:
@smarnach Oh yeah, that's kind of dumb for me to ask :) Sorry, I was blindly installing Ocim requirements on the host virtualenv as well, and noticed the discrepancy so mentioned it, but it's clearly not needed. Thanks for upgrading anyway :)

PR #305, opened 2018-05-01
Document Ansible installation for Vagrant.

smarnach:
@UmanShahzad This is a small documentation fix I made while addressing review comments on #304, but it somehow didn't make it into the repo – I must have messed up with amending commits and pushing.  Could you please take a quick look?


smarnach:
@UmanShahzad You are right – I originally used `mkvirtualenv`, which creates the environment and acitvates it in a single step, but I the realized that not everyone has virtualenvwrapper installed.  Will ifx.

PR #310, opened 2018-05-04
Change from `DEBUG` to `INFO` log level.

UmanShahzad:
We're getting absolutely spammed with useless DEBUG logs on ELK.

(Note I made this by editing on GitHub and have not tested it just yet).

### DIFF
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -230,7 +230,7 @@ HUEY = {
     'always_eager': env.bool('HUEY_ALWAYS_EAGER', default=False),
 
     # Options to pass into the consumer when running ``manage.py run_huey``
-    'consumer': {'workers': 1, 'loglevel': logging.DEBUG},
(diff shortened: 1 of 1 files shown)


opencraft/settings.py
```diff
@@ -579,7 +579,7 @@
 
 if 'file' in HANDLERS:
     LOGGING['handlers']['file'] = {
-        'level': 'DEBUG',
+        'level': 'INFO',
```
smarnach: This is the only log level we care about.  I'd prefer if we only change this one, and maybe the Huey log level, since the Huey debug log messages are really not that useful.  (They actually were quite useful when I had to debug a Huey bug, but we haven't run into any in a while.)

UmanShahzad: Oh yep, sorry, didn't think to just change the file handler since that's what Filebeat will interact with. Will leave the Huey log level at `INFO` then too.

UmanShahzad:
@smarnach Looks good to you?

PR #311, opened 2018-05-06
Catch any exception if an appserver fails to spawn.

UmanShahzad:
Otherwise workers will completely drop the task and the instance redeployment management command will not be able to recognize if an ongoing instance has actually failed, since the ongoing tag will remain.

Test instructions:

Bring this to prod where we're currently facing issues with provisioning S3 on the first try, run the instance redeployment command, and notice that now the low prio workers no longer completely drop all work after failing to provision S3.

### DIFF
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -67,7 +67,10 @@ def spawn_appserver(
         instance = OpenEdXInstance.objects.get(ref_set__pk=instance_ref_id)
 
         instance.logger.info('Spawning new AppServer, attempt %d of %d', i, num_attempts)
-        appserver_id = instance.spawn_appserver()
+        try:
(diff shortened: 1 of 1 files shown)


smarnach:
@UmanShahzad This solution is somewhat incomplete, since the instance itself can't respond to the failure.  In general, the instance may want to perform cleanup operations and send a failure email when a deployment fails, but if we only catch the exception in the Huey task, none of this will happen.  A proper solution would require some refactoring, but if we stick to our plan of eventually factoring the builder out, it may not be worth it to implement this now.

This change is an improvement in any case, and will help to solve our current problem, so :+1: on merging this.  I haven't performed any manual tests, but the change is trivial, and the code is covered at least by the integration tests.

PR #313, opened 2018-05-09
Add retries to S3 bucket creation

rocioar:
I've added retries to the s3 bucket creation, IAM keys may take a bit to be propagated across amazon servers after they are created.

Testing
-------

1- Create a new instance with AWS enabled and a spawn a new app server, check that bucket is created (check the number of attempts that were done, there should be logs about this).

### DIFF
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -22,6 +22,7 @@ Instance app model mixins - Database
 
 # Imports #####################################################################
 import json
+import time
 
 import boto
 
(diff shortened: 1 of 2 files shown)


instance/models/mixins/storage.py
```diff
@@ -211,14 +212,30 @@ def get_s3_connection(self):
...
             bucket = s3.create_bucket(self.s3_bucket_name)
             bucket.set_cors(get_s3_cors_config())
+        except boto.exception.S3ResponseError:
+            if attempts == attempt:
+                raise
+            self.logger.exception(
+                'Retrying bucket creation. IAM keys are not propagated yet, attempt %s of %s',
+                attempt, attempts
+            )
+            time.sleep(retry_delay)
```
bradenmacdonald: This is always running in a worker process, never a web process, right? Want to make sure `time.sleep` won't block anything else.

bradenmacdonald: Also, if this is known to fail from time to time, perhaps we should always sleep first, to reduce the number of warnings/errors showing up in the logs, instead of expecting to log errors from time to time?

rocioar: agree on adding a sleep first. This runs when a new app server is spawned, so it should be ok.

instance/models/mixins/storage.py
```diff
@@ -211,14 +212,30 @@ def get_s3_connection(self):
...
+            self._create_bucket(attempt + 1)
```
UmanShahzad: This will pass `2` into `attempts` on the first retry, rather than update `attempt`. Can be fixed by specifying `attempt` as the key for this value.

However you can probably do without a separate `attempt` key. Just pass in `attempts - 1`, which will take the 4 to a 3, and check if `attempts == 0` to decide to re-raise the error. Maybe made easier to understand by using the name `attempts_remaining`.

rocioar: Good catch, just fixed it. Thanks!

rocioar:
@viadanna thanks for your feedback! I've fixed your suggestions, and also included tests.

rocioar:
@UmanShahzad I've just squashed the commits, merging now

UmanShahzad:
@rocioar There seems to be flakiness with the tests here unfortunately :\ The flakiness seems related to the PR though?

rocioar:
@UmanShahzad fixed!

PR #314, opened 2018-05-10
Fix: don't try to provision a bucket if bucket name is empty

xitij2000:
Currently, it is not possible to save an instance with an empty bucket name since Ocim will try to provision an empty bucket name and fail. This fixes that issue.

### DIFF
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -215,7 +215,7 @@ class S3BucketInstanceMixin(models.Model):
         """
         Create S3 Bucket if it doesn't exist
         """
-        if self.s3_access_key and self.s3_secret_access_key:
+        if self.s3_access_key and self.s3_secret_access_key and self.s3_bucket_name:
             s3 = self.get_s3_connection()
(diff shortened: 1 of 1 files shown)


mtyaka:
Did not actually test this but the fix makes sense 👍

xitij2000:
@mtyaka  I tested it on stage: https://stage.console.opencraft.com/instance/2100/edx-appserver/606/
It at least lets you save without an s3 bucket name. If there are issues after that, they will need to be looked at separately.

xitij2000:
@mtyaka  Actually that doesn't test it at all cause it'll just generate a new name, I need to enable swift.

rocioar:
Once SWIFT_ENABLE=True this should work ok, if SWIFT_ENABLE is set to false then this will still generate a new bucket name, as you said @xitij2000. This is a good quick fix for the current issue, which also needs SWIFT_ENABLE set on True.

xitij2000:
@rocioar  Thanks! Yes, I agree it's for the SWIFT_ENABLE=true case only, but without it, you simply can't launch a new appserver if the s3 bucket name is blank, so it's needed in any case.

PR #316, opened 2018-05-11
Check that the heartbeat is functioning before activating.

UmanShahzad:
There are issues where an appserver is very quickly made active behind the load balancer before it has a steady heartbeat. In that case, HAProxy takes it off the map because it assumes the server is down. Do not put a server in a healthy state on our side unless the heartbeat is functioning.

### DIFF
--- a/instance/models/openedx_appserver.py
+++ b/instance/models/openedx_appserver.py
@@ -21,6 +21,7 @@ Instance app models - Open EdX AppServer models
 """
 import yaml
 
+import requests
 from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
@@ -456,6 +457,10 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
             """ Does server accept SSH commands? """
             return self.server.status.accepts_ssh_commands
 
+        def heartbeat_active():
+            """Is this server's /heartbeat returning 200s?"""
+            return requests.options('http://{}/heartbeat'.format(self.server.public_ip)).ok
+
         try:
             self.server.start(
                 security_groups=self.security_groups,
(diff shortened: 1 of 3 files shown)


instance/models/openedx_appserver.py
```diff
@@ -21,6 +21,7 @@
 """
 import yaml
 
+import requests
```
rocioar: nit: what do you think about moving this requests import next to the yaml import and leaving a space before the django imports? It think it makes imports more clear, and easier to read.

instance/models/openedx_appserver.py
```diff
@@ -456,6 +457,10 @@ def accepts_ssh_commands():
             """ Does server accept SSH commands? """
             return self.server.status.accepts_ssh_commands
 
+        def heartbeat_active():
+            """Is this server's /heartbeat returning 200s?"""
```
UmanShahzad: Note that this comment is inaccurate: see [here](https://github.com/requests/requests/blob/e4fc3539b43416f9e9ba6837d73b1b7392d4b242/requests/models.py#L689-L695).

rocioar: @UmanShahzad you think we should check specifically for 200 here? or just update the comment?

UmanShahzad: @rocioar I wanted to mimic what HAProxy does with `option httpchk`, and [according to the documentation](https://github.com/haproxy/haproxy/blob/master/doc/configuration.txt#L3785-L3786), it seems to do  what `.ok` does, so probably just change the comment.

instance/models/openedx_appserver.py
UmanShahzad: :+1: we should probably use `isort` if we aren't already, or something equivalent.

instance/models/openedx_appserver.py
```diff
@@ -490,6 +495,7 @@ def accepts_ssh_commands():
             self.logger.info('Rebooting server %s...', self.server)
             self.server.reboot()
             self.server.sleep_until(accepts_ssh_commands)
+            self.server.sleep_until(heartbeat_active)
```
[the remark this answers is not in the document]
rocioar: I've updated it to 30 min, @smarnach let me know if that's ok.

PR #317, opened 2018-05-16
Disable New Relic and RabbitMQ by default

rocioar:
Disables New relic and RabbitMQ depends on: https://github.com/open-craft/configuration/pull/43/

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -234,7 +234,7 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             "EDXAPP_WORKER_DEFAULT_STOPWAITSECS": 1200,
 
             # Monitoring
-            "COMMON_ENABLE_NEWRELIC": bool(settings.NEWRELIC_LICENSE_KEY),
+            "COMMON_ENABLE_NEWRELIC": False,
             "COMMON_ENABLE_NEWRELIC_APP": bool(settings.NEWRELIC_LICENSE_KEY),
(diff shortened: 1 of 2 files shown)


bradenmacdonald:
Are we still using New Relic on some instances?

clemente:
@bradenmacdonald It seems this only disables the [newrelic role](https://github.com/edx/configuration/blob/master/playbooks/roles/newrelic/tasks/main.yml), which is for `nrsysmond` which was deprecated.
`COMMON_ENABLE_NEWRELIC` is used in many places, but all seem to relate to the `newrelic` role.

bradenmacdonald:
I see, thanks.

instance/models/mixins/openedx_config.py
```diff
@@ -246,6 +246,9 @@ def _get_configuration_variables(self):
             "DISCOVERY_NGINX_PORT": 80,
             "DISCOVERY_SSL_NGINX_PORT": 443,
 
+            # RabbitMQ disabled locally
+            "SANBOX_ENABLE_RABBITMQ": False,
```
clemente: SANBOX → SANDBOX

UmanShahzad:
@rocioar @clemente Shouldn't this be merged soon?

pomegranited:
@rocioar CC @UmanShahzad  Looks like the [`SANBOX_ENABLE_RABBITMQ` typo](https://github.com/open-craft/opencraft/pull/317/files/5f7c4a39073f56f9abb4be0dec998b551702d83f#diff-b3cc82fdb60b23de6fff2d4252452568) that @clemente  noted didn't get fixed before this was merged 😦

PR #321, opened 2018-05-22
Check that the heartbeat is functioning before activating

rocioar:


rocioar:
@UmanShahzad  could you review this PR? I updated the way we check for heartbeat_active, and added tests. This should avoid the issue of it being on Ready state before it being actually active.

rocioar:
@UmanShahzad big face palm here, I'm updating the code

UmanShahzad:
@rocioar No worries, it happens :) I'm updating stage and re-spawning an appserver with this, hopefully it does it.

rocioar:
@UmanShahzad  I still need to mock heartbeat_active on tests, will do this after dinner.

PR #322, opened 2018-05-23
WIP: Fix S3 provisioning for new appservers when Swift is enabled

viadanna:
[OC-4586](https://tasks.opencraft.com/browse/OC-4586)

S3 is being provisioned for new app servers which were running on Swift. S3 should only be provisioned for new instance, or when migrating an appserver to use it, shouldn't do this automatically.


viadanna:
Changed the approach here, implementing INSTANCE_STORAGE_TYPE environment variable to be set one of ('s3', 'swift' or 'filesystem') as suggested by @pomegranited

README.md
```diff
@@ -335,7 +335,7 @@ Permissions required for master AWS account are:
 * iam:DeleteUserPolicy
 
 Required settings:
-* `SWIFT_ENABLE`: Default is True, it should be set to False when using s3.
+* `INSTANCE_STORAGE_TYPE`: A choice between ('s3', 'swift', 'filesystem').
```
rocioar: nit: can you add here which one is the default one?

instance/models/mixins/storage.py
```diff
@@ -56,6 +56,24 @@ def get_s3_cors_config():
...
+    storage_type = models.CharField(
+        max_length=16,
+        blank=True,
+        default=default_setting("INSTANCE_STORAGE_TYPE")
+    )
+
+    class Meta:
+        abstract = True
+
+
```
rocioar: Could you generate the migration file for this?

.env.test
```diff
@@ -3,7 +3,8 @@ SECRET_KEY='tests'
 DATABASE_URL='postgres://localhost/opencraft'
 DEFAULT_INSTANCE_MYSQL_URL='mysql://root@127.0.0.1'
 DEFAULT_INSTANCE_MONGO_URL='mongodb://127.0.0.1'
-SWIFT_ENABLE=true
+# SWIFT_ENABLE=true  # Not used any longer
```
rocioar: What do you think of changing this comment to "Deprecated in favor of INSTANCE_STORAGE_TYPE setting"

rocioar:
The PR is looking good, just left some tiny comments. 

Will run tests and approve if everything goes well.

.env.test
viadanna: I forgot to remove this after testing.

instance/models/mixins/storage.py
```diff
@@ -235,14 +253,23 @@ def provision_s3(self):
...
+        self._create_bucket()
 
     def deprovision_s3(self):
         """
         Deprovision S3 by deleting S3 bucket and IAM user
         """
-        if not(self.s3_access_key or self.s3_secret_access_key or self.s3_bucket_name):
+        if not self.storage_type == self.S3_STORAGE or \
+                not(self.s3_access_key or self.s3_secret_access_key or self.s3_bucket_name):
```
rocioar: @viadanna maybe to avoid this long line, that is a bit hard to understand, we could add a private method:

def _is_s3_setup(self):
    return self.s3_access_key and self.s3_secret_access_key and self.s3_bucket_name

what do you think?

viadanna: That's an excellent idea!

viadanna:
So I dropped the broken test, as S3 provision was already tested somewhere else. What do you say, @rocioar ?

instance/models/mixins/storage.py
```diff
@@ -207,7 +208,7 @@ def create_iam_user(self):
         """
         Create IAM user with access only to the s3 bucket set in s3_bucket_name
         """
-        if not(settings.AWS_ACCESS_KEY_ID or settings.AWS_SECRET_ACCESS_KEY):
+        if not (settings.AWS_ACCESS_KEY_ID or settings.AWS_SECRET_ACCESS_KEY):
```
rocioar: @viadanna just wondering why the added space here?

rocioar:
- [x] I tested this:
  - I respawned an instance with this branch, with storage_type = swift
  - I respawned an instance with this branch, with storage_type = s3
- [x] I read through the code
- [ ]  ~~I checked for accessibility issues~~
- [x] Includes documentation

PR #324, opened 2018-06-04
Dummy commit to trigger CircleCI

viadanna:
Checking if CircleCI builds master correctly.


viadanna:
Not needed any longer

PR #325, opened 2018-07-03
Mark an instance as 'terminated' only on successful deletion

lgp171188:
## Short description of the changes

If there are errors while making the OpenStack delete API call,
mark the instance as 'unknown'. In case the server is not found,
mark it as 'terminated'.

### Server not found scenario
* Launch an AppServer instance and wait till the OpenStack VM is up and running.
* Delete the VM from the OpenStack console bypassing OCIM.
* Try to terminate the VM from OCIM (I called the `terminate()` method on the instance from the shell)
* The OpenStack API should return a 'not found' status and the status of the AppServer should change to `terminated`.

### OpenStack API issues scenario
* Launch an AppServer and wait till the OpenStack VM is up and running.
* Change the OpenStack API endpoint to gibberish. (I tested by disconnecting the network on my computer).
* Try to terminate the VM.
* Since the OpenStack API is unreachable, there will be a `NewConnectionError` exception thrown (caught by `requests.RequestException`).
* The status of the server should now change to `unknown`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ jobs:
           LOAD_BALANCER_FRAGMENT_NAME_PREFIX: 'integration-'
           TEST_RUNNER: 'opencraft.tests.utils.CircleCIParallelTestRunner'
       - image: redis
-      - image: mongo
+      - image: mongo:3.2-jessie
       - image: "circleci/mysql:5"
         environment:
           MYSQL_ALLOW_EMPTY_PASSWORD: True
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -436,10 +436,16 @@ class OpenStackServer(Server):
             self._status_to_terminated()
             return
 
-        self._status_to_terminated()
         try:
             self.os_server.delete()
         except novaclient.exceptions.NotFound:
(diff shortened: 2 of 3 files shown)


lgp171188:
@UmanShahzad, the CI failure seems to be due to an error which is unrelated to my changes and seems to have been there before as well. What should I do now?

UmanShahzad:
@lgp171188 I've restarted the build; let's see what happens.

UmanShahzad:
@lgp171188 This seems to be happening because we don't have the `mongo` image for CircleCI pinned to a version, and version 4.0, which is the latest, has changes that are breaking things. Can you pin to `3.2-jessie`? It will match the major and minor version we use on our live server which should be enough:

```
$ mongod --version
db version v3.2.19
```

lgp171188:
@UmanShahzad, I have pinned the mongo image version to 3.2-jessie now. Will monitor the CI run to verify whether the issue is fixed or not.

UmanShahzad:
@lgp171188 Thanks! I decided to merge this before squashing since the commits are really different so it'd be nice to have them separate.

PR #326, opened 2018-07-04
Set DISCOVERY_VERSION to a correct default value

lgp171188:
## Short Description

Set the default value of `DISCOVERY_VERSION` to the openedx release
being installed to avoid the master branch of discovery
being used as the default value. The master version of discovery
doesn't work well with other `gingko`-based services.

### Normal Scenario:
1. Create an instance with a specific openedx release (for example, `open-release/ginkgo.2`)
1. Spawn an AppServer.
1. Once the AppServer has started spawning, check the configuration value used for `DISCOVERY_VERSION`. This can be checked from the `Configuration` section of the AppServer in the web UI from the value of `DISCOVERY_VERSION` in `Combined ansible vars:`
1. The value should match the openedx release with which the instance was created.

### With configuration override:
1. Create an instance with a specific openedx release and extra configuration overriding `DISCOVERY_VERSION` to `master`.
1. Repeat steps 2, 3 from the previous scenario.
1. The value of `DISCOVERY_VERSION` should be set to `master` and not the openedx release with which the instance was created.

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -245,6 +245,7 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             "DISCOVERY_HOSTNAME": '~{}'.format(self.instance.discovery_domain_nginx_regex),
             "DISCOVERY_NGINX_PORT": 80,
             "DISCOVERY_SSL_NGINX_PORT": 443,
+            "DISCOVERY_VERSION": self.openedx_release,
 
             # RabbitMQ disabled locally
             "SANBOX_ENABLE_RABBITMQ": False,
--- a/instance/tests/models/test_openedx_appserver.py
+++ b/instance/tests/models/test_openedx_appserver.py
@@ -47,6 +47,8 @@ from instance.tests.utils import patch_services
 
 
 # Tests #######################################################################
+
+@ddt  # pylint: disable=too-many-public-methods
 class OpenEdXAppServerTestCase(TestCase):
     """
     Test cases for OpenEdXAppServer objects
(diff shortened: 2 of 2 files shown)


instance/models/mixins/openedx_config.py
```diff
@@ -245,6 +245,10 @@ def _get_configuration_variables(self):
             "DISCOVERY_HOSTNAME": '~{}'.format(self.instance.discovery_domain_nginx_regex),
             "DISCOVERY_NGINX_PORT": 80,
             "DISCOVERY_SSL_NGINX_PORT": 443,
+            # set to the openedx release being installed since the
+            # default is to install the master version which doesn't work well
+            # with other ginkgo-based services.
```
xitij2000: I don't think this comment is necessary, this is true of all ``_VERSION`` variables.

instance/tests/models/test_openedx_appserver.py
```diff
@@ -201,6 +201,18 @@ def test_launch_in_other_region(self, mock_get_nova_client):  # pylint: disable=
         make_test_appserver(instance)
         mock_get_nova_client.assert_called_once_with("elsewhere")
 
+    def test_default_discovery_version(self):
```
xitij2000: I don't see tests for any of the other ``*_VERSION`` variables. So can you make this more extensive so it checks all configured versions using ddt?

instance/models/mixins/openedx_config.py
lgp171188: Removed in c3b1ae

instance/tests/models/test_openedx_appserver.py
lgp171188: Added tests for all the `*_VERSION` variables in c3b1abe2a2944ebfd2301d1b69eec472d7bd301c

xitij2000:
@lgp171188  I'll test the change and then give final approval, after that you can merge.

lgp171188:
@xitij2000 should I squash all the commits into one before merge?

xitij2000:
@lgp171188  Yup, but wait till I've has a chance to test this.

lgp171188:
As the review is now complete and the changes have been approved, I am squashing the commits before merging.

lgp171188:
@xitij2000, I pushed a squashed commit and the checks have failed due to an error unrelated to my changes - `instance.models.load_balancer.ReconfigurationFailed`. What should I do now?

lgp171188:
@xitij2000, the tests have passed. So I am merging the changes now.

PR #329, opened 2018-07-11
[SE-122] Management command to migrate static data from SWIFT to S3

clemente:
This PR adds a management code to migrate an instance from using SWIFT for files (attachments, forum uploads, exams, …) to using S3. It moves the files (with rclone) and then changes the instance's storage setting.
It includes Rocío's code.

It has already been used to migrate many trial instances.
It will be used very few more items and doesn't need fancy features. It doesn't have unit tests since it does so much remote communication that we would have to mock rclone and AWS, and it's not worth it for a few-times script. It has been tested with real servers and verified to work there.

The script fixes some of the filename bugs we found in servers (e.g. `{submissions_attachmentsuser_tasks/ ---> submissions_attachments/user_tasks`). There may be more edge cases but the current fixes cover all of our servers.

To choose the set of servers you want, you need to change some line.
You can also easily enable/disable particular sections (useful if you need to retry without creating the user/bucket each time).

To run it: `honcho run ./manage.py migrate_swift_to_s3` (from Ocim). Or better, follow the included instructions at the top of the file.

### DIFF
--- a/instance/management/commands/migrate_swift_to_s3.py
+++ b/instance/management/commands/migrate_swift_to_s3.py
@@ -1,0 +1,347 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2019 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Migrate SWIFT assets to S3.
+One-time script.
+
+
+Detailed process to migrate 1 server:
+- download all SWIFT files to your local computer with rclone, as a reference, as a backup and as a guide to know what
+  you'll be copying. Take note of particular patterns in the file names and the types of files used
+- check the current swift/s3 storage settings of the instance to migrate (verify that S3 is still not set up)
+- manually modify the .filter() in this script to find the IDs of the instances to migrate
+- also set the region (use Ireland)
+- enable all conditional parts (change the "if 0:" to "if 1:") the first time. When you re-run it you may use this to
+  skip some migration steps or for debugging
+- run script, from Ocim production (just place the file and run it, no need to change branches). The command is:
+  honcho -e .env run ./manage.py migrate_swift_to_s3
+- let the script copy everything, watch for errors, re-run if it fails (some AWS things take time and you may need
+  to wait some seconds between re-runs). If it fails, it's safe to retry it, since it won't delete the old data:
+  it will create S3 resources and copy data to S3, but still won't use it until you deploy new servers. You can
+  safely remove the new S3 files and re-run the script in case something breaks in the middle, or just re-run to
+  make rclone copy the missing files
+- when the script successfully copies files, let it save() the instance with the new settings
+- check settings and deploy a new server with the blue button
+- wait 2 h
+- test the new server through the internal link
+- to be extra careful: download the files from SWIFT again and compare them with the first download (you can use a tool
+  like dirdiff). If they differ, it means that files changed during those 2 h. You can re-run this script (you can
+  skip some steps in re-runs, like the bucket creation and user creation) and this will cause a 2nd SWIFT→S3
+  synchronization and it will keep files up to date. Verify that files match now. Test the server and activate it
+- if the server doesn't work, don't activate it; restore storage_type to swift (you may delete the IAM user and bucket
+  too, from the AWS web interface)
+- if the instance has high activity or constantly moving files, do a final sync after you activated the new server (from
+  the deactivated server to the activated one). Or you could have scheduled a downtime for this
+- don't delete old files from the SWIFT container (from OVH) and don't delete the container. The container could have
+  been used to create URLs for uploaded images that are linked from forum posts. These URLs still point to SWIFT, and
+  you don't want to go through mongo fixing the posts. Don't delete SWIFT settings from the OpenEdXInstance object
+  either; it's better to leave them as they were, to show that the SWIFT container still exists
+- delete ~/.config/rclone/ if still there
+"""
+
+# Imports #####################################################################
+
+import logging
+import re
+import subprocess
+import swiftclient
+
+from django.core.management.base import BaseCommand
+from instance.models.openedx_instance import OpenEdXInstance
+
+LOG = logging.getLogger(__name__)
+
+# Desired destination region for S3 bucket.
+# This will be saved in instance.s3_region
+# Different regions have different requirements
+# see https://docs.google.com/document/d/1H8iUa05nSD6puQQoUf3DTfPb3gtUFGKtKjUNMKoWAAg/edit#
+#
+# Recommended: '' (default) and 'eu-west-1' (Ireland)
+S3_REGION = 'eu-west-1'
+
+# Classes #####################################################################
+
+# We don't have unit tests for this command ("no cover" disables test coverage in the class) because it uses
+# too many network tools and processes involved that aren't worth mocking: rclone file transfers,
+# creation of rclone config files, IAM user and S3 bucket creation. There are already some "assert" that do
+# basic sanity checks.
+# The best way to test the script is by running it against some servers and verifying that it moves all files.
+
+
+class Command(BaseCommand):  # pragma: no cover
+    """
+    Migrate SWIFT to S3.
+    """
+    help = (
+        'Migrate a server\'s data from SWIFT to S3.'
+    )
+
+    def __init__(self, *args, **kwargs):
+        """
+        Some unused options.
+        """
+        super(Command, self).__init__(*args, **kwargs)
+        self.options = {}
+        self.retried = {}
+
+    @staticmethod
+    def _get_swift_connection(instance):
+        """
+        Get Connection object.
+        """
+        return swiftclient.Connection(
+            user=instance.swift_openstack_user,
+            key=instance.swift_openstack_password,
+            authurl=instance.swift_openstack_auth_url,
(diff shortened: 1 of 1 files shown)


instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        LOG.info("%s ---> %s" % (key_name, new_name))
+        command = "rclone copy ocim-swift:%s/%s ocim-s3:%s/%s" % (swift_container, key_name, s3_bucket, new_name)
+        print("Would run this command: %s.       ← FIXME test it and run it" % command)
+        return
+        subprocess.Popen(command, shell=True)
+
+    def _migrate_swift_to_s3(self, instance):
+        # FIXME Will these 2 values already be stored in the instance? If not, where to get them from? (Or should we set them and save?)
+        # FIXME and in particular: will the S3 region match the SWIFT region? (some clients don't want their data in a particular region; see comments on ticket)
```
rocioar: No, there is no support for specifying region now, we will have to skip cloudera this time.

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+        Copy subdirectory using rclone
+        """
+        LOG.info("%s ---> %s" % (key_name, new_name))
+        command = "rclone copy ocim-swift:%s/%s ocim-s3:%s/%s" % (swift_container, key_name, s3_bucket, new_name)
+        print("Would run this command: %s.       ← FIXME test it and run it" % command)
+        return
+        subprocess.Popen(command, shell=True)
+
+    def _migrate_swift_to_s3(self, instance):
+        # FIXME Will these 2 values already be stored in the instance? If not, where to get them from? (Or should we set them and save?)
```
rocioar: you mean the iam keys and the s3 bucket? No, they will need to be created. You can do so by running:

```python
instance.s3_bucket_name = instane.bucket_name
instance.create_iam_user()
instance._create_bucket()
instance.save()
```

There has to be some wait time after creating the iam keys, before creating the bucket, cause the IAM keys don't get propagated immediately, you can add some seconds of sleep time here.

Also instance.storage_type has to be updated after doing the migration to s3, and a new appserver needs to be spawned for it to take the new bucket.

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+            if key_name in copied:
+                continue
+
+            # replace new name if they contain submissions_attachments (there is a bug on swift implementation that makes
+            # files be uploaded to the wrong directory)
+            # The changes to do are like:
+            #   submissions_attachmentsbadges/ ---> submissions_attachments/badges/
+            #   submissions_attachmentsuser_tasks ---> submissions_attachments/user_tasks/
+            # etc.
+            # FIXME what about submissions_attachmentssubmissions_attachments? In OC-4395 we needed: rclone-sync someserver/submissions_attachmentssubmissions_attachments/ someserver/submissions_attachments/
```
rocioar: @clemente not sure what you mean with this comment, could you explain a bit more?

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+        self.options = options
+
+        instances = OpenEdXInstance.objects.all()  # FIXME filter a little (e.g. exclude old/archived or those already in S3)
+
+        # Limit to a subset. Debugging only
+        instances = instances.filter(ref_set__name__contains='SWIFT') # FIXME disable
+
+        LOG.info("Will migrate %i instances", instances.count())
+        for instance in instances:
+            # FIXME add some failure handling to display errores and revert things if one fails (e.g. delete the rclone config, etc.)
```
rocioar: this is probably a good place to create the iam keys, and bucket

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+    def handle(self, *args, **options):
+        """
+        Finds all instances and migrate them.
+        """
+        self.options = options
+
+        instances = OpenEdXInstance.objects.all()  # FIXME filter a little (e.g. exclude old/archived or those already in S3)
+
+        # Limit to a subset. Debugging only
+        instances = instances.filter(ref_set__name__contains='SWIFT') # FIXME disable
```
rocioar: we also have to filter by storage_type='swift'

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+# Imports #####################################################################
+
+import re
+import subprocess
+import logging
+
+from django.core.management.base import BaseCommand
+
+from instance.models.openedx_instance import OpenEdXInstance
+import swiftclient
```
rocioar: nit: this is a third party import, can it be between django import and the standard library imports?

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+    def _create_rclone_config(self, access_key, secret_key):
+        """
+        Creates rclone config, to ease the process of copying
+        """
+        rclone_config = "rclone config create ocim-s3 s3 access_key_id %s secret_access_key %s" % (access_key, secret_key)
+        print("Would run this command: %s.       ← FIXME test it and run it" % rclone_config)
+        return
+
+        subprocess.Popen(rclone_config % (access_key, secret_key), shell=True)
+
```
rocioar: nit: leave only 2 spaces between methods.

instance/management/commands/migrate_swift_to_s3.py
```diff
@@ -0,0 +1,135 @@
...
+        copied = []  # FIXME describe variable
+        swift = self._get_swift_connection(instance)
+
+        container = instance.swift_container_name
+        s3_bucket = instance.s3_bucket_name  # FIXME will this instance already have the right bucket name? If not, we must set it and save the instance
+        LOG.info("Migrating container: %s" % container)
+        for item in swift.get_container(container)[1]:
+            LOG.info("Seeing file: %s", item['name'])
+
+            # FIXME test this part; will it also copy files in the root? (in addition to directories):
```
rocioar: Agree that this is something to test, I think it will copy them, and if it does I don't think it's needed to filter by directory as you are doing below

instance/management/commands/migrate_swift_to_s3.py
smarnach: How much work would it be to hard-code the region to Paris?  Our instances are hosted in France, so it would make sense to have the object storage in France as well, and that would also work for Cloudera.

instance/management/commands/migrate_swift_to_s3.py
smarnach: It would be nice to be able to pass a custom filter on the command line.  We may have different maintenance windows for different instances, since we need to coordinate the migration with clients.  If you want to get fancy, you can copy the logic for the `--filter` and `--exclude` parameters from the `instance_redeploy` management command, but it would be enough to accept a single instance id on the command line, and only migrate a single instance at a time.  We could then create the list of instance ids manually and use a shell loop to migrate multiple instances at once.  For paid instances, we may want to migrate them individually anyway, to make it easier to react to problems.

instance/management/commands/migrate_swift_to_s3.py
rocioar: Shouldn't be much work to hardcode that.

smarnach: Then let's do this please.  It would allow us to completely drop Swift support.  Swift has required significant maintenance to keep working in the past, because it's not tested by upstream, and I would like to get rid of that maintenance effort.  Moreover, we already had at least two customers requiring storage within the EU, so storing the data in Paris by default seems like a good idea.

instance/management/commands/migrate_swift_to_s3.py
clemente: I meant the „submissions_attachment“ directory itself (or its contents), it seems it's a special case (at least in Jill's example it was a special case requiring an extra line). Maybe the files inside `submissions_attachmentssubmissions_attachments` must be moved to `submissions_attachments`.
I should check it.

instance/management/commands/migrate_swift_to_s3.py
clemente: Would that mean: the same S3 user and password for all instances?

instance/management/commands/migrate_swift_to_s3.py
clemente: @rocioar @smarnach is this script also responsible for spawning a new server?
In that case, the script should be run carefully, and do much better error handling, and use some tags etc.
And run either with few instances, or interactively („press enter to continue“). Or use tags that later can be used by instance_redeploy (in case nothing else needs to be done with the servers).

Yes, it makes sense to change .storage_type to s3; it seems the script should do this.

instance/management/commands/migrate_swift_to_s3.py
rocioar: Good catch! I missed that, and it seems yes, it was a special case, and looking at the command she run it seems they have to be moved to just submissions_attachments/

instance/management/commands/migrate_swift_to_s3.py
clemente: (Yes, same user/pass).

However, there's a problem: if I try to upload things to a bucket without knowing the region, I get errors like `BucketRegionError: incorrect region, the bucket is not in 'eu-central-1' region`. This means that either we must store the region of each instance's S3 (same as we store SWIFT's region) or we change some default in AWS web so that everything goes to Paris, *and* then we migrate the existing buckets from non-Paris to Paris so that all of them be in Paris (and then we can't have buckets in US).

I think we should store the region, but we need a new field for that, near `s3_bucket_name`; I can create a ticket.

DevanR:
@clemente There is quite a bit to take in here, but I can understand the the objective. Given the nature of the script, I think this looks ok. The only thing I would maybe add in the description is maybe what to do if the script fails. Is it indempotent? Is there any rollback instructions or do you just re-try again?

clemente:
@DevanR I added it:
```
- let the script copy everything, watch for errors, re-run if it fails (some AWS things take time and you may need
  to wait some seconds between re-runs). If it fails, it's safe to retry it, since it won't delete the old data:
  it will create S3 resources and copy data to S3, but still won't use it until you deploy new servers. You can
  safely remove the new S3 files and re-run the script in case something breaks in the middle, or just re-run to
  make rclone copy the missing files
```

The tests were blocked some days ago. Now it's fixed. I re-based on master and I think this time the tests will pass; I'll merge when it does. Thanks.

clemente:
Tests pass. Merged

PR #330, opened 2018-07-12
Fix s3 bucket creation wait more for IAM keys propagation, save IAM keys for instance

rocioar:
S3 bucket creation is failing due to Amazon's delay on propagating the keys. I increased the retry delay to fix this.

IAM keys were not being saved when s3 bucket creation failed, fixed this too.

### DIFF
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -221,6 +221,7 @@ class S3BucketInstanceMixin(models.Model):
         keys = key_response['create_access_key_response']['create_access_key_result']['access_key']
         self.s3_access_key = keys['access_key_id']
         self.s3_secret_access_key = keys['secret_access_key']
+        self.save()
 
     def get_s3_connection(self):
         """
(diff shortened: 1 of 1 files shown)


instance/models/mixins/storage.py
```diff
@@ -231,7 +232,7 @@ def get_s3_connection(self):
             self.s3_secret_access_key
         )
 
-    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=1):
+    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=2):
```
smarnach: Have you measured how long we typically need to wait?  If 4 attempts with a one-second interval are not enough, wouldn't it be better to at least increse the delay to 5 seconds?

rocioar: I will deploy to staging now, and test.

rocioar: tested on staging, went through without any retry.

clemente:
@rocioar You will need to pass the `retry_delay` as a parameter [here](https://github.com/open-craft/opencraft/blob/ca3e37dd44ed70e8bf6dc40cd5bbf0f822642a45/instance/models/mixins/storage.py#L252), otherwise it only waits the specified amount before the first attempt.

rocioar:
@clemente it will wait the 4 seconds anyway cause it's the default for the method, but you are right that we need to add it there too. Otherwise if we customize it, it won't take it for the following retries

PR #331, opened 2018-07-13
Don't merge: Add a Nomad jobspec for memcached jobs.

smarnach:
We probably won't need the spec in its current form, since we will create the jobs via the Nomad API.  I'm just adding it to the repo so it can be reviewed and used as a basis for the API call.

### DIFF
--- a/instance/templates/instance/memcached/memcached.nomad
+++ b/instance/templates/instance/memcached/memcached.nomad
@@ -1,0 +1,53 @@
+# Jobs need a globally consistent name, so we need to include the instance id.
+job "memcached-1234" {
+  datacenters = ["dc1"]
+  type = "service"
+  constraint {
+    distinct_hosts = true
+  }
+  update {
+    max_parallel = 1
+  }
+  migrate {
+    max_parallel = 1
+  }
+  group "memcached" {
+    count = 3    # Only 1 for stage
+    task "memcached" {
+      driver = "exec"
+      config {
+        command = "/usr/bin/memcached"
+        args    = [
+          # TCP port
+          "-p", "${NOMAD_PORT_cache}",
+          # Memory limit.  The daemon won't use more than the limit, but may use significantly less.
+          "-m", "64", # MB
+        ]
+      }
+      resources {
+        cpu    = 25 # MHz
+        # The daemon may grow bigger for some instances, capped at 64 MB, but will probably remain
+        # smaller for most sandboxes and instances, so we don't want to make excessive allocations.
+        memory = 32 # MB
+        network {
+          # Should be easily enough on average.  We don't want this to become a limiting factor for
+          # scheduling.
+          mbits = 1
+          port "cache" {}
+        }
+      }
+      # Register with Consul.
+      service {
+        name = "memcached-1234"
+        tags = ["memcached"]
+        port = "cache"
+        check {
+          name     = "alive"
+          type     = "tcp"
+          interval = "10s"
+          timeout  = "1s"
+        }
+      }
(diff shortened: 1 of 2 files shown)


instance/templates/instance/memcached/memcached.nomad
```diff
@@ -0,0 +1,53 @@
...
+    distinct_hosts = true
+  }
+  update {
+    max_parallel = 1
+  }
+  migrate {
+    max_parallel = 1
+  }
+  group "memcached" {
+    count = 3    # Only 1 for stage
```
UmanShahzad: Does this mean 3 memcached processes per instance? Wouldn't it be enough to go with 1, and have Nomad reschedule the job in case the node dies? (I'm assuming that's how things work).

instance/templates/instance/memcached/memcached.nomad
```diff
@@ -0,0 +1,53 @@
...
+    task "memcached" {
+      driver = "exec"
+      config {
+        command = "memcached"
```
UmanShahzad: Apparently the docs say this value should be absolute:

> If executing a binary that exists on the host, the path must be absolute.

But somehow this worked? (I don't know why it has to be absolute though TBH, that isn't explained).

instance/templates/instance/memcached/memcached.nomad
UmanShahzad: Oh wait, I realized that may not be practical because it'd mean reprovisioning the appserver with new settings since the ports will likely change.

smarnach: The goal is high availability.  With just a single process, the instance would be down while Nomad reschedules the process when it goes down, so we need at least two.  To still retain this resilience even during a rolling update or other maintenance work, three processes are what is generally recommended.  I'm fine with using either two or three processe, but I'd start with three.  If resource usage becomes an issue, we can reduce this to two.

The app server needs to be able to deal with changing ports.  The solution suggested in the original design is to use three fixed local ports on each app server that are dynamically forwarded to the correct ports via an iptables hack, with Consul watches in place to update the port forwards as needed.  However, now that Consul has the nice connect feature, we may be able to use that feature instead – I haven't looked into the details yet, but since this is exactly what it is meant to be used for, I'm optimistic that it will provide a nicer solution for the problem, with the added benefit of using TLS.

instance/templates/instance/memcached/memcached.nomad
smarnach: Obviously, this _isn't_ required to be an absolute path, but if that's what's documented, I'll change it.  (I've force-pushed the change over the old commit, so you can't see the diff, but it's the only thing I changed.)

UmanShahzad: Hmm, for some reason I don't actually see any difference though anyway o_0

instance/templates/instance/memcached/memcached.nomad
UmanShahzad: OK, yeah I get the HA part now, not sure why I thought 1 was going to be okay :)

And yeah the proxy-that-we-manage solution seems a little complex, so I also hope Connect works just fine here.

instance/templates/instance/memcached/memcached.nomad
smarnach: My bad, I changed the semantics of one of my git aliases before my vacation, and forgot about the change in the last few weeks.  Now it's _actually_ updated.

pomegranited:
Closing PR, unneeded now.

PR #333, opened 2018-07-15
Make notifier refresh the DB connection less frequently

clemente:
This sets a new value for the `NOTIFIER_POLLING_INTERVAL` variable introduced in https://github.com/open-craft/configuration/pull/49

The previous value was 5 seconds, which was too often (each server refreshed the DB connection every 5 seconds). The new value 60 seconds should reduce this, and is fast enough for what notifier does (sending e-mails once a day).

Testing:
1. Test https://github.com/open-craft/configuration/pull/49 with the instruction upstream
2. Try to create a server in Ocim and verify that the new variable `NOTIFIER_POLLING_INTERVAL` is added and set to 60
3. Tests should pass


tomaszgy:
@clemente The PR has one CircleCI test failing but it doesn't seem to have much to do with the code you've introduced. I started a rebuild of the CI to see if it solves the problem.

As `NOTIFIER_POLLING_INTERVAL` has been tested in configuration PRs, I'm going to give a conditional 👍 given the tests are green here.

Conditional :+1: (given the CI is green)

- [x] I tested this:
1. I tested https://github.com/open-craft/configuration/pull/49 and saw sandboxes provision with newly introduced variable.
- [x] I read through the code
~- [ ] I checked for accessibility issues~
~- [ ] Includes documentation~

clemente:
Ok, this can be merged just before starting the redeployment of all instances (OC-4954).

clemente:
@smarnach we can merge it, and new instances will use the new setting. But there's the part about checking that everything is working fine and that the DB load is reduced; and we did OC-4954 to make it in one go (measure, merge, redeploy all, measure). It's fine to merge it now and see the new settings progressively; then new instances will have interval 5 and old ones will have 60. Eventually (after an eventual redeployment) all will move to interval 60.

PR #327, opened 2018-07-06
Remove ephemeral databases support

viadanna:
This pull request removes support for ephemeral databases.

**Testing instructions**:

1. Spin up an OCIM devstack
2. Create an instance, checking that the `use_ephemeral_databases` option disappeared.
3. Spin up an AppServer for given instance.
4. Check that the AppServer makes use of external databases.

**Author notes and concerns**:

~1. Some unit tests are failing as the provided .env.test still does not have external servers configured. This will be fixed asap.~

**Reviewers**
- [ ] @itsjeyd 

**Changes included**
Modify models to remove field
Provides migration
Fix templates and serialization
Remove stale tests for ephemeral support
Fix remaining tests
Adjust README for given changes

PR #336, opened 2018-07-25
Add migrate_ephemeral management command

viadanna:
@itsjeyd 

I decided to extract the management command to a separate branch to make it easier to run the tests.

This PR is just for code reviewing, as I have moved from synchronous to async provisioning.

### DIFF
--- a/instance/tests/management/test_migrate_ephemeral.py
+++ b/instance/tests/management/test_migrate_ephemeral.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2016 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Instance - migrate_ephemeral unit tests
+"""
+from unittest.mock import patch
+
+from django.core.exceptions import FieldDoesNotExist
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils.six import StringIO
+from testfixtures import LogCapture
+
+from instance.models.openedx_instance import OpenEdXInstance
+
+
+class MigrateEphemeralTestCase(TestCase):
+    """
+    Test cases for the `migrate_ephemeral` management command.
+    """
+
+    def test_no_instances(self):
+        """
+        Verify that the command correctly notifies the user that there are no instances for migration.
+        """
+        try:
+            with LogCapture() as captured_logs:
+                call_command(
+                    'migrate_ephemeral',
+                    stdout=StringIO(),
+                )
+                # Verify the logs
+                self.assertIn(
+                    'Found "0" instances using ephemeral databases',
+                    [l[2] for l in captured_logs.actual()])
+        except FieldDoesNotExist:
+            # Field already removed from database
+            pass
+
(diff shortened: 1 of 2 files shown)


instance/tests/management/test_migrate_ephemeral.py
```diff
@@ -0,0 +1,67 @@
...
+            )
+            # Verify the logs
+            self.assertIn(
+                'Found "0" instances using ephemeral databases',
+                [l[2] for l in captured_logs.actual()])
+
+    @patch('instance.tasks.spawn_appserver')
+    def test_migrate(self, mock_spawn_appserver):
+        """
+        Verify that the command correctly notifies the user that there are no instances for migration.
```
itsjeyd: @viadanna Copy-pasta 🍝 :) This will need to be updated to mention the real purpose of this test.

instance/tests/management/test_migrate_ephemeral.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        OpenEdXInstance.objects.create(sub_domain='test_migrate', use_ephemeral_databases=True)
```
itsjeyd: @viadanna Nit: Can we set a name for this instance that the check in L64 can reference (instead of the generic "Instance")?

instance/tests/management/test_migrate_ephemeral.py
```diff
@@ -0,0 +1,67 @@
...
+        with LogCapture() as captured_logs:
+            call_command(
+                'migrate_ephemeral',
+                stdout=StringIO(),
+            )
+            # Verify the logs
```
itsjeyd: @viadanna In addition to the check below, I think it would be worth verifying that the management command found the right number of instances (i.e., check for the presence of a log entry that says:

> Found "1" instances using ephemeral databases

instance/tests/management/test_migrate_ephemeral.py
viadanna: 🤦‍♂️

instance/tests/management/test_migrate_ephemeral.py
viadanna: Sure 👍

instance/tests/management/test_migrate_ephemeral.py
viadanna: Alright.

instance/tests/management/test_migrate_ephemeral.py
```diff
@@ -0,0 +1,71 @@
...
+                'Found "0" instances using ephemeral databases',
+                [l[2] for l in captured_logs.actual()])
+
+    @patch('instance.tasks.spawn_appserver')
+    def test_migrate(self, mock_spawn_appserver):
+        """
+        Verify that the command correctly migrate an instance to use external databases.
+        """
+        OpenEdXInstance.objects.create(
+            sub_domain='test_migrate', use_ephemeral_databases=True, name='test_migrate instance')
```
itsjeyd: @viadanna To make this test pass with the changes from #327 in place, could you just restructure the body using a try-except-else block that checks for `FieldError`? Like this:

```python
    @patch('instance.tasks.spawn_appserver')
    def test_migrate(self, mock_spawn_appserver):
        """
        Verify that the command correctly migrate an instance to use external databases.
        """
        try:
            OpenEdXInstance.objects.create(
                sub_domain='test_migrate', use_ephemeral_databases=True, name='test_migrate instance'
            )
        except FieldError:
            print('`use_ephemeral_databases` field no longer exists. Skipping migration test.')
        else:
            with LogCapture() as captured_logs:
                call_command(
                    'migrate_ephemeral',
                    stdout=StringIO(),
                )
                # Verify the logs
                actual = set(l[2] for l in captured_logs.actual())
                expected = {
                    'Found "1" instances using ephemeral databases',
                    'Migrated and started provisioning a new app server for test_migrate instance',
                }
                self.assertTrue(expected <= actual)
    
            self.assertTrue(mock_spawn_appserver.called)
```

You could also use [`skipUnless`](https://docs.python.org/3/library/unittest.html#unittest.skipUnless) together with a check that provides information about whether the `use_ephemeral_databases` field exists ([example](https://stackoverflow.com/q/29034721/1199226)). But that would probably require a bit more effort than the approach mentioned above.

viadanna: Yeah, I actually had the tests checking for FieldDoesNotExist but it felt hacky. That's why I created a separate branch, but I'll will revert to that.

PR #334, opened 2018-07-20
Fix playbooks repo typo in README.md

ahmedaljazzar:
This is to fix a typo in README.md file which leads to a 404 page when trying to access the [playbooks](https://github.com/open-craft/ansible-playbooks) repo.

PR #338, opened 2018-07-26
Store instance information in Consul

ahmedaljazzar:
This change will help us store and update instance information in Consul. It introduces a mixin that can be used in any model we desire to add this feature to.

### WIP
- [x] Management Commands.
- [x] Tests.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -58,6 +58,14 @@ jobs:
             sudo apt-get install mysql-client
             pip install -r requirements.txt
             sudo apt-get install postgresql-client
+            sudo apt-get install unzip
+            sudo wget -P /tmp https://releases.hashicorp.com/consul/1.2.1/consul_1.2.1_linux_amd64.zip
+      - run:
+          name: Run Consul
+          command: |
+            sudo unzip /tmp/consul_1.2.1_linux_amd64.zip -d /usr/local/bin
+            consul agent -dev
+          background: true
       - save_cache:
           key: dependencies-{{ checksum "requirements.txt" }}
           paths:
--- a/instance/management/commands/update_metadata.py
+++ b/instance/management/commands/update_metadata.py
@@ -1,0 +1,138 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2018 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Instance app - instances' metadata update management command
+"""
+
+# Imports #####################################################################
+
+import sys
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import consul
+
+from instance.models.openedx_instance import OpenEdXInstance
+
+
+# Classes #####################################################################
+
+
+class Command(BaseCommand):
+    """
+    management command class for updating instances metadata
+    """
+    help = 'Updates the Consul metadata for all instances or just a single instance.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--skip-clean',
+            action='store_true',
+            help='Will skip cleaning metadata for archived instances.'
+        )
+        parser.add_argument(
+            '--skip-update',
+            action='store_true',
+            help='Will skip updating metadata for instances.'
+        )
+
+    def handle(self, *args, **options):
+        if not settings.CONSUL_ENABLED:
+            self.stdout.write(self.style.WARNING(
+                'This command does nothing unless you enable Consul in the configuration. Exiting...'
+            ))
+            sys.exit(0)
+
+        if not options.get('skip_update'):
+            self.update_all_instances_metadata()
+
+        if not options.get('skip_clean'):
+            self.clean_consul_metadata()
+
+    def update_all_instances_metadata(self):
+        """
+        This method will iterate over all non-archived instances to update
+        their metadata in Consul.
+        """
+        instances = self.get_running_instances()
+
+        self.stdout.write('Updating {} instances\' metadata...'.format(instances.count()))
+        for instance in instances:
+            instance.update_consul_metadata()
+
+        self.stdout.write(self.style.SUCCESS('Successfully updated instances\' metadata'))
+
+    def clean_consul_metadata(self):
+        """
+        This method will iterate over all archived instances and clean their
+        metadata from Consul.
+        """
+        client = consul.Consul()
+        instances_ids = self.get_archived_instances()
+        self.stdout.write('Cleaning metadata for {} archived instances...'.format(len(instances_ids)))
+
+        for instances_id in instances_ids:
+            prefix = settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=instances_id)
+            client.kv.delete(prefix, recurse=True)
+        self.stdout.write(self.style.SUCCESS('Successfully cleaned archived instances\' metadata'))
+
+    @staticmethod
+    def get_running_instances():
+        """
+        Generates a queryset of the active instances. Didn't put it as a
+        class variable to avoid multiple processes mutations.
+
+        :return: A Queryset of all active OpenEdXInstances
+        """
+        return OpenEdXInstance.objects.filter(ref_set__is_archived=False)
+
+    def get_archived_instances(self):
+        """
+        This management command gets the list of instance ids in Consul and the
+        non-archived instance ids from the database. Each instance only occurring
+        in Consul, but has been archived since should be considered archived.
+
+        :return: A set of all archived OpenEdXInstances
+        """
+        agent = consul.Consul()
+        archived_instances_ids = set()
+
+        instances_prefix = '{ocim}/instances/'.format(ocim=settings.OCIM_ID)
+        _, consul_instances_keys = agent.kv.get(instances_prefix, recurse=True, keys=True)
+
+        if not consul_instances_keys:
+            self.stdout.write('Consul does not contain data yet')
+            return archived_instances_ids
+
+        running_instances = self.get_running_instances()
+        running_instances_ids = running_instances.values_list('id', flat=True)
+
+        for key in consul_instances_keys:
+            id_elements = key.split('/')
+            try:
+                instance_id = int(id_elements[2])
+            except (IndexError, ValueError):
+                self.stdout.write(self.style.WARNING('A Consul key found with an unknown structure: {}'.format(key)))
+                continue
+
+            if instance_id not in running_instances_ids:
+                archived_instances_ids.add(instance_id)
+
+        return archived_instances_ids
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -24,6 +24,7 @@ import string
 from django.conf import settings
 from django.db import models, transaction
 from django.db.backends.utils import truncate_name
+from django.db.models import F
 from django.template import loader
 from django.utils import timezone
 
@@ -41,7 +42,7 @@ from instance.models.mixins.openedx_storage import OpenEdXStorageMixin
 from instance.models.mixins.openedx_theme import OpenEdXThemeMixin
 from instance.models.mixins.secret_keys import SecretKeyInstanceMixin
 from instance.models.openedx_appserver import OpenEdXAppConfiguration
-from instance.models.utils import WrongStateException
+from instance.models.utils import WrongStateException, ConsulAgent
 from instance.utils import sufficient_time_passed
 
 
@@ -101,7 +102,9 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
             self.edx_platform_commit = self.openedx_release
         if self.storage_type is None:
             self.storage_type = settings.INSTANCE_STORAGE_TYPE
+
         super().save(**kwargs)
+        self.update_consul_metadata()
 
     def get_load_balancer_configuration(self, triggered_by_instance=False):
         """
@@ -354,7 +357,8 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
 
     def archive(self):
         """
-        Shut down this instance's app servers and mark it as archived.
+        Shut down this instance's app servers, mark it as archived and
+        remove its metadata from Consul.
         """
         self.disable_monitoring()
         self.remove_dns_records()
@@ -365,6 +369,7 @@ class OpenEdXInstance(DomainNameInstance, LoadBalancedInstance, OpenEdXAppConfig
             self.reconfigure_load_balancer(load_balancer)
         for appserver in self.appserver_set.iterator():
             appserver.terminate_vm()
+        self.purge_consul_metadata()
         super().archive()
 
     @staticmethod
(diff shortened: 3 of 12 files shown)


instance/models/mixins/consul.py
```diff
@@ -0,0 +1,73 @@
...
+                 changes or not.
+        """
+        agent = consul.Consul()
+        includes_changes = False
+
+        for field in self._meta.get_fields():
+            value = getattr(self, field.name)
+
+            index, _ = agent.get(field.name)
+            changed = agent.put(field.name, value, cas=index)
```
smarnach: We need to include the path `ocim/instances/{instance_id}/` in the key, since otherwise the instances would override each others data.

instance/models/mixins/consul.py
```diff
@@ -0,0 +1,73 @@
...
+        updated the field then we update the configuration version
+        number in Consul.
+
+        :return: The current version number, and the if they include
+                 changes or not.
+        """
+        agent = consul.Consul()
+        includes_changes = False
+
+        for field in self._meta.get_fields():
```
smarnach: You chose to simply store all fields in Consul, rather than the data listed in the ticket.  This is definitely an approach worth considering, but I'm not sure it's going to work well, since we also need to store some data that comes from related models, e.g. the list of active app servers.

instance/models/mixins/consul.py
ahmedaljazzar: Didn't catch that. Thanks @smarnach

ahmedaljazzar:
This is ready for another pass, @smarnach.

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
         self.deprovision_rabbitmq()
         super().delete(*args, **kwargs)
+
+    def update_consul_metadata(self):
+        """
+        This method is going over all fields in this model to reflect
+        their values on Consul.
+        Values on consul will be fetched first, then updated only if
+        they changed in this version of the model using agent's `cas`
+        parameter (Check-And-Set). If PUT operation successfully
```
smarnach: It appears this isn't accurate anymore.  Is there any reason you abandoned that approach?

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
+            'domains': str(self.get_load_balanced_domains()),
+            'health_checks_enabled': str(int(active_servers_num > 1)),
+            'basic_auth': self.http_auth_info_base64(),
+            'active_app_servers': str([
+                server.name for server in self.get_active_appservers()
+            ]),
+        }
+
+        # Reflect new configurations on Consul
+        includes_changes = True
```
smarnach: Shouldn't this be `False`?  Otherwise I can't see how this will ever have any other value than `True`.

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
+        _, consul_conf = agent.kv.get(consul_prefix, recurse=True)
+        consul_conf = {
+            conf['Key']: conf['Value']
+            for conf in consul_conf
+        } if consul_conf else {}
```
smarnach: I also liked the approach you originally had, using check-and-set, which may work without first reading all the data.  If you get that approach working, that would be great, too.

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
+
+        for key, value in new_conf.items():
+            consul_key = consul_prefix + key
+            # This will handle the not-exist case
+            if consul_conf.get(consul_key) != new_conf[key]:
```
smarnach: Instead of looking up `key` in `new_conf`, you could simply use `value`.

opencraft/settings.py
```diff
@@ -52,6 +52,10 @@
 ENABLE_DEBUG_TOOLBAR = env.bool('ENABLE_DEBUG_TOOLBAR', default=False)
 
 
+# Consul #########################################################################
+CONSUL_ENABLED = env.bool('CONSUL_ENABLED', default=False)
+OCIM_ID = env('OCIM_ID', default='ocim-1')
```
smarnach: Let's make the default just `ocim`.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+
+        instance.update_consul_metadata()
+        self.stdout.write(self.style.SUCCESS(
+                'Successfully update instances {}\' metadata'.format(
+                    instance_id)))
+
+    def update_all_instances_metadata(self):
+        self.stdout.write('Updating all instances\' metadata...')
+        for instance in OpenEdXInstance.objects.all():
+            instance.update_consul_metadata()
```
smarnach: This does not remove archived instances from the list in Consul – see my comment on the ticket for more details.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+            instance = OpenEdXInstance.objects.get(id=instance_id)
+        except:
+            self.stdout.write(self.style.ERROR(
+                'Couldn\'t find an instance with id: {}'.format(
+                    instance_id)))
+            sys.exit(1)
+
+        instance.update_consul_metadata()
+        self.stdout.write(self.style.SUCCESS(
+                'Successfully update instances {}\' metadata'.format(
```
smarnach: Suggested text: `'Successfully updated metadata for instance {}.'`

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+# Classes #####################################################################
+
+
+class Command(BaseCommand):
+    """
+    management command class for updating instances metadata
+    """
+    help = 'Updates the Consul metadata for all instances.'
```
smarnach: …or just a single instance.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Instance app - Activity report management command
```
smarnach: Copy-and-paste error – please update.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def update_instance_metadata(self, instance_id):
+        self.stdout.write('Updating instance {}\' metadata...'.format(
+            instance_id))
+
+        try:
+            instance = OpenEdXInstance.objects.get(id=instance_id)
+        except:
+            self.stdout.write(self.style.ERROR(
+                'Couldn\'t find an instance with id: {}'.format(
```
smarnach: Style note: We generally format the code to somewhat longer lines.  The linter complains only at 120 characters; you don't need to use long lines, of course, but you can if you think it improves readability.  If we do break lines, we generally put the closing parens on a separate line, in line with the edX Python style guide.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+    def handle(self, *args, **options):
+        instance_id = options.get('instance')
+
+        if instance_id:
+            self.update_instance_metadata(instance_id)
+        else:
+            self.update_all_instances_metadata()
+
+    def update_instance_metadata(self, instance_id):
+        self.stdout.write('Updating instance {}\' metadata...'.format(
```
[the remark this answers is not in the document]
ahmedaljazzar: It's a possessive apostrophe, I'll emit it now.

instance/management/commands/update_metadata.py
ahmedaljazzar: I used to use 72 characters per line. Just edited the IDE to break on 120 characters.

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
+        :note: This still doesn't apply removed-configurations case.
+
+        :return: The current version number, and the if they include
+                 changes or not.
+        """
+        if not settings.CONSUL_ENABLED:
+            return 0, False
+
+        agent = consul.Consul()
+        active_servers_num = len(self.get_active_appservers())
```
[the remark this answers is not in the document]
ahmedaljazzar: Good catch 👍. I also just noticed that `get_active_appservers` actually returns a Queryset. So I'll use `active_servers.count()` instead of `len(active_servers)`

instance/models/openedx_instance.py
ahmedaljazzar: Outdated documentation, yes. I didn't (some how) abandoned the approach, but instead, I handled it in Python instead of Consul. The idea here now instead of getting the keys one-by-one which hits the consul multiple times, I fetched them all at once, and did Check-And-Set in a python for-loop that compares the checked value with the currently calculated one.

ahmedaljazzar: This is also gonna be helpful in removing unnecessary fields that might go stale.

instance/models/openedx_instance.py
ahmedaljazzar: Debugging leftovers.

instance/models/openedx_instance.py
```diff
@@ -392,3 +396,63 @@ def delete(self, *args, **kwargs):
...
+        consul_prefix = '{ocim}/instances/{instance}/'.format(
+            ocim=settings.OCIM_ID, instance=self.pk)
+
+        # Collect the current version's configurations
+        new_conf = {
+            'id': str(self.pk),
+            'ocim': settings.OCIM_ID,
+            'name': self.name,
+            'domains': str(self.get_load_balanced_domains()),
```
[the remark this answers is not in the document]
smarnach: Yes, that's what I meant.

smarnach:
@ahmedaljazzar Thanks for the updates – I'll look into the details later.  The code still has a few linter violations – see the failed CircleCI tests for details.

ahmedaljazzar:
Yes, just started working on tests :D

ahmedaljazzar:
@smarnach, this is ready for another pass.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        parser.add_argument(
+            '--skip-clean',
+            action='store_true',
+            help='Will skip cleaning metadata for archived instances.'
+        )
+        parser.add_argument(
+            '--skip-update',
+            action='store_true',
+            help='Will skip cleaning metadata for archived instances.'
```
smarnach: Copy-paster error – please update the help string.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+    def handle(self, *args, **options):
+        if not settings.CONSUL_ENABLED:
+            self.stdout.write(self.style.WARNING('Consider Enable Consul before using this command.'))
```
smarnach: Suggested text:

> This command does nothing unless you enable Consul in the configuration. Exiting.

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+
+        instance_id = options.get('instance')
+
+        if not options.get('skip_update'):
+            if instance_id:
+                self.update_instance_metadata(instance_id)
+            else:
+                self.update_all_instances_metadata()
+
+        if not options.get('skip_clean'):
```
smarnach: I think it's somewhat unexpected that old data of unrelated instances will be cleaned when specifying an instance id on the command line.

(There isn't a strong need for all the command line flags anyway – in general, we always want all metadata to be up to date.  So it's a bit hard to argue about the semantics without having a use case in mind.)

instance/management/commands/update_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+        This method will iterate over all non-archived instances to update
+        their metadata in Consul.
+        """
+        instances = self.get_running_instances()
+
+        self.stdout.write('Updating {} instances\' metadata...'.format(instances.count()))
+        for instance in instances:
+            instance.update_consul_metadata()
+
+        self.stdout.write(self.style.SUCCESS('Successfully update instances\' metadata'))
```
smarnach: Should be "updated".

instance/models/openedx_instance.py
```diff
@@ -389,3 +394,117 @@ def delete(self, *args, **kwargs):
...
+    def _generate_consul_configurations(self):
+        """
+        Collects required configurations for this instance to reflect on Consul.
+        You can add, delete or alter configurations keys and values here which are
+        going to ve reflected in Consul immediately.
+
+        :return: A dict of the configurations.
+        """
+        active_servers = self.get_active_appservers()
+        active_servers_names = [server.name for server in active_servers]
```
smarnach: The information we need for each app server is the id (`appserver.pk`) and the public IP address (`appserver.server.public_ip`).  We eventually want to build a load balancer configuration from this information.  See [this code](https://github.com/open-craft/opencraft/blob/d5393bcda6fb26395357efb7ebbfc01eddf1bce5/instance/models/openedx_instance.py#L119-L131) for additional context.  I think a list of dictionaries would work here.

As a performance note, the current code will query the database for all appserver fields, build an `OpenEdXAppServer` instance for each of them and then extracts the `name` field.  A more efficient solution that only retrieves the required fields from the database is to use the `values()` methos, e.g.

    active_server_data = active_servers.values('pk', 'server__public_ip')

(I didn't verify the result for this particular example, but in general this results in close to optimal SQL queries, without making the code less readable.)

smarnach: Another side note:  We are trying to keep the information in Consul in sync with the information in the database.  Both the id and the IP address of an app server don't ever change, so we don't need any provisions to keep them in sync; we only need to update the list if an app server gets activated or deactivated, so the code we have should be enough.

instance/models/openedx_instance.py
```diff
@@ -389,3 +394,117 @@ def delete(self, *args, **kwargs):
...
+        """
+        This is a property that helps determining a specific instance's rpefix in
+        Consul. The prefix helps in keeping different instances' configurations
+        separate from overridings and modifications.
+
+        :return: A unique Key-Value prefix for this instance in Consul.
+        """
+        return '{ocim}/instances/{instance}/'.format(ocim=settings.OCIM_ID, instance=self.pk)
+
+    def _generate_consul_configurations(self):
```
smarnach: Nit: I'd prefer `_generate_consul_metadata` for more consistent naming.

instance/models/openedx_instance.py
```diff
@@ -389,3 +394,117 @@ def delete(self, *args, **kwargs):
...
+        _, version_data = agent.kv.get(version_key)
+        version_number = int(version_data.get('Value')) if version_data else 0
+
+        for key, value in configurations.items():
+            consul_key = self.consul_prefix + key
+            index, version_data = agent.kv.get(consul_key)
+            cas = index if version_data else 0
+            agent.kv.put(consul_key, value, cas=cas)
+
+            is_updated = not self.compare(value, version_data['Value']) if version_data else True
```
smarnach: I find it hard to understand what exactly is going on here.  I don't like the `OpenEdXInstance.compare()` static method, since I think its name doesn't give any hint as to what it does.  If anything, this should be a function in a utils module.  But prefereably, we don't need this function altogether.  Can't we convert all versions to numbers, and then compare them?

instance/management/commands/update_metadata.py
ahmedaljazzar: You are right, this is why I left it like this without adding a check for the instance flag, and I also thought that having Consul clean is good in all cases. However, it's really confusing to see that you're touching multiple instances if you specified a single instance to work with. I'll update it now.

ahmedaljazzar: I removed the support for a single instance in this command, because a user who's updating the metadata for an instance doesn't expect the data to go through the purge functionality and vice versa. Until we have a real need for it with a clear requirement, we can re-do it again.

instance/models/openedx_instance.py
ahmedaljazzar: I wrote this method to manage the expectations of getting binary/str value back from the agent to allow us comparing mixed types together. I removed this unclear code, and created a helper class that gets and puts the values as expected.

instance/models/openedx_instance.py
```diff
@@ -389,3 +394,97 @@ def delete(self, *args, **kwargs):
...
+        active_servers = self.get_active_appservers()
+        basic_auth = self.http_auth_info_base64()
+        enable_health_checks = int(active_servers.count() > 1)
+        active_servers_data = list(active_servers.annotate(public_ip=F('server___public_ip')).values('id', 'public_ip'))
+
+        configurations = {
+            'name': self.name,
+            'domains': self.get_load_balanced_domains(),
+            'health_checks_enabled': enable_health_checks,
+            'basic_auth': basic_auth.decode('utf-8'),
```
smarnach: Just as a side note, the UTF-8 encoding is the default in Python 3, so just `decode()` will do the trick.  In some cases it may be desirable to be explicit about the encoding anyway, but in this particular case the encoding is rather irrelevant.

circle.yml
```diff
@@ -58,6 +58,20 @@ jobs:
             sudo apt-get install mysql-client
             pip install -r requirements.txt
             sudo apt-get install postgresql-client
+            sudo apt-get install unzip
+            sudo wget -P /tmp https://releases.hashicorp.com/consul/1.2.1/consul_1.2.1_linux_amd64.zip
+      - run:
+          name: Wait for Consul
+          command: |
+            while [ ! -f /tmp/consul_1.2.1_linux_amd64.zip ]
```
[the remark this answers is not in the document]
ahmedaljazzar: I'm not sure, will try to remove it and see.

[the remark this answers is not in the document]
ahmedaljazzar: Yes, you are right. I didn't know that.

ahmedaljazzar:
Done with the changes. For the tests, I don't exactly what's happening, I had the same issue in w/ #334 where it's fixed when someone from the organization pinged the tests. I still don't know if that's the reason or not, but I'll ping the firefighter.

ahmedaljazzar:
@smarnach, I'm done with the changes. For the tests, I don't exactly what's happening, I had the same issue in w/ #334 where it's fixed when someone from the organization pinged the tests. I still don't know if that's the reason or not, but I'll ping the firefighter.

ahmedaljazzar:
@smarnach After some debugging, the issues was that the instance ref's name in the tests is hardcoded, so I made it dynamic and pushed the changes. The tests are passing now on circle.

ahmedaljazzar:
It also seems that I don't have an access to the deployment document you shared with me, so maybe you can give me one? Or should we wait to see who can deploy it?

smarnach:
@ahmedaljazzar Thanks for addressing the comments!  I tested the changes on stage, and everything is looking good to me.  The management command is working fine, and activating and deactivating app servers results in the expected changes in Consul.  Thanks a lot for seeing this through!

PR #340, opened 2018-08-08
Add configuration variables for enabling forum heartbeat

lgp171188:
This PR adds the default configuration variables for enabling forum heartbeat for OCIM installs. https://github.com/open-craft/edx-platform/pull/135 contains the backported changes for the heartbeat checks from the upstream master branch to the `opencraft-release/ginkgo.1` branch. It has to be applied to all client ginkgo branches before redeploying all the managed appservers running code from those branches.


clemente:
Since this is the same feature, could you squash commits? Please rebase too.

:+1:

  - [X] I tested this:
    - the right variables are set up
    - HEARTBEAT_EXTENDED_CHECKS doesn't include the problematic `check_celery` (which was included by default but failed due to an import; see below)
    - setting up these variables, the server has the forum heartbeat enabled (I checked the one already spawned in Ocim)
    - if the forum heartbeat fails, the whole heartbeat returns 503 (and some JSON)
  - [X] I read through the code
  - [X] Includes documentation

Attention: only merge this after the „customizable heartbeats“ patch (https://github.com/open-craft/edx-platform/pull/135/) has been merged *to all our branches* in Ocim.
Otherwise, the next time we spawn a server, New Relic will add a monitor to access the *extended heartbeat* URL and it will fail in servers that don't have that patch (PR 135), because they don't have the right *extended heartbeat* settings, they have a failing one (`check_celery`). The extended heartbeat will fail with this error:
```
{u'utc': True, u'chord': None, u'args': [], u'retries': 0, u'expires': u'2018-08-12T22:24:44.534963', u'task': u'openedx.core.djangoapps.heartbeat.tasks.sample_task', u'callbacks': None, u'errbacks': None, u'timelimit': [None, None], u'taskset': None, u'kwargs': {}, u'eta': None, u'id': u'd4f3b60f-f830-4470-9b98-269f692c9223'} (312b)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/celery/worker/consumer.py", line 455, in on_task_received
    strategies[name](message, body,
KeyError: u'openedx.core.djangoapps.heartbeat.tasks.sample_task'
```
and then New Relic would alert about a failed heartbeat.
This PR also passes an ansible variable to set the right heartbeat settings, but the ansible variable won't be read until https://github.com/open-craft/edx-platform/pull/135/ is merged.

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -152,10 +153,11 @@ def test_update_monitoring_additional_email(self, mock_newrelic):  # pylint: dis
         'additional_monitoring_emails' to the existing monitors.
         """
         instance = OpenEdXInstanceFactory()
-        existing_monitor_ids = [str(uuid4()) for i in range(3)]
+        existing_monitor_ids = [str(uuid4()) for i in range(4)]
         existing_monitors = {}
-        existing_monitor_urls = [instance.url, instance.studio_url, instance.lms_preview_url]
-        for i in range(3):
+        existing_monitor_urls = [instance.url, instance.studio_url, instance.lms_preview_url,
+                                 instance.lms_extended_heartbeat_url]
```
rocioar: @lgp171188 nit picking here, we usually split lists like this:

```
existing_monitor_urls = [
    instance.url,
    instance.studio_url,
    instance.lms_preview_url,
    instance.lms_extended_heartbeat_url
]
```

lgp171188: @rocioar, fixed it with 803676c

lgp171188:
Rebasing on top of the latest changes in the master branch.

smarnach:
@lgp171188 Is there any reason we can't merge this PR?

smarnach:
@ahmedaljazzar I guess rebasing and merging this PR will be your job – it's part of [SE-191](https://tasks.opencraft.com/browse/SE-191).

PR #341, opened 2018-08-12
Add example Nomad jobspec for a Redis process.

smarnach:
This adds a simple Redis jobspec, which is virtually identical to the Memcache jobspec.  The main differences are naming and a count of 1 instead of 3 – as discussed on the ticket, we won't be able to use a highly available Redis deployment in this iteration.

This PR is not meant to be merged.  We will add the ability to automatically launch the jobs to Ocim in a future PR, which will likely build on this jobspec, but put it in a Python dictionary rather than a separate file.

I started a job using this spec on the production Nomad cluster to verify the spec.  Given that it is virtually identical to the Memcache spec, I don't think this needs separate end-to-end testing – simply sanity-checking the jobspec should be enough for review.

### DIFF
--- a/instance/templates/instance/redis/redis.nomad
+++ b/instance/templates/instance/redis/redis.nomad
@@ -1,0 +1,66 @@
+# Jobs need a globally consistent name, so we need to include the instance id.
+job "redis-1234" {
+  datacenters = ["dc1"]
+  type = "service"
+  constraint {
+    distinct_hosts = true
+  }
+  update {
+    max_parallel = 1
+  }
+  migrate {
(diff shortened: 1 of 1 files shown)


smarnach:
@UmanShahzad @taranjeet Thanks for reviewing!

> I was just wondering what is the reason of mixing cases in variable names like `NOMAD_PORT_redis`.

These variables are provided by Nomad, so I don't really have control over them.  See the [documentation on interpolation|https://www.nomadproject.io/docs/runtime/interpolation.html#environment-variables-] for further details.

PR #344, opened 2018-08-15
Add the missing Consul decorator to UpdateMetadataTestCase

ahmedaljazzar:
JIRA: [OC-4288](https://o/OC-4288)

This is to add the missing `skip_unless_consul_running` that makes the tests fail if the devstack doesn't contain Consul. This is not the ideal solution to continue with, but it's a work around until we have Consul up and running in the devstack.

Thanks @lgp171188 for pointing this out.


smarnach:
@ahmedaljazzar Thanks for fixing this!

PR #346, opened 2018-08-22
Link developers and organizations to instances

rocioar:
Main changes:

- Adds Organization model
- Adds github_handle and organization to UserProfile
- Adds organization owner and user creator to the instanceReference
- Updated watch PRs task to watch PRs for all users registered in OCIM on the Watched forks.

Reviewers:

@itsjeyd


userprofile/models.py
```diff
@@ -31,12 +31,34 @@
 
 # Models ######################################################################
 
+
+class Organization(ValidateModelMixin, TimeStampedModel):
+    """
+    Organizations model
+    """
+    name = models.CharField(max_length=255)
+    github_handle = models.CharField(unique=True, max_length=255)
```
itsjeyd: @rocioar This field isn't used anywhere at the moment, so I think it would be fine to make it optional. Right now I'm thinking that we'll mostly be using it for informational purposes (and won't strictly need it to be able to implement the first batch of access control and billing features). But that might change in the future.

On a related note, it would be good to establish a direct connection between watched forks and organizations (by changing `WatchedFork.organization` to point to the `Organization` model), to prepare for implementing the billing features. Right now, the only way for us to relate an organization to a watched fork would be via the `WatchedFork.organization` and `Organization.github_handle` fields. But that will break down if an organization (a/k/a a client) would like to use PR watch functionality for multiple GitHub organizations.

pr_watch/tasks.py
```diff
@@ -50,8 +52,15 @@ def watch_pr():
...
-            team_username_list = get_username_list_from_team(watched_fork.organization)
-            for pr in get_pr_list_from_usernames(team_username_list, watched_fork.fork):
+            usernames = list(
+                UserProfile.objects.filter(
+                    organization__pr_watch_active=True
+                ).values_list(
+                    'github_username',
+                    flat=True
+                )
+            )
```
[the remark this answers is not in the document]
rocioar: @itsjeyd one thing I'm wondering here, if we filter by only the organization that owns the fork, then wouldn't we be limiting the capabilities? Let's say we are watching "edx-platform" (the main repository), then one of us makes a PR, we are not the owners of the fork, so the OCIM wouldn't see it and there will not be any sandbox created.

userprofile/models.py
```diff
@@ -31,12 +31,34 @@
...
+    pr_watch_active = models.BooleanField(
+        default=True,
+        help_text="Is PR watch active for this organization?"
+    )
```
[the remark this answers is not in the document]
rocioar: Makes sense, thanks for the detailed explanation.

userprofile/models.py
rocioar: I made the field optional.

Make sense to update the WatchedFork to point to the new organization model, I will update it.

pr_watch/models.py
```diff
@@ -297,6 +300,10 @@ def update_instance_from_pr(self, pr):
             self.watched_fork.configuration_extra_settings,
             pr.extra_settings
         )
+        if not instance.ref.creator or instance.ref.owner:
```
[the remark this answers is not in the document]
rocioar: Good catch! fixed

instance/migrations/0106_auto_20180821_1605.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
rocioar: 👍 done

userprofile/migrations/0003_auto_20180821_1605.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
rocioar: 👍 done

pr_watch/tasks.py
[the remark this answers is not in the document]
itsjeyd: @rocioar Good question! If you look at the [list of watched forks](https://console.opencraft.com/admin/pr_watch/watchedfork/), you'll see that there are multiple ones that target `edx/edx-platform`. Each one of those maps `edx/edx-platform` to a different organization, but none of these organizations actually owns `edx/edx-platform`. Conceptually it doesn't matter who owns a given fork -- OCIM just needs to know which PRs against that fork it should create sandboxes for. And we store that information in the `organization` field of the `WatchedFork` model. In other words, the way to "read" a `WatchedFork` is: OCIM will create sandboxes for PRs against `WatchedFork.fork` for all members of `WatchedFork.organization` (*not* "`WatchedFork.organization` owns `WatchedFork.fork`") 🙂

pr_watch/models.py
```diff
@@ -56,11 +56,14 @@ class WatchedFork(models.Model):
...
-    organization = models.CharField(
-        max_length=200,
+    organization = models.ForeignKey(
+        Organization,
+        on_delete=models.CASCADE,
         db_index=True,
-        help_text=("Watched GitHub organization. E.g.: open-craft. PRs against the watched fork "
-                   "made by members of this organization will trigger a sandbox build"),
+        help_text=(
+            "Watched GitHub organization. E.g.: open-craft. PRs against the watched fork "
```
itsjeyd: @rocioar Could you update this to reflect the conceptual changes that we are making to the logic for watching PRs? You could simplify the first part to something like:

> Organization to watch. PRs against against the watched fork ..."

pr_watch/migrations/0009_organization_db_index.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
itsjeyd: @rocioar Would it be possible to merge this migration into the previous one? (`update_organization`)?

rocioar: @itsjeyd actually this was done because there was an issue with postgres when changing the type of a db_index from a char to an int, see: https://code.djangoproject.com/ticket/27860

pr_watch/models.py
rocioar: 👍  done

rocioar:
@itsjeyd I've addressed your feedback, and now fixed the issue with the buckets (reason why the build was failing, we were on the max amount of buckets that we could have).

pr_watch/migrations/0009_organization_db_index.py
itsjeyd: @rocioar Ah, OK. Totally fine to keep both of these migrations, then :)

pr_watch/models.py
```diff
@@ -61,7 +61,7 @@ class WatchedFork(models.Model):
         on_delete=models.CASCADE,
         db_index=True,
         help_text=(
-            "Watched GitHub organization. E.g.: open-craft. PRs against the watched fork "
+            "Organization to watch. PRs against against the watched fork."
```
itsjeyd: @rocioar Ack... You'll need to remove the full stop at the end of this line. Sorry if my [earlier comment](https://github.com/open-craft/opencraft/pull/346#discussion_r212939789) was confusing in that regard?

Also, changing the help text should generate a new migration?

rocioar: @itsjeyd my mistake here, I did it in a rush. Fixed it now, and updated the migrations too (squashed with the other migration)

pr_watch/models.py
```diff
@@ -53,11 +56,13 @@ class WatchedFork(models.Model):
...
-    organization = models.CharField(
-        max_length=200,
+    organization = models.ForeignKey(
+        Organization,
+        on_delete=models.CASCADE,
         db_index=True,
-        help_text=("Watched GitHub organization. E.g.: open-craft. PRs against the watched fork "
-                   "made by members of this organization will trigger a sandbox build"),
+        help_text=(
+            "Organization to watch. PRs against against the watched fork"
```
itsjeyd: @rocioar But now this is missing the second part of the sentence... To be completely clear, here's what I think this help text should say:

> Organization to watch. PRs against the watched fork made by members of this organization will trigger a sandbox build.

rocioar:
@itsjeyd fixed tests. Will deploy to stage once tests finish running.

pr_watch/migrations/0008_update_organization.py
```diff
@@ -0,0 +1,26 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='watchedfork',
+            name='organization'
+        ),
+        migrations.AddField(
+            model_name='watchedfork',
+            name='organization',
+            field=models.ForeignKey(null=True, help_text='Organization to watch. PRs against the watched fork made by members of this organization will trigger a sandbox build.', on_delete=django.db.models.deletion.CASCADE, to='userprofile.Organization'),
```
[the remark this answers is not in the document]
rocioar: This was failing on stage, because there was data for the index. So I had to alter the migrations, so that it first removes the organization field completely, and then adds it again.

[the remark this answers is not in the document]
itsjeyd: @rocioar Ah, OK -- so this migration (`pr_watch/migrations/0008_update_organization.py`) now takes care of removing the `organization` field from the `WatchedFork` model and adding it back`. What's the purpose of `pr_watch/migrations/0009_organization_db_index.py`, do we still need it?

PR #289, opened 2018-03-23
Set background colors for header/footer in hawthorn compatible way in openedx_theme mixin

akovari:
Set background colors for header/footer in hawthorn compatible way in openedx_theme mixin

**Dependencies**: https://github.com/edx/configuration/pull/4396

**Merge deadline**: None

**Testing instructions**:

1. Deploy a new instance and set theme colors

**Reviewers**
- [ ] clemente

Rebased d36cc1f77cdfc38f1b27cae1188c41d96f3f3d54 onto master.

PR #347, opened 2018-08-22
Ginkgo-compatible simple_theme changes for Hawthorn

lgp171188:
This PR rebases https://github.com/open-craft/opencraft/pull/289/ on top of the latest changes in the master branch and retains ginkgo-specific changes removed in that PR, for backward compatibility.

### DIFF
--- a/instance/models/mixins/openedx_theme.py
+++ b/instance/models/mixins/openedx_theme.py
@@ -72,6 +72,17 @@ class OpenEdXThemeMixin(models.Model):
                     "variable": "link-color",
                     "value": application.link_color,
                 },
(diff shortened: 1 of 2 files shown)


clemente:
The Ginkgo instance that @lgp171188 deployed has the right colors.
So if this works in Hawthorn too, we can merge. (Or if this blocks anything, merge now and do the Hawthorn-related changes in another PR).

clemente:
@lgp171188 The Hawthorn instance was deployed and works well. So :+1: 
This is safe to merge now, so that it doesn't block the upgrade.

CC @pomegranited if you wanted to review (pre- or post-merge).

PR #349, opened 2018-08-28
[OC-5074] Store AppServer termination date

ahmedaljazzar:
Task: [OC-5074](https://tasks.opencraft.com/browse/OC-5074)
This PR is to add the termination and the creation time for AppServers.


ahmedaljazzar:
All comments have been addressed:
- Change the PR title.
- Remove the field from admin.
- Add the field to the API

For the following:
> the default value (null) is ok: servers prior to this change won't have their termination date stored, but that should be fine because they don't use billing. In case we need to set the field, we could get it from the logs (but it's more complex) or set it from the „last modified“ date.

We can make a data migration script that runs on all instances and updates the date in terminated instances to match the date in `modified` field, but as you mentioned since these servers don't use billing, then that might not worth the effort.

clemente:
@ahmedaljazzar I checked that the field appears in the API.
However, this test fails due to that change:
`honcho -e .env.test run ./manage.py test instance.tests.api.test_openedx_instance.OpenEdXInstanceAPITestCase`
It says `AssertionError: 11 != 9 : 11 queries executed, 9 expected`

The new field may need to be added to a list of columns to prefetch (~ `appservers_few_columns`).

clemente:
:+1:, everything in the last review was addressed. Thanks!

clemente:
Deployed.

PR #350, opened 2018-08-30
[OC-4962] Add field to store the S3 region used by each instance

lgp171188:
This PR adds a field to store the S3 region used by each instance. Some clients are particular about which regions can be used to store their data and the data stored in this field will help in easily identifying the S3 region used by an instance.

**Dependencies**: None

**Testing instructions**:
1. Pull the code changes in the source branch of this PR.
1. Create a new instance.
1. Add the `AWS_S3_DEFAULT_REGION` configuration setting to the `.env` file and specify a value for that since the setting will otherwise use `''` as the default value. This has to be added to the `.env` file on OCIM production and stage environments via a separate PR to `ansible-secrets` repository.
1. The S3 region field should have the `s3_region` field set to the configured value or the default value `AWS_S3_DEFAULT_REGION`. Verify this in the admin interface form for creating an instance as well.
1. Provision an appserver for the instance. It should use S3 bucket(s) created in the configured region.
1. Archive an instance and verify that the S3 field still retains its value. The same behaviour should be exhibited when the `deprovision_s3()` method is directly called.

**Reviewers**
- [x] @itsjeyd


instance/models/mixins/openedx_storage.py
```diff
@@ -39,6 +41,7 @@ class Meta:
     s3_access_key = models.CharField(max_length=50, blank=True)
     s3_secret_access_key = models.CharField(max_length=50, blank=True)
     s3_bucket_name = models.CharField(max_length=50, blank=True)
+    s3_region = models.CharField(max_length=50, blank=True, default=default_setting('AWS_S3_DEFAULT_REGION'))
```
itsjeyd: @lgp171188 It might make sense to move these fields to `S3BucketInstanceMixin`? I'm not sure why they ended up in this class in the first place; it seems like `S3BucketInstanceMixin` would be a better fit for them.

instance/models/mixins/storage.py
```diff
@@ -232,14 +232,23 @@ def get_s3_connection(self):
             self.s3_secret_access_key
         )
 
-    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=4):
+    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=4, location=None):
         """
         Create bucket, retry up to defined attempts if it fails
+        If you specify a location (e.g. 'EU', 'us-west-1'), it will use it.
```
itsjeyd: @lgp171188 Maybe mention what the method falls back to if caller doesn't specify `location`.

instance/models/mixins/storage.py
```diff
@@ -232,14 +232,23 @@ def get_s3_connection(self):
...
         """
         time.sleep(retry_delay)
+        location = location or self.s3_region
         try:
             s3 = self.get_s3_connection()
-            bucket = s3.create_bucket(self.s3_bucket_name)
+            try:
+                bucket = s3.create_bucket(self.s3_bucket_name, location=location)
```
itsjeyd: @lgp171188 Will this part of the code also need to include special handling for specific regions?

> Different zones have different settings, e.g. Paris [requires AWSv4 signatures](https://stackoverflow.com/questions/27400105/using-boto-for-aws-s3-buckets-for-signature-v4).

instance/tests/models/test_openedx_storage_mixins.py
```diff
@@ -380,15 +386,18 @@ def test_deprovision_s3_delete_bucket_fails(self, s3_connection, connect_iam):
         instance.s3_access_key = 'test'
         instance.s3_secret_access_key = 'test'
         instance.s3_bucket_name = 'test'
+        instance.s3_region = 'test'
         with self.assertLogs("instance.models.instance"):
             instance.deprovision_s3()
         # Since it failed deleting the bucket, s3_bucket_name should not be empty
         self.assertEqual(instance.s3_bucket_name, "test")
+        self.assertEqual(instance.s3_region, "test")
```
itsjeyd: @lgp171188 You could add another comment above this line that says something like:

```python
# We always want to preserve information about a client's preferred region, so s3_region should not be empty.
```

opencraft/settings.py
```diff
@@ -677,6 +677,7 @@
 AWS_SECRET_ACCESS_KEY = env('AWS_SECRET_ACCESS_KEY', default=None)
 AWS_ACCESS_KEY_ID = env('AWS_ACCESS_KEY_ID', default=None)
 AWS_S3_BUCKET_PREFIX = env('S3_BUCKET_PREFIX', default='ocim')
+AWS_S3_DEFAULT_REGION = env('AWS_S3_DEFAULT_REGION', default='')
```
itsjeyd: @lgp171188 You can set the default to `us-east-1` since that [seems to be the default region](https://tasks.opencraft.com/browse/OC-4962?focusedCommentId=84576&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-84576).

instance/models/mixins/openedx_storage.py
lgp171188: @itsjeyd, `S3BucketInstanceMixin` was the first place I checked for these fields and since I didn't find them there, I searched further and found them in this class. I will move these fields to the mixin and test if everything works okay.

instance/models/mixins/storage.py
lgp171188: @itsjeyd, I am not sure if `boto` handles that transparently. Will check this on staging and post an update.

opencraft/settings.py
lgp171188: @itsjeyd, wouldn't it be better to leave the default here to `''` which will use the default region of the account and use `us-east-1` as the value for `AWS_S3_DEFAULT_REGION` in the `.env` file?

lgp171188:
@itsjeyd, addressed all the review comments except https://github.com/open-craft/opencraft/pull/350#discussion_r214302905 for which I have asked a question and https://github.com/open-craft/opencraft/pull/350#discussion_r214300065 for which some testing on OCIM stage is required.

opencraft/settings.py
itsjeyd: @lgp171188 Ah, yes -- that's a better pattern 👍

instance/models/mixins/storage.py
itsjeyd: @lgp171188 OK, sounds good.

lgp171188:
@itsjeyd, now that the CI tests are passing for other blocked PRs, I will rebase and squash this PR and deploy it on stage for verification. I also intend to test with regions which only support v4 signatures.

lgp171188:
@itsjeyd, the CI checks have passed now. I will deploy the changes to OCIM stage and post an update after verification.

instance/tests/models/test_openedx_storage_mixins.py
```diff
@@ -275,6 +276,16 @@ def test_get_s3_connection(self):
...
+        self.assertEqual(s3_connection.host, settings.AWS_S3_DEFAULT_HOSTNAME)
+
+        instance2 = OpenEdXInstanceFactory()
+        instance2.s3_region = 'eu-central-1'
+        s3_connection = instance2.get_s3_connection()
+        self.assertIsInstance(s3_connection, boto.s3.connection.S3Connection)
+        self.assertEqual(
+            s3_connection.host,
+            settings.AWS_S3_CUSTOM_REGION_HOSTNAME.format(region=instance2.s3_region)
+        )
```
itsjeyd: @lgp171188 If you override `settings.AWS_S3_DEFAULT_HOSTNAME` and `settings.AWS_S3_CUSTOM_REGION_HOSTNAME` via the decorator, it should be possible to reduce duplication here by using `ddt` to test different regions:

```python
    @override_settings(
        INSTANCE_STORAGE_TYPE=StorageContainer.S3_STORAGE,
        AWS_ACCESS_KEY='test',
        AWS_SECRET_ACCESS_KEY_ID='test',
        AWS_S3_DEFAULT_HOSTNAME='s3.test.host',
        AWS_S3_CUSTOM_REGION_HOSTNAME='s3.{region}.test.host',
    )
    @ddt.data(
        ('', 's3.test.host'),  # Default region (us-east-1)
        ('eu-central-1', 's3.eu-central-1.test.host'),
        # More regions ...
    )
    @ddt.unpack
    def test_get_s3_connection(self, s3_region, expected_hostname):
        """
        Test get_s3 connection returns right instance
        """
        instance.s3_region = s3_region
        s3_connection = instance.get_s3_connection()
        self.assertIsInstance(s3_connection, boto.s3.connection.S3Connection)
        self.assertEqual(s3_connection.host, expected_hostname)
```

opencraft/settings.py
```diff
@@ -684,6 +684,9 @@
 AWS_SECRET_ACCESS_KEY = env('AWS_SECRET_ACCESS_KEY', default=None)
 AWS_ACCESS_KEY_ID = env('AWS_ACCESS_KEY_ID', default=None)
 AWS_S3_BUCKET_PREFIX = env('S3_BUCKET_PREFIX', default='ocim')
+AWS_S3_CUSTOM_REGION_HOSTNAME = 's3.{region}.amazonaws.com'
+AWS_S3_DEFAULT_HOSTNAME = 's3.amazonaws.com'
```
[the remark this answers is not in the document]
lgp171188: I thought about this but decided against doing it because the values are not environment-specific (imho) unless we start using S3-compatible backends which work with `boto`.

instance/tests/models/test_openedx_storage_mixins.py
lgp171188: Ah, this makes it cleaner. Will implement the changes for this :+1:

instance/models/mixins/storage.py
```diff
@@ -229,17 +244,30 @@ def get_s3_connection(self):
...
+            # The host parameter is required when connecting to non-default regions and using
+            # AWS signature version 4.
+            host=self.s3_hostname
         )
 
-    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=4):
+    def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=4, location=None):
         """
         Create bucket, retry up to defined attempts if it fails
+        If you specify a location (e.g. 'EU', 'us-west-1'), it will use it. If location is
```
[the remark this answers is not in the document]
lgp171188: @itsjeyd, will address this - the existing message was a copy-paste from a different related PR and I must have missed reading it thoroughly.

lgp171188:
@itsjeyd, I have addressed all your review comments except [this one](https://github.com/open-craft/opencraft/pull/350#discussion_r218054103) for which I need your confirmation on whether to leave the configuration setting as is or if I should make the suggested change.

instance/models/mixins/openedx_storage.py
```diff
@@ -101,6 +95,13 @@ def _get_s3_settings(self):
             "AWS_S3_LOGS_SECRET_KEY": self.s3_secret_access_key,
         }
 
+        if self.s3_region:
+            s3_settings.update({
+                "aws_region": self.s3_region,
```
symbolist: Needed by https://github.com/edx/configuration/blob/open-release/hawthorn.1/playbooks/roles/aws/templates/send-logs-to-s3.j2#L107

clemente:
:+1: I tested it again with new servers, with the simplified code (no EDXAPP_AWS_S3_HOST, no XQUEUE_AWS_S3_HOST) and the {{aws_region}} override for the log sync script. As explained in the ticket, we're not supposed to override lowercase variables, but upstreaming a rename to uppercase would make the situation more complex and cause temporary drift, and I see there's already some mess with a pair called {{vhost_name}}/{{VHOST_NAME}}. It wouldn't change how it works, so I'd prefer the very simple approach. This is modifying the variable globally (not only in the role to sync logs) but that affects just 1 other rare role and we would also want to use our region.
More comments in the ticket.

Could you rebase, rearrange migrations if needed, squash if you want (not required), merge, deploy etc.?

PR #351, opened 2018-08-30
[OC-5177] Add a new configuration setting OPENSTACK_PRODUCTION_INSTANCE_FLAVOR

lgp171188:
Add a new configuration setting `OPENSTACK_PRODUCTION_FLAVOR`. Use it in the `production_instance_factory()` method to specify the
flavor to use for the production instances. Currently, all the
instances use the value of the `OPENSTACK_SANDBOX_FLAVOR` variable
by default. This change will facilitate using different flavors
for production and sandbox instances.

**Testing instructions**
* Deploy the changes in the source branch of this PR.
* Deploy the updated `.env` file from https://github.com/open-craft/ansible-secrets/pull/80 and restart OCIM.
* Create a new production instance and verify that it uses the flavor specified in `OPENSTACK_PRODUCTION_FLAVOR`.
* Create a non-production instance and verify that it uses the flavor specified in `OPENSTACK_SANDBOX_FLAVOR`.

**Reviewer**
- [ ] @itsjeyd

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -145,6 +145,9 @@ def production_instance_factory(**kwargs):
         configuration_version=settings.STABLE_CONFIGURATION_VERSION,
         openedx_release=settings.OPENEDX_RELEASE_STABLE_REF,
         configuration_extra_settings=extra_settings,
+        # Allow production instances to use a different OpenStack instance flavor by default.
+        # This allows using a larger instance flavor for production instances.
+        openstack_server_flavor=settings.OPENSTACK_PRODUCTION_INSTANCE_FLAVOR,
     )
     instance_kwargs.update(kwargs)
 
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -260,6 +260,10 @@ OPENSTACK_SANDBOX_FLAVOR = env.json('OPENSTACK_SANDBOX_FLAVOR', default={"ram": 
 OPENSTACK_SANDBOX_BASE_IMAGE = env.json('OPENSTACK_SANDBOX_BASE_IMAGE', default={"name": "Ubuntu 16.04"})
 OPENSTACK_SANDBOX_SSH_KEYNAME = env('OPENSTACK_SANDBOX_SSH_KEYNAME', default='opencraft')
 OPENSTACK_SANDBOX_SSH_USERNAME = env('OPENSTACK_SANDBOX_SSH_USERNAME', default='ubuntu')
+OPENSTACK_PRODUCTION_INSTANCE_FLAVOR = env.json(
+    'OPENSTACK_PRODUCTION_INSTANCE_FLAVOR',
+    default={"ram": 8192, "disk": 80}
+)
 
 # Separate credentials for Swift.  These credentials are currently passed on to each instance
 # when Swift is enabled.
--- a/instance/tests/test_factories.py
+++ b/instance/tests/test_factories.py
@@ -52,11 +52,13 @@ class FactoriesTestCase(TestCase):
         "configuration_version": settings.DEFAULT_CONFIGURATION_VERSION,
         "openedx_release": settings.DEFAULT_OPENEDX_RELEASE,
         "configuration_extra_settings": "",
+        "openstack_server_flavor": settings.OPENSTACK_SANDBOX_FLAVOR,
     }
     PRODUCTION_DEFAULTS = {
         "configuration_version": settings.STABLE_CONFIGURATION_VERSION,
         "openedx_release": settings.OPENEDX_RELEASE_STABLE_REF,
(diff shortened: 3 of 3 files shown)


opencraft/settings.py
```diff
@@ -261,6 +261,7 @@
 OPENSTACK_SANDBOX_SSH_KEYNAME = env('OPENSTACK_SANDBOX_SSH_KEYNAME', default='opencraft')
 OPENSTACK_SANDBOX_SSH_USERNAME = env('OPENSTACK_SANDBOX_SSH_USERNAME', default='ubuntu')
 
+OPENSTACK_PRODUCTION_FLAVOR = env.json('OPENSTACK_PRODUCTION_FLAVOR', default={"ram": 8192, "disk": 80})
```
itsjeyd: @lgp171188 Should this default to `{"name": "vps-ssd-3"}`?

instance/factories.py
```diff
@@ -145,6 +145,7 @@ def production_instance_factory(**kwargs):
         configuration_version=settings.STABLE_CONFIGURATION_VERSION,
         openedx_release=settings.OPENEDX_RELEASE_STABLE_REF,
         configuration_extra_settings=extra_settings,
+        openstack_server_flavor=settings.OPENSTACK_PRODUCTION_FLAVOR,
```
itsjeyd: @lgp171188 It would be nice to add a comment above this line that explains the purpose of setting that field here. Something like:

```python
# Allow production instances to use larger instance types by default.
```

opencraft/settings.py
itsjeyd: @lgp171188 Also, let's call this `OPENSTACK_PRODUCTION_INSTANCE_FLAVOR` like @pomegranited suggested; that's a bit more specific than the current name :)

lgp171188: @itsjeyd, the default value for the existing `OPENSTACK_SANDBOX_FLAVOR` is set to `{"ram": 4096, "disk": 40}` and that is why I setup the default value similarly for the production instance flavor. Should I still change it to `vps-ssd-3`? `vps-ssd-3` seems to be specific to a particular OpenStack provider and I think it is best defined in the `.env` file.

lgp171188:
@itsjeyd, I have addressed all your review comments except https://github.com/open-craft/opencraft/pull/351#discussion_r214274188 as I am waiting for a response to my question in the reply.

opencraft/settings.py
itsjeyd: @lgp171188 

> ... the default value for the existing `OPENSTACK_SANDBOX_FLAVOR` is set to `{"ram": 4096, "disk": 40}` and that is why I setup the default value similarly for the production instance flavor.

Ah, I see -- makes sense to go with the original approach, then.

A follow-up question, though: Does `vps-ssd-3` come with a disk that's twice the size of `vps-ssd-2`? I'm asking because we really only wanted to increase the amount of RAM that each production instance has available, not necessarily partition size.

itsjeyd:
@lgp171188 I just replied to your question, see my [comment](https://github.com/open-craft/opencraft/pull/351#discussion_r214384013) above.

opencraft/settings.py
lgp171188: @itsjeyd, `vps-ssd-2` has 4 GB RAM with 1 vCPU and 20 GB of disk space and `vps-ssd-3` has 8 GB RAM with 2 vCPUs and 40 GB of disk space. I don't see a flavor which just offers more RAM.

lgp171188:
@itsjeyd, the CI checks are failing due to what appears to be an unrelated error. I have restarted the CI build. Once it succeeds, I will deploy the changes on OCIM stage.

instance/factories.py
```diff
@@ -145,6 +145,9 @@ def production_instance_factory(**kwargs):
         configuration_version=settings.STABLE_CONFIGURATION_VERSION,
         openedx_release=settings.OPENEDX_RELEASE_STABLE_REF,
         configuration_extra_settings=extra_settings,
+        # Allow production instances to use a different OpenStack instanceflavor by default.
```
itsjeyd: @lgp171188 Typo: "instanceflavor"

lgp171188:
@itsjeyd, updated the PR to address your review comment. Waiting for the CI checks to pass before deploying it on stage. Will ensure that the changes of this PR do not undo the changes of PR #346.

itsjeyd:
@lgp171188 Sounds good :) Feel free to squash your commits before deploying them to stage -- this will need to happen prior to merging this PR anyway, and I'm done reviewing the individual commits.

lgp171188:
@itsjeyd, all the CI checks have passed. I will deploy the changes on stage by cherry-picking the commit in this PR and https://github.com/open-craft/ansible-secrets/pull/80 and test it.

PR #354, opened 2018-09-13
SE-87 Limit SSH access to sandboxes and self-hosted instances

ahmedaljazzar:
Task: [OC-5077](https://tasks.opencraft.com/browse/OC-5077).

This code change is to limit the sandbox ssh access to the Github users' of the organization that owns the Sandbox. We can  restrict the access more to the github users that are also users on OpenCraft, but I found it out-of-scope so I didn't add it.
This PR left some stale methods and removed some tests in the same time, these to be addressed in [OC-5237](https://tasks.opencraft.com/browse/OC-5237).

### Test Instructions
1. Create an Organization and a corresponding WatchedFork object. 
1. Make sure to specify a valid `github_username` for your OCIM user (via the corresponding UserProfile).
1. Add your OCIM user to the Organization created in the previous step. 
1. Set the fork field of the WatchedFork created previously to the GitHub name of the fork against which you'll be creating PRs.
1. Check that OCIM automatically granted SSH access to the created instance to all members of the Organization.
1.  Check that OCIM automatically granted SSH access to the created instance to all OCIM users with superuser privileges.
1. Check the new AppServer configuration for `COMMON_USER_INFO` and make sure that there're no users from outside the organization and the admin users.
1. Create a new instance that's not linked with a WatchedFork and make sure that OCIM admins are the only ones who can access that server.


clemente:
@ahmedaljazzar Thanks, I'll review it soon (maybe in some days). The changes in OC-5072 have already been merged, so you'll be able to rebase. I'll be checking the last commit anyway.
Could you also include some test instructions in the PR? Some scenarios, or some ways in which I can test it.
It's also good to summarize in the description the changes that your PR is doing (on a functional level) and how they relate to SSH access.

clemente:
(@ahmedaljazzar Please ping @rocioar for this review)

instance/models/openedx_appserver.py
```diff
@@ -361,13 +361,30 @@ def smtp_relay_settings(self):
...
+    @property
+    def organization_users(self):
+        """
+        :return: Github usernames of the organization this instance owned by.
+        """
+        users = []
+        organization = self.instance.ref.owner
+        if organization:
+            org_github_handle = organization.github_handle
+            usernames = get_username_list_from_team(org_github_handle)
```
rocioar: @ahmedaljazzar  get_username_list_from_team is obsolete, we should instead use the users that are registered on ocim under the organization that owns the instance.

So something like:

`python
UserProfile.objects.filter(organization=organization).values_list('github_username', flat=True)
`

rocioar: We should also remove the function `get_username_list_from_team` together with `github_admin_username_list`

instance/tests/models/test_openedx_appserver.py
```diff
@@ -132,6 +133,30 @@ def test_provision_unhandled_exception(self, mocks):
         self.assertFalse(result)
         mocks.mock_provision_failed_email.assert_called_once_with("AppServer deploy failed: unhandled exception")
 
+    @patch('instance.models.openedx_appserver.get_username_list_from_team')
```
rocioar: Update the tests according to the feedback above (re: removing the `get_username_list_from_team` function)

ahmedaljazzar:
@rocioar This is ready for another pass. I'm sorry I had to force push the first commit so that git doesn't add a record for .DS_Store file in the repo and since I rebased my changes on master.

userprofile/factories.py
```diff
@@ -36,7 +37,7 @@ class Meta:
         model = User
         django_get_or_create = ('username',)
 
-    username = "edx"
+    username = factory.Sequence(lambda n: 'user_%d' % n)
```
rocioar: @ahmedaljazzar instead of doing this, can we just use the github_username parameter on make_user_and_organization function be the username?

rocioar:
@ahmedaljazzar I left a couple more comments, once that's address you can run it on stage, and point me to the instance, so that I can confirm everything is running as expected.

instance/models/openedx_appserver.py
```diff
@@ -362,18 +362,19 @@ def smtp_relay_settings(self):
...
-        admin_users = []
-        for org in self.github_admin_organizations:
-            admin_users += get_username_list_from_team(org)
-        admin_users += self.github_admin_users
-        return admin_users
+        users = []
+        organization = self.instance.ref.owner
+        if organization:
+            usernames = UserProfile.objects.filter(
+                organization__github_handle=organization.github_handle
```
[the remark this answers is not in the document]
ahmedaljazzar: I'm not sure, because the primary key for `Organization` is integer while the `github_handle` is a string.

userprofile/factories.py
ahmedaljazzar: Yes we can do that, but don't you think this might confuse stuff? I think this is okay as we can create as many object as we want, and we can override its value whenever we want.

rocioar: @ahmedaljazzar confuse stuff in which way? I'm not sure I follow.

I think it's more readable if it's a parameter, you can re-use the github_username parameter or if you think that's confusing it can be another parameter "username".

instance/models/openedx_appserver.py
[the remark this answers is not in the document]
rocioar: @ahmedaljazzar yes, you can do UserProfile.objects.filter(organization=organization), can you update the code with this? It makes it more readable.

instance/models/openedx_appserver.py
```diff
@@ -362,18 +337,29 @@ def smtp_relay_settings(self):
...
+            ).values_list('github_username', flat=True)
+            users += usernames if usernames else []
+
+        # Add superusers to the list of users that can administer this instance
+        usernames = UserProfile.objects.filter(
+            user__is_superuser=True
+        ).values_list('github_username', flat=True)
+        users += usernames if usernames else []
+
+        return users
```
clemente: If some users (organization or superuser) doesn't have github_username, this list contains `None`s and could fail in a later step. Could you filter them out? You can either in QuerySets or in a Python filter at the end.

instance/static/html/instance/appserver.html
```diff
@@ -75,12 +75,12 @@ <h4>{{ appserver.name }}</h4>
...
         <tr>
-          <td>GitHub admin organizations</td>
-          <td>{{ appserver.github_admin_organizations }}</td>
+          <td>Owner organization</td>
+          <td>{{ instance.owner_organization }}</td>
         </tr>
         <tr>
-          <td>GitHub admin users</td>
-          <td>{{ appserver.github_admin_users }}</td>
+          <td>configuration_source_repo_url</td>
```
clemente: This seems a typo, we don't want `configuration_source_repo_url` here. I think we wanted to show the github admin users. Is this even possible? Do we store them? Note that this (if we have it) should show the *appserver* admins, not the instance admins (same as the other details, it relates to the appserver).
If we don't *store* the appserver's admins, then I think that we could add the admin users to the instance view (instance admins). But this gets more complex and we don't have time to test this. So I'd prefer not to show this field.

userprofile/factories.py
```diff
@@ -65,9 +65,12 @@ class Meta:
 
 def make_user_and_organization(organization_name="Test Org", github_handle="test-org", github_username="edx"):
     """
-    Create user, userprofile and organization needed for reference on instances
+    Create user, userprofile and organization needed for reference on instances.
+    We're using the github_username passed to this method to create User and
```
clemente: I understand that you experienced this case (some test failing because of this), but there's no need to mention it; it doesn't directly help the next person who uses the code. Better explain what the function does, and what is expected from the caller. E.g. „If provided, it also sets the user's GitHub username (which must be unique)“. And if you prefer, we don't even need to say that the github username should be unique (it's like User.username): someone who writes tests will notice if a test tries to reuse an existing github_username and fails.

instance/models/openedx_appserver.py
clemente: I tested this and found that it also adds a username twice if it's superuser and belongs to an Organization. You may want to use list(set(…)) or something similar at the end so that each username appears once. I don't know the effects of being there twice, but maybe SSH later can complain.

instance/static/html/instance/appserver.html
ahmedaljazzar: @taranjeet can you take a look into this?

itsjeyd: @clemente @ahmedaljazzar @taranjeet Oops, my bad -- I had suggested in [this comment](https://github.com/open-craft/opencraft/pull/357#discussion_r223975991) to show the creator of the app server here (which should really be the creator of the parent instance), but ended up not updating the code snippet appropriately before posting the comment. Here's an updated version of that snippet:

```html
        <tr>
          <td>Owner organization</td>
          <td>{{ instance.owner_organization }}</td>
        </tr>
        <tr>
          <td>Instance creator (Ocim user)</td>
          <td>{{ instance.creator_username }}</td>
        </tr>
```

I agree that we shouldn't look into displaying the list of users that have SSH access to the app server here. Simply updating the diff to match the snippet above will be enough.

instance/tests/models/test_openedx_appserver.py
```diff
@@ -132,35 +134,48 @@ def test_provision_unhandled_exception(self, mocks):
         self.assertFalse(result)
         mocks.mock_provision_failed_email.assert_called_once_with("AppServer deploy failed: unhandled exception")
 
-    def test_github_admin_username_list_default(self):
+    def test_admin_users(self):
         """
-        By default, no admin should be configured
+        By default, all users that belong to an organization that owns the server
+        have access to the sandbox.
```
itsjeyd: @ahmedaljazzar This comment should also mention superusers.

instance/tests/models/test_openedx_appserver.py
```diff
@@ -132,35 +134,48 @@ def test_provision_unhandled_exception(self, mocks):
...
+        ]
+        admin_users = [
+            ('admin1', 'admin1', admin_org_handle),
+            ('admin3', 'admin3', 'another_org'),
+            ('admin4', 'admin4', ''),
+            ('admin5', 'admin5', None),
+            ('admin_no_org', '', admin_org_handle),
+        ]
+
+        valid_users = ['user', 'admin1', 'admin2', 'admin3', 'admin4', 'admin5']
```
itsjeyd: @ahmedaljazzar Let's call this `expected_admin_users` (`valid_users` is more ambiguous -- reading this line for the first time made me ask "valid according to what criteria"?)

userprofile/factories.py
```diff
@@ -63,18 +63,24 @@ class Meta:
...
+def make_user_and_organization(username="testuser", org_name="Test Org", org_handle="test-org", github_username="edx"):
     """
-    Create user, userprofile and organization needed for reference on instances
+    Create user, userprofile and organization needed for reference on instances.
+
+    :param username: The username of User object
+    :param org_name: The organization name the user belongs to
+    :param org_handle: The organization github handle the user belongs to
+    :param github_username: The user's username on github
+    :return: A pair of the created userprofile and the organization that profile belongs to
```
itsjeyd: @ahmedaljazzar Thanks for documenting these parameters! I had some small suggestions for additional polish. Here's what the L70-L74 would look like with that polish applied:

```
    Create User, UserProfile and Organization to reference from individual instances.

    :param username: The username of the User to create.
    :param org_name: The name of the Organization to create. (The user to create will be added to that organization.)
    :param org_handle: The GitHub handle of the organization to create.
    :param github_username: The GitHub username of the user to create.
    :return: A tuple consisting of the created UserProfile and Organization.
```

instance/models/openedx_appserver.py
```diff
@@ -340,22 +340,26 @@ def smtp_relay_settings(self):
     @property
     def admin_users(self):
         """
-        Return GitHub usernames, if exists, of admin users for this instance.
+        Retrieve all users that fulfill requirements defined via `admin_user_query`,
+        excluding any users that don't have a meaningful GitHub username.
+        (Without a meaningful GitHub username it won't be possible
+        to correctly set up SSH access to a given sandbox.
```
itsjeyd: @ahmedaljazzar Shoot, this last sentence is missing a closing paren. Looks like I forgot to include that [here](https://github.com/open-craft/opencraft/pull/354#discussion_r225066048).

instance/static/html/instance/appserver.html
```diff
@@ -79,8 +79,12 @@ <h4>{{ appserver.name }}</h4>
           <td>{{ instance.owner_organization }}</td>
         </tr>
         <tr>
-          <td>configuration_source_repo_url</td>
-          <td>{{ instance.configuration_source_repo_url }}</td>
+          <td>Owner organization</td>
+          <td>{{ instance.owner_organization }}</td>
+        </tr>
```
itsjeyd: @ahmedaljazzar This info is now duplicated (cf. L77-L80 above), please remove one of two occurrences.

clemente:
:+1: I re-tested everything, specially the functional part: I made an instance belong to Ahmed's organization, I made some non-admin user belong to that organization, and checked that in new servers, the list of admins included all superusers plus the list.

I'll merge and deploy.

PR #356, opened 2018-09-21
(DON'T MERGE) [OC-5408] Add support for agent cluster leave

ahmedaljazzar:
This change introduces two things:
1. The ability to communicate with Consul through APIs directly without a need for external library. This is useful when we run into advanced cases where the library we're using doesn't support them.
1. Consul leave functionality which asks the agent to leave the cluster gracefully, or to instruct the agent to force a node into the left state if the agent couldn't leave its cluster gracefully.

I wanted to implement a multi-(threading|processing) approach here since the operation takes about 5-8 seconds to complete, then I backed of when I saw there's no such usage in the whole project for any of the approaches.

- [x] Tests

### DIFF
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -28,7 +28,6 @@ from weakref import WeakKeyDictionary
 from django.conf import settings
 
 import requests
-from requests import exceptions as requests_exceptions
 
 import consul
 
@@ -526,7 +525,7 @@ class ConsulClient(object):
 
         try:
             response = requests.request(method, url)
-        except requests_exceptions.ConnectionError:
+        except requests.exceptions.ConnectionError:
             return False
 
         return 200 <= response.status_code < 300
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -27,9 +27,10 @@ from weakref import WeakKeyDictionary
 
 from django.conf import settings
 
-import consul
 import requests
-from requests.exceptions import ConnectionError
+from requests import exceptions as requests_exceptions
+
+import consul
 
 # Exceptions ##################################################################
 
@@ -484,7 +485,7 @@ class ConsulAgent(object):
         """
         return self._client.kv.delete(self.prefix, recurse=True)
 
-    def leave(self, force=True):
+    def leave(self, force=False):
         """
         This method will request the agent to leave the cluster, if the
         operation wasn't completed successfully then it'll force nodes to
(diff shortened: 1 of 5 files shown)


instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
...
+        This method will force the agent nodes to leave the cluster.
+        """
+        _, nodes = self._client.catalog.nodes()
+        for node in nodes:
+            self._client.agent.force_leave(node['Node'])
+
+    def _api_request(self, path, method='get'):
+        """
+        This method serves a good need of making API requests to Consul
+        directly. This good in casae we need to make some advanced
```
UmanShahzad: Nit: `casae` -> `case`.

instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
...
+        operations that are not supported from the library.
+        :param path:
+        :param method:
+        :return:
```
UmanShahzad: These params and return value docs need to be filled in still.

instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
...
+        if not is_left and force:
+            return self.force_leave()
+
+    def force_leave(self):
+        """
+        This method will force the agent nodes to leave the cluster.
+        """
+        _, nodes = self._client.catalog.nodes()
+        for node in nodes:
+            self._client.agent.force_leave(node['Node'])
```
UmanShahzad: Why do we make every single node leave the cluster in this function? That would be catastrophic. We only want a single node to leave, namely the one we specify. And since the version we're using doesn't have the `get` vs `put` fix yet this should actually just be:

```python
request_path = '/v1/agent/force-leave'
return self._api_request(request_path, method='put')
```

otherwise as it currently is it wouldn't work.

instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
         """
         return self._client.kv.delete(self.prefix, recurse=True)
 
+    def leave(self, force=True):
```
UmanShahzad: This is dangerous to be `True` by default. If we somehow managed to run this on the Consul server agents, we could have a problem. See 2nd paragraph in [the docs](https://www.consul.io/api/agent.html#graceful-leave-and-shutdown).

instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
...
+        """
+        This method will request the agent to leave the cluster, if the
+        operation wasn't completed successfully then it'll force nodes to
+        leave the cluster if force option is set to True.
+
+        :param force: Will fall back to the force mode if the operation
+                      wasn't completed successfuly.
+        :return: True if the operation completed successfully, False otherwise.
```
UmanShahzad: It's not currently doing that, because if the server `is_left` (successful graceful leave), we skip the next `if` check and just return `None` by default. We can assign `is_left` to `self.force_leave` in that conditional and then ultimately return `is_left`, or whatever gets the job done.

instance/models/utils.py
```diff
@@ -482,6 +484,49 @@ def purge(self):
         """
         return self._client.kv.delete(self.prefix, recurse=True)
 
+    def leave(self, force=True):
+        """
+        This method will request the agent to leave the cluster, if the
```
UmanShahzad: I'm confused as to how these new functions will actually request a _particular_ agent to leave the cluster. There seems to be no option in these functions so far to allow that.

instance/models/utils.py
ahmedaljazzar: Sending a force-leave request using this path only `/v1/agent/force-leave` won't put any node in a left state; we have to send a node name in this API for Consul to start the leave procedure. So what I did is using `settings.OCIM_ID`, that's used for Consul to identify a server, to remove this node. However, this might require some changes in our playbooks in running the consul agent to change the node name from `ansible_nodename` to `ocim_id` [[here](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/consul/defaults/main.yml#L7)]. The main point I saw here is that our playbooks doesn't set `OCIM_ID` env variable as expected so far.

ahmedaljazzar:
Closing in favor of #359

PR #357, opened 2018-09-28
Add Opencraft admins's SSH Keys to client sandboxes.

taranjeet:
This PR adds the functionality of adding opencraft admin's ssh keys to client sandboxes.

**JIRA tickets**: [SE-129](https://tasks.opencraft.com/browse/SE-129)

**Discussions**: 

**Dependencies**: Depends on PR #354 

**Merge deadline**: None

This PR builds on PR #354 and adds an incremental support, so first, this pr should be merged and then the PR 354 in master.

**Reviewers**
- [x] @itsjeyd 
- [ ] @ahmedaljazzar


instance/models/openedx_appserver.py
```diff
@@ -364,7 +364,10 @@ def smtp_relay_settings(self):
     @property
     def organization_users(self):
```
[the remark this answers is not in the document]
taranjeet: I have updated it to `admin_users`. @itsjeyd

instance/models/openedx_appserver.py
```diff
@@ -362,18 +349,28 @@ def smtp_relay_settings(self):
...
+        # Add members of owning organization to the list of users that can administer this instance
         organization = self.instance.ref.owner
         if organization:
             usernames = UserProfile.objects.filter(
                 organization=organization
             ).values_list('github_username', flat=True)
             users += usernames if usernames else []
 
+        # Add superusers to the list of users that can administer this instance
+        usernames = UserProfile.objects.filter(
```
ahmedaljazzar: This will hit the database twice for the same query but for different filtering, we can reduce the hits to one if we have on QuerySet and we iterate over it once. Something like the following would be more efficient: (Didn't test it though)
```PYTHON
users_qs = UserProfile.objects.filter(user__is_superuser=True)

organization = self.instance.ref.owner
if organization:
    users_qs = users_qs.filter(organization=organization)

usernames = list(users_qs.values_list('github_username', flat=True))
```

clemente: I didn't read all the code, but the people we want in a sandbox are: everyone from the organization *plus* all of our superusers.

ahmedaljazzar: @taranjeet That's true. I didn't change the end result in my suggestion, I only changed the approach.

clemente: But we don't require people in a sandbox organization to have a superuser account in our Ocim. So @ahmedaljazzar I think that the code you proposed with 1 query won't work. It's not a problem to have 2 queries either.

instance/models/openedx_appserver.py
```diff
@@ -151,19 +151,6 @@ class Meta:
     )
 
     # Misc settings:
-    github_admin_organizations = JSONField(
-        max_length=256,
-        blank=True,
-        default=default_admin_organizations,
```
clemente: @taranjeet @ahmedaljazzar If you remove this field, then the function `default_admin_organizations` is left alone (noone calls it), you should delete it too. Then you'll find that you'll need to delete a setting too (`DEFAULT_ADMIN_ORGANIZATION`). This setting is actually defined in our repo `ansible-secrets` and it's used by ansible to put it into the `.env` when deploying, and we'll need to delete it too (just to keep it synchronized). I can do that last step (it needs access) but we should remove the unused code here too.

instance/models/openedx_appserver.py
ahmedaljazzar: True, @clemente. A thing we can do here is to use [Q objects](https://docs.djangoproject.com/en/1.10/topics/db/queries/#complex-lookups-with-q-objects).
```Python
users_qs = UserProfile.objects.filter(Q(user__is_superuser=True) | Q(organization=organization))
```
However, it's up to @taranjeet to use it or not.

instance/models/openedx_appserver.py
taranjeet: @clemente : I did remove that function and the related settings, but then the migrate crashes. You can read more about it [here|https://stackoverflow.com/a/40994166/2534102]

instance/models/openedx_appserver.py
taranjeet: Keeping the current code, hence marking this as resolved.

instance/models/openedx_appserver.py
ahmedaljazzar: We can replace this function in the migration files with `[]`. I think it's safe to do that as it's a data-migration that only expect a given result from some function Django doesn't know how it works. For previous migrations these have been already migrated and won't be affected, and for the new ones the value is just an empty list.

itsjeyd:
@taranjeet Could you please rebase this PR on the current base branch (`jazzar/limit-ssh-access`) to clean up the diff? That will make it easier to (re-)review the changes :)

instance/models/openedx_appserver.py
clemente: I saw they are used in migrations… I don't know what will acceptable to Django. You can try what @ahmedaljazzar suggested. Noone should run this function in the future even if they migrate Ocim from 0 to the current migration, so it should be fine to delete it.

taranjeet:
@itsjeyd : I have rebased the PR and removed everything as requested in the review.

PR #358, opened 2018-09-29
(SE-125) Sandbox billing: basic report

ahmedaljazzar:
TASK: [SE-125](https://tasks.opencraft.com/browse/SE-125)

### Requirements
In this PR you need to test the following:

Description | Developed | Tested
:---|:---:|:---:
New Django app and URLs for issuing reports. | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul> |
Custom links for users in the Admin panel to view their reports. | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul> |
Filtering data for an organization, and for a particular month (URL only in this task). | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul> 
Generate invoices, by computing the number of days to bill. | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul> 
Show only relevant AppServers (The ones that were active in the period). | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul>
Simple HTML report, that can be copy+pasted into invoices. | <ul><li>- [x] </li></ul> | <ul><li>- [ ] </li></ul>

### Test instructions
- Create an Organization
- Create OpenEdxInstance
- Create WatchedFork and link it to that organization.
- Create Watched PR and link them with WatchedFork and the created instance.
- Spawn new AppServers
- Go to the [Organizations](http://localhost:5000/admin/userprofile/organization/) page in the admin panel
- Click on [Invoice Report]() link that will take you to the Invoice page.
- Repeat the process above to cover all test casses (AppServer that's terminated in the previous billing month, AppServer that's still running the current billing month, AppServer that's created after the billing month, ..)
- Check that the number of computed working days for each AppServer are correct.
- Check the number of computed charge per each AppServer, and fork, and all forks are computed correctly.
- Try copying and pasting the table in different platforms and check whether it keeps the correct format after the paste.
- Try changing the month and the year from the url and check the above again.

### Screenshots
screen shot 2018-09-30 at 3 29 52 am
screen shot 2018-09-30 at 3 29 47 am
screen shot 2018-10-01 at 4 37 47 am

### DIFF
--- a/instance/tests/models/factories/openedx_appserver.py
+++ b/instance/tests/models/factories/openedx_appserver.py
@@ -22,15 +22,28 @@ OpenEdXAppServer model - Factories
 
 # Imports #####################################################################
 
+from instance.models.appserver import Status as AppServerStatus
 from instance.models.load_balancer import LoadBalancingServer
 from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
 
 # Functions ###################################################################
 
 
-def make_test_appserver(instance=None, s3=False, server=None):
+def make_test_appserver(instance=None, s3=False, server=None, status=None):
     """
     Factory method to create an OpenEdXAppServer (and OpenStackServer).
+
+    Note that this method does not set the status of the VM (OpenStackServer)
+    that is associated with the app server.
+    Client code is expected to take care of that itself (if necessary).
+
+    :param instance: The OpenEdx instance to create an AppServer for, if not
+                     given will create a new instance.
+    :param s3: Will configure S3 storage for the OpenEdXInstance the AppServer
+               belongs to.
+    :param server: The OpenStackServer to associate with this AppServer.
+    :param status: Will move an AppServer to the specified state
+    :return: appserver for `instance`
     """
     if not instance:
         instance = OpenEdXInstanceFactory()
@@ -49,4 +62,49 @@ def make_test_appserver(instance=None, s3=False, server=None):
         appserver.server = server
         appserver.save()
 
+    if status == AppServerStatus.Running:
+        _set_appserver_running(appserver)
+    if status == AppServerStatus.ConfigurationFailed:
+        _set_appserver_configuration_failed(appserver)
+    elif status == AppServerStatus.Error:
+        _set_appserver_errored(appserver)
+    elif status == AppServerStatus.Terminated:
+        _set_appserver_terminated(appserver)
+
     return appserver
+
+
+def _set_appserver_terminated(appserver):
+    """
+    Transition `appserver` to AppServerStatus.Terminated.
+    """
+    appserver._status_to_waiting_for_server()
+    appserver._status_to_configuring_server()
+    appserver._status_to_running()
+    appserver._status_to_terminated()
+
+
+def _set_appserver_running(appserver):
+    """
+    Transition `appserver` to AppServerStatus.Running.
+    """
+    appserver._status_to_waiting_for_server()
+    appserver._status_to_configuring_server()
+    appserver._status_to_running()
+
+
+def _set_appserver_configuration_failed(appserver):
+    """
+    Transition `appserver` to AppServerStatus.ConfigurationFailed.
+    """
+    appserver._status_to_waiting_for_server()
+    appserver._status_to_configuring_server()
+    appserver._status_to_configuration_failed()
+
+
+def _set_appserver_errored(appserver):
+    """
+    Transition `appserver` to AppServerStatus.Error.
+    """
+    appserver._status_to_waiting_for_server()
+    appserver._status_to_error()
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -536,103 +536,32 @@ class OpenEdXInstanceTestCase(TestCase):
             appserver.refresh_from_db()
 
     @staticmethod
-    def _set_appserver_terminated(appserver):
-        """
-        Transition `appserver` to AppServerStatus.Terminated.
-        """
-        appserver._status_to_waiting_for_server()
-        appserver._status_to_configuring_server()
-        appserver._status_to_running()
-        appserver._status_to_terminated()
-
-    @staticmethod
-    def _set_appserver_running(appserver):
-        """
-        Transition `appserver` to AppServerStatus.Running.
-        """
-        appserver._status_to_waiting_for_server()
-        appserver._status_to_configuring_server()
-        appserver._status_to_running()
-
-    @staticmethod
-    def _set_appserver_configuration_failed(appserver):
-        """
-        Transition `appserver` to AppServerStatus.ConfigurationFailed.
-        """
-        appserver._status_to_waiting_for_server()
-        appserver._status_to_configuring_server()
-        appserver._status_to_configuration_failed()
-
-    @staticmethod
-    def _set_appserver_errored(appserver):
-        """
-        Transition `appserver` to AppServerStatus.Error.
-        """
-        appserver._status_to_waiting_for_server()
-        appserver._status_to_error()
-
-    @staticmethod
-    def _set_server_ready(server):
-        """
-        Transition `server` to Status.Ready.
-        """
-        server._status_to_building()
-        server._status_to_booting()
-        server._status_to_ready()
-
-    @staticmethod
-    def _set_server_terminated(server):
-        """
-        Transition `server` to Status.Terminated.
-
-        Note that servers are allowed to transition to ServerStatus.Terminated from any state,
-        so it is not necessary to transition to another status first.
-        """
-        server._status_to_terminated()
-
-    def _create_appserver(self, instance, status):
-        """
-        Return appserver for `instance` that has `status`
-
-        Note that this method does not set the status of the VM (OpenStackServer)
-        that is associated with the app server.
-
-        Client code is expected to take care of that itself (if necessary).
-        """
-        appserver = make_test_appserver(instance)
-        if status == AppServerStatus.Running:
-            self._set_appserver_running(appserver)
-        if status == AppServerStatus.ConfigurationFailed:
-            self._set_appserver_configuration_failed(appserver)
-        elif status == AppServerStatus.Error:
-            self._set_appserver_errored(appserver)
-        elif status == AppServerStatus.Terminated:
-            self._set_appserver_terminated(appserver)
-        return appserver
-
-    def _create_running_appserver(self, instance):
+    def _create_running_appserver(instance):
         """
         Return app server for `instance` that has `status` AppServerStatus.Running
         """
-        return self._create_appserver(instance, AppServerStatus.Running)
+        return make_test_appserver(instance, status=AppServerStatus.Running)
 
-    def _create_failed_appserver(self, instance, with_running_server=False):
+    @staticmethod
+    def _create_failed_appserver(instance):
         """
         Return app server for `instance` that has `status` AppServerStatus.ConfigurationFailed
         """
-        return self._create_appserver(instance, AppServerStatus.ConfigurationFailed)
+        return make_test_appserver(instance, status=AppServerStatus.ConfigurationFailed)
 
-    def _create_errored_appserver(self, instance):
+    @staticmethod
+    def _create_errored_appserver(instance):
         """
         Return app server for `instance` that has `status` AppServerStatus.Error
         """
-        return self._create_appserver(instance, AppServerStatus.Error)
+        return make_test_appserver(instance, status=AppServerStatus.Error)
 
-    def _create_terminated_appserver(self, instance):
+    @staticmethod
+    def _create_terminated_appserver(instance):
         """
         Return app server for `instance` that has `status` AppServerStatus.Terminated
         """
-        return self._create_appserver(instance, AppServerStatus.Terminated)
+        return make_test_appserver(instance, status=AppServerStatus.Terminated)
 
     def _assert_status(self, appservers):
         """
--- a/reports/helpers.py
+++ b/reports/helpers.py
@@ -1,0 +1,261 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2018 OpenCraft <xavier@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Helpers functions for Invoices Reports
+"""
+import calendar
+
+from datetime import timedelta
+from collections import defaultdict
+
+from django.conf import settings
+from django.utils import timezone
+
+from instance.models.appserver import AppServer
+from pr_watch.models import WatchedPullRequest
+
+
+def _normalize_date(date):
+    """
+    Manipulates the given date and normalizes it to a standard day in a month
+    to help in generating and comparing billing periods.
+    The standardization process keeps the values of year and month as is in the given
+    date and defaults the following parameters to these values:
+    day: 1, hour: 0, minute: 0, second: 0, and microsecond: 0.
+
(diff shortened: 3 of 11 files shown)


reports/tests.py
```diff
@@ -0,0 +1,504 @@
...
+
+        self.assertIsInstance(watched_forks[fork2], list)
+        self.assertEqual(len(watched_forks[fork2]), 1)
+        for instance in watched_forks[fork1]:
+            self.assertIsInstance(instance, OpenEdXInstance)
+
+    @mock.patch('reports.helpers.get_instance_charges')
+    def test_generate_charges(self, instance_charges_mock):
+        """
+        This method is to make sure the we're able to generate the charges
```
clemente: - Simplify the wording (you can remove „This method is to…“)
- the→that.
- the description doesn't describe much more than the name. The description could explain what the test actually does, or could put the focus in some particular thing that is being tested (e.g. I think the point here is that it includes _all_ servers?).

reports/tests.py
```diff
@@ -0,0 +1,504 @@
...
+        self.assertIn(str(charges_mock), str(appservers_charges['fork2']['instances']))
+
+    def _setup_watched_forks(self):
+        """
+        This is a helper method that will generate the mesh of forks,
+        instances, and pull requests linked together
+        """
+        watched_fork1 = WatchedForkFactory.create(
+            fork='fork/repo1',
+            organization=self.organization)
```
clemente: `)` should be in next line

reports/tests.py
```diff
@@ -0,0 +1,504 @@
...
+        watched_fork2 = WatchedForkFactory.create(
+            fork='fork/repo2',
+            organization=self.organization)
```
clemente: Same: `)` should be in next line

reports/views.py
```diff
@@ -0,0 +1,60 @@
...
+from django.shortcuts import get_object_or_404, render
+from django.utils import timezone
+
+from instance.views import instance_manager_required
+from reports.helpers import generate_watched_forks_instances, generate_charges
+from userprofile.models import Organization
+
+
+@login_required
+@instance_manager_required(redirect_to='registration:register')
```
clemente: No, it should require superuser, not instance manager.

reports/helpers.py
```diff
@@ -0,0 +1,252 @@
...
+                'fork2': [instance 4, instance 7, instance 6],
+            }
+    """
+    watched_forks = defaultdict(list)
+    pull_requests = WatchedPullRequest.objects.filter(
+        watched_fork__organization=organization
+    ).select_related('watched_fork', 'instance')
+
+    for pull_request in pull_requests:
+        watched_forks[pull_request.watched_fork].append(pull_request.instance)
```
clemente: In the edge case when a WatchedPullRequest doesn't have an instance (i.e. its instance is None), this will add None to the list. It shouldn't; you can add an `if pull_request.instance:`.
(This „edge case“ happens in Ocim stage and production; we have around 15 WPR like that.)

userprofile/admin.py
```diff
@@ -21,26 +21,57 @@
...
+        :param obj: The instance we this action is related to
+        :return: An HTML of the actions we want to display.
+        """
+        now = timezone.now()
+
+        if obj:
+            return format_html(
+                '<a class="button" target="_blank" href="{}">Invoice Report</a>',
+                reverse('reports:report', kwargs={
+                    'organization': obj.github_handle,
```
clemente: If an organization doesn't have a github handle, could you avoid the link? (otherwise it includes `/None/2018/10/`). You can add it to the if.

reports/helpers.py
```diff
@@ -0,0 +1,252 @@
...
+                    }
+                ]
+    """
+    appservers_charges = []
+    appservers_total = 0
+
+    if invoice_month > timezone.now():
+        return appservers_charges, appservers_total
+
+    for server in instance.appserver_set.all():
```
clemente: What happens with servers that failed configuration? They will still be billed, right? They were created but were not „terminated“, so now they are accounted as active and billed.
They should be excluded. We don't need „active appservers“; I think it's best to explicitly get „terminated“ + „running“ appservers.
From this `reports` app, you can filter by AppServer status, in order to get only the `Running` and `Terminated` ones. You can do it in a QuerySet if it's clean, otherwise [here's how](https://github.com/open-craft/opencraft/blob/70235cd077eee7c44b6bb4785bd9a76dad4efd57/registration/approval.py#L66). I think we don't need to bill the other statuses (Error, ConfiguringServer, …) though we may rethink this in the future.

This will also require a test…

reports/templates/report.html
```diff
@@ -0,0 +1,68 @@
...
+            <tbody>
+            {% for instance, appservers in fork_data.instances.items %}
+              <tr>
+                <td style="text-align: left;">{{ instance }}</td>
+                <td></td>
+                <td></td>
+                <td></td>
+                <td></td>
+              </tr>
+              {% for appserver in appservers %}
```
clemente: Appservers with 0 days (e.g. because they're much newer than the report's month) will be shown as 0€, but it's informational and for now we can remove it manually from the invoice if it makes it too long. We could add an `if` here later if we want to disable them.

reports/helpers.py
```diff
@@ -0,0 +1,252 @@
...
+from datetime import timedelta
+from collections import defaultdict
+
+from django.conf import settings
+from django.utils import timezone
+
+from pr_watch.models import WatchedPullRequest
+
+
+def _trunc_datetime(date):
```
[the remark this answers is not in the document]
ahmedaljazzar: It's `normalize_month` now.

reports/helpers.py
```diff
@@ -0,0 +1,252 @@
...
+    return date.replace(day=1, hour=0, minute=0, second=0, microsecond=0)
+
+
+def get_billing_dates(appserver, invoice_month):
+    """
+    This function will be responsible of fetching the first and the last
+    billing days in a given month for a given server.
+    This can be tricky as there are a lot of cases to cover everything, but
+    basically here are the general rules that we follow to generate this.
+
```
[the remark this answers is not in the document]
ahmedaljazzar: True, I tried to minimize the number of cases as well, but it'll be reduced to either longer statements or more nested blocks, so I kept them like this to make the code more readable without the need of adding a comment next to each statement to describe it.

reports/tests.py
```diff
@@ -0,0 +1,504 @@
...
+            branch_name='new-tag',
+            ref_type='tag',
+            watched_fork=watched_fork2,
+            instance=instance3
+        )
+
+        return watched_fork1, watched_fork2
+
+    @staticmethod
+    def _generate_invoice_month(year=2018, month=1, this_month=False):
```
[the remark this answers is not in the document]
ahmedaljazzar: I updated the documentation as suggested, but kept it generate because I don't want it to imply that we're getting a date that already exists in somewhere else.

reports/tests.py
```diff
@@ -0,0 +1,504 @@
...
+                'days': expected_days,
+                'charge': settings.BILLING_RATE * expected_days,
+            })
+
+    def get_instance_charges_future_month(self):
+        """
+        Test that the instance will return empty charges and 0 total for its
+        AppServers if the invoice month is in future.
+        """
+        invoice_month = self._generate_invoice_month(year=2019)
```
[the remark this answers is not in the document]
ahmedaljazzar: That just slipped from me, thanks for the catch.

instance/tests/models/factories/openedx_appserver.py
```diff
@@ -49,4 +50,68 @@ def make_test_appserver(instance=None, s3=False, server=None):
...
+        set_appserver_configuration_failed(appserver)
+    elif status == AppServerStatus.Error:
+        set_appserver_errored(appserver)
+    elif status == AppServerStatus.Terminated:
+        set_appserver_terminated(appserver)
+
     return appserver
+
+
+def set_appserver_terminated(appserver):
```
clemente: Moving these functions from `test_openedx_instance.py` to `factories/openedx_appserver.py` feels bad for many reasons:
they seem to still be test functions only, with no other real use outside tests (we don't want servers moving so fast between states for no reason).
In addition, they are moved into the `factories` folder, but they are not factories; a better place for these functions would be directly somewhere in the server objects: `appserver.set_configuration_failed`.
I'd argue that the existing `make_test_appserver` that we already have at `factories/openedx_appserver.py` isn't at the right place either, because it's used for tests only, and should be somewhere in the tests.
By the way, there's 8-space indentation.
Could you leave these functions near the tests instead of near the implementation?

reports/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+        'billing_end': billing_end,
+        'days': days,
+        'charge': charge,
+    }
+
+
+def get_instance_charges(instance, invoice_period):
+    """
+    Generates the charges for the given instance's terminated and running
+    AppServers in the given period of the invoice date.
```
clemente: „in the given period of the invoice date“: That's not clear. `invoice_period` isn't a period (it's not a tuple which specifies a start and an end), it's a date but it represents a month. All 3 names are confusing.
You could just leave it as „in the given month“.

reports/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+
+        appservers_charges.append(charge_details)
+        appservers_total += charge_details['charge']
+
+    return appservers_charges, appservers_total
+
+
+def generate_watched_forks_instances(organization):
+    """
+    Extracts all of the watched forks for a specific organization overtime
```
clemente: I think this just needs to be `over time`

reports/tests.py
```diff
@@ -0,0 +1,521 @@
...
+        # Anonymous user
+        response = self.client.get(invoice_url)
+        self.assertRedirects(
+            response, expected_redirect,
+            status_code=302,
+            target_status_code=200,
+            msg_prefix=''
+        )
+
+        # Normal User
```
clemente: Could you use ddt to avoid repeating code? There are [other tests that try several unauthorized users](https://github.com/open-craft/opencraft/blob/fbd28cea4962056a70078098f90a0f5e7f8b01f5/instance/tests/api/test_openedx_appserver.py#L111). Could you also add user4 (instance manager) to the test? (instance managers shouldn't see the reports; it's important to include this in the checks).

reports/tests.py
```diff
@@ -0,0 +1,521 @@
...
+
+        # Watched Fork 3 has no instances
+        self.assertIsInstance(watched_forks[fork3], list)
+        self.assertEqual(len(watched_forks[fork3]), 0)
+
+    @mock.patch('reports.helpers.get_instance_charges')
+    def test_generate_charges(self, instance_charges_mock):
+        """
+        Make sure that we're able to generate charges for all of the appservers
+        that belongs to any fork under a given organization.
```
clemente: that belong

reports/tests.py
```diff
@@ -0,0 +1,521 @@
...
+        The test here will pass if all forks with their AppServers are included
+        in the billing data each associated with all charges that belong to its instances.
```
clemente: Should there be a comma before `each`? If you mean that „each [should be] associated with all charges that belong to the instances“, you can also add an `and` („the test will pass if […] and each one is associated with all charges […]“).

instance/tests/models/factories/openedx_appserver.py
ahmedaljazzar: I moved them to the factories module as they help to creating a test instance at a particular state. I also checked that all modules inside `factories` are not being used outside the test context in any way.

> By the way, there's 8-space indentation.
This has been fixed and that what caused the commits difference between this and jazzar/billing-reports-all-commits branch

reports/helpers.py
ahmedaljazzar: True, I only tried to get a good name for it to not make it confusing more as it's a date in a specific month of an invoice that's going to be normalized. But I was thinking since the view is going to generate a `datetime` object for the invoice start date, we can name it `invoice_start_date` which might reduce the confusion.

instance/tests/models/factories/openedx_appserver.py
```diff
@@ -49,4 +50,68 @@ def make_test_appserver(instance=None, s3=False, server=None):
...
+
+def _set_appserver_errored(appserver):
+    """
+    Transition `appserver` to AppServerStatus.Error.
+    """
+    appserver._status_to_waiting_for_server()
+    appserver._status_to_error()
+
+
+def _set_server_ready(server):
```
clemente: Noone is using this function so I think it should be removed. It's probably not even about AppServer but maybe about Server.

instance/tests/models/factories/openedx_appserver.py
```diff
@@ -49,4 +50,68 @@ def make_test_appserver(instance=None, s3=False, server=None):
...
+    """
+    Transition `server` to Status.Ready.
+    """
+    server._status_to_building()
+    server._status_to_booting()
+    server._status_to_ready()
+
+
+def _set_server_terminated(server):
```
clemente: Same: Noone is using this function so I think it should be removed. It's probably not even about AppServer but maybe about Server.

instance/tests/models/test_openedx_instance.py
```diff
@@ -536,103 +536,32 @@ def test_delete_instance(self, mocks, delete_by_ref, *mock_methods):
...
-        Note that servers are allowed to transition to ServerStatus.Terminated from any state,
-        so it is not necessary to transition to another status first.
-        """
-        server._status_to_terminated()
-
-    def _create_appserver(self, instance, status):
-        """
-        Return appserver for `instance` that has `status`
-
-        Note that this method does not set the status of the VM (OpenStackServer)
```
clemente: This comment should be kept. Please include it in the documentation of make_test_appserver (document the new `status` parameter).

instance/tests/models/test_openedx_instance.py
```diff
@@ -536,103 +536,32 @@ def test_delete_instance(self, mocks, delete_by_ref, *mock_methods):
...
+    def _create_running_appserver(instance):
         """
         Return app server for `instance` that has `status` AppServerStatus.Running
         """
-        return self._create_appserver(instance, AppServerStatus.Running)
+        return make_test_appserver(instance, status=AppServerStatus.Running)
 
-    def _create_failed_appserver(self, instance, with_running_server=False):
+    @staticmethod
+    def _create_failed_appserver(instance, with_running_server=False):
```
clemente: `, with_running_server=False` seems that can be deleted (noone uses it)

reports/tests.py
```diff
@@ -0,0 +1,508 @@
...
+
+        self.username = 'user3'
+        self.password = 'pass'
+        self.organization = OrganizationFactory.create(
+            name='Organization Name',
+            github_handle='organization_github_handle'
+        )
+
+    @ddt.data(
+        'user0',  # Anonymous User
```
clemente: user0 isn't a valid user (see the users at base.py), I don't know how is this even working, maybe it redirects to some login form when it detects that the user doesn't exist.
You can use the same format used by the tests I linked to earlier, where they use `None` for anonymous and then do „if user: then login()“ (and if user is None, then there's no login and thus anonymous).

reports/tests.py
```diff
@@ -0,0 +1,508 @@
...
+    )
+    @ddt.unpack
+    def test_normalize_date(self, year, month, day, hour, minute, second, microsecond):  # pylint: disable=too-many-arguments
+        """
+        Will test the default case and the edge cases to truncate the date.
+        """
+        date = datetime(year, month, day=day, hour=hour, minute=minute, second=second, microsecond=microsecond)
+
+        new_date = _normalize_date(date)
+        expected_date = datetime(date.year, date.month, 1)  # Should keep the same year and month
```
clemente: This is running some Python code to get the expected value. In a test it would be better to explicitly say which value is expected, for instance passing it in the ddt parameters.
But this function is simple and we don't have much time, so we can leave it like this.

instance/tests/models/factories/openedx_appserver.py
```diff
@@ -22,13 +22,14 @@
...
 from instance.models.load_balancer import LoadBalancingServer
 from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
 
 # Functions ###################################################################
 
 
-def make_test_appserver(instance=None, s3=False, server=None):
+def make_test_appserver(instance=None, s3=False, server=None, status=None):
     """
     Factory method to create an OpenEdXAppServer (and OpenStackServer).
```
clemente: Documentation about `status` (see comment from the original function).

clemente:
:+1:

@ahmedaljazzar could you squash? Then you can merge and deploy to production.
Other additions (removing github handle etc.) can be done in a later PR.

  - [X] I tested this: many servers, different dates, different states running, some edge cases. I saw that many servers are excluded from the report (e.g. failed servers), that's the right behaviour though it can be a bit confusing if we don't find some servers in the report 
  - [X] I read through the code: a lot
  - [ ] Includes documentation: not much, only the name „reports“. But we don't require it

PR #359, opened 2018-10-02
Shutdown a server before termination.

UmanShahzad:
TASK: [OC-5408](https://tasks.opencraft.com/browse/OC-5408)

This PR is made in an effort to make OCIM AppServer closes gracefully which was not the case. The approach followed here is to shut the AppServer off and then terminating it rather than doing a hard termination.

### DIFF
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -427,17 +427,25 @@ class OpenStackServer(Server):
 
     def terminate(self):
         """
-        Terminate the server
+        Stop and terminate the server.
+
+        We explicitly stop and wait for a graceful shutdown of the server before terminating it.
+        This ensures any daemons that need to perform cleanup tasks can do so, or that any
+        shutdown scripts/tasks get executed.
         """
-        self.logger.info('Terminating server (status=%s)...', self.status)
         if self.status == Status.Terminated:
             return
-        elif not self.vm_created:
+
+        self.logger.info('Terminating server (status=%s)...', self.status)
+        if not self.vm_created:
+            self.logger.info('Note: server was not created when terminated.')
             self._status_to_terminated()
             return
 
         try:
-            self.os_server.delete()
+            os_server = self.os_server
+            self._shutdown(os_server)
+            os_server.delete()
         except novaclient.exceptions.NotFound:
             self.logger.error('Error while attempting to terminate server: could not find OS server')
             self._status_to_terminated()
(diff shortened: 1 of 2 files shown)


instance/models/server.py
```diff
@@ -449,3 +457,17 @@ def terminate(self):
...
+
+    def _shutdown(self, os_server, poll_interval=10, max_wait=600):
+        """
+        Shutdown the server and wait to return until shutdown or wait threshold reached.
+
+        We don't have an explicit state for this and don't catch exceptions, which is why this is private.
+        The caller is expected to handle any exceptions and retry if necessary.
+        """
+        os_server.stop()
+        while max_wait and os_server.status != 'SHUTOFF':
```
ahmedaljazzar: @UmanShahzad Just noticed this, why don't we post a thread for [terminate()](https://github.com/open-craft/opencraft/pull/359/files#diff-c40bbeb7cb644b7f9da62d92619b34e6R428) instead of keeping the process waiting for it to be completed after a certain amount of time?

instance/models/server.py
```diff
@@ -449,3 +457,17 @@ def terminate(self):
                 self._status_to_unknown()
         else:
             self._status_to_terminated()
+
+    def _shutdown(self, os_server, poll_interval=10, max_wait=600):
```
ahmedaljazzar: Do you think it's better to have the `max_wait` as system-wide variable to control the grace period from the settings instead of here?

instance/models/server.py
UmanShahzad: Commented on the ticket.

instance/models/server.py
UmanShahzad: That wouldn't hurt, yep!

instance/models/server.py
```diff
@@ -458,14 +458,16 @@ def terminate(self):
...
 
-    def _shutdown(self, os_server, poll_interval=10, max_wait=600):
+    def _shutdown(self, os_server, poll_interval=10):
         """
         Shutdown the server and wait to return until shutdown or wait threshold reached.
 
         We don't have an explicit state for this and don't catch exceptions, which is why this is private.
         The caller is expected to handle any exceptions and retry if necessary.
         """
+        max_wait = settings.SHUTDOWN_TIMEOUT
```
UmanShahzad: @ahmedaljazzar We should still keep this as an argument to allow it to be modified on-call, but it can default to this value in there.

instance/models/server.py
```diff
@@ -427,17 +427,25 @@ def reboot(self, reboot_type='SOFT'):
...
+        self.logger.info('Terminating server (status=%s)...', self.status)
+        if not self.vm_created:
+            self.logger.info('Note: server was not created when terminated.')
             self._status_to_terminated()
             return
 
         try:
-            self.os_server.delete()
+            os_server = self.os_server
+            self._shutdown(os_server)
```
UmanShahzad: We can do a small check here to see if it wasn't shut down, so we can log a message notifying the user:

```python
if not self._shutdown(os_server):
    self.logger.warning("Server has not reached SHUTOFF state after max wait time; terminating forcefully.")
```

PR #360, opened 2018-10-09
SE-224 Move to using django.contrib.postgres.fields.JSONField

earaujoassis:
TASK: [SE-224](https://tasks.opencraft.com/browse/SE-224)

This PR explicitly uses `django.contrib.postgres.fields` module for
the `JSONField` model's fields class, replacing the `django_extensions`
solution and definitions.

It runs a migration in order to alter the column type from `TEXT` to
Postgresql's `JSONB` (which is used by the `JSONField` class from the
`django.contrib.postgres.fields` module).

### Test Instructions
- First, you need to apply the last migration from this PR
- Run the project's tests
- Make sure it doesn't trigger any error

### Reviewers
- [ ] @clemente


instance/admin.py
```diff
@@ -24,7 +24,7 @@
 
 from django.contrib import admin
 from django.contrib.contenttypes.admin import GenericStackedInline
-from django_extensions.db.fields.json import JSONField
+from django.contrib.postgres.fields import JSONField
```
UmanShahzad: Won't we need a new migration that updates the fields for existing users of Ocim who've already run all of these migrations?

earaujoassis: Hi, @UmanShahzad. I'm going to move to a new migration scheme and test for it in new test cases. From what I can see, it shouldn't have any major difference between these two approaches, but architecturally, it makes sense to create a new migration. Thank you for pointing that out.

UmanShahzad: If Postgres has to run an `ALTER TABLE` command to make the change internally, then we have no choice but to make a new migration for current users of Ocim.

earaujoassis: Hi, @UmanShahzad. I had to explicitly convert the VARCHAR/TEXT field into a JSON hstore through Postgresql, using the original field/column (just like described in the migration file). I'm finishing tests for that; once I have those finished, I'll let you know. Thanks.

instance/migrations/0109_alter_json_fields.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0108_manage_own_permission'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "ALTER TABLE instance_openedxappserver ALTER COLUMN openstack_server_base_image TYPE JSON USING openstack_server_base_image::JSON;",
```
smarnach: Just wondering here – I expected that Django's built-in `JSONField` uses `JSONB` database fields, whereas this command ssems to set the column type to `JSON`.  See also the [`JSONField` documentation](https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#jsonfield).

To be clear, I haven't tested whether this works as intended; I'm just wondering why the column type isn't what I expect here.

UmanShahzad: Also in case this fails, do we have/need a migration going backwards?

smarnach: We sometimes have migrations that only work one way, but since it's easy to add the backwards migrations here, I'd be in favour of doing so.

earaujoassis: @smarnach I'll update it to use JSONB instead and create a backwards migration as well. Thank you!

smarnach: @earaujoassis I'm not sure what is correct, but I believe that only the correct version should work.  So you should be able to find out what right by simply trying it.

Another thing I just noticed – don't we need to modify Django's notion of the models in the migration as well?  Otherwise Django believes we are still using the old type.

I don't really know what the migration code is supposed to look like, so I'm asking you to figure it out. :)

earaujoassis: @smarnach Django's models using that field were updated as well. We just need to update which class field is being used by the model-mapping, which is done in the following file: `instance/models/openedx_appserver.py`

earaujoassis: Please correct me if I misunderstood your message. Thank you!

earaujoassis:
Please, reviewer(s), double check the column's and table's names. It worked fine here, but I'm explicitly asking for a double check. Many thanks!

instance/migrations/0109_alter_json_fields.py
smarnach: @earaujoassis Yeah, my wording was probably confusing.  Django tracks the state of each model in the migrations, and compares the state to what's in the code.  The `RunSQL` directives you added are opaque to Django, so when resolving the state in the migration, it will still end up with the old type of the fields.  So if you run `makemigrations` on your branch, Django still generates `AlterField` directives.

I just did this locally and looked at the SQL output of the automatically generated migrations.  It looks like Django is able to do these migrations automatically, and you don't need the manually written SQL.  I haven't actually tested it, but on the surface it looks promising.

So I believe all you need to do is

* remove the manually written migration.
* run `honcho run ./manage.py makemigrations`.
* test whether it actually works correctly.

smarnach:
@earaujoassis Thanks for the updates.  This looks good to me now, but I'll leave it to @clemente to test this.

clemente:
:+1:

`JSONField` is used only in these 2 fields (both fields are in OpenEdXAppServer and OpenEdXInstance):
- openstack_server_flavor
- openstack_server_base_image

I verified that migrations affect these two and these two only.

- [X] I tested this: (describe what you tested)
  - If I input invalid JSON (e.g. `xxxxx`, or normal JSON but with single quotes instead of double quotes): with the **old field**, I get `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` and a 500 error (not nice). With the **new one**, I got a form validation error (so this is better).
  - If I leave the field blank: with the old one, it switches it to `{}` (fine). With the new one, to `null` (yes, those 4 letters). *This is a behaviour change*. The new one differentiates between `{}` and `null` (both work). I checked from the shell that Python considers both as not equal (`null` is `None`, `{}` isn't). Makes sense because we requested `null=True`. For the use that we do in those two fields, this new distinction is fine (though we don't need it because we never leave those fields empty)
  - I saved the settings in an instance, then spawned an appserver. The values were transferred to the appserver
  - I saw the values from the UI, in a server's „Configuration“ tab. I tried with `{}`, `null` (which shows empty) and a typical value like `{"name":"vps-ssd-3"}`, all are shown correctly
  - I started deploying a new server from my local machine, and iff I set the right JSON values (OpenStack flavour, Ubuntu image) it started the deployment and went past the `Booting` status
- [X] I read through the code
- [X] I checked for accessibility issues: none
- [X] Includes documentation: none required. It deletes the right code.

@earaujoassis Could you rebase on top of master? After that, you can merge (I can deploy later).

earaujoassis:
@clemente it is ready for deployment. thank you!

clemente:
Thanks, deployed

PR #361, opened 2018-10-12
Minor fixes: organization regexp, documentation typo in "manage_own"

clemente:
- accept dashes in organization slug in URL. The regexp comes from documentation like https://simpleisbetterthancomplex.com/references/2016/10/10/url-patterns.html
- fix typo (manage_all doesn't exist, now it's manage_own)

### DIFF
--- a/instance/models/instance.py
+++ b/instance/models/instance.py
@@ -111,7 +111,7 @@ class InstanceReference(TimeStampedModel):
 
         Instance managers are those users that can see a list of instances (at least their own).
         Superusers are automatically instance managers and will see all instances.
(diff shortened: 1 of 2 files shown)


ahmedaljazzar:
LGTM 👍Though the actual pattern for Github usernames is `^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$` which doesn't allow some use cases like:
- `ahmed-jazzar-`
- `-ahmed-jazzar`
- `ahmed--jazzar`

However, this change is not about validating the github username field, but to let the reports url able to recognize it.

- [x] I tested this: verified the new pattern on the different organizations' names.
- [x] I read through the code
- [x] Includes documentation: n/a

PR #366, opened 2018-11-23
[BB-550] Add generation of asymmetric JWT keys

giovannicimolin:
This PR adds the automatic generation of asymmetric JWT keys when deploying appservers.

It working by creating and storing a RSA key for each instance and generating public and private JWK sets for each AppServer by setting the following Ansible variables: `COMMON_JWT_PUBLIC_SIGNING_JWK_SET` and `EDXAPP_JWT_PRIVATE_SIGNING_JWK`.

This is needed because [EdX recently switched from HMAC signed JWT keys to asymmetrically encrypted JWT keys.](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0008-use-asymmetric-jwts.rst)

**Dependencies**: ~https://github.com/edx/configuration/pull/4871 [Already merged].~

**Sandbox URL**: TBD - This PR will be deployed for testing on Ocim stage.

**Merge deadline**: Sprint 182.

**Testing instructions**:

1. Set up the Ocim devstack with this branch.
2. Deploy an instance with this configuration:
```
configuration_source_repo_url   https://github.com/edx/configuration
configuration_version                 master
edx_platform_repository_url       https://github.com/edx/edx-platform
edx_platform_commit                 release
openedx_release                       master
```
3. Create a new AppServer and check on it's configurations if `COMMON_JWT_PUBLIC_SIGNING_JWK_SET` and `EDXAPP_JWT_PRIVATE_SIGNING_JWK` are set.
4. Wait for the end of the provisioning and try to log in the LMS, if the configuration worked correctly, you should be able to login.

1. This PR includes a migration for storing the RSA key used to generate the JWT keys, because the keys for each AppServer must be compatible in order to preserve interoperability between them.
2. I've stored the RSA private key but I'm unsure that is the best option for this case.

**Reviewers**
- [ ] @lgp171188


requirements.txt
```diff
@@ -85,18 +88,19 @@ pep8==1.7.0
...
 prospector==0.12.4
 psycopg2==2.7.3.2
 ptyprocess==0.5.1
 pycodestyle==2.0.0
+pycryptodomex==3.7.0
 pydocstyle==1.0.0
 pyflakes==1.3.0
 Pygments==2.1.3
-# Pylint 1.6.* does not work with prospector 0.12.4, so we have to use 1.5.6 for now.
-# See: https://github.com/landscapeio/prospector/issues/183
```
lgp171188: @giovannicimolin, why have you removed this comment? This needs to be present just above the line with the pylint requirement.

requirements.txt
```diff
@@ -70,6 +72,7 @@ mysqlclient==1.3.9
 netaddr==0.7.19
 netifaces==0.10.5
 oauthlib==2.0.0
+olefile==0.46
```
lgp171188: @giovannicimolin, why is `olefile` needed for this PR? It appears to be totally unrelated.

requirements.txt
```diff
@@ -110,6 +114,7 @@ python-consul==1.1.0
 python-dateutil==2.6.0
 python-glanceclient==2.5.0
 python-keystoneclient==3.8.0
+python-magic==0.4.15
```
lgp171188: @giovannicimolin, why is `python-magic` needed? It is unused in this PR and appears to be unrelated.

requirements.txt
giovannicimolin: I've ran pip freeze > requirements.txt, just fixed it by only adding the necessary requirements at the bottom of the file.

instance/management/commands/instance_redeploy.py
```diff
@@ -225,6 +225,8 @@ def _pending_instances(self):
                 self.success_tag,
                 self.failure_tag,
             ]
+        ).order_by(
```
UmanShahzad: This very likely doesn't need to have separated lines. The others were separated to allow the whole chain to separate with adding `\`, but if the ending query is so small now that it can probably go on one line.

UmanShahzad: And thanks for the fix :)

UmanShahzad: Oh, @lgp171188 is going to fix this in https://github.com/open-craft/opencraft/pull/368/files.

giovannicimolin: @UmanShahzad I've seem that too. I'll remove these modifications when the PR is approved so I rebase the repo only once.

giovannicimolin:
I've removed the fixes for the failing tests as they are not related to this PR.
Rebased from 11ece7a to 0db4f28.

instance/models/mixins/secret_keys.py
```diff
@@ -104,6 +120,17 @@ class SecretKeyInstanceMixin(models.Model):
         ),
     )
 
+    secret_key_rsa_private = models.TextField(
+        default=functools.partial(generate_rsa_key, 1024),
```
lgp171188: @giovannicimolin, is there a reason why used `1024` bits for the RSA key? 1024 bits are no longer considered secure and the recommendation is to use 2048 bits or more.

instance/models/mixins/secret_keys.py
```diff
@@ -114,12 +141,47 @@ def secret_key(self):
         """
         return b64decode(self.secret_key_b64encoded)
 
+    @property
+    def rsa_key(self):
+        """
+        Return the RSA key as a Cryptodome.RSA object.
+        """
+        if not self.secret_key_rsa_private:
```
lgp171188: @giovannicimolin, doesn't it make sense to generate it if it doesn't exist?

lgp171188:
@giovannicimolin, I did another review pass and left a couple of comments. Please check those out. In the mean time, I will test this and approve the PR after that.

instance/models/mixins/secret_keys.py
giovannicimolin: Yes! I'll make those changes.

instance/models/mixins/secret_keys.py
giovannicimolin: I'll change that!

giovannicimolin:
@lgp171188 I've just resolved the PR comments and commited new code. I've also added a new test since a key can be generated when `self.rsa_key` is called.

instance/migrations/0113_auto_20181129_1622.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
```
lgp171188: @giovannicimolin, since this change is still in development, you should squash these migrations into one.

instance/models/mixins/secret_keys.py
```diff
@@ -51,6 +55,18 @@ def generate_secret_key(char_length):
     return random_string
 
 
+def generate_rsa_key(key_size):
+    """
+    Creates a key_size-bit RSA key and returns private key
+    """
+    if key_size <= 1023 or key_size % 1024:
```
lgp171188: @giovannicimolin, is this check needed?

instance/models/mixins/secret_keys.py
```diff
@@ -114,12 +141,45 @@ def secret_key(self):
...
         return b64decode(self.secret_key_b64encoded)
 
+    @property
+    def rsa_key(self):
+        """
+        Return the RSA key as a Cryptodome.RSA object.
+        """
+        if not self.secret_key_rsa_private:
+            self.secret_key_rsa_private = generate_rsa_key(2048)
+            self.secret_key_rsa_private.save()
```
lgp171188: @giovannicimolin, is there a `self.<field>.save()` method? How does it differ from `self.save()`. Also when calling this from an instance retrieved from the database, does it persist changes to the database and also update the instance from which it was called?

instance/migrations/0113_auto_20181129_1622.py
giovannicimolin: @lgp171188 ok!

instance/models/mixins/secret_keys.py
lgp171188: This is not needed.

lgp171188:
@giovannicimolin, this is ready for merge once the CI checks pass. Please squash all the commits into one before merging and make sure the squash+merge commit has a meaningful commit message describing the feature implemented.

giovannicimolin:
Rebased from **a60fd15** to **c6fabd2**. Squashing commits and writing a good commit message.

giovannicimolin:
Fixed typos and added more information in the commit (**c6fabd2** to **f029055**).

PR #367, opened 2018-11-26
Add expiration rule to S3 bucket at creation.

Metfriet:
This PR adds code to make S3 buckets versioned and have the old versions expire after 30days.

**Jira Ticket**: BB-582

### DIFF
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -26,6 +26,8 @@ import time
 
 import boto
 from boto.s3.lifecycle import Lifecycle, Expiration, Rule
+import boto3
+
 
 from django.db.backends.utils import truncate_name
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


instance/models/mixins/storage.py
```diff
@@ -276,6 +277,14 @@ def _create_bucket(self, attempts=4, ongoing_attempt=1, retry_delay=4, location=
                 else:
                     raise e
             bucket.set_cors(get_s3_cors_config())
+            bucket.configure_versioning(versioning=True)
```
UmanShahzad: @Metfriet Just to be sure, this function is called for any instance upon provisioning a VM, even if the instance already existed with VMs in the past?

Metfriet: @UmanShahzad, anytime the spawn_appserver method in OpenEdXInstance gets called for an instance with s3 storage this function gets called (via some other methods) and that should happen regardless of whether or not there was a VM before right?

PR #370, opened 2018-12-05
Configure PROFILE_IMAGE_BACKEND when using AWS s3

GustavoKatel:
See SE-559

Turns out that to upload profile images to aws, the profile images backend has to be configured similar to the default file storage. This PR adds the ansible variable to the `configuration_storage_settings` entry in OCIM.

Ref: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L688

**Sandbox URL**: 
Sandbox with the current change in the extra settings: https://gustavo-profile-image-test.stage.opencraft.hosting/

**Testing instructions:**

1. Set up the Ocim devstack with this branch.
2. Deploy an instance with default configurations and using s3.
3. Create a new AppServer and check on its configurations if `EDXAPP_PROFILE_IMAGE_BACKEND` is set and matches:

```yaml
EDXAPP_PROFILE_IMAGE_BACKEND:
  class: storages.backends.s3boto.S3BotoStorage
  options:
    headers:
      Cache-Control: max-age-"{{ EDXAPP_PROFILE_IMAGE_MAX_AGE }}"
    location: '{{ EDXAPP_AWS_LOCATION }}/profile-images'
```
4. After the provision, set a profile picture for one of its user and check if the url points to aws s3.

**Reviewers**
* [ ] @symbolist


instance/models/mixins/openedx_storage.py
```diff
@@ -48,6 +48,13 @@ def swift_container_names(self):
         """
         return [self.swift_container_name]
 
+    @property
+    def s3_custom_domain(self):
```
symbolist: Can you move this to `/mixins/storage.py#S3BucketInstanceMixin`? There are a few other related properties there.

instance/models/mixins/openedx_storage.py
```diff
@@ -61,13 +68,29 @@ def _get_s3_settings(self):
             # using shared s3 buckets
             "EDXAPP_AWS_LOCATION": self.swift_container_name,
             "EDXAPP_DEFAULT_FILE_STORAGE": 'storages.backends.s3boto.S3BotoStorage',
+
+            # Set up S3 image backend
+            # https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L688
```
symbolist: This reference will get out of date so we can leave it out. :)

instance/models/mixins/openedx_storage.py
```diff
@@ -61,13 +68,29 @@ def _get_s3_settings(self):
...
+            "EDXAPP_PROFILE_IMAGE_BACKEND": {
+                "class": "storages.backends.s3boto.S3BotoStorage",
+                "options": {
+                    "bucket": self.s3_bucket_name,
+                    "location": '{}/{}'.format(self.swift_container_name, 'profile-images'),
+                    "custom_domain": self.s3_custom_domain,
+                    "access_key": self.s3_access_key,
+                    "secret_key": self.s3_secret_access_key,
```
symbolist: Hmm. Since django-storages [defaults to the global ones](https://github.com/jschneier/django-storages/blob/1.4.1/storages/backends/s3boto.py#L206) can we get away with not passing the ones which are already set? I think only `location` and `headers` are unique here.

instance/tests/models/test_openedx_storage_mixins.py
```diff
@@ -21,18 +21,22 @@
...
 from django.test.utils import override_settings
 
-from instance.models.mixins.storage import get_s3_cors_config, get_master_iam_connection, StorageContainer
+from instance.models.mixins.storage import (StorageContainer,
+                                            get_master_iam_connection,
+                                            get_s3_cors_config)
 from instance.tests.base import TestCase
-from instance.tests.models.factories.openedx_appserver import make_test_appserver
-from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
+from instance.tests.models.factories.openedx_appserver import \
```
symbolist: edX uses 120 as line length so you may want to set that in your IDE and undo these formatting changes.

instance/models/mixins/openedx_storage.py
GustavoKatel: You're right! Thanks for the tip :)

PR #371, opened 2018-12-07
Set CONN_MAX_AGE default to 0

kaizoku:
This sets both the CONN_MAX_AGE where manually specified  to 0,
and uses the `edx_django_service_default_db_conn_max_age` variable
to set the default CONN_MAX_AGE for all edx django services.

### DIFF
--- a/instance/models/mixins/openedx_database.py
+++ b/instance/models/mixins/openedx_database.py
@@ -250,7 +250,7 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
                     "HOST": "{{ ECOMMERCE_DATABASE_HOST }}",
                     "PORT": self.mysql_server.port,
                     "ATOMIC_REQUESTS": True,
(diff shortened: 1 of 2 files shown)


instance/models/mixins/openedx_database.py
```diff
@@ -345,6 +345,7 @@ def _get_mysql_settings(self):
             "COMMON_MYSQL_READ_ONLY_PASS": self._get_mysql_pass(self.read_only_user),
             "COMMON_MYSQL_ADMIN_USER": self.admin_user,
             "COMMON_MYSQL_ADMIN_PASS": self._get_mysql_pass(self.admin_user),
+            "edx_django_service_default_db_conn_max_age": 0,
```
UmanShahzad: This doesn't look like it should be under the "Common users" header. Maybe we'll have to make a new header for it and similar vars?

kaizoku: Good point, it didn't exactly fit anywhere else.
I created a new section for this variable and other global options.

PR #372, opened 2018-12-07
BB-690 - Add new integration cleanup script WIP

giovannicimolin:
Add new integration cleanup script to delete all AWS provisioned resources for CircleCI build that where left behind. Also cleans up OpenStack instances and DNS records.

Features:
- [x] Deletion of S3 buckets, policies, keys and users
- [x] Shutdown of VM's older than 3 days
- [x] Deletion of DNS records associated to integration 

**JIRA tickets**: [BB-690](https://tasks.opencraft.com/browse/BB-690)

**Merge deadline**: Sprint 183.

**Testing instructions**:

1. Clone this branch of OCIM
2. Create a `virtualenv` and install requirements with `pip install -r requirements.txt`
3. Make sure you have the following environment variables set (from CircleCI env):
```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
OPENSTACK_AUTH_URL
OPENSTACK_USER
OPENSTACK_PASSWORD
OPENSTACK_TENANT
OPENSTACK_REGION
GANDI_ZONE_ID
GANDI_API_KEY
```
4. To test the cleanup tool without actually deleting anything, run:
```
# This will execute the tool, query every old S3 bucket, user, DNS records and OVH machines running that are related to CircleCI and are older than 3 days but not delete anything
./integration_cleanup.py --dry_run
```

1. I'm aware that this cleanup tool doesn't clean up all DNS records.

**Reviewers**
- [ ] @lgp171188


integration_cleanup/dns_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+              skipped_count
+        ))
+
+        print("Found {} entries related to old instances. Starting deletion process...".format(
+            len(records_to_delete)
+        ))
+
+        # Delete entries
+        for record in records_to_delete:
+            print("  > DNS Entry {}...".format(record['id']))
```
lgp171188: @giovannicimolin, I think it is better to use the `logging` module for logging the information and configure it to print it on the console and also write it to a file which can be an artifact of the build.

integration_cleanup/aws_cleanup.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Deletes a S3 bucket and all of it's files
+        """
+        try:
+            print("Deleting bucket {}.".format(bucket_name))
+            bucket = self.s3_resource.Bucket(bucket_name)
+            bucket.object_versions.delete()
+            bucket.objects.all().delete()
+            bucket.delete()
+        except self.s3_client.exceptions.NoSuchBucket:
```
lgp171188: @giovannicimolin, is this the only exception to be handled? In Ocim code, we handle the `boto.exception.S3ResponseError` while trying to delete a bucket. Can you check and add a handler + logging statement to log the error?

integration_cleanup/dns_cleanup.py
```diff
@@ -0,0 +1,126 @@
+import xmlrpc.client
+
+
+BATCH_SIZE = 100
+
+
+class DnsCleanupInstance:
```
lgp171188: @giovannicimolin, can't you reuse the code in `gandi.py`?

giovannicimolin: Just did this. I'll test it now.

integration_cleanup/aws_cleanup.py
giovannicimolin: I've put `botocore.exceptions.ClientError` to be a little more generic when catching errors relaetd to the deletion of a bucket

cleanup_utils/aws_cleanup.py
```diff
@@ -0,0 +1,227 @@
...
+            bucket.object_versions.delete()
+            bucket.objects.all().delete()
+            bucket.delete()
+        except self.s3_client.exceptions.NoSuchBucket:
+            # Ignore if the bucket doesn't exist
+            logger.warning("The bucket %s did not exist, skipping.", bucket_name)
+        except botocore.exceptions.ClientError as e:
+            # Ignore if the bucket doesn't exist
+            logger.error(
+                "Untreated error when deleting %s: %s",
```
lgp171188: @giovannicimolin `s/Untreated/Unhandled/`

cleanup_utils/aws_cleanup.py
```diff
@@ -0,0 +1,227 @@
...
+            logger.info("Deleting bucket %s.", bucket_name)
+            bucket = self.s3_resource.Bucket(bucket_name)
+            bucket.object_versions.delete()
+            bucket.objects.all().delete()
+            bucket.delete()
+        except self.s3_client.exceptions.NoSuchBucket:
+            # Ignore if the bucket doesn't exist
+            logger.warning("The bucket %s did not exist, skipping.", bucket_name)
+        except botocore.exceptions.ClientError as e:
+            # Ignore if the bucket doesn't exist
```
lgp171188: @giovannicimolin, this comment seems out of place since the 'bucket doesn't exist' exception is handled in the previous `except` clause.

cleanup_utils/aws_cleanup.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def get_iam_users(self):
+        """
+        Lists all IAM users
+        """
+        users = []
+        paginator = self.iam_client.get_paginator('list_users')
+        for page in paginator.paginate():
+            for user in page['Users']:
+                users.append(user)
```
lgp171188: @giovannicimolin, unless `page['Users']` is a generator, can't you just use `users.extend(page['Users'])`?

lgp171188: Similar changes can be done for all the pagination code.

cleanup_utils/aws_cleanup.py
```diff
@@ -0,0 +1,227 @@
...
+        been used in at least age_limit
+        """
+        old_keys = []
+        user_access_keys = []
+
+        # Get all user access keys
+        paginator = self.iam_client.get_paginator('list_access_keys')
+        for page in paginator.paginate(UserName=username):
+            for access_key in page['AccessKeyMetadata']:
+                user_access_keys.append(access_key)
```
lgp171188: @giovannicimolin, you could avoid using the `.extend()` method here and combine the checks in the next for loop into this one itself.

lgp171188: You can also create a nested function which returns `True` or `False` based on whether a key is old or not and use a list comprehension in the for loop to construct the list. Check whichever way is more concise and pythonic and use it.

giovannicimolin: @lgp171188 Something like this?
```
        old_keys = []

        def is_key_old(access_key):
            last_used = self.iam_client.get_access_key_last_used(AccessKeyId=access_key['AccessKeyId'])
            last_used_date = last_used.get('AccessKeyLastUsed', {}).get('LastUsedDate')
            if last_used_date and (last_used_date < self.age_limit):
                return True
            return False

        # Get all user access keys
        paginator = self.iam_client.get_paginator('list_access_keys')
        for page in paginator.paginate(UserName=username):
            old_keys2.extend([key for key in page['AccessKeyMetadata'] if is_key_old(key)])

        return old_keys
```

cleanup_utils/openstack_cleanup.py
```diff
@@ -0,0 +1,135 @@
...
+
+            # Check if it's a valid date and add UTC timezone
+            try:
+                instance_age_unaware = datetime.strptime(
+                    instance.created,
+                    '%Y-%m-%dT%H:%M:%SZ'
+                )
+                instance_age = instance_age_unaware.replace(tzinfo=pytz.UTC)
+            except ValueError:
+                instance_age = None
```
lgp171188: @giovannicimolin, it might be better to log a warning as the logging code below in the else block will happily convert `None` to `'None'` and it might be difficult to trace the log in case of issues.

lgp171188: @giovannicimolin, it will get to the `else` block because `instance_age = None` is falsey.

cleanup_utils/openstack_cleanup.py
```diff
@@ -131,9 +131,15 @@ def run_cleanup(self):
...
-            else:
+            elif instance_age:
                 logger.info(
                     "    * SKIPPING: Instance is only %s seconds old (age threshold is %s seconds).",
                     instance_age,
                     self.age_limit
                 )
+            else:
+                logger.info(
+                    "    * SKIPPING: Couldn't parse instance age.",
```
lgp171188: @giovannicimolin, it would be better use format strings to include the variables in the message or not include them at all since we know that `instance_age` is going to be `None`.

cleanup_utils/aws_cleanup.py
```diff
@@ -0,0 +1,226 @@
...
+                return True
+            return False
+
+        # Get all user access keys
+        try:
+            paginator = self.iam_client.get_paginator('list_access_keys')
+            for page in paginator.paginate(UserName=username):
+                old_keys.extend([key for key in page['AccessKeyMetadata'] if is_key_old(key)])
+        except botocore.exceptions.NoSuchEntityException:
+            logger.info("Error retrying user keys, user %s will be skipped...", username)
```
lgp171188: @giovannicimolin, `s/retrying/retrieving/` and the log level should be error.

cleanup_utils/dns_cleanup.py
```diff
@@ -0,0 +1,133 @@
...
+            # Delete record
+            try:
+                self.delete_dns_record(
+                    zone_id=self.zone_id,
+                    zone_version_id=new_zone_version,
+                    record_name=record
+                )
+            except xmlrpc.client.Fault as e:
+                logger.info("  > FAILED Deleting DNS entries for %s...", record)
+                logger.info("  > ERROR: %s", e)
```
lgp171188: @giovannicimolin, shouldn't the logging level be `error` for both lines since this is a failure of an action that has to be done?

cleanup_utils/openstack_cleanup.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Check if it's a valid date and add UTC timezone
+            try:
+                instance_age_unaware = datetime.strptime(
+                    instance.created,
+                    '%Y-%m-%dT%H:%M:%SZ'
+                )
+                instance_age = instance_age_unaware.replace(tzinfo=pytz.UTC)
+            except ValueError:
+                logger.info(
```
lgp171188: @giovannicimolin, log level should be warning or error (if you think so).

cleanup_utils/integration_cleanup.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def main():
+    """
+    Main function responsible for the cleanup
+    """
+    parser = argparse.ArgumentParser(
+        description="Cleanup all unused resources from integration runs that "
+                    "have been cancelled or failed."
+    )
```
lgp171188: @giovannicimolin, this script should cleanup all old instances irrespective of whether they were created in cancelled or failed or successful runs.

cleanup_utils/openstack_cleanup.py
```diff
@@ -0,0 +1,139 @@
...
+                    self.age_limit
+                )
+                if not self.dry_run:
+                    instance.delete()
+
+            else:
+                logger.info(
+                    "    * SKIPPING: Instance is only %s seconds old (age threshold is %s seconds).",
+                    instance_age,
+                    self.age_limit
```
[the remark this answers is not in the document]
giovannicimolin: I just changed this to:
`SKIPPING: Instance was created at %s but the cut-off age threshold is %s.`

PR #373, opened 2018-12-11
BB-692 Fix errors when deprovisioning resources that has been already deprovisioned.

xirdneh:
- Add flags to ignore errors when deleting an OpenEDX Instance.
- Add log messages to show what's going on when deleting an instance.
- Add unit test for deleting an instance ignoring errors.
- Fix some deprovision tests.
- Update responses library to mock response with 'reason' attribute.

When this task was created we thought the issue was that `OpenEDXInstance.delete()` was trying to deprovision resources that were already deprovisioned before.
This is partially true because the `delete()` method was trying to get rid of things that were not there anymore, but those resources were not "correctly" deprovisioned. Meaning, checking for `self.<resource>_provisioned` would not give us the correct state if the resource was deleted manually or using another method.
Also, the class methods implemented currently in the codebase they do check if the `<resouce>_provisioned` flag is `False` and if it so then the codebase do not try to delete that resource, e.g. [MySQL](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/database.py#L232), [Mongo](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/database.py#L343), [Swift](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/storage.py#L138), [RabbitMQ](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/rabbitmq.py#L164)

This PR implements a few flags to ignore errors raised when using `OpenEdxInstance.delete()`.
Errors will be logged but not raised.

**Test instructions:**

1. ssh to [ocim stage](https://github.com/open-craft/ansible-secrets/blob/master/host_vars/stage.console.opencraft.com.yml) and apply patch.
2. Create an instance using Django's shell:
```
from instance.models.openedx_instance import OpenEdXInstance

    name='Dogwood sandbox',
    sub_domain='dogwood',
    # The rest of the parameters are all optional:
    email='myname@opencraft.com',
    openedx_release='named-release/dogwood',
    configuration_version='named-release/dogwood',
    configuration_extra_settings='',
)
```
3. Manually delete resources:
```
instance.deprovision_mysql()
instance.deprovision_mongo()
instance.deprovision_rabbitmq()
```
4. Set `<resource>_provisioned` flag to `True`.
```
instance.mysql_provisioned = True
instance.mongo_provisioned = True
instance.rabbitmq_provisioned = True
instance.save()
```
5. Delete instance:
```
instance.delete(ignore_errors=True)
```

**Expected behavior:**

Every other resource/config should be deleted correctly (e.g. DNS records) as well as database records. There should be a few MySQL, Mongo and RabbitMQ messages printed to the output since the code will try to delete databases and/or users that are no longer there but the `delete()` method should not exit with an error.

### DIFF
--- a/instance/models/mixins/database.py
+++ b/instance/models/mixins/database.py
@@ -26,14 +26,22 @@ import functools
 import inspect
 import string
 import warnings
+import logging
 
 from django.conf import settings
 from django.db import models
 from django.utils.crypto import get_random_string
 import MySQLdb as mysql
+from MySQLdb import Error as MySQLError
 import pymongo
+from pymongo.errors import PyMongoError
 
 from instance.models.database_server import MySQLServer, MongoDBServer, MongoDBReplicaSet
+
+
+# Logging #####################################################################
+
+logger = logging.getLogger(__name__)
 
 
 # Functions ###################################################################
@@ -62,12 +70,16 @@ def _get_mysql_cursor(mysql_server):
     """
     Get a database cursor.
     """
-    connection = mysql.connect(
-        host=mysql_server.hostname,
-        user=mysql_server.username,
-        passwd=mysql_server.password,
-        port=mysql_server.port,
-    )
+    try:
+        connection = mysql.connect(
+            host=mysql_server.hostname,
+            user=mysql_server.username,
+            passwd=mysql_server.password,
+            port=mysql_server.port,
+        )
+    except MySQLError as exc:
+        logger.exception('Cannot get MySQL cursor: %s. %s', mysql_server, exc)
+        raise
     return connection.cursor()
 
 
@@ -76,6 +88,7 @@ def _create_database(cursor, database):
     """
     Create MySQL database, if it doesn't already exist.
     """
+    logger.info('Creating MySQL database: %s', database)
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
         cursor.execute('CREATE DATABASE IF NOT EXISTS `{db}` DEFAULT CHARACTER SET utf8'.format(db=database))
@@ -90,6 +103,7 @@ def _create_user(cursor, user, password):
     # so we need to use a different approach for now:
     user_exists = cursor.execute("SELECT 1 FROM mysql.user WHERE user = %s", (user,))
     if not user_exists:
+        logger.info('Creating mysql user: %s', user)
         cursor.execute('CREATE USER %s IDENTIFIED BY %s', (user, password,))
 
 
@@ -111,7 +125,12 @@ def _drop_database(cursor, database):
     """
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        cursor.execute('DROP DATABASE IF EXISTS `{database}`'.format(database=database))
+        logger.info('Dropping mysql db: %s', database)
+        try:
+            cursor.execute('DROP DATABASE IF EXISTS `{db}`'.format(db=database))
+        except MySQLError as exc:
+            logger.exception('Cannot drop MySQL database: %s. %s', database, exc)
+            raise
 
 
 def _drop_user(cursor, user):
@@ -123,7 +142,12 @@ def _drop_user(cursor, user):
     # so we need to use a different approach for now:
     user_exists = cursor.execute("SELECT 1 FROM mysql.user WHERE user = %s", (user,))
     if user_exists:
-        cursor.execute('DROP USER %s', (user,))
+        logger.info('Dropping mysql user: %s.', user)
+        try:
+            cursor.execute('DROP USER %s', (user,))
+        except MySQLError as exc:
+            logger.exception('Cannot drop MySQL user: %s. %s', user, exc)
+            raise
 
 
 def select_random_mysql_server():
@@ -225,25 +249,30 @@ class MySQLInstanceMixin(models.Model):
             self.mysql_provisioned = True
             self.save()
 
-    def deprovision_mysql(self):
+    def deprovision_mysql(self, ignore_errors=False):
         """
         Drop all MySQL databases and users.
         """
+        self.logger.info('Deprovisioning MySQL started.')
         if self.mysql_server and self.mysql_provisioned:
-            cursor = _get_mysql_cursor(self.mysql_server)
+            try:
+                cursor = _get_mysql_cursor(self.mysql_server)
 
-            # Drop default databases and users
-            for database in self.mysql_databases:
-                database_name = database["name"]
-                _drop_database(cursor, database_name)
-                _drop_user(cursor, database["user"])
-
-            # Drop users with global privileges
-            for user in self.global_users:
-                _drop_user(cursor, user)
+                # Drop default databases and users
+                for database in self.mysql_databases:
+                    database_name = database["name"]
+                    _drop_database(cursor, database_name)
+                    _drop_user(cursor, database["user"])
+                # Drop users with global privileges
+                for user in self.global_users:
+                    _drop_user(cursor, user)
+            except MySQLError:
+                if not ignore_errors:
+                    raise
 
             self.mysql_provisioned = False
             self.save()
+        self.logger.info('Deprovisioning MySQL finished.')
 
 
 class MongoDBInstanceMixin(models.Model):
@@ -331,19 +360,28 @@ class MongoDBInstanceMixin(models.Model):
             mongo = pymongo.MongoClient(database_url)
             for database in self.mongo_database_names:
                 # May update the password if the user already exists
+                self.logger.info('Creating mongo db: %s', database)
                 mongo[database].add_user(self.mongo_user, self.mongo_pass)
             self.mongo_provisioned = True
             self.save()
 
-    def deprovision_mongo(self):
+    def deprovision_mongo(self, ignore_errors=False):
         """
         Drop Mongo databases.
         """
+        self.logger.info('Deprovisioning Mongo started.')
         database_url = self._get_main_database_url()
         if database_url and self.mongo_provisioned:
             mongo = pymongo.MongoClient(database_url)
             for database in self.mongo_database_names:
                 # Dropping a non-existing database is a no-op.  Users are dropped together with the DB.
-                mongo.drop_database(database)
+                self.logger.info('Dropping mongo db: %s.', database)
+                try:
+                    mongo.drop_database(database)
+                except PyMongoError as exc:
+                    self.logger.exception('Cannot drop Mongo database: %s. %s', database, exc)
+                    if not ignore_errors:
+                        raise
             self.mongo_provisioned = False
             self.save()
+        self.logger.info('Deprovisioning Mongo finished.')
--- a/instance/models/mixins/rabbitmq.py
+++ b/instance/models/mixins/rabbitmq.py
@@ -40,9 +40,8 @@ def generate_random_vhost():
     """
     Helper function for the default value of the field `rabbitmq_vhost`.
     """
-    return '/{id}'.format(
-        id=get_random_string(length=14, allowed_chars=string.ascii_lowercase)
-    )
+    vhost_id = get_random_string(length=14, allowed_chars=string.ascii_lowercase)
+    return '/{id}'.format(id=vhost_id)
 
 
 def new_rabbitmq_user():
@@ -134,8 +133,37 @@ class RabbitMQInstanceMixin(models.Model):
                 action,
                 response.status_code
             )
-            raise RabbitMQAPIError
+            raise RabbitMQAPIError('{reason}: {content}'.format(reason=response.reason, content=response.text))
         return response
+
+    def _delete_rabbitmq_user(self, user, ignore_errors=False):
+        """Deletes a rabbitmq user.
+
+        Arguments:
+            user (RabbitMQUser): User instance.
+            ignore_errors (bool): If True RabbitMQAPIError exceptions will not be raised.
+        """
+        self.logger.info('Deleting rabbitmq user: %s', user.username)
+        try:
+            self._rabbitmq_request('delete', 'users', user.username)
+        except RabbitMQAPIError as exc:
+            self.logger.exception('Cannot delete rabbitmq user: %s. %s', user.username, exc)
+            if not ignore_errors:
+                raise
+
+    def _delete_rabbitmq_vhost(self, ignore_errors=False):
+        """Delete rabbitmq vhost for this instance.
+
+        Arguments:
+            ignore_errors (bool): If true RabbitMQAPIError exceptions will not be raised.
+        """
+        self.logger.info('Deleting rabbitmq vhost: %s', self.rabbitmq_vhost)
+        try:
+            self._rabbitmq_request('delete', 'vhosts', self.rabbitmq_vhost)
+        except RabbitMQAPIError as exc:
+            self.logger.exception('Cannot delete rabbitmq vhost: %s. %s', self.rabbitmq_vhost, exc)
+            if not ignore_errors:
+                raise
 
     def provision_rabbitmq(self):
         """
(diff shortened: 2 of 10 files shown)


instance/models/mixins/database.py
```diff
@@ -76,9 +77,13 @@ def _create_database(cursor, database):
     """
     Create MySQL database, if it doesn't already exist.
     """
+    logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
xirdneh: I added the logger in that local scope because I noticed the loggers used by Models are instantiated as instance attributes rather than on the module level and wasn't sure if it was going to mess something up.
But it seems like that should be ok. I'll move this to the module level.

instance/models/mixins/database.py
```diff
@@ -76,9 +77,13 @@ def _create_database(cursor, database):
...
+    logger.info('database: %s', database)
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        cursor.execute('CREATE DATABASE IF NOT EXISTS `{db}` DEFAULT CHARACTER SET utf8'.format(db=database))
+        cursor.execute(
+            'CREATE DATABASE IF NOT EXISTS `{db}` DEFAULT CHARACTER SET utf8'.format(db=database)
+        )
```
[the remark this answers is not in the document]
xirdneh: This was more muscle memory than anything. I'm more used to read strings/parameters like this. But I'll change it back :)

instance/models/mixins/database.py
```diff
@@ -101,17 +106,22 @@ def _grant_privileges(cursor, database, user, privileges):
         tables = "*.*"
     else:
         tables = "`{database}`.*".format(database=database)
-    cursor.execute('GRANT {privileges} ON {tables} TO %s'.format(privileges=privileges, tables=tables), (user,))
+    grant_statement = 'GRANT {privileges} ON {tables} TO %s'.format(
+        privileges=privileges, tables=tables
+    )
+    cursor.execute(grant_statement, (user,))
```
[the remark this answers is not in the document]
xirdneh: Sure, np. I'll change it back.

instance/models/mixins/database.py
```diff
@@ -76,9 +77,13 @@ def _create_database(cursor, database):
     """
     Create MySQL database, if it doesn't already exist.
     """
+    logger = logging.getLogger(__name__)
+    logger.info('database: %s', database)
```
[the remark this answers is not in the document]
xirdneh: Yes, the message should be more descriptive. Sorry about that.

requirements.txt
```diff
@@ -124,7 +124,7 @@ requests-oauthlib==0.7.0
 requests-toolbelt==0.7.0
 requestsexceptions==1.1.3
 requirements-detector==0.5.2
-responses==0.5.1
+responses==0.10.4
```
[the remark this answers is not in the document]
xirdneh: I updated the method that does a request to rabbitmq so it will [raise a RabbitMQAPIError](https://github.com/open-craft/opencraft/blob/josuebc/BB-692/instance/models/mixins/rabbitmq.py#L155) with a descriptive message because before there wasn't a message and it can become a bit hard to debug something like that. The old `responses` library did not mock the `reason` attribute in a `requests` Response instance so the test would fail with an `AttributeError`.
And `requests` is pinned to `2.11.1` which has a [Response model](https://github.com/requests/requests/blob/v2.11.1/requests/models.py#L545) with the `reason` attribute.

instance/models/openedx_instance.py
```diff
@@ -386,13 +399,68 @@ def delete(self, *args, **kwargs):
...
+        self.logger.info('Deprovision S3 END.')
+
+        self.logger.info('Deprovision RabbitMQ START.')
+        try:
+            self.deprovision_rabbitmq()
+        except RabbitMQAPIError as exc:
+            if ignore_errors or kwargs.pop('ignore_rabbitmq_errors', False):
+                self.logger.exception(exc)
+            else:
+                raise
```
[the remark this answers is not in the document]
xirdneh: I think we can do that. The only issue here is that would mean we'd have to pass the `ignore_errors` flag to the `deprovision_<resource>` method and wasn't sure if that was OK to change. But I'm up for moving this to the actual method :)

instance/openstack_utils.py
```diff
@@ -265,5 +265,5 @@ def delete_swift_container(container_name, **kwargs):
     with swift_service(**kwargs) as service:
         # Service delete may yield arbitrary large amount of results,
         # and since it yields we should read all results.
-        for result in service.delete(container_name):  # pylint: disable=unused-variable
-            pass
+        results = [result for result in service.delete(container_name)]
+        del results[:]
```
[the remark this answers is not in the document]
xirdneh: This is just a nitpick of mine since now we don't have to disable a pylint warning and also list comprehensions are faster than for loops. Since the whole point of this line is to consume the messages then it felt like a more efficient way to do it. Also, we can use `del` instead of letting the garbage collector free the memory used here.

instance/tests/integration/integration_instance.py
```diff
@@ -267,6 +267,7 @@ def test_spawn_appserver(self):
         # Add an lms user, as happens with beta registration
         user, _ = get_user_model().objects.get_or_create(username='test', email='test@example.com')
         instance.lms_users.add(user)
+        instance.iam_user = 'test'
```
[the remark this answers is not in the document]
xirdneh: This is used when [deprovision S3](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/storage.py#L334) and the test was failing without it.

instance/tests/models/test_openedx_instance.py
```diff
@@ -533,6 +539,63 @@ def test_delete_instance(self, mocks, delete_by_ref, *mock_methods):
...
+        Test that an instance can be deleted directly or by its InstanceReference.
+        """
+        mock_methods = [
+            mock_deprovision_rabbitmq,
+            mock_deprovision_s3,
+            mock_deprovision_swift,
+            mock_deprovision_mongo,
+            mock_deprovision_mysql,
+            mock_archive,
+        ]
```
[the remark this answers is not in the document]
xirdneh: It's just that this test mocks when these methods raises an exception and I prefer to use the name of the mock than `mock_methods[0]` to set the mocked exception.
But if it's preferred that way, it's OK. I can change it to `*mock_methods`

instance/tests/models/test_openedx_instance.py
```diff
@@ -533,6 +539,63 @@ def test_delete_instance(self, mocks, delete_by_ref, *mock_methods):
...
+        mock_deprovision_rabbitmq = RabbitMQAPIError('RabbitMQ error')
+        mock_deprovision_mongo = PyMongoError('Mongo error')
+        mock_deprovision_mysql = MySQLError('MySQL error')
```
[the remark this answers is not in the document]
xirdneh: You mean define the exceptions in the `patch` decorator?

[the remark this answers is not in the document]
xirdneh: In reality this change was wrong, we have to use `side_effect` to make sure the call to `deprovision_<resource>` will raise an error.

xirdneh:
@Agrendalath PR updated. I'll squash the commits once you finish the review that way we don't loose the conversations. Thanks.

instance/tests/integration/integration_instance.py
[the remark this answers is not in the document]
xirdneh: Sorry, this was a typo.

instance/openstack_utils.py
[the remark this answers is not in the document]
xirdneh: Yes, you are correct. It is not necessary to change it to a list comprehension.
Usually list comprehensions are faster than for loops when creating a list.
I also thought it would be easier to read by re-writing it like that since the whole point of the for loop is to just consume the response. There's nothing done with the response so it is explicitly deleted instead of waiting for the garbage collector to free up that memory. But by using a for loop and doing assigning the consumed response to a single variable the code doesn't pollute the memory and still the garbage collection will only have to free up the memory used by the last result in the list.
I changed it back to how it was.

Agrendalath:
@xirdneh, could you please resolve the conflicts between this branch and `master`?

instance/models/mixins/rabbitmq.py
```diff
@@ -157,13 +156,27 @@ def provision_rabbitmq(self):
...
+                    raise
             for user in [self.rabbitmq_consumer_user, self.rabbitmq_provider_user]:
-                self._rabbitmq_request('delete', 'users', user.username)
+                self.logger.info('Deleting rabbitmq user: %s', user.username)
+                try:
+                    self._rabbitmq_request('delete', 'users', user.username)
+                except RabbitMQAPIError as exc:
+                    self.logger.exception('Cannot delete rabbitmq user: %s. %s', user.username, exc)
+                    if not ignore_errors:
+                        raise
```
Agrendalath: Could we move it to `_rabbitmq_request` to avoid code repetition?

On the second thought, we could also simplify the whole exception handling by adding decorator for exception logging and optionally reraising the exception. Please see [this link](https://www.blog.pythonlibrary.org/2016/06/09/python-how-to-create-an-exception-logging-decorator/) for the extended idea. It would greatly increase the code readability, because we won't need to dig through the error handling here.

instance/models/mixins/database.py
```diff
@@ -225,25 +249,30 @@ def provision_mysql(self):
...
+                # Drop users with global privileges
+                for user in self.global_users:
+                    _drop_user(cursor, user)
+            except MySQLError:
+                if not ignore_errors:
+                    raise
 
             self.mysql_provisioned = False
             self.save()
+        self.logger.info('Deprovisioning MySQL finished.')
```
Agrendalath: Please see [this comment](https://github.com/open-craft/opencraft/pull/373#discussion_r241937262)
It would be much easier (and more reusable) to use something like this for logging:
```python
def action_logger(action: str):
    def decorator(f):
        def wrapper(self, *args, **kwargs):  
            self.logger.info('{} started.'.format(action)  
            f()
            self.logger.info('{} finished.'.format(action)  
        return wrapper
    return decorator
```

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -372,6 +375,35 @@ def test_ansible_settings_no_mysql_server(self):
...
+    @patch('instance.models.mixins.database._get_mysql_cursor', return_value=Mock())
+    @patch('instance.models.mixins.database._drop_database')
+    @patch('instance.models.mixins.database._drop_user')
+    def test_ignore_errors_deprovision_mysql(self, mock_drop_user, mock_drop_database, mock_get_cursor):
+        """
+        Test mysql is set as deprovision when ignoring errors.
+        """
+        self.instance = OpenEdXInstanceFactory()
+        self.instance.mysql_provisioned = True
+        mock_drop_database.side_effect = MySQLError()
```
Agrendalath: Could we move it to the decorator?
```python
@patch('instance.models.mixins.database._drop_database', side_effect=MySQLError())
```

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -688,6 +744,27 @@ def test_string_representation(self, fields, representation):
...
+        mock_rabbitmq_request.assert_any_call('delete', 'users', self.instance.rabbitmq_consumer_user.username)
+        mock_rabbitmq_request.assert_any_call('delete', 'users', self.instance.rabbitmq_provider_user.username)
+
+    @patch('instance.models.mixins.rabbitmq.RabbitMQInstanceMixin._rabbitmq_request')
+    def test_ignore_errors_deprovision_rabbitmq(self, mock_rabbitmq_request):
+        """
+        Test rabbitmq is set as deprovision when ignoring errors.
+        """
+        self.instance.rabbitmq_provisioned = True
+        mock_rabbitmq_request.side_effect = RabbitMQAPIError()
```
Agrendalath: Could we move it to the `@patch`?

instance/tests/models/test_openedx_database_mixins.py
```diff
@@ -566,6 +598,30 @@ def test__get_main_database_url(self):
...
+
+    @patch('instance.models.mixins.database.MongoDBInstanceMixin._get_main_database_url')
+    @patch('instance.models.mixins.database.pymongo.MongoClient', autospec=True)
+    def test_ignore_errors_deprovision_mongo(self, mock_mongo_client_cls, mock_get_main_db_url):
+        """
+        Test mongo is set as deprovision when ignoring errors.
+        """
+        self.instance = OpenEdXInstanceFactory()
+        self.instance.mongo_provisioned = True
+        mock_mongo_client_cls().drop_database = Mock(side_effect=PyMongoError())
```
Agrendalath: Why don't we just patch this method directly?

instance/tests/models/test_openedx_instance.py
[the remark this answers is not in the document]
Agrendalath: I meant using named arg of the `patch` decorator for defining that the method should raise specific exception.

instance/models/mixins/database.py
xirdneh: Yes, this could be easily put into a decorator. But I think that would add an unnecessary layer of abstraction when looking at the logs.
The logs will tell us that the line where this log message comes from is not where the actual deprovision code is but where the decorator is implemented. To me, this is counterintuitive to what logs should be.
Although, if this is a standard we follow then I can move this to a decorator.

instance/tests/models/test_openedx_database_mixins.py
xirdneh: I think it reads better to write it explicitly in the test since that's the point of the test. To handle a `MySQLError`.
But I can move it to the `@patch` decorator.

instance/models/mixins/database.py
Agrendalath: > The logs will tell us that the line where this log message comes from is not where the actual deprovision code is but where the decorator is implemented. To me, this is counterintuitive to what logs should be.
Do you think it's important for info logs? We have the module name for them.

instance/tests/models/test_openedx_database_mixins.py
Agrendalath: It's nice to clearly separate data preparation from actual testing.

instance/models/mixins/rabbitmq.py
xirdneh: I think it would indeed increase readability but I'm not sure about what it would do to debugging.
If there's anything other error that needs to be followed through the stacktrace will, unnecessarily, have another line where the actual exception was logged. And that extra set of lines will probably not even live in the same module.
If we're ok with that then the decorator should probably live in a more general `utils.py` or `decorators.py` module and we could use it in other methods that catch specific exceptions.
The decorator should catch specific exceptions so a `exc_class` argument should be there as well as the message we wish to log.
But there's another issue here. These exceptions are not in `_rabbitmq_request` because that method is too broad for what we want to log. We want to specify either which user or vhost we were not able to delete. This would also be the issue if we move error handling into a decorator since we won't necessarily have access to the variables in the method's inner scope.

instance/tests/models/test_openedx_database_mixins.py
xirdneh: I see, let me move it to the `@patch` decorator :)

instance/models/mixins/database.py
xirdneh: Since the message is not general, as in "Deprovisioned started", but more specific, as in "Deprovisioning <resource> started/finished" I think it is important.
But it's not something that would break anything so I'm also OK to move it into a decorator.

instance/tests/models/test_openedx_instance.py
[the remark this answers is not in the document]
xirdneh: Yeap, sorry about the misunderstanding here. I updated the test signature.

instance/tests/models/test_openedx_database_mixins.py
xirdneh: Ah, that's true. Let me modify this. Thanks

instance/models/mixins/database.py
Agrendalath: @xirdneh, I'm not sure what you're referring to. The example decorator above works like:

```python
@action_logger('Deprovisioning MySQL')
def deprovision_mysql(...):
    ...
```

So it's generic enough to handle all specific messages for our use cases, right? Am I missing something here?

xirdneh: Yes, exactly. The log messages are not really generic because the actual message is passed as an argument into the decorator. If the message was generic then it wouldn't need to have a configurable action and could work for any deprovision method. So, really we're only saving a few characters an introducing something that is not clear what'll do right away (a decorator). Until you look at the decorator implementation that is.
Also, even if this is an info log message there's some context derived from the record, the module and line where the message is logged. If I were trying to get a sense about what is happening from the logs I would be directed to a function that is not where the actual deprovision is being implemented.

Now, these points of view are more preferences than anything and either way doesn't affect performance or clarity too much, at least not to my perception. If this is something that would align better with the project I'm up for it. What do you think, @Agrendalath ?

Agrendalath: Ok, let's leave it this way.

PR #376, opened 2019-01-07
BB-756 - Improve fault tolerance of cleanup script and run it only after all tests are done

giovannicimolin:
This PR add two main changes to the integration cleanup script:

1. Improved fault tolerance by catching untreated exceptions
2. Added CircleCI workflows configuration to run cleanup only after all tests are done.

## Testing

1. Checkout this branch
2. `vagrant up && vagrant ssh` on Ocim's devstack
3. Export environment variables needed for cleanup:
```
export GANDI_API_KEY=
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
export GANDI_ZONE_ID=
export OPENSTACK_USER=
export OPENSTACK_TENANT=
export OPENSTACK_REGION=
export OPENSTACK_AUTH_URL=
export OPENSTACK_PASSWORD=
```
4. Run `PYTHONPATH=$PYTHONPATH:$(pwd) honcho -e .env.integration run python3 cleanup_utils/integration_cleanup.py --dry_run` and check that there were no errors on the output.
5. (OPTIONAL) Run `make test.integration_cleanup` to run the actual cleanup and check that there were no errors on the output.
6. Check the latest CI run and see that the workflows are working correctly. Latest commit:
**Workflow**: https://circleci.com/workflow-run/8fa529e2-2703-4293-991e-93f7e09d5fd8
**CI Run**: https://circleci.com/gh/open-craft/opencraft/4249

### DIFF
--- a/cleanup_utils/aws_cleanup.py
+++ b/cleanup_utils/aws_cleanup.py
@@ -203,9 +203,8 @@ class AwsCleanupInstance:
                 # If user policy exists
                 if user_policy:
                     buckets_to_delete = self.get_bucket_names_from_policy(user_policy)
-
                     logger.info("  > Cleaning up stuff from user %s.", user['UserName'])
-
+                    
                     # Delete buckets, user policy, access keys and the iam user
                     for bucket_name in buckets_to_delete:
                         logger.info("    * Deleting bucket %s.", bucket_name)
@@ -217,23 +216,29 @@ class AwsCleanupInstance:
                         user['UserName']
                     )
                     self.delete_user_policy(user['UserName'], DEFAULT_POLICY_NAME)
-
-                    for access_key in old_keys:
-                        logger.info(
-                            "    * Deleting access key %s  from user %s.",
-                            access_key['AccessKeyId'],
-                            user['UserName']
-                        )
-                        self.delete_user_access_key(
-                            username=user['UserName'],
-                            access_key=access_key
-                        )
-
-                    logger.info("    * Deleting user %s.", user['UserName'])
-                    self.delete_user(username=user['UserName'])
-
-                    # Saves hashes from user name
-                    # ocim-HASH_integration_plebia_net.
-                    self.cleaned_up_hashes.append(
-                        user['UserName'].split('_')[0][5:]
+                else:
+                    logger.warning(
+                        "    * WARNING: The user %s doesn't have %s policy. Skipping bucket and policy deletion...",
+                        user['UserName'],
+                        DEFAULT_POLICY_NAME
                     )
+
+                for access_key in old_keys:
+                    logger.info(
+                        "    * Deleting access key %s  from user %s.",
+                        access_key['AccessKeyId'],
+                        user['UserName']
+                    )
+                    self.delete_user_access_key(
+                        username=user['UserName'],
+                        access_key=access_key
+                    )
+
+                logger.info("    * Deleting user %s.", user['UserName'])
+                self.delete_user(username=user['UserName'])
+
+                # Saves hashes from user name
+                # ocim-HASH_integration_plebia_net.
+                self.cleaned_up_hashes.append(
+                    user['UserName'].split('_')[0][5:]
+                )
--- a/cleanup_utils/aws_cleanup.py
+++ b/cleanup_utils/aws_cleanup.py
@@ -204,7 +204,7 @@ class AwsCleanupInstance:
                 if user_policy:
                     buckets_to_delete = self.get_bucket_names_from_policy(user_policy)
                     logger.info("  > Cleaning up stuff from user %s.", user['UserName'])
-                    
+
                     # Delete buckets, user policy, access keys and the iam user
                     for bucket_name in buckets_to_delete:
                         logger.info("    * Deleting bucket %s.", bucket_name)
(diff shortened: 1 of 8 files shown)


cleanup_utils/aws_cleanup.py
```diff
@@ -183,10 +199,7 @@ def run_cleanup(self):
             )
             # If the user has any old keys, proceed with deletion
             if old_keys:
-                user_policy = self.iam_client.get_user_policy(
-                    UserName=user['UserName'],
-                    PolicyName=DEFAULT_POLICY_NAME
-                )
+                user_policy = self.get_user_policy(user['UserName'])
                 # If user policy exists
                 if user_policy:
```
lgp171188: @giovannicimolin, should some thing be logged when `user_policy` is `None`?

cleanup_utils/dns_cleanup.py
```diff
@@ -79,6 +86,37 @@ def get_dns_record_list(self, zone_id):
...
+        except xmlrpc.client.Fault as e:
+            logger.error("  > FAILED Deleting DNS entries for %s...", record_name)
+            logger.error("  > ERROR: %s", e)
+
+    def create_new_zone_version(self, zone_id):
+        """
+        Create a new version of the domain, based on the current version
+        Returns the `version_id` of the version
+        """
+        return self.client_zone.version.new(self.api_key, zone_id)
```
lgp171188: @giovannicimolin, in case of concurrent operations, can this become stale before the operations that use this actually run? If that is possible, can we handle the error and retry?

cleanup_utils/dns_cleanup.py
```diff
@@ -79,6 +86,37 @@ def get_dns_record_list(self, zone_id):
...
+
+    def set_zone_version(self, zone_id, zone_version_id):
```
lgp171188: @giovannicimolin, similar concerns to the comment on creating a new zone version.

cleanup_utils/aws_cleanup.py
giovannicimolin: Yes, and even when the user doesn't have that policy, but has `integration` on the name, it should be deleted.
I'll implement this behaviour.

cleanup_utils/dns_cleanup.py
giovannicimolin: This creates a new zone version and deletes all cleaned up hashes on that version. 
Then, only after completing all operations, `self.set_zone_version` is ran, so, if two cleanup scripts run concurrently, each of them will create a new zone and the one that finishes last will set the domains DNS records.
There's no way to fix this unless we either run this separately (outside the CI) or make some kind of atomic operation for all the cleanup runs.
Since this PR also introduces workflows to avoid running multiple cleanup instances at the same time we shouldn't have issues with this. Also, the entire cleanup run in pretty fast (<3minutes), so it's gonna be somewhat hard to have a concurrent execution.

cleanup_utils/dns_cleanup.py
giovannicimolin: Same as the comment above. 
We either move this away from the CI or make this an atomic transaction.

cleanup_utils/dns_cleanup.py
giovannicimolin: Also, between putting this outside the CI and making it an "atomic transaction" between all concurrent instances, I think that hosting this outside the CI is the best option. 
However, both of these options are more complex that the solution that we have today, and might be a huge over engineering. 
But there's a plus side: if this is hosted outside the CI, we can make this cleanup tool watch over all Opencraft's resources and prune them according to our necessities, not only the integration resources.

cleanup_utils/dns_cleanup.py
giovannicimolin: @lgp171188

cleanup_utils/dns_cleanup.py
giovannicimolin: @lgp171188

lgp171188: @giovannicimolin, the only rare chance for concurrency issues is when some one edits a zone while the CI checks are running. But a re-run should probably fix it. So I guess we can live with that.

lgp171188:
@giovannicimolin, I see that the status of CircleCI is stuck on `Expected — Waiting for status to be reported`. Can you check why and verify if this is due to running 2 separate jobs?

giovannicimolin:
This is because we introduced workflows to our CI runs. Now, the CI will run the two tasks as separate CI checks. 

> If you have implemented workflows on a branch in your GitHub repository, but the status check never completes, there may be status settings in GitHub that you need to deselect. For example, if you choose to protect your branches, you may need to deselect the ci/circleci status key as this check refers to the default CircleCI 1.0 check, as follows:
github_branches_status

So in order to merge this, we'll have to disable that ci/circleci check. All the other PR's must be rebased in order to be mergeable after this.

lgp171188:
@giovannicimolin, I disabled the `ci/circleci` check and enabled the `ci/circleci: build` check now. The status now shows as `All checks have passed`. We need to confirm that the checks don't show the status as `failed` if the `cleanup` task fails.

lgp171188:
@giovannicimolin, I have reverted the changes made for the new workflow changes because it is causing the checks on other PRs to wait for the non-existent `build` step key. We can re-enable it after merging this PR.

PR #377, opened 2019-01-07
Upgrade to boto3

viadanna:
Description
------------
This PR upgrades the requirement from boto2 to boto3.

It also implements versioning and lifecycle management of s3 buckets

Related tickets: BB-719 BB-329

Testing
-------

1. Checkout the changes and load the shell using `make shell` (this can be done using the admin as well).

2. Create an instance using the shell as mentioned in the README:
```python
from instance.factories import instance_factory, production_instance_factory

# Creating an instance with defaults appropriate for sandboxes:
instance = instance_factory(name="Sandbox instance", sub_domain="sandbox")

# Creating an instance with defaults appropriate for production:
production_instance = production_instance_factory(name="Production instance", sub_domain="production")
```

3. Spawn an AppServer using on `stage.console.opencraft.com`

4. Use either the shell or `awscli` to query the bucket exists, CORS and lifecycle are correct, i.e:
```python
instance.s3.get_bucket_cors(Bucket=instance.s3_bucket_name)
instance.s3.get_bucket_lifecycle(Bucket=instance.s3_bucket_name)
instance.s3.get_bucket_lifecycle_configuration(Bucket=instance.s3_bucket_name)
```

5. Delete the instance and make sure the bucket is deleted as well, i.e:
```python
bucket = instance.s3_bucket_name
instance.delete()
boto3.client('s3').list_objects(Bucket=bucket)  # Should not exist
```

Management Command
------------------------

A management command called `reprovision_buckets` was added to reprovision and configure iam, versioning and lifecycle on existing instance.

This needs to be tested in stage using:
```bash
$ make manage reprovision_buckets
```

This can be tested with the step `4` shown above.

### DIFF
--- a/instance/management/commands/reprovision_buckets.py
+++ b/instance/management/commands/reprovision_buckets.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2019 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This management command will reprovision the S3 bucket for each available
+instance to have its configuration updated on AWS.
+"""
+
+import logging
+
+from botocore.exceptions import ClientError
+from django.core.management.base import BaseCommand
+
+from instance.models.instance import InstanceReference
+
+LOG = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    This management command will reprovision the S3 bucket for each available
+    instance to have its configuration updated on AWS.
+    """
+    help = (
+        'Reprovision S3 buckets for all available instances'
+    )
+
+    def handle(self, *args, **options):
+        """
+        Reprovisions the S3 bucket for each available instance to have its
+        configuration updated on AWS.
+        """
+
+        # Identify instances that need updating
+        refs = InstanceReference.objects.filter(is_archived=False)
+        LOG.info('Found "%d" active instances', len(refs))
+
+        for ref in refs:
+            LOG.info('Reprovisioning %s S3 bucket to update bucket and iam configuration', ref.instance.name)
+
+            # Reprovisioning will update the necessary configuration
+            try:
+                ref.instance.provision_s3()
+            except ClientError as e:
+                LOG.error(
+                    'Failed to reprovision bucket for "%s" due to: %s',
+                    ref.instance.name,
+                    e
+                )
+            else:
+                LOG.info('Reprovisioned bucket for %s', ref.instance.name)
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -24,7 +24,8 @@ Instance app model mixins - Database
 import json
 import time
 
-import boto
+import boto3
+from botocore.exceptions import ClientError
 
 from django.db.backends.utils import truncate_name
 from django.conf import settings
@@ -34,24 +35,28 @@ from swiftclient.exceptions import ClientException as SwiftClientException
 from instance import openstack_utils
 from instance.models.utils import default_setting
 
+S3_LIFECYCLE = {
+    'Rules': [
+        {
+            'NoncurrentVersionExpiration': {
+                'NoncurrentDays': settings.S3_VERSION_EXPIRATION
+            },
+            'Prefix': '',
+            'Status': 'Enabled',
+        },
+    ]
+}
 
-def get_master_iam_connection():
-    """
-    Create connection to IAM service
-    """
-    return boto.connect_iam(
-        settings.AWS_ACCESS_KEY_ID,
-        settings.AWS_SECRET_ACCESS_KEY
-    )
+S3_CORS = {
+    'CORSRules': [{
+        'AllowedHeaders': ['*'],
+        'AllowedMethods': ['GET', 'PUT'],
+        'AllowedOrigins': ['*'],
+        'ExposeHeaders': ['GET', 'PUT'],
+    }]
+}
 
-
-def get_s3_cors_config():
-    """
-    Create CORS config needed for ORA2 File uploads
-    """
-    cors_config = boto.s3.cors.CORSConfiguration()
-    cors_config.add_rule(allowed_method=['GET', 'PUT'], allowed_header=["*"], allowed_origin=["*"])
-    return cors_config
+USER_POLICY_NAME = 'allow_access_s3_bucket'
 
 
 # Classes #####################################################################
@@ -175,16 +180,33 @@ class S3BucketInstanceMixin(models.Model):
         )
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._iam_client = None
+        self._s3_client = None
+
     def get_s3_policy(self):
         """
         Return s3 policy with access to create and update bucket
         """
-        policy = {
+        return {
             "Version": "2012-10-17",
             "Statement": [
                 {
                     "Effect": "Allow",
-                    "Action": ["s3:ListBucket", "s3:CreateBucket", "s3:DeleteBucket", "s3:PutBucketCORS"],
+                    "Action": [
+                        "s3:CreateBucket",
+                        "s3:DeleteBucket",
+                        "s3:DeleteObjects",
+                        "s3:GetBucketCORS",
+                        "s3:GetBucketVersioning",
+                        "s3:GetLifecycleConfiguration",
+                        "s3:ListBucket",
+                        "s3:ListBucketVersions",
+                        "s3:PutBucketCORS",
+                        "s3:PutBucketVersioning",
+                        "s3:PutLifecycleConfiguration",
+                    ],
                     "Resource": ["arn:aws:s3:::{}".format(self.s3_bucket_name)]
                 },
                 {
@@ -196,7 +218,6 @@ class S3BucketInstanceMixin(models.Model):
                 }
             ]
         }
-        return json.dumps(policy, indent=2)
 
     @property
     def bucket_name(self):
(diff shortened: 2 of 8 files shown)


instance/models/mixins/storage.py
```diff
@@ -34,24 +35,38 @@
 from instance import openstack_utils
 from instance.models.utils import default_setting
 
+USER_POLICY_NAME = 'allow_access_s3_bucket'
 
-def get_master_iam_connection():
+
+def get_s3_cors_config():
```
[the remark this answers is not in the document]
viadanna: Good point, I just refactored the functions but it makes sense 👍

instance/models/mixins/storage.py
```diff
@@ -34,24 +35,38 @@
...
+    Default bucket lifecycle for expiring old versions of objects
     """
-    cors_config = boto.s3.cors.CORSConfiguration()
-    cors_config.add_rule(allowed_method=['GET', 'PUT'], allowed_header=["*"], allowed_origin=["*"])
-    return cors_config
+    return {
+        'Rules': [
+            {
+                'NoncurrentVersionExpiration': {
+                    'NoncurrentDays': 30  # settings.S3_VERSION_EXPIRATION
```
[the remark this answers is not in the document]
viadanna: Yeah, I'm looking into creating a new setting for this. The expiration is actually related to another ticket but was gobbled by this work. I'll get both done in this PR and fix this 👍

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,100 @@ def s3_custom_domain(self):
         """
         return "{}.s3.amazonaws.com".format(self.s3_bucket_name)
 
+    @property
+    def iam(self):
```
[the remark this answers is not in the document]
viadanna: 👍

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,100 @@ def s3_custom_domain(self):
...
+            PolicyDocument=json.dumps(self.get_s3_policy())
+        )
+        access_key = self.iam.create_access_key(UserName=self.iam_username)['AccessKey']
+        self.s3_access_key = access_key['AccessKeyId']
+        self.s3_secret_access_key = access_key['SecretAccessKey']
         self.save()
 
-    def get_s3_connection(self):
+    @property
+    def s3(self):  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
viadanna: 👍

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,100 @@ def s3_custom_domain(self):
...
         Create bucket, retry up to defined attempts if it fails
         If you specify a location (e.g. 'EU', 'us-west-1'), this method will use it. If the location is
         not specified, the value of the instance's 's3_region' field is used.
         """
-        time.sleep(retry_delay)
-        location = location or self.s3_region
-        try:
-            s3 = self.get_s3_connection()
+        attempt, bucket = 1, None
+        location_constraint = location or self.s3_region or None
```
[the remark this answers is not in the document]
viadanna: An empty string `""` for self.s3_region  won't work here, it must be None in such case.

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,100 @@ def s3_custom_domain(self):
...
+        location_constraint = None if location_constraint == 'us-east-1' else location_constraint
```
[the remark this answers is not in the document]
viadanna: Fair enough 👍

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,100 @@ def s3_custom_domain(self):
...
-                location=location
-            )
+                    bucket = self.s3.create_bucket(
+                        Bucket=self.s3_bucket_name,
+                        CreateBucketConfiguration={
+                            'LocationConstraint': location_constraint
+                        },
+                    )
+            except ClientError as e:
+                if e.response.get('Error', {}).get('Code') == 'EntityAlreadyExists':
```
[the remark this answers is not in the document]
viadanna: The idea was to keep the same behavior as before. It's not really a problem, but yeah, logging won't hurt. 👍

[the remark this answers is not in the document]
viadanna: The idea was to keep the same behavior as before. It's not really a problem, but yeah, logging won't hurt. 👍

instance/models/mixins/storage.py
```diff
@@ -325,28 +376,40 @@ def deprovision_s3(self):
...
+                    Bucket=self.s3_bucket_name,
+                    Delete={
+                        'Objects': [{'Key': d['Key'], 'VersionId': d['VersionId']} for d in to_delete]
+                    }
+                )
+            # Remove bucket
+            self.s3.delete_bucket(Bucket=self.s3_bucket_name)
+            self.s3_bucket_name = ""
+            self.save()
+        except ClientError as e:
```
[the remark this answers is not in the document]
viadanna: Personally, I think this is more readable than adding an `else` clause, let's see how it looks.

instance/tests/models/test_openedx_storage_mixins.py
```diff
@@ -305,145 +328,234 @@ def test_ansible_settings_s3(self, s3_region):
...
-
-    @patch('boto.s3.connection.S3Connection.create_bucket')
-    def test__create_bucket_fails(self, create_bucket):
+        if not provision_iam:
+            instance.s3_access_key = 'test'
+            instance.s3_secret_access_key = 'test'
+
+        iam = get_session().create_client('iam')
+        s3 = get_session().create_client('s3')
+        with patch('instance.models.mixins.storage.S3BucketInstanceMixin.iam', iam):
```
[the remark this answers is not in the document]
viadanna: Problem is, `@patch` will not return the client and it must be used with the stubber.

I'll experiment using a constant to move to the decorator.

instance/tests/models/test_openedx_storage_mixins.py
```diff
@@ -305,145 +328,234 @@ def test_ansible_settings_s3(self, s3_region):
...
+            with patch('instance.models.mixins.storage.S3BucketInstanceMixin.s3', s3):
+                with S3Stubber(s3) as stubber:
```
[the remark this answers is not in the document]
viadanna: 👍

instance/models/mixins/storage.py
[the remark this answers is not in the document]
viadanna: That's the behaviour previously implemented, as seen [here](https://github.com/open-craft/opencraft/pull/377/files#diff-3906999c5f272a75fc92df1d9991a030L336), I'm trying to keep the same.

instance/models/mixins/storage.py
```diff
@@ -325,28 +376,40 @@ def deprovision_s3(self):
...
-                    key.delete()
-                s3.delete_bucket(self.s3_bucket_name)
-                self.s3_bucket_name = ""
-                self.save()
-            except boto.exception.S3ResponseError:
-                self.logger.exception('There was an error trying to remove S3 bucket "%s".', self.s3_bucket_name)
         try:
-            self.logger.info('Deleting IAM user: %s', self.iam_username)
-            iam = get_master_iam_connection()
+            while True:
```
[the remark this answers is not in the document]
viadanna: Oh, I misread your example, makes sense.

instance/models/mixins/storage.py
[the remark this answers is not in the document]
viadanna: Ok

instance/tests/management/test_reprovision_buckets.py
```diff
@@ -0,0 +1,87 @@
...
+        Verify that the command correctly reprovision the bucket for an instance.
+        """
+        OpenEdXInstance.objects.create(
+            sub_domain='test', name='test instance', storage_type=OpenEdXInstance.S3_STORAGE)
+        with LogCapture() as captured_logs:
+            call_command(
+                'reprovision_buckets',
+                stdout=StringIO(),
+            )
+            # Verify the logs
```
jamestait: I think these tests should be outside the context manager, to ensure `captured_logs` isn't being modified after the call to `call_command()`.

instance/tests/management/test_reprovision_buckets.py
```diff
@@ -0,0 +1,87 @@
...
+            actual = set(l[2] for l in captured_logs.actual())
+            expected = {
+                'Found "1" active instances',
+            }
+            self.assertTrue(expected <= actual)
```
jamestait: Could we not use something like:

```self.assertIn('Found "1" active instances', captured_logs.actual())```

which might be a bit more self explanatory?

File: instance/tests/management/test_reprovision_buckets.py [hunk 14]
```diff
@@ -0,0 +1,87 @@
...
+
+        mock_create_iam_user.assert_called_once_with()
+        mock_update_iam.assert_called_once_with()
+        mock_create_bucket.assert_called_once_with('')
+        mock_update_cors.assert_called_once_with()
+        mock_update_lifecycle.assert_called_once_with()
+        mock_enable_versioning.assert_called_once_with()
```
jamestait: While these tests look technically correct, I'm not sure about the utility of them. It's more of a philosophical argument, really. It seems to be testing the implementation rather than the intended behaviour. I think elements of the underlying calls are tested elsewhere, perhaps, but really what I'd like to see here is that calling the management command results in a call to `StorageContainer._s3_client.put_bucket_lifecycle_configuration()`, `StorageContainer._iam_client.put_user_policy()`, etc.

This is a difficult one, because those clients aren't our own code (at least I don't think they are), and we'd have to mock their public interfaces - but those interfaces shouldn't change unless we change which implementation of the client we use, e.g. by upgrading it, in which case if the public API has changed, we *want* the test to fail.

instance/models/mixins/storage.py
```diff
@@ -307,53 +307,68 @@ def _create_bucket(self, max_tries=4, retry_delay=4, location=None):
...
+                Bucket=self.s3_bucket_name,
+                CreateBucketConfiguration={
+                    'LocationConstraint': location_constraint
+                },
+            )
+        return bucket
+
     def _update_iam_policy(self):
         """
         Update S3 IAM user policy
```
jamestait: Not actually shown in this diff, but a few lines below this we set `self._s3_client` to `None` - it'd be useful to have a comment explaining why we do this.

instance/models/mixins/storage.py
```diff
@@ -307,53 +307,68 @@ def _create_bucket(self, max_tries=4, retry_delay=4, location=None):
...
+        self._enable_bucket_versioning()
+
+    def _enable_bucket_versioning(self):
+        self.s3.put_bucket_versioning(
             Bucket=self.s3_bucket_name,
-            CORSConfiguration=S3_CORS
+            VersioningConfiguration={
+                'Status': 'Enabled'
+            }
         )
```
jamestait: Do we need to catch any errors here?

instance/models/mixins/storage.py
```diff
@@ -307,53 +307,68 @@ def _create_bucket(self, max_tries=4, retry_delay=4, location=None):
...
 
-        # Enable bucket lifecycle
+    def _update_bucket_lifecycle(self):
         self.s3.put_bucket_lifecycle_configuration(
             Bucket=self.s3_bucket_name,
             LifecycleConfiguration=S3_LIFECYCLE
         )
```
jamestait: Do we need to catch any errors here?

instance/models/mixins/storage.py
```diff
@@ -307,53 +307,68 @@ def _create_bucket(self, max_tries=4, retry_delay=4, location=None):
...
-        # Enable bucket versioning
-        self.s3.put_bucket_versioning(
+    def _update_bucket_cors(self):
+        self.s3.put_bucket_cors(
             Bucket=self.s3_bucket_name,
-            VersioningConfiguration={
-                'Status': 'Enabled'
-            }
+            CORSConfiguration=S3_CORS
         )
```
jamestait: Do we need to catch any errors here?

instance/tests/management/test_reprovision_buckets.py
viadanna: The command must be inside the context manager as it's actually verifying the output o the command.

instance/tests/management/test_reprovision_buckets.py
viadanna: 👍

instance/tests/management/test_reprovision_buckets.py
viadanna: During the migration to `boto3` the testing of the actual calls to the API are being tested by stubbing.

The idea here is testing the behavior of the command, making sure it perform all the updates necessary. I was actually a bit overzealous here as there's no need to test create_iam and create_bucket.

instance/models/mixins/storage.py
viadanna: 👍

instance/models/mixins/storage.py
viadanna: Those errors are bubbling up to the `_create_bucket` where it's being retried. In case of failure, the redeployment/provisioning must fail.

instance/models/mixins/storage.py
viadanna: ☝️

instance/models/mixins/storage.py
viadanna: ☝️

instance/tests/management/test_reprovision_buckets.py
jamestait: I'm always wary of creating brittle tests that make it difficult to refactor the implementation later - if I have to change tests to change the implementation, I tend to find I've lost some of the benefit of having the tests in the first place.

In any case, the tests do look correct, and it's hugely subjective, so I can live with this. :smile:

instance/tests/management/test_reprovision_buckets.py
jamestait: Sorry, I put the comment in a weird place. The `call_command` command needs to be inside the context manager to capture its output. But I think the assertion can be outside the context manager, because `captured_logs` is still in scope, but then `LogCapture` isn't still appending new logs to it?

instance/tests/management/test_reprovision_buckets.py
viadanna: I agree with that, but with  `boto3` it got out of hand given there's no possibility of mocking it. Stubbing is required and it's an order of magnitude more complex and [verbose](https://github.com/open-craft/opencraft/blob/615f99ac9ac64c308c168623de2f31873aa602b8/instance/tests/models/utils.py).

That's why I chose to encapsulate the calls to `boto3`, assuming these are correct, since these are being tested [here](https://github.com/open-craft/opencraft/pull/377/files#diff-337408cef240bc3cb85bbc4d3088bb6cR373).

Mainly, I'm checking that the management command has the appropriate behavior of updating the user policy, bucket cors, versioning and lifecycle. I have removed the unnecessary assertions.

instance/tests/management/test_reprovision_buckets.py
viadanna: 👍

jamestait:
Assuming CI passes, I'm happy with the code changes. @viadanna - when CI is happy, if you could set up stage for me to test, then I can approve.

instance/management/commands/reprovision_buckets.py
```diff
@@ -53,6 +54,13 @@ def handle(self, *args, **options):
...
 
             # Reprovisioning will update the necessary configuration
-            ref.instance.provision_s3()
-
-            LOG.info('Reprovisioned bucket for %s', ref.instance.name)
+            try:
+                ref.instance.provision_s3()
+            except ClientError as e:
+                LOG.error(
+                    'Failled to reprovision bucket for "%s" due to: %s',
```
jamestait: Absolute nittiest of nits: "Failed" should only have one "l".

jamestait:
#
:+1:

- [x] I tested this:
 * Verified output on the Jira ticket provided by @viadanna.
- [x] I read through the code

instance/management/commands/reprovision_buckets.py
viadanna: 🤦‍♂️

viadanna:
@kaizoku @xirdneh @jamestait 

This is finally ready for merge and deployment. I had to make changes after finding edge cases during testing on stage, mainly retrying configuration of the buckets due to delays in IAM permission propagation, and catching errors during reprovision.

Do you want to take a look here before I merge? I have already rebased and don't expect further changes before merging and deploying, which I intend to do over the weekend.

jamestait:
:+1: LGTM.

instance/models/mixins/storage.py
```diff
@@ -236,70 +257,161 @@ def s3_custom_domain(self):
         """
         return "{}.s3.amazonaws.com".format(self.s3_bucket_name)
 
+    @property
+    def iam(self):
```
kaizoku: Will these connections reconnect after a timeout? Should or could we test if they're still connected?

viadanna: There's a [default 60 second timeout](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html) and 4 retries before raising, so I think we are good here.

kaizoku: Cool, thanks for verifying. That should be plenty.
Just wanted to make sure we weren't passing back stale connections.

kaizoku:
Thanks @viadanna , the code and tests look comprehensive to me. :+1:

PR #378, opened 2019-01-11
[BB-783] Mock the appserver spawn in the betatest accepted integration test

lgp171188:
The test tries to test the betatest application status and the actual
provisioning of the appserver is not needed at all, particularly since
it takes a long time.


lgp171188:
Thanks for the review @clemente! I have rebased the PR branch and will merge after the CI checks pass 🤞

PR #379, opened 2019-01-16
BB-804 - Clean up old MySQL instances as part of CI cleanup

jamestait:
This PR adds an extra cleanup instance to the Integration Cleanup script, to remove MySQL databases that are older than three days.

## Testing

You can test the cleanup function on a local OCIM instance:

    $ vagrant ssh
    (opencraft) vagrant@ubuntu-xenial:~$ cd opencraft
    (opencraft) vagrant@ubuntu-xenial:~/opencraft$ mysql -u root -h localhost
    Welcome to the MySQL monitor.  Commands end with ; or \g.
    Your MySQL connection id is 20                                                                                                                      
    Server version: 5.7.24-0ubuntu0.16.04.1 (Ubuntu)
    
    Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.
    
    Oracle is a registered trademark of Oracle Corporation and/or its
    affiliates. Other names may be trademarks of their respective
    owners.
    
    Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.
    
    mysql> CREATE DATABASE f4ad7be3_integration_plebia_net_edxapp;
    Query OK, 1 row affected (0.01 sec)
    
    mysql> use f4ad7be3_integration_plebia_net_edxapp
    Database changed
    mysql> CREATE TABLE foo (id int);
    Query OK, 0 rows affected (0.04 sec)
    
    mysql> CREATE DATABASE unrelated)db;
    Query OK, 1 row affected (0.01 sec)
    
    mysql> use unrelated_db
    Database changed
    mysql> CREATE TABLE foo (id int);
    Query OK, 0 rows affected (0.04 sec)
    
    mysql> \q
    Bye
    (opencraft) vagrant@ubuntu-xenial:~/opencraft$ make shell
    ...
    In [1]: from cleanup_utils import mysql_cleanup
    
    In [2]: my_cleanup = mysql_cleanup.MySqlCleanupInstance(0, settings.DEFAULT_INSTANCE_MYSQL_URL, 'plebia.net', False)
    
    In [3]: my_cleanup._get_old_databases()
    Out[3]:
    (('f4ad7be3_integration_plebia_net_edxapp',
      datetime.datetime(2019, 1, 18, 11, 23, 58)),)
    
    In [4]: my_cleanup.run_cleanup()
    
    In [5]: my_cleanup._get_old_databases()
    Out[5]: ()
    
    In [6]: my_cleanup.cleaned_up_hashes
    Out[6]: ['f4ad7be3']
    
    In [7]: 
    Do you really want to exit ([y]/n)? 
    (opencraft) vagrant@ubuntu-xenial:~/opencraft$ mysql -u root -h localhost
    mysql> show DATABASES;
    +-------------------------------------------------+
    | Database                                        |
    +-------------------------------------------------+
    | information_schema                              |
    | ...                                             |
    | mysql                                           |
    | performance_schema                              |
    | sys                                             |
    | unrelated_db                                    |
    +-------------------------------------------------+
    38 rows in set (0.00 sec)

    mysql>

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ Before activating a server, there's the option to test it through a
 basic-auth password-protected link in the "Authenticated Link" section
 (the username and password are embedded in the link).
 
-Sometimes Open edX playbook fails, and then you need to read the log, which is
-shown in real-time in the web console.  You can fix the settings and then spawn
-another server.  Failed and old inactive servers and MySQL databases are
-automatically cleaned up after three days.  An important feature is that Ocim
-*grants SSH access* to members of a configurable GitHub organization, so you
-can always SSH to an appserver's IP, *even if Open edX's deployment failed*,
-and then debug it.  You can use your GitHub username and key.
+Sometimes Open edX playbook fails, and then you need to read the log,
+which is shown in real-time in the web console.
+You can fix the settings and then spawn another server.
+Failed and old inactive servers are automatically cleaned up after some configurable amount of days.
+An important feature is that Ocim *grants SSH access* to members of a configurable GitHub organization,
+so you can always SSH to an appserver's IP, *even if Open edX's deployment failed*, and then debug it.
+You can use your GitHub username and key.
 
 To create a new instance, you use Django's admin and you need to fill in the domain name,
 the prefixed domain names (for Studio, e-commerce, etc.), the edx-platform/configuration branches to use,
--- a/README.md
+++ b/README.md
@@ -639,10 +639,12 @@ your development environment is likely a good starting point:
     cp .env .env.integration
 
 
-There is also a cleanup routine intended for use by CI services to check for and
-clean up any dangling OpenStack VMs past a certain age threshold. While it isn't
-necessary in the usual case, old integration tests that were killed without cleanup
-can be cleaned up after four hours by running the make target:
+There is also a cleanup routine intended for use by CI services to check for
+and clean up any dangling OpenStack VMs and MySQL databases past a certain age
+threshold. While it isn't necessary in the usual case, old integration tests
+that were killed without cleanup that are older than four hours, and old MySQL
+databases that are older than three days can be cleaned up by running the make
+target:
 
     make test.integration_cleanup
 
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ Before activating a server, there's the option to test it through a
 basic-auth password-protected link in the "Authenticated Link" section
 (the username and password are embedded in the link).
 
-Sometimes Open edX playbook fails, and then you need to read the log,
-which is shown in real-time in the web console.
-You can fix the settings and then spawn another server.
-Failed and old inactive servers are automatically cleaned up after some configurable amount of days.
-An important feature is that Ocim *grants SSH access* to members of a configurable GitHub organization,
-so you can always SSH to an appserver's IP, *even if Open edX's deployment failed*, and then debug it.
-You can use your GitHub username and key.
+Sometimes Open edX playbook fails, and then you need to read the log, which is
+shown in real-time in the web console.  You can fix the settings and then spawn
+another server.  Failed and old inactive servers and MySQL databases are
+automatically cleaned up after three days.  An important feature is that Ocim
+*grants SSH access* to members of a configurable GitHub organization, so you
+can always SSH to an appserver's IP, *even if Open edX's deployment failed*,
+and then debug it.  You can use your GitHub username and key.
 
 To create a new instance, you use Django's admin and you need to fill in the domain name,
 the prefixed domain names (for Studio, e-commerce, etc.), the edx-platform/configuration branches to use,
@@ -638,24 +638,13 @@ your development environment is likely a good starting point:
 
     cp .env .env.integration
 
-
 There is also a cleanup routine intended for use by CI services to check for
 and clean up any dangling OpenStack VMs and MySQL databases past a certain age
 threshold. While it isn't necessary in the usual case, old integration tests
-that were killed without cleanup that are older than four hours, and old MySQL
-databases that are older than three days can be cleaned up by running the make
-target:
+that were killed without cleanup and old MySQL databases that are older than
+three days can be cleaned up by running the make target:
 
     make test.integration_cleanup
-
-The age threshold for the cleanup script defaults to four hours, but this can be
-adjusted by setting `INSTANCE_AGE_THRESHOLD` to a number (in seconds) in the
-`.env.integration` file.
-
-Note that, if executing locally and with the same environment as Circle CI, setting
-the `INSTANCE_AGE_THRESHOLD` to a number too low may result in interrupted
-integration test builds on Circle CI.
-
 
 Debug
 -----
--- a/cleanup_utils/integration_cleanup.py
+++ b/cleanup_utils/integration_cleanup.py
@@ -41,7 +41,8 @@ from cleanup_utils.openstack_cleanup import OpenStackCleanupInstance
 # Default age at which things should be cleaned up, in days
 DEFAULT_AGE_LIMIT = 3
 DEFAULT_CUTOFF_TIME = (
-    datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=DEFAULT_AGE_LIMIT))
+    datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=DEFAULT_AGE_LIMIT)
+)
 
 
 # Logging #####################################################################
@@ -118,7 +119,8 @@ def run_integration_cleanup(dry_run=False):
     # the deletion_blacklist
     hashes_to_clean = (
         aws_cleanup.cleaned_up_hashes + os_cleanup.cleaned_up_hashes +
-        mysql_cleanup.cleaned_up_hashes)
+        mysql_cleanup.cleaned_up_hashes
+    )
     dns_cleanup.run_cleanup(
         hashes_to_clean=hashes_to_clean
     )
--- a/cleanup_utils/integration_cleanup.py
+++ b/cleanup_utils/integration_cleanup.py
@@ -38,7 +38,10 @@ from cleanup_utils.openstack_cleanup import OpenStackCleanupInstance
 
 # Constants ###################################################################
 
-DEFAULT_AGE_LIMIT = datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=3)
+# Default age at which things should be cleaned up, in days
+DEFAULT_AGE_LIMIT = 3
+DEFAULT_CUTOFF_TIME = (
+    datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=DEFAULT_AGE_LIMIT))
 
 
 # Logging #####################################################################
@@ -84,7 +87,7 @@ def main():
 
     # Clean up AWS
     aws_cleanup = AwsCleanupInstance(
-        age_limit=DEFAULT_AGE_LIMIT,
+        age_limit=DEFAULT_CUTOFF_TIME,
         aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
         aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
         dry_run=args.dry_run
@@ -100,7 +103,7 @@ def main():
         'region_name': os.environ['OPENSTACK_REGION'],
     }
     os_cleanup = OpenStackCleanupInstance(
-        age_limit=DEFAULT_AGE_LIMIT,
+        age_limit=DEFAULT_CUTOFF_TIME,
         openstack_settings=openstack_settings,
         dry_run=args.dry_run
     )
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -49,13 +49,13 @@ Before activating a server, there's the option to test it through a
 basic-auth password-protected link in the "Authenticated Link" section
 (the username and password are embedded in the link).
 
-Sometimes Open edX playbook fails, and then you need to read the log,
```
giovannicimolin: These changes are outside the scope of BB-804. 
Please remove them from this PR.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Handles the cleanup of old MySQL databases
+    """
+    def __init__(self, age_limit, url, domain, dry_run):
+        """
+        Set up variables needed for cleanup
+        """
+        self.age_limit = age_limit
+        self.cleaned_up_hashes = []
+        self.cursor = _get_cursor(url)
```
giovannicimolin: If we pass a URL that is either broke or unaccessible, this will throw an exception. 
Can you add a handler for this to skip cleaning up MySQL if a connection error happens?
I've done something similar [here](https://github.com/open-craft/opencraft/blob/master/cleanup_utils/openstack_cleanup.py#L51).

We want this script to be as error tolerant as possible. This is needed because if this module throws an exception, we will lose all those hashes from the cleanups that ran before this one.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,124 @@
...
+    database_url_obj = urlparse(url)
+    try:
+        connection = mysql.connect(
+            host=database_url_obj.hostname,
+            user=database_url_obj.username or '',
+            passwd=database_url_obj.password or '',
+            port=database_url_obj.port or 3306,
+        )
+    except MySQLError as exc:
+        logger.excecption('Cannot get MySQL cursor: %s', exc)
```
giovannicimolin: There's a typo here: `excecption`

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,124 @@
...
+                )
+                continue
+
+            logger.info(
+                '    * Dropping MySQL DB `%s` created at %s', database,
+                datetime.strftime(create_date, '%Y-%m-%dT%H:%M:%SZ'))
+            if not self.dry_run:
+                try:
+                    self.cursor.execute(
+                        'DROP DATABASE IF EXISTS {}'.format(database))
```
giovannicimolin: Nit: Add a line break on the parenthesis.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,124 @@
...
+                    self.cleaned_up_hashes.append(match.groups()[0])
+                except MySQLError as exc:
+                    logger.exception(
+                        'Unable to remove MySQL DB: %s. %s', database, exc)
```
giovannicimolin: Nit: Add a line break on the parenthesis.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.dry_run:
+            logger.info("Running in DRY_RUN mode, no actions will be taken.")
+
+        databases = self._get_old_databases()
+        logger.info('Found %d old databases', len(databases))
+
+        for (database, create_date) in databases:
+            logger.info('  > Considering database %s', database)
+            instance_database_re = r'^([0-9a-f]{8})([_0-9a-z]+)?_%s' % (
+                self.domain_suffix,)
```
giovannicimolin: You don't need this line break here, this will improve readability.

cleanup_utils/integration_cleanup.py
```diff
@@ -31,13 +31,17 @@
...
+from cleanup_utils.openstack_cleanup import OpenStackCleanupInstance
 
 
 # Constants ###################################################################
 
-DEFAULT_AGE_LIMIT = datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=3)
+# Default age at which things should be cleaned up, in days
+DEFAULT_AGE_LIMIT = 3
+DEFAULT_CUTOFF_TIME = (
+    datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=DEFAULT_AGE_LIMIT))
```
lgp171188: @jamestait, don't you think the below style of brackets is more readable and looks cleaner?

```python
DEFAULT_CUTOFF_TIME = (
    datetime.utcnow().replace(tzinfo=UTC) - timedelta(days=DEFAULT_AGE_LIMIT)
)
```

cleanup_utils/integration_cleanup.py
```diff
@@ -103,7 +116,9 @@ def run_integration_cleanup(dry_run=False):
     )
     # Run DNS cleanup erasing all integration entries except for those on
     # the deletion_blacklist
-    hashes_to_clean = aws_cleanup.cleaned_up_hashes + os_cleanup.cleaned_up_hashes
+    hashes_to_clean = (
+        aws_cleanup.cleaned_up_hashes + os_cleanup.cleaned_up_hashes +
+        mysql_cleanup.cleaned_up_hashes)
```
lgp171188: @jamestait, similar to a previous comment, it is a good idea to move the closing parenthesis to a new line.

```python
hashes_to_clean = (
    aws_cleanup.cleaned_up_hashes + os_cleanup.cleaned_up_hashes +
    mysql_cleanup.cleaned_up_hashes
)
```

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,136 @@
...
+        query = """SELECT table_schema, MAX(create_time) AS create_time
+            FROM information_schema.tables
+            WHERE create_time <= DATE_SUB(NOW(), INTERVAL %(age_limit)s DAY)
+            AND table_schema LIKE %(domain_filter)s
+            GROUP BY table_schema ORDER BY create_time DESC"""
+        params = {
+            'age_limit': self.age_limit,
+            'domain_filter': '%_{}_%'.format(self.domain_suffix)
+        }
+        self.cursor.execute(query, params)
```
lgp171188: @jamestait, you have to wrap this in a `try-except` block and then log an error message if something goes wrong here.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if self.dry_run:
+            logger.info("Running in DRY_RUN mode, no actions will be taken.")
+
+        databases = self._get_old_databases()
+        logger.info('Found %d old databases', len(databases))
+
+        for (database, create_date) in databases:
+            logger.info('  > Considering database %s', database)
+            instance_database_re = r'^([0-9a-f]{8})([_0-9a-z]+)?_%s' % (self.domain_suffix,)
```
lgp171188: @jamestait, this may not be required because the databases are already filtered using the base domain name and it is unlikely to have a database on this RDS server which matches the domain name but doesn't match the pattern. All the databases in this MySQL server are used solely for the CI checks.

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,136 @@
...
+                    instance_database_re
+                )
+                continue
+
+            logger.info(
+                '    * Dropping MySQL DB `%s` created at %s', database,
+                datetime.strftime(create_date, '%Y-%m-%dT%H:%M:%SZ'))
+            if not self.dry_run:
+                try:
+                    self.cursor.execute(
```
lgp171188: @jamestait, I am wondering if we need to do this one by one instead of using `executemany()` to do it in one shot. I know that the performance of the `executemany()` method varies from driver to driver and also based on the backend, but can you check and find out if it makes sense to use it?

jamestait: From the docstring for `MySQLdb.cursors.executemany()`:

    """Execute a multi-row query.

    :param query: query to execute on server
    :param args:  Sequence of sequences or mappings.  It is used as parameter.
    :return: Number of rows affected, if any.

    This method improves performance on multiple-row INSERT and
    REPLACE. Otherwise it is equivalent to looping over args with
    execute().
    """

So I don't think we gain anything with `executemany()` but we'd lose the ability to report individually on each operation.

cleanup_utils/mysql_cleanup.py
jamestait: This is serving a dual purpose: it's an additional precaution against dropping unrelated databases, but it's also extracting the hash to return to the main script so it can be used to clean up old DNS entries. We could, perhaps, go ahead with the removal even if the RE doesn't match, in which case no hash would be returned for that database, but I've erred on the side of caution in case this is used elsewhere, where our assumptions don't hold.

lgp171188: @jamestait, it will be good to mention the gist of this explanation in a comment before this block of code.

cleanup_utils/integration_cleanup.py
jamestait: Yes, thank you - fixed this and the other instances. :)

cleanup_utils/mysql_cleanup.py
jamestait: Done. In addition, I've returned an empty list in that instance, so the remainder of the cleanup will complete normally - we'll log the exception, log that no databases will be cleaned up, and the DNS cleanup will proceed as normal.

cleanup_utils/mysql_cleanup.py
jamestait: You read my mind - added. :)

README.md
giovannicimolin: Ah, I've just asked because you also updated this block of text here, which isn't related to the cleanup.
```
Sometimes Open edX playbook fails, and then you need to read the log, which is
shown in real-time in the web console.  You can fix the settings and then spawn
another server.  Failed and old inactive servers and MySQL databases are
automatically cleaned up after three days.  An important feature is that Ocim
*grants SSH access* to members of a configurable GitHub organization, so you
can always SSH to an appserver's IP, *even if Open edX's deployment failed*,
and then debug it.  You can use your GitHub username and key.
```

jamestait: There are changes that are related to the PR:

 * "old inactive server **and MySQL databases** are"
 * "automatically cleaned up after **three days**" rather than "**some configurable amount of days**".

But then I re-flowed the text for better readability, which isn't related.

Would you prefer if I reverted these changes and made them in a separate PR?

giovannicimolin: @jamestait This paragraph is talking about Ocim automatic cleanup of Appservers and isn't related to the integration cleanup.
If you leave an old Appserver or a deployment through Ocim fails, it'll be automatically cleaned after 3 days, and this process isn't related to the CI runs :grin:

jamestait: Ah, got it! I thought it was actually the CI cleanup process that was responsible for this, sorry! In that case, I'll revert this change.

giovannicimolin:
@jamestait I think you are only missing the MySQL requirement on `cleanup_utils/requirements.txt`.

jamestait:
Thanks for the nudge, @giovannicimolin - I've added the requirement. :smiley_cat:

cleanup_utils/integration_cleanup.py
```diff
@@ -89,12 +94,21 @@ def run_integration_cleanup(dry_run=False):
...
         dry_run=dry_run
     )
     os_cleanup.run_cleanup()
 
+    # Run MySQL cleanup
+    mysql_cleanup = MySqlCleanupInstance(
+        age_limit=DEFAULT_AGE_LIMIT,
+        url=os.environ['DEFAULT_INSTANCE_MYSQL_URL'],
+        domain=os.environ['DEFAULT_INSTANCE_BASE_DOMAIN'],
+        dry_run=args.dry_run
```
lgp171188: @jamestait, there is no variable `args` in this method - it was removed in @giovannicimolin's recent PR which you had to rebase your changes on. So how did this even work? 🤔

```suggestion
        dry_run=dry_run
```

cleanup_utils/mysql_cleanup.py
```diff
@@ -0,0 +1,146 @@
...
+            FROM information_schema.tables
+            WHERE create_time <= DATE_SUB(NOW(), INTERVAL %(age_limit)s DAY)
+            AND table_schema LIKE %(domain_filter)s
+            GROUP BY table_schema ORDER BY create_time DESC"""
+        params = {
+            'age_limit': self.age_limit,
+            'domain_filter': '%_{}_%'.format(self.domain_suffix)
+        }
+        try:
+            self.cursor.execute(query, params)
```
lgp171188: @jamestait, wouldn't this statement throw an unhandled `AttributeError: 'NoneType' object has no attribute 'execute'` exception when the code to create a connection and then retrieve the cursor throws an exception, thereby causing the `self.cursor` value to be set to `None`?

cleanup_utils/integration_cleanup.py
giovannicimolin: @lgp171188 As the args definition happens outside a function, it becomes a global variable. Nevertheless, I've created `run_integration_cleanup` method so we can run the cleanup from python shell if needed, and this `args.dry_run` would throw an error, so it must be changed to the parameter `dry_run` as suggested.

jamestait: Great catch, thanks! Fixed.

cleanup_utils/mysql_cleanup.py
jamestait: Indeed it would. This wouldn't be a problem in the normal case, as `run_cleanup` would have already caught that and skipped the cleanup, but this method should be robust against such a case. I've moved the check to here, and just return an empty list in the event the cursor is `None`.

PR #380, opened 2019-01-23
BB-209: Add setting to disable celery heartbeats

giovannicimolin:
This PR adds a default ansible variable to disable celery heartbeats on supported configurations.

**Testing instructions:**
1. Check out this branch on your Ocim devstack
2. Create a new OpenEdX instance on Django Shell:
```
make shell
> instance = OpenEdXInstance.objects.create(        
>     name='Test Sandbox',                                                               
>     sub_domain='test-heartbeats',
> )
```
3. Spawn a new appserver from the web interface (you might have to do this twice, the first time it'll fail with a load balancer error)
4. Check on the appserver configuration if the ansible variable was correctly set: 
**Combined ansible vars:**
```
worker_django_enable_heartbeats: false
```

This is just to check if the variable wa correctly set, no need to wait for the AppServer to completely spawn (I'll set up Ocim stage for that).
Don't forget to delete your instance with `instance.delete()` after testing locally (you should never delete instances on stage or production.

**Reviewers:**
- [ ] @jamestait 
- [ ] @lgp171188

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,7 +495,6 @@ set the following configuration variables:
 * `OPENEDX_APPSERVER_SECURITY_GROUP_RULES`: This specifies the firewall rules
   that the above security group will have. The default allows ingress on ports
   22, 80, and 443 only.
-* `EDX_WORKERS_ENABLE_CELERY_HEARTBEAT`: Switch to enable/disable celery
 * `EDX_WORKERS_ENABLE_CELERY_HEARTBEATS`: Switch to enable/disable celery
   heartbeats used to detect connection drops. Disabling heartbeats can have a
   drastic reduction RabbitMQ usage. This setting sets
--- a/README.md
+++ b/README.md
@@ -495,6 +495,10 @@ set the following configuration variables:
 * `OPENEDX_APPSERVER_SECURITY_GROUP_RULES`: This specifies the firewall rules
   that the above security group will have. The default allows ingress on ports
   22, 80, and 443 only.
+* `EDX_WORKERS_ENABLE_CELERY_HEARTBEAT`: Switch to enable/disable celery
+  heartbeats used to detect connection drops. Disabling heartbeats can have a
+  drastic reduction RabbitMQ usage. This setting sets
+  `worker_django_enable_heartbeats` on supported playbooks. Defaults to `False`.
 
 Migrations
 ----------
--- a/README.md
+++ b/README.md
@@ -496,6 +496,7 @@ set the following configuration variables:
   that the above security group will have. The default allows ingress on ports
   22, 80, and 443 only.
 * `EDX_WORKERS_ENABLE_CELERY_HEARTBEAT`: Switch to enable/disable celery
+* `EDX_WORKERS_ENABLE_CELERY_HEARTBEATS`: Switch to enable/disable celery
   heartbeats used to detect connection drops. Disabling heartbeats can have a
   drastic reduction RabbitMQ usage. This setting sets
   `worker_django_enable_heartbeats` on supported playbooks. Defaults to `False`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -495,6 +495,11 @@ set the following configuration variables:
 * `OPENEDX_APPSERVER_SECURITY_GROUP_RULES`: This specifies the firewall rules
   that the above security group will have. The default allows ingress on ports
   22, 80, and 443 only.
+* `EDX_WORKERS_ENABLE_CELERY_HEARTBEAT`: Switch to enable/disable celery
```
jamestait: This line is unnecessary.

instance/models/mixins/openedx_config.py
```diff
@@ -297,7 +297,17 @@ def _get_configuration_variables(self):
...
-            "SANDBOX_ENABLE_INSIGHTS": False # set to true to enable insights
+            "SANDBOX_ENABLE_INSIGHTS": False, # set to true to enable insights
+
+            # Disable celery heartbeats
+            # The current default setup of Celery in Open edX uses something called
+            # heartbeats to detect connection drops to the celery broker (RabbitMQ).
+            # With as many 15 celery processes running on Open edX servers this can mean
+            # a lot of the RabbitMQ usage is used up just to check for connection drops,
+            # and that too mostly to get around issues with RabbitMQ behind a load-balancer,
+            # which is not our setup.
```
jamestait: Is this comment talking about a specific instance of Open edX, or all instances that are installed following the default installation guide?

giovannicimolin: By default, all instances provisioned using the EdX configuration repo use the heartbeats.
I've improved the text a little. :)

giovannicimolin: Answering your PR question, you can deploy a RabbitMQ server in some ways depending on your need. 
The heartbeats used by Celery are mainly there to check for connection blocks, and are critical when using RabbitMQ behind a load balancer, which is not our setup. You can read up more about this and how the decision was made by reading the discovery ticket. :grin:

jamestait: Ah, my question was more related to the "behind a load balancer, which is not our setup" part. I'm not sure who "our" refers to here - if not being behind a load balancer is the general case with all OCIM setups, something that's enforced, something that's a default setup but could be modified, or if we should reword it slightly to something like "which is not the standard setup".

giovannicimolin: I'm talking about only the RabbitMQ server: in our case (OpenCraft), we don't use the RabbitMQ server behind a load balancer.

giovannicimolin: I've reworded the comment :grin:

jamestait: Thanks, I think that reads better. I'm conscious that OCIM is open source AGPL software, so potentially is being used by others, which is why I'm a bit cautious about the use of "our" in documentation. :)

jamestait:
:+1:

- [x] I tested this:
 * Applied the changes to my local OCIM instance.
 * Verified the default case: `worker_django_enable_heartbeats` is `false`.
 * Verified that setting `EDX_WORKERS_ENABLE_CELERY_HEARTBEATS` sets `worker_django_enable_heartbeats` correctly.
- [x] I read through the code
- [x] Includes documentation

opencraft/settings.py
```diff
@@ -479,6 +479,9 @@
     },
 ]
 
+# Enable or disable celery heartbeats on instances managed by Ocim
+EDX_WORKERS_ENABLE_CELERY_HEARTBEATS = env.bool('EDX_WORKERS_ENABLE_CELERY_HEARTBEATS', default=False)
```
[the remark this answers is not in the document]
giovannicimolin: @lgp171188 I'll make a PR, can you give me permissions to the repo?

PR #384, opened 2019-02-06
BB-834: Fix simple-theme deployments

giovannicimolin:
This PR fixes a lot of styling errors and improves readability for simple-theme deployments on Ocim.

**Testing instructions:**
Note: These changes are deployed to Ocim stage.
1. Use Ocim stage to deploy an instance using [the beta testing application form](https://stage.console.opencraft.com/registration/) and this [configuration branch](https://github.com/open-craft/configuration/pull/87).
2. Check if the theme is correctly deployed in the appserver.

I already done this with 2 sets of colors, so you can check the result at:
Light colors: https://stage.console.opencraft.com/instance/3230/edx-appserver/1022/
Darker colors: https://stage.console.opencraft.com/instance/3230/edx-appserver/1020/

**Reviewers:**
- [ ] @lgp171188

### DIFF
--- a/instance/models/mixins/openedx_theme.py
+++ b/instance/models/mixins/openedx_theme.py
@@ -73,8 +73,6 @@ class OpenEdXThemeMixin(models.Model):
                     "variable": "link-color",
                     "value": application.link_color,
                 },
-                # TODO: These are specific to Ginkgo and can be removed
-                # after Hawthorn upgrade
                 {
                     "variable": "header-bg",
                     "value": application.header_bg_color,
@@ -83,7 +81,6 @@ class OpenEdXThemeMixin(models.Model):
                     "variable": "footer-bg",
                     "value": application.footer_bg_color,
                 },
-                # END TODO
                 {
                     "variable": "button-color",
                     "value": application.main_color,
@@ -102,16 +99,6 @@ class OpenEdXThemeMixin(models.Model):
                         primary=application.main_color,
                         secondary=application.main_color
                     ),
-                },
-            ],
-            "SIMPLETHEME_SASS_BOOTSTRAP_OVERRIDES": [
-                {
-                    "variable": "primary",
-                    "value": application.main_color,
-                },
-                {
-                    "variable": "secondary",
-                    "value": application.main_color,
                 },
             ],
             "SIMPLETHEME_STATIC_FILES_URLS": [
--- a/instance/models/mixins/openedx_theme.py
+++ b/instance/models/mixins/openedx_theme.py
@@ -120,9 +120,9 @@ class OpenEdXThemeMixin(models.Model):
                 $main-color: {main_color};
                 $link-color: {link_color};
                 $header-bg: {header_bg};
-                $header-font-color {header_font_color};
+                $header-font-color: {header_font_color};
                 $footer-bg: {footer_bg};
-                $footer-font-color {footer_font_color};
+                $footer-font-color: {footer_font_color};
             """.format(
                 link_color=application.link_color,
                 main_color=application.main_color,
--- a/instance/models/mixins/openedx_theme.py
+++ b/instance/models/mixins/openedx_theme.py
@@ -22,6 +22,7 @@ Open edX instance theme mixin, e.g. for simple_theme related settings
 import yaml
 
 from django.db import models
+from django.conf import settings
 
 
 # Classes #####################################################################
@@ -65,7 +66,7 @@ class OpenEdXThemeMixin(models.Model):
             return ""
 
         # These settings set the values required by simple_theme
-        settings = {
+        theme_settings = {
             # This block defines our theme by applying the chosen colors to SASS-defined color variables
             "SIMPLETHEME_SASS_OVERRIDES": [
                 {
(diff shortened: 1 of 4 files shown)


instance/models/mixins/openedx_theme.py
```diff
@@ -117,12 +127,94 @@ def get_theme_settings(self):
             "EDXAPP_DEFAULT_SITE_THEME": "simple-theme",
             "SIMPLETHEME_EXTRA_SASS": """
                 .global-header {{
-                    background: {header_bg};
+                    background: {header_bg}; // Header color
```
lgp171188: @giovannicimolin, is it possible to avoid adding all the SASS/CSS code in Ocim and instead use it from an external skeleton theme? That way, Ocim code will be cleaner and separate.

instance/models/mixins/openedx_theme.py
```diff
@@ -117,12 +127,94 @@ def get_theme_settings(self):
...
+
+                    span {{
+                      color: {link_color}; // Link Color
+                    }}
+
+                    .nav-links .secondary {{
+                      a.sign-in-btn,
+                      a.register-btn:hover
+                      {{
+                        color: {header_bg} !important; // Header color
```
lgp171188: @giovannicimolin, did you check if you could add more specific selectors instead of using `!important` which is generally not a good thing to do?

instance/models/mixins/openedx_theme.py
```diff
@@ -65,15 +67,13 @@ def get_theme_settings(self):
...
         # These settings set the values required by simple_theme
-        settings = {
+        theme_settings = {
             # This block defines our theme by applying the chosen colors to SASS-defined color variables
             "SIMPLETHEME_SASS_OVERRIDES": [
                 {
                     "variable": "link-color",
                     "value": application.link_color,
                 },
-                # TODO: These are specific to Ginkgo and can be removed
```
lgp171188: @giovannicimolin, you can remove the enclosed styles as per the removed comment. :smiley:

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -85,13 +85,10 @@ def test_colors_and_images_applied(self):
                 'SIMPLETHEME_SASS_OVERRIDES': [
                     {'variable': 'link-color',
                      'value': '#003344', },
-                    # TODO: These are specific to Ginkgo and can be removed
```
lgp171188: @giovannicimolin, the enclosed styles can be removed when the corresponding ones above are removed.

instance/models/mixins/openedx_theme.py
```diff
@@ -115,14 +114,48 @@ def get_theme_settings(self):
...
+                    application.footer_bg_color
+                ),
+            )
         }
 
-        return yaml.dump(settings, default_flow_style=False)
+        return yaml.dump(theme_settings, default_flow_style=False)
+
+    @staticmethod
+    def get_contrasting_font_color(background_color, delta=0.5):
```
lgp171188: @giovannicimolin, can you write some tests for this method just to make sure it gives the expected results for a few colors?

opencraft/settings.py
```diff
@@ -482,6 +482,13 @@
 # Enable or disable celery heartbeats on instances managed by Ocim
 EDX_WORKERS_ENABLE_CELERY_HEARTBEATS = env.bool('EDX_WORKERS_ENABLE_CELERY_HEARTBEATS', default=False)
 
+# Open EdX Instance custom theme for clients ##################################
+EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO = env(
```
lgp171188: @giovannicimolin, just asking to be sure. These can be overridden in the extra settings right? Also, can the settings variables be renamed to indicate that they are related to the simple-theme? Right now it is not clear what they are used for and where.

instance/models/mixins/openedx_theme.py
giovannicimolin: Sure, done!

opencraft/settings.py
giovannicimolin: @lgp171188 Yes, we can always override those. 
I don't think it's good for us to change the names because those are the same that Ansible uses. I can change them to `CUSTOM_THEME_REPO` and `CUSTOM_THEME_VERSION` if you'd like, but those are new variable names. Do you have any better suggestion on this?

giovannicimolin: @lgp171188 I changed the variable names to `SIMPLE_THEME_SKELETON_THEME_REPO` and `SIMPLE_THEME_SKELETON_THEME_VERSION` and added a small description to the variables.

instance/models/mixins/openedx_theme.py
```diff
@@ -115,14 +106,52 @@ def get_theme_settings(self):
...
+            "EDXAPP_COMPREHENSIVE_THEME_VERSION": settings.SIMPLE_THEME_SKELETON_THEME_VERSION,
             "SIMPLETHEME_EXTRA_SASS": """
-                .global-header {{
-                    background: {header_bg};
-                }}
-                .wrapper-footer {{
-                    background: {footer_bg};
-                }}""".format(header_bg=application.header_bg_color,
-                             footer_bg=application.footer_bg_color)
+                $main-color: {main_color};
```
lgp171188: @giovannicimolin, since this is only defining new variables, can't these be set in the `SIMPLETHEME_SASS_OVERRIDES` above?

opencraft/settings.py
```diff
@@ -482,6 +482,17 @@
 # Enable or disable celery heartbeats on instances managed by Ocim
 EDX_WORKERS_ENABLE_CELERY_HEARTBEATS = env.bool('EDX_WORKERS_ENABLE_CELERY_HEARTBEATS', default=False)
 
+# Open EdX Instance custom theme for Ocim Managed client instances
+# The `simple-theme` ansible role uses the below repository as a skeleton
```
lgp171188: @giovannicimolin, how about making this
```
# This repository is used as a skeleton theme for the betatest instances
# to apply the specified theming configuration through the `simple-theme` role.
```

opencraft/settings.py
```diff
@@ -482,6 +482,17 @@
 # Enable or disable celery heartbeats on instances managed by Ocim
 EDX_WORKERS_ENABLE_CELERY_HEARTBEATS = env.bool('EDX_WORKERS_ENABLE_CELERY_HEARTBEATS', default=False)
 
+# Open EdX Instance custom theme for Ocim Managed client instances
```
lgp171188: This can be removed if the below suggestion is accepted

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -146,3 +140,31 @@ def test_simpletheme_optout(self):
...
+    def test_get_contrasting_font_color(self):
+        """
+        Tests if the automatic font color selection is working properly
+        """
+        # Create objects
+        OpenEdXInstanceFactory(name='Integration - test_colors_applied', deploy_simpletheme=True)
+        instance = OpenEdXInstance.objects.get()
+
+        # Test if the font color is correctly returned depending on the background color
+        test_colors = [
```
lgp171188: @giovannicimolin, use `ddt.data` for this.

instance/models/mixins/openedx_theme.py
giovannicimolin: @lgp171188 I can't do that, because I'm not able to access those values on the overrides when I tried at first.
I'll give this a couple more tries, but if I don't succeed I'll leave this as is :grin:

opencraft/settings.py
giovannicimolin: :+1:

opencraft/settings.py
giovannicimolin: :+1:

instance/tests/models/test_openedx_theme_mixins.py
giovannicimolin: :+1:

instance/models/mixins/openedx_theme.py
giovannicimolin: @lgp171188 I had to leave them as is. :disappointed:

PR #385, opened 2019-02-15
[BB-835] Use a sub-domain instead of a prefixed domain for Studio, Preview etc.

lgp171188:
**Description**
From the Ironwood release, it is a [requirement](https://groups.google.com/d/msg/openedx-ops/521im9B92Xo/nS2aLWOGHwAJ) to serve the LMS and Studio from cookie-compatible domains. So the studio domain is now a sub-domain of the LMS domain. To allow session sharing, the `EDXAPP_SESSION_COOKIE_DOMAIN` ansible variable has been set to `.<LMS domain>`. For consistency, all the other domains - preview, discovery, ecommerce are updated to be sub-domains of the LMS domain.

This also adds a new ansible variable `EDXAPP_LOGIN_REDIRECT_WHITELIST` which allows the redirect to Studio after logging in via the LMS.

**Testing instructions**
1. Run Ocim with the PR branch.
1. Apply the migrations.
1. Create a new instance. Verify that the internal domains for Studio, Preview, Discovery and E-Commerce are now sub-domains of the internal LMS domain and that the `enable_prefix_domains_redirect` field is set to `False`.
1. Launch a new appserver for the instance and activate it after the provisioning completes successfully.
1. Verify that the load-balancer configuration is added correctly for the internal domains and that there are no redirects setup for the instance.
1. Add external domains for the same instance and launch a new appserver. Verify that the configuration is added correctly for the internal and external domains and that there are no redirects setup for the instance.
1. Create a new instance and set the `enable_prefix_domains_redirect` field to `True`.
1. Launch a new appserver for the instance and activate it after the provisioning completes successfully.
1. Verify that the load-balancer configuration is added correctly for the internal domains and that there are redirects from the prefix domains to the corresponding sub-domains. Verify that the DNS records have been added for the prefix domains and that the redirect works as expected.
1. Add external domains for the same instance and launch a new appserver. Verify that the configuration is added correctly for the internal and external domains and that the redirects are setup only for the prefix domains to the corresponding internal sub-domains. Verify that the internal domains, prefix domains and the external domains work as expected.
1. Optionally, repeat the same steps for the previous instance on an existing instance after updating its internal domains to be sub-domains.
1. Archive the instance and verify that the DNS records for the prefix domains are deleted.
1. Optionally, verify that these changes do not impact the existing hawthorn instances.

**Reviewers**
- [ ] @viadanna


instance/models/mixins/openedx_config.py
```diff
@@ -75,7 +75,8 @@ def _get_configuration_variables(self):
 
             # Set this to a string such as ".myinstance.org" to enable session sharing between LMS and the Studio.
             # We cannot do this on OC IM for security reasons (we don't want different *instances* to share cookies).
-            "EDXAPP_SESSION_COOKIE_DOMAIN": '',
+            "EDXAPP_SESSION_COOKIE_DOMAIN": '.{}'.format(self.instance.domain),
```
viadanna: During my testing, I found that `instance.domain` and `instance.internal_lms_domain` are the same for instances created using the default subdomains.

These seem to be used interchangeably thorough the code, do you know if these can take different values? I couldn't find one such example.

instance/tests/models/test_openedx_instance.py
```diff
@@ -426,18 +426,37 @@ def _check_load_balancer_configuration(self, backend_map, config, domain_names,
...
 
+    def _check_load_balancer_configuration_prefix_domains(self,  # pylint: disable=invalid-name
+                                                          backend_map,
+                                                          config,
+                                                          domain_names,
+                                                          ip_address):
+        """
+        Verify the load balancer configuration given in backend_map and config when the instance has prefix domains.
+        """
+        default_backend_map, _ = backend_map[:-4], backend_map[-4:]
```
viadanna: Nit. This can be simplified by removing the second term being assigned to `_`

instance/tests/test_factories.py
```diff
@@ -39,7 +39,7 @@
 class FactoriesTestCase(TestCase):
     """
     Test cases for functions in the factories module
-    """
+    # """
```
viadanna: Nit, please remove the `# `

instance/models/mixins/openedx_config.py
lgp171188: @viadanna, based on my understanding, `instance.domain` maps to the internal LMS domain or the external LMS domain when it is defined. But I am not able to find out the exact location where it is assigned. Will have to check more.

viadanna: 👍

lgp171188:
@viadanna, I tested this PR on stage and have listed my findings on the ticket.

PR #386, opened 2019-02-21
[BB-857] Falsy values stored in Consul are still values.

UmanShahzad:
When we had an actual value at some key path in Consul, such as `[]` for `ocim/instances/<ID>/active_app_servers`, Ocim would bring it in and assume that because it's a falsy value that it shouldn't be updated. So if the instance had any new active appservers, nothing would get updated.

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -479,7 +479,7 @@ class OpenEdXInstance(
         version_number = agent.get('version') or 0
         for key, value in configurations.items():
             index, stored_value = agent.get(key, index=True)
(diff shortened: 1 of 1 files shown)


UmanShahzad:
@SSPJ Yeah the only things using the KV store are Ocim (for this), HAProxy (for this ticket's work), and Terraform for LabXchange.

UmanShahzad:
@SSPJ I realized my earlier message may not have been clear: Ocim only uses this code for sending info to Consul where the only client is currently HAProxy. Since this ticket is the first time we finally make use of it, there's nothing relying on this bug. :+1:

PR #388, opened 2019-02-27
BB-962: Make base playbook configurable

giovannicimolin:
This PR adds support for changing the name of the base Open edX playbook. This is needed because edX recently changed the name of the base playbook from `edx_sandbox.yml` to `openedx_native.yml`. 
[More info](https://github.com/edx/configuration/pull/5025)

**Testing instructions:**
1. Using `master`, create a new OpenEdXInstance for `ginkgo`, `hawthorn`, `ironwood` and `master` on your Ocim devstack.
2. Pull this branch and apply migrations.
3. Check that all the instances that you created have `OpenEdXInstance.configuration_playbook_name` set to the correct playbook name:
* `ginkgo`: `playbooks/edx_sandbox.yml`
* `hawthorn`: `playbooks/edx_sandbox.yml`
* `ironwood`: `playbooks/openedx_native.yml`
* `master`: `playbooks/openedx_native.yml`
4. Create new instances for each release we support (`ginkgo`, `hawthorn`, `ironwood`, `master`) and leave the `configuration_playbook_name` empty, check if the correct playbook name was filled.
5. Click on **Launch New Appserver** and check if `configuration_playbook_name` was correctly set there too.
6. Check the instances on Ocim stage that were deployed for every type of deployment we support. The instances don't need to provision successfully, you just need to check if the correct playbook was used, and you can do that by checking for this line on the logs:
```
Running playbook "playbooks/openedx_native.yml" from "https://github.com/edx/configuration.git"
```
Deployed instances:
* [Master](https://stage.console.opencraft.com/instance/3280/) 
* [Hawthorn](https://stage.console.opencraft.com/instance/3290/) 
* [Ironwood](https://stage.console.opencraft.com/instance/3300/)
* [Beta test instance ](https://stage.console.opencraft.com/instance/3310/)
* [PR watcher sandbox ](https://stage.console.opencraft.com/instance/3330/)

**Reviewers:**
- [ ] @lgp171188


instance/models/utils.py
```diff
@@ -79,6 +79,25 @@ def default_setting(name):
     return functools.partial(_get_setting, name)
 
 
+def get_base_playbook_name(openedx_release):
+    """
+    This functions returns the correct main playbook name for the given Open EdX
+    release
+
+    This is needed because OpenEdX changed the main playbook name from
+    `edx_sandbox.yml` to `openedx_native.yml`. This is needed because the old
```
lgp171188:
```suggestion
    `edx_sandbox.yml` to `openedx_native.yml` and the old
```

instance/models/utils.py
```diff
@@ -79,6 +79,25 @@ def default_setting(name):
...
+    releases still use the old playbook name.
```
lgp171188:
```suggestion
    releases before Ironwood still use the old playbook.
```

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
```
lgp171188: @giovannicimolin, can the file be named meaningfully instead of using an auto-generated name?

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
...
+    # OpenEdXInstance model
+    queryset = OpenEdXInstance.objects.all()
+    # Make all saves() in a single transaction
+    with transaction.atomic():
+        for item in queryset:
+            item.configuration_playbook_name = get_base_playbook_name(item.openedx_release)
+            item.save()
+
+def backwards_func(apps, schema_editor):
+    pass
```
lgp171188: @giovannicimolin, doesn't it make sense to throw error or at least log a warning when a backward migration is attempted since that will result in loss of data? I know the data for this field can be calculated, but this will still be good to have.

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2019-02-27 15:34
+from __future__ import unicode_literals
+
+from django.db import migrations, models, transaction
+
+
+def forwards_func(apps, schema_editor):
```
lgp171188: @giovannicimolin, wouldn't it be better to name the function based on what it actually does. Ditto for the reverse data migration function.

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
...
+            'hawthorn',
+        ]
+        if any([release_name in openedx_release for release_name in old_playbook_releases]):
+            return 'playbooks/edx_sandbox.yml'
+        return 'playbooks/openedx_native.yml'
+
+    # We get the model from the versioned app registry;
+    # if we directly import it, it'll be the wrong version
+    OpenEdXInstance = apps.get_model("instance", "OpenEdXInstance")
+    db_alias = schema_editor.connection.alias
```
lgp171188: @giovannicimolin, this is unused.

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
...
+    # We only need to update the OpenEdXInstance, because these AppServer
+    # should be immutable, and new AppServers will use the correct values from
```
lgp171188:
```suggestion
    # are immutable, and new AppServers will use the correct values from
```

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+        if any([release_name in openedx_release for release_name in old_playbook_releases]):
+            return 'playbooks/edx_sandbox.yml'
+        return 'playbooks/openedx_native.yml'
+
+    # We get the model from the versioned app registry;
+    # if we directly import it, it'll be the wrong version
+    OpenEdXInstance = apps.get_model("instance", "OpenEdXInstance")
+    db_alias = schema_editor.connection.alias
+    # We only need to update the OpenEdXInstance, because these AppServer
```
lgp171188: Since the `configuration_playbook_name` is a field in the `OpenEdXInstance` model, I don't think the below explanation is needed. If you think it is still needed, can you check the below suggestions out?
```suggestion
    # We only need to update the OpenEdXInstance, because AppServers
```

instance/migrations/0114_auto_20190227_1534.py
```diff
@@ -0,0 +1,59 @@
...
+    # should be immutable, and new AppServers will use the correct values from
+    # OpenEdXInstance model
+    queryset = OpenEdXInstance.objects.all()
+    # Make all saves() in a single transaction
+    with transaction.atomic():
+        for item in queryset:
```
lgp171188: @giovannicimolin, can't you just use the `.update()` method on the queryset using the `F()` method to pass the value to the `get_base_playbook_name()` method?

instance/models/openedx_appserver.py
```diff
@@ -163,6 +167,17 @@ class Meta:
         )
     )
 
+    def get_configuration_playbook_name(self):
```
lgp171188: @giovannicimolin, wouldn't it be cleaner to make this a property?

instance/tests/integration/integration_instance.py
```diff
@@ -279,6 +278,7 @@ def test_spawn_appserver(self):
         """
         OpenEdXInstanceFactory(
             name='Integration - test_spawn_appserver',
+            configuration_playbook_name='playbooks/opencraft_integration.yml',
```
lgp171188: @giovannicimolin, wouldn't it be better to mock the `get_base_playbook_name()` method to return this value for all the tests instead of repeating this argument in every test for every case where we do not expect the playbook to fail?

instance/tests/models/test_openedx_ansible_mixins.py
```diff
@@ -90,7 +90,7 @@ def test_provisioning(
             inventory_str=mock_inventory,
             vars_str=appserver.configuration_settings,
             playbook_path='{}/playbooks'.format(working_dir),
-            playbook_name='edx_sandbox.yml',
+            playbook_name='openedx_native.yml',
```
lgp171188: @giovannicimolin, it might be worth testing for the ginkgo, hawthorn releases by passing the release and the expected value via ddt.

instance/tests/models/test_utils.py
```diff
@@ -685,53 +687,53 @@ def test_put_no_prefix(self):
         value = 'value'
         agent.put(key, value)
 
-        _, data = self.client.kv.get(key)
-        fetched_value = data['Value'].decode()
+        _, test_data = self.client.kv.get(key)
```
lgp171188: @giovannicimolin, why are these changes needed and how are they related to this PR?

giovannicimolin: The name of the variable was `data`, so I kept getting pylint errors because I was overriding `ddt.data`. That's why I've changed these names.

lgp171188: @giovannicimolin, we usually just import `ddt` and use it as `ddt.data`, `ddt.unpack` :smiley:

instance/migrations/0114_auto_20190227_1534.py
giovannicimolin: ooops, missed that when querying.

instance/migrations/0114_auto_20190227_1534.py
giovannicimolin: Makes sense, I'll remove this comment.

instance/tests/models/test_utils.py
giovannicimolin: I've reverted these changes

instance/migrations/0114_auto_20190227_1534.py
giovannicimolin: We can't do this because  `F()` is a operation that is performed on the database, so we can pass this values through any function. Only arithmetic operations are allowed. [ref](https://docs.djangoproject.com/en/2.1/ref/models/expressions/)

instance/models/openedx_appserver.py
giovannicimolin: Yes, and done.

instance/tests/integration/integration_instance.py
giovannicimolin: Done!

instance/migrations/0114_introduce_base_playbook_name.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2019-02-27 15:34
+from __future__ import unicode_literals
+
+import logging
+
+from django.db import migrations, models, transaction
+from django.db.models import F
```
lgp171188: @giovannicimolin, this is an unused import

giovannicimolin: Done

giovannicimolin: Fixed.

giovannicimolin:
@lgp171188 The latest commit is deployed on stage. Do you need me to deploy any other test instances there?

cleanup_utils/aws_cleanup.py
```diff
@@ -129,7 +129,7 @@ def get_user_policy(self, username):
                 UserName=username,
                 PolicyName=DEFAULT_POLICY_NAME
             )
-        except botocore.exceptions.NoSuchEntityException:
+        except self.iam_client.exceptions.NoSuchEntityException:
```
lgp171188: @giovannicimolin, didn't we discuss and decide to make a separate PR for this? If so, can you remove it?

instance/tests/integration/integration_instance.py
```diff
@@ -362,7 +361,9 @@ def test_external_databases(self):
         if not settings.DEFAULT_INSTANCE_MYSQL_URL or not settings.DEFAULT_INSTANCE_MONGO_URL:
             print('External databases not configured, skipping integration test')
             return
-        OpenEdXInstanceFactory(name='Integration - test_external_databases')
+        OpenEdXInstanceFactory(
```
lgp171188: @giovannicimolin , can you revert this and the similar changes to formatting now that we do not require an additional argument to be passed?

cleanup_utils/aws_cleanup.py
giovannicimolin: Yes, I missed it. I'll update the PR.

instance/tests/integration/integration_instance.py
giovannicimolin: Done!

cleanup_utils/aws_cleanup.py
giovannicimolin: Done

giovannicimolin:
@lgp171188 I"ve addressed the nits and will merge this as soon as the CI checks are green.

PR #390, opened 2019-03-01
[BB-856] Use locks when updating Consul.

UmanShahzad:
Test instructions:

1. Go to https://stage.console.opencraft.com/instance/3360/edx-appserver/1064/
1. SSH into haproxy-stage-1.net.opencraft.hosting and haproxy-stage-2.net.opencraft.hosting
1. On one server run `consul watch -type=keyprefix -prefix=lock/ consul kv get -recurse -detailed lock/`
1. On the other server run `consul watch -type=keyprefix -prefix=ocim/instances/362/ consul kv get -recurse ocim/instances/362`
1. Deactivate one server, activate another
1. After some time passes, check the spam from the watches - the lock is being worked on by both Ocim and consul-template across the HAProxy servers.
1. After it's settled down, check `consul kv export lock/` and notice no lock is taken anymore by anyone.


instance/models/utils.py
```diff
@@ -423,9 +433,48 @@ class ConsulAgent(object):
...
+        self.lock_wait_sleep = lock_wait_sleep
+
+    def __enter__(self):
+        # Any `put` operation using a session with "delete" behavior is ephemeral.
+        # So if we create the lock below with this session, when the session is
+        # invalidated (TTL completed, session was destroyed, etc.) the lock key
+        # is also deleted. However there is a default lock delay of 15 seconds, so
+        # new locks have to wait for that timer even if the lock key is physically gone.
+        # See https://www.consul.io/docs/internals/sessions.html for details.
+        self.session_id = self._client.session.create(behavior="delete")
```
dupainaulevain: shouldn't node be provided as well?

UmanShahzad: @dupainaulevain https://python-consul.readthedocs.io/en/latest/#consul.base.Consul.Session.create - it'll use the current node.

dupainaulevain:
Since the session may be invalidated at any time, the ConsultAgent.put also needs to set the  within the context manager also needs to set the session_id and handle the **session invalid** failure  somehow. Which makes me wonder if a context manager is the best way to ensure the data is consistent.

dupainaulevain:
Note to self: data is read from consul, modified, and written back to consul. What happens if data is modified in consul **after** it is being read?

UmanShahzad:
@dupainaulevain The way we're doing locks is using them to essentially emulate a transaction. I realize in retrospect that since Consul already has a transaction endpoint we should just use that, but I'm not sure how mature that functionality is, and unfortunately it'd require a pretty big refactor to the current Ocim-Consul integration that was built some time ago by people who aren't here now / on a long vacation.

So to explain it, the idea is to take a common lock on `/lock/ocim/.lock`, between Ocim, the cert manager, and `consul watch` on the HAProxy servers which will run `consul lock` and consequently render HAProxy configuration. The reason is that we don't want a weird situation where Consul gets a single update from Ocim to an instance configuration, and the consul-template daemon on the load balancing servers suddenly updates configuration without waiting for Ocim to have completed _all_ of its updates. That could lead to a flawed configuration and potentially downtime if it's severely bad enough.

Regarding how this lock tactic works: when you acquire a lock on `/lock/ocim/.lock`, obviously no one else can write to it, while still being able to write anywhere else. But we program all clients who need to write to paths with prefix `ocim/` to respect the lock at `/lock/ocim/.lock`. A stray client doesn't have to, and could just write to `ocim/` anyway, but at least the automation we program will respect those locks. And performance is not an issue at all, since writes to that path are highly infrequent.

So because of this, we don't need to use locks on a per-put basis - we create a session, grab a lock on `/lock/ocim/.lock` with it, and go about our updates as usual, with the understanding that clients will respect our lock because they were programmed to do so. Per-`put`-level locks don't help at all here because they're locking something for no reason.

UmanShahzad:
I think one addition is to put at least some TTL on the sessions, if there isn't already a default. If Ocim has an error that crashes the process before we get to release the session, and it has no TTL, we would have to manually erase the lock to allow other writes to continue.

instance/models/openedx_instance.py
```diff
@@ -483,19 +480,17 @@ def _write_metadata_to_consul(self, configurations):
...
-            if not version_updated and value != stored_value:
-                version_updated = True
-                version_number += 1
-                agent.put('version', version_number)
+        with ConsulAgent(prefix=self.consul_prefix) as agent:
+            version_updated = False
+            version_number = agent.get('version') or 0
+            for key, value in configurations.items():
+                index, stored_value = agent.get(key, index=True)
+                cas = index if stored_value is not None else 0
```
dupainaulevain: what is this **cas** guarding against? While I would understand getting the index when the value is first read, then transformed and eventually written back iff the index did not change, I don't get what is the value here.

dupainaulevain:
@UmanShahzad thanks for the explanations. I think I'm now in a good position to usefully review this PR when you need me to :-)

> I think one addition is to put at least some TTL on the sessions

Agreed! Are you thinking 30 seconds ? Should \_\_exit__ check if the session is still valid before destroying it? So it raises in the case the session expired before the context manager had a chance to complete?

dupainaulevain:
@UmanShahzad the context manager looks like something python-consul would accept, don't you think? And cert-manager/cert_manager/backends.py in cert-manager could use it too. Otherwise the code would need to be duplicated. Or... there is something I'm not seeing maybe?

instance/models/openedx_instance.py
dupainaulevain: Ok, I get it now :)

UmanShahzad:
> Agreed! Are you thinking 30 seconds ? Should __exit__ check if the session is still valid before destroying it? So it raises in the case the session expired before the context manager had a chance to complete?

I was thinking 20 minutes! The work done to update Consul wasn't done in a way that allows it to be highly performant, so it can take some time to complete. I'd prefer to be on the safe side, because there's usually no rush here. The TTL is just to make sure it doesn't go on forever and require manual intervention.

With that much time, I think we should be alright and not need to check for pre-mature expiration. Even if it did happen, throwing an exception wouldn't fix anything, because we would have already done all of our writes.

> the context manager looks like something python-consul would accept, don't you think? And cert-manager/cert_manager/backends.py in cert-manager could use it too. Otherwise the code would need to be duplicated. Or... there is something I'm not seeing maybe?

Yeah I'm not sure what to do about that yet. The context manager as it currently stands has I think the minimum logic needed to get locks working the way we want, but I don't know if upstream would accept it, and we don't have a big budget to rest on for this project.

I think what I'll do for the cert-manager is use transactions so we don't have to care about locks there - there's only a single place where we use `.put`, inside of `deploy_cert`, so we can just change callers of `deploy_cert` to use a transaction and change `deploy_cert` so it doesn't actually `put` but returns what is needed for transactions. I'll see.

UmanShahzad:
Oh you know what, for the cert-manager we might not even need to do any locking within the code nor transactions! I'll comment on the ticket.

instance/models/utils.py
```diff
@@ -423,9 +432,48 @@ class ConsulAgent(object):
...
     """
-    def __init__(self, prefix=''):
+
+    # We use the `consul lock` CLI to acquire locks on the server side.
+    # Those use a "magic flag" to help indicate conflicts with semaphores.
+    # We want to compete with those locks though, so we use this flag too.
+    # Comes from https://github.com/hashicorp/consul/blob/master/api/lock.go#L35-L38
+    LOCK_MAGIC_FLAG = 0x2ddccbc058a50c18
+
+    def __init__(self, prefix='', lock_wait=100, lock_wait_sleep=1):
```
dupainaulevain: Is it ok for whatever thread is using the context manager to block up to 100 seconds?

instance/models/utils.py
```diff
@@ -423,9 +432,48 @@ class ConsulAgent(object):
     managing prefixes, and reduces the call size to the main needed things with a
     possibility to expand it for more advanced queries.
     """
-    def __init__(self, prefix=''):
+
+    # We use the `consul lock` CLI to acquire locks on the server side.
+    # Those use a "magic flag" to help indicate conflicts with semaphores.
+    # We want to compete with those locks though, so we use this flag too.
+    # Comes from https://github.com/hashicorp/consul/blob/master/api/lock.go#L35-L38
+    LOCK_MAGIC_FLAG = 0x2ddccbc058a50c18
```
dupainaulevain: What sequence of event does this protect against?

dupainaulevain:
@UmanShahzad I apologize for questioning the foundations of this change. I know it can be frustrating. Unless I'm completely mistaken, which would be the best outcome ;-) Here it is.

I thought there both the cert-manager and ocim put at least one key for which they could race. However I found cert-manager/cert_manager/backends.py has {prefix}/{name}.pem and opencraft/instance/models/openedx_instance.py _generate_consul_metadata has no key with .pem, therefore I conclude there is no risk for them to race.

If I understand correctly, the lock is only used to ensure the data read from consul and written by ocim are consistent (i.e. the consul-template is not reading while ocim is in the middle of the _write_metadata_to_consul loop with only half the values updated. If that's the case, it wouldn't it be simpler to replace the loop with [a transaction](https://www.consul.io/api/txn.html)? The `clean_consul_metadata` function runs an atomic recursive delete for each instance and does not need a transaction: the only way consul-template could get inconsistent data is if it was deleting each key individually.

I would be  disappointed to be right because I look forward to writing tests for the context manager. Could you please tell me that I got it backwards and that the lock is necessary? :-P

dupainaulevain:
Assuming the lock & context manager are the way to go, I think there is no need for the CAS to be set. The only way a k/v is updated is via the `_write_metadata_to_consul` function which is now protected with a lock. If two Ocim workers try to race against each other, they will be guarded by the lock. Removing the CAS also means the underlying function (ConsulAgent.get()) can be removed and the Consul.get function used instead.

dupainaulevain:
I overlooked the fact that two Ocim workers may race against each other when updating the database. It is my (noob) understanding that views which need to be protected against such races have two choices:

* either they are simple and a database transaction is all they need, or
* the code path is paved with read/writes and the view itself must be guarded with [DATABASE-ATOMIC_REQUESTS](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-DATABASE-ATOMIC_REQUESTS)  or something similar so two workers never go into the same code path concurrently

Is `_write_metadata_to_consul` part of a code path protected in this way? If could there be a race by which the data in consul does not match the data in the database?

instance/models/utils.py
UmanShahzad: No sequence of events - it ensures we're playing the same game as the `consul lock` command.

instance/models/utils.py
UmanShahzad: Yeah I think the majority of cases where this would execute is in a background process, of which there are many. If it blocks for that long, it means there was seriously something wrong with Consul or similar and we need to look into it.

UmanShahzad:
> I thought both the cert-manager and ocim put() at least one key for which they could race. However I found cert-manager/cert_manager/backends.py has {prefix}/{name}.pem and opencraft/instance/models/openedx_instance.py _generate_consul_metadata has no key with .pem, therefore I conclude there is no risk for them to race.

There is no data race between the cert-manager and Ocim, no, but rather between these (or more obviously, Ocim) and consul-template. But consul-template can only use one lock in the scheme we used, so we use the same lock across all clients making updates to Ocim in ways that affect HAProxy.

> If I understand correctly, the lock is only used to ensure the data read from consul and written by ocim + cert-manager are consistent (i.e. the consul-template is not reading while ocim is in the middle of the _write_metadata_to_consul loop with only half the values updated). If that's the case, it wouldn't it be simpler to replace the loop with a transaction?

Yeah, but that would require a refactor and then probably a ton of changes to the mass unit tests (which I complained about in the ticket). And that's not the only place, there's also the management command. The main issue is that I found out that Consul _had_ transactions kind of late unfortunately (see above), but at that point I decided to stick with this because it seemed like the shortest path given the non-technical restrictions on us. I would certainly use transactions if having to do things over, or if I had more time right now, or rather if I were designing the integration from the beginning myself! (I didn't).

> The clean_consul_metadata function runs an atomic recursive delete for each instance and does not need a transaction: the only way consul-template could get inconsistent data is if it was deleting each key individually.

It is atomic indeed, and wouldn't cause inconsistencies, but would cause plenty of HAProxy reloads (each render causes a reload, and since each deletion would cause a render there will be multiple consecutive HAProxy reloads within a short period of time without the lock - I don't know if that's a problem in practice though, I'd have to read the HAProxy code and see the potential side effects of spam reloads, but I'm assuming they're no free lunch). This isn't actually too important because purging a bunch of instances is a rare operation, i.e. maybe once a month.

> Assuming the lock & context manager are the way to go, I think there is no need for the CAS to be set. The only way a k/v is updated is via the _write_metadata_to_consul function which is now protected with a lock. If two Ocim workers try to race against each other, they will be guarded by the lock.

Erased it.

> Removing the CAS also means the underlying function (ConsulAgent.get()) can be removed and the Consul.get function used instead.

The underlying function seems to be for some casting magic unrelated to check-and-set; I hesitate to change that right now.

> If could there be a race by which the data in consul does not match the data in the database?

I'd have to look at this separately, I'm not really sure.

UmanShahzad:
@dupainaulevain All in all I think you shouldn't bother with tests for that context manager, but I won't stop you :P There's a lot of refactoring to do of the Ocim codebase in general which we will get to soon with the Terraform epic and other followups, and that's where most of our brains and money should be spent.

instance/models/utils.py
UmanShahzad: To quote the docs from the link there:

```go
	// LockFlagValue is a magic flag we set to indicate a key
	// is being used for a lock. It is used to detect a potential
	// conflict with a semaphore.
```

But we don't want it to think we're holding a semaphore.

UmanShahzad: Yes I know, seems like hacks galore :P If we had budget we could do a lot better...

dupainaulevain: Oh, **dang** I get it now. If not set [it will fail](https://github.com/hashicorp/consul/blob/master/api/lock.go#L198). Not sure how you figured that bit but well done :+1:

dupainaulevain:
Test instructions:

- [x] Go to https://stage.console.opencraft.com/instance/3360/edx-appserver/1064/
- [x] SSH into haproxy-stage-1.net.opencraft.hosting and haproxy-stage-2.net.opencraft.hosting
- [x] On one server run `consul watch -type=keyprefix -prefix=lock/ consul kv get -recurse -detailed lock/`
- [x] On the other server run `consul watch -type=keyprefix -prefix=ocim/instances/362/ consul kv get -recurse ocim/instances/362`
- [x] Deactivate one server, activate another
- [x] After some time passes, check the spam from the watches - the lock is being worked on by both Ocim and consul-template across the HAProxy servers.
- [x] After it's settled down, check `consul kv export lock/` and notice no lock is taken anymore by anyone. **I verified this with a caveat: `watch -n1 consul kv export lock/` was not fast enough to ever display anything. But that's ok because the other watch on lock/ did**

PR #391, opened 2019-03-04
improve the consistency of the values stored in consul

dupainaulevain:
This is a followup to [the subtle bug fixed a few days ago](https://github.com/open-craft/opencraft/pull/386). It addresses border cases that are unlikely to be an issue right now but may create problems similarly difficult diagnose problems later on.

It would also allow functions that use **get** to differentiate key containing None from non existent keys. For instance:

```
modified   instance/models/openedx_instance.py
@@ -489,13 +489,13 @@ class OpenEdXInstance(
         version_number = agent.get('version') or 0
         for key, value in configurations.items():
             index, stored_value = agent.get(key, index=True)
-            cas = index if stored_value is not None else 0
-            agent.put(key, value, cas=cas)
+            if index == '0' or value != stored_value:
+                agent.put(key, value, cas=index)
 
-            if not version_updated and value != stored_value:
-                version_updated = True
-                version_number += 1
-                agent.put('version', version_number)
+                if not version_updated:
+                    version_updated = True
+                    version_number += 1
+                    agent.put('version', version_number)
 
         return version_number, version_updated
```

### DIFF
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -460,7 +460,7 @@ class ConsulAgent(object):
         :return: Either True or False. If False is returned, then the update has not taken place.
         """
         consul_key = self.prefix + key
-        value = json.dumps(value) if self._is_json_serializable(value) else str(value)
+        value = json.dumps(value)
 
         return self._client.kv.put(consul_key, value, **kwargs)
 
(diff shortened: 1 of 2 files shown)


UmanShahzad:
@dupainaulevain I guess we should make a ticket for this?

dupainaulevain:
@UmanShahzad I think this can be made redundant if a single value is used instead of many and there already is a ticket for that, linking to this PR. Would that be enough?

UmanShahzad:
@dupainaulevain sounds good :+1:

dupainaulevain:
closing to avoid cluttering the pull request list

PR #399, opened 2019-03-07
BB-999 Change expected error code during bucket creation

viadanna:
Description
=========

This is intended to fix the `BucketAlreadyOwnedByYou` error that's blocking deployments using AWS S3 zones other than US-EAST-1.

Testing
--------

1. Create an instance and set the AWS zone to something other than us-east-1, e.g. us-west-1.
2. Provision an app server or run the `reprovision_buckets` management command and make sure S3 is provisioned correctly.
3. Provision another app server or run the management command again, making sure it does not fail while reprovisioning the bucket. Instead it should mention `bucket <bucket name>  already exists.


instance/models/mixins/storage.py
```diff
@@ -313,8 +313,9 @@ def _create_bucket(self, max_tries=4, retry_delay=15, location=None):
                 )
                 break
             except ClientError as e:
-                if e.response.get('Error', {}).get('Code') == 'EntityAlreadyExists':
+                if e.response.get('Error', {}).get('Code') == 'BucketAlreadyOwnedByYou':
```
lgp171188: @viadanna, will there be any scenarios where handling the `EntityAlreadyExists` exception will be needed?

viadanna: Not in this context, as after the refactor it just tries to create a bucket. In `us-east-1` it's simply ignored and on other zones it returns the value tested when the bucket already exists.

PR #402, opened 2019-03-08
[BB-200] add validators for color fields in the registration form

dupainaulevain:
### Implementation notes

The regex validation only happens when the form is submitted for the same reason [subdomain needs a special patch|https://github.com/open-craft/opencraft/blob/914b759abd4e99c48f3514177b34e0436898dc47/registration/forms.py#L180]. This is not a problem however since this validation is only a safeguard against the **very** rare case where someone (or a program) fills the registration form without the help of a color picker.

### Instructions to test manually:

* make run.dev
* firefox http://localhost:5000/registration
* remove **type="color"** from one color field (right click + inspect)
* enter **#1234** in the field revealed
* enter all required values in the form
* click the submit button
* check there is an error message under the field looking like this:

z

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -36,6 +36,16 @@ from instance.models.utils import ValidateModelMixin
 
 # Models ######################################################################
 
+def validate_color(color):
+    """
+    Check the color is either #123 or #123456
+    """
+    validators.RegexValidator(
+        r'^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$',
+        '{} is not a valid color, it must be either #123 or #123456'.format(color),
+    )(color)
+
+
 def validate_available_subdomain(subdomain):
     """
     Check that the given subdomain is not blacklisted.
@@ -132,6 +142,7 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
                   'It is used as filler for buttons.',
         # #126f9a == $m-blue-d3 in variables.scss. It's rgb(18,111,154)
         default='#126f9a',
+        validators=[validate_color],
     )
     link_color = models.CharField(
         max_length=7,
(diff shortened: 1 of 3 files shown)


dupainaulevain:
The browser tests fail although the unit tests succeed, with: **AssertionError: defaultdict(<class 'list'>, {}) != {'main_color': ['#1234 is not a valid col[288 chars]56']}** as if the string was interpreted somewhere.

dupainaulevain:
It turned out the test utils failed to set an invalid value in a type="color" field in the browser. I chose to [convert the field to type="text"](https://github.com/open-craft/opencraft/pull/402/commits/fb8c404d72430dfd23886ae9db06d759948cbb6c) but I'm not sure this is the right thing to do.

There is no need for browser side testing of this server side validation because it cannot happen in the browser. Not with firefox anyways and there are no tests for any other browser. The border case it addresses is if someone uses **curl** or scripting to submit the form with incorrect values. So... another way to fix the problem would be to exclude fields of type=color from browser testing entirely.

What do you think is better?

viadanna:
@dupainaulevain Given the browser forces the use of valid color codes on `type="color"` and there's validation in Python code, I'm ok with excluding this from browser testing.

dupainaulevain:
@viadanna thanks for the feedback, I'll do that instead of implementing this bizarre and kind of self defeating conversion of a type=color into a type=text.

dupainaulevain:
@viadanna BTW, any idea why the tests do not run on this PR?

viadanna:
@dupainaulevain Interesting, missed that.

I see you are present in the org. Do you have an account in CircleCI? You can login using your Github account.

dupainaulevain:
@viadanna I logged in CircleCI, closing & re-opening to see if that pleases the pipeline. Thanks for the advice :100:

viadanna:
@dupainaulevain Hum, no go. I'll have to check if its restricted to certain groups and get back to you.

dupainaulevain:
Moved the test so it does not run on the browser and successfully ran `make test.unit test.browser` locally

dupainaulevain:
My affiliation with the OpenCraft organization was [made public](https://help.github.com/en/articles/publicizing-or-hiding-organization-membership). I closed and re-opened in the hope CircleCI will now agree to run.

PR #403, opened 2019-03-10
[BB-197] misc pylint fixes

dupainaulevain:


cleanup_utils/mysql_cleanup.py
```diff
@@ -52,7 +52,7 @@ def _get_cursor(url):
         )
     except MySQLError as exc:
         logger.exception('Cannot get MySQL cursor: %s', exc)
-        raise
+        return None
```
viadanna: Nice 🙂

pylintrc
```diff
@@ -89,10 +89,15 @@ confidence=
 disable=
     fixme,
     locally-disabled,
+    no-self-use,
```
viadanna: Do you think it would be possible to apply `@staticmethod` to those methods instead of ignoring every occurrence?

dupainaulevain: Yes it would. I thought best to just ignore them all because:
* The dozens of exceptions in the codebase show the developers do not generally value this
* I don't see a value in this stylistic convention, i.e. what exactly is the problem with a self that's not used?
* The edx-lint has [no-self-use](https://github.com/edx/edx-lint/blob/master/edx_lint/files/pylintrc#L303) disabled which probably has an influence in the fact that developers do not value this stylistic convention in the Open edX ecosystem

That bein said, if you think it must be enforced I will remove this exception and fix all issue.

viadanna: No need to, I'm convinced 😛

viadanna:
@dupainaulevain Can you push this branch to `open-craft/opencraft` repo and create another PR to get the CircleCI builds as well?

dupainaulevain:
I think there is a switch in CircleCI to allow runs from forked repository. It would help with external contributors who do not have write access to the repository. I don't have permissions to configure CircleCI though.

antoviaque:
@dupainaulevain Yes, but if you allow this, you give unrestricted access to the build environment (since any code can be added on the fork), including to secrets stored there - such as OpenStack credentials. I would also prefer to allow it for anyone, but I'm not sure how to do that properly.

dupainaulevain:
>  including to secrets stored there - such as OpenStack credentials. 

It's a risk indeed. FWIW Ceph took that risk and nothing happened. I take that risk every day and nothing happened either. The worst that can happen with OpenStack credentials, assuming you have a dedicated project where nothing else is running, is that some malicious person max out your quotas during a month and you loose X euros when you receive the bill at the end of the month. Assuming you don't notice before. With my quotas it is a risk in the order of ~100 euros, because I keep the quotas of this tenant low.

antoviaque:
@dupainaulevain Good point - if after reviewing the list of secrets that can be accessed from the CI we deem the risk low (making sure we don't use the OpenStack account for other work/VMs, and keeping quotas low), we could do that too. It might indeed make sense to wait until we have a few more contributors though yes ;p In the meantime, as long as we have a way to whitelist individual contributors that can probably be good enough.

dupainaulevain:
> as long as we have a way to whitelist individual contributors that can probably be good enough.

Do you mean that it would be possible to run the CI from forks but only for whitelisted contributors? That would work. I thought it was an all or nothing thing. @lgp171188 Do you know how to do that by any chance? You mentionned in the chat that activating the CI for forks is a few clicks away for someone who has admin privileges but I don't know if whitelisting forks is possible...

lgp171188:
> @lgp171188 Do you know how to do that by any chance? You mentioned in the chat that activating the CI for forks is a few clicks away for someone who has admin privileges but I don't know if whitelisting forks is possible

@dupainaulevain, that wasn't me, but I can check this. :stuck_out_tongue:

dupainaulevain:
@lgp171188 gahh, sorry for the confusion :blush:

antoviaque:
@dupainaulevain I meant more whitelisting a specific contributor, by trying to add the person to the OpenCraft org, like I tried to do with @singuliere . Did that end up working actually?

dupainaulevain:
> Did that end up working actually?

It is my understanding that there are two permissions:

* people who are allowed to run the CI (singuliere is in the list)
* running the CI from forks (that's not allowed)

Maybe it's possible to also run CircleCI from forks and restrict that ability to designated people.

antoviaque:
@dupainaulevain Oh, I see - that's how we differentiated between the two. Then instead can we allow specific contributors to push their branch to the fork? If we keep master protected (ie requiring a PR to make changes to it), then that's probably enough protection for whitelisted contributors, since we would be notified for any change made there?

dupainaulevain:
I don't know if it's possible to allow someone to push a branch but not push to master.

antoviaque:
@dupainaulevain It is possible yes - actually I have just checked, and master is already protected, so all changes going there have to go through PRs. So I have given write access to the contributors team, to which @singuliere belongs, so he should now be able to push his branches to the official repo.

> How do you mean? Which fork? /me confused :-)

The main/official one, ie https://github.com/open-craft/opencraft -- sorry, I call all repos "forks", even the main one ;p

PR #405, opened 2019-03-12
Don't grant SSH privileges to inactive users [MNG-918]

bradenmacdonald:
Our code to grant SSH access was mistakenly including user accounts that were deactivated. This is a small fix to exclude them. I have already updated accounts in the DB so that even without this fix, they aren't included.

Testing instructions: deactivate a user via the django admin, then use the django shell to query an appserver's `admin_users` list. Or just run the before + after versions of this ORM statement (`users = UserProfile.objects. ... .values_list('github_username', flat=True)`) in the django shell. Confirm that deactivated users aren't in the list.

### DIFF
--- a/instance/models/openedx_appserver.py
+++ b/instance/models/openedx_appserver.py
@@ -371,6 +371,7 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
 
         users = UserProfile.objects \
             .filter(admin_user_query) \
+            .filter(user__is_active=True) \
             .exclude(Q(github_username__isnull=True) | Q(github_username__exact='')) \
             .values_list('github_username', flat=True)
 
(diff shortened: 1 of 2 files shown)


lgp171188:
> This is a small fix to exclude them. I have already updated accounts in the DB so that even without this fix, they aren't included.

@bradenmacdonald, without the fix, the `admin_users` list still contains the inactive users.

bradenmacdonald:
> without the fix, the admin_users list still contains the inactive users.

@lgp171188 Do you mean on stage? It shouldn't on prod.

lgp171188:
> @lgp171188 Do you mean on stage? It shouldn't on prod.

@bradenmacdonald :+1: I will check this on prod once it is deployed

bradenmacdonald:
@lgp171188 I've added a test via 48714855224ab937966b76a9d48138db383349ed, so merging this now.  Thanks for the review!

PR #408, opened 2019-03-18
[BB-1069] Add Privacy Policy and Terms of Use

jamestait:
* Add a checkbox to the registration form to indicate acceptance of the Privacy Policy, without which the application will not be allowed.
* Add templates for the Privacy Policy and Terms of Use, and link these from the descriptions of the relevant registration checkboxes.
* Add a field to the `Instance` model indicating the timestamp at which the Privacy Policy was accepted.
* Populate this field with the timestamp of the `BetaTestApplication` creation when creating an `Instance` in response to email validation, which is effectively the time at which we can be certain the Privacy Policy was accepted.

## Testing Instructions

* Merge this branch to your local devstack and apply migrations.
* Navigate to the registration form and click the links for the Privacy Policy and Terms of Use.
* Attempt to submit without accepting the Privacy Policy.
  * The submission should be denied, and an error message displayed.
* Accept the privacy policy and resubmit.
  * The submission should proceed, and you will see that it is under review.
* In the DJango admin, locate the application and set its status to `Approved`.
* In the Django console output should be an email to validate your email address; logout of Django admin and then navigate to the link in the email.
* In the Django admin, observe the newly-created OpenedX Instance.
* Navigate to the instance and observe the `Accepted Privacy Policy` field.


registration/templates/registration/terms.html
```diff
@@ -0,0 +1,514 @@
...
+    <h1 class="page-title__text">Terms of Use</h1>
+</div>
+
+<div class="row page-content clearfix">
+    <div class="page-content__inner small-12 medium-10 medium-offset-1 float-left background-white">
+        <div class="page-content__inner__wrapper">
+            <section>
+                <h3>General</h3>
+                <p>
+                    Headquarters for the National Digital Israel Initiative in
```
lgp171188: @jamestait, is this the correct ToS? If not, can we add the correct text?

registration/templates/registration/terms.html
```diff
@@ -0,0 +1,514 @@
+{% extends "base.html" %}
+{% load staticfiles %}
+
+{% block title %}Terms of Use{% endblock title %}
```
lgp171188: @jamestait, I think the name we are using is `Terms of Service`.

instance/models/instance.py
```diff
@@ -157,6 +157,11 @@ class Instance(ValidateModelMixin, models.Model):
         blank=True,
         help_text='Custom tags associated with the instance.',
     )
+    accepted_privacy_policy = models.DateTimeField(
```
lgp171188: @jamestait, just thinking out loud, while we could block the spawning of new appservers unless the privacy policy is accepted, I am not sure if that is the best way forward because the customers almost never spawn new appservers after the initial trials to get their theming right.

lgp171188: @jamestait, this field should be read-only in the admin interface.

registration/templates/registration/terms.html
jamestait: @lgp171188, I believe so. The document wasn't linked from the ticket, so I went searching for it. @viadanna confirmed I'd got the correct documents, and was looking for an email confirmation to make doubly sure.

instance/models/instance.py
jamestait: The approach that appears to have been decided upon is to ask existing customers to agree to the Privacy Policy via email.

jamestait: Given we're intending to ask customers to agree to the Privacy Policy by email, do we want to keep this editable, or perhaps editable if empty, so we can add this date manually when we get customers' acceptance?

registration/templates/registration/terms.html
lgp171188: @jamestait, this mentions some organization in Israel and has nothing to do with it. @viadanna

instance/models/instance.py
lgp171188: @jamestait, we could do that via the CLI and keep the admin interface read-only. But that would mean Gabriel won't be able to update this if he wants to.

lgp171188: @jamestait, this field should be in the registration model because that imho is the correct place. What do you think?

jamestait: Yes, that does seem like a more natural place for it. I'll move it.

jamestait: Done in the latest commit. :)

registration/templates/registration/terms.html
viadanna: @jamestait There's been some confusion, I guess this might be a sample document. The correct ones are linked in the ticket.

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -22,6 +22,8 @@
 
 # Imports #####################################################################
 import ddt
+from freezegun import freeze_time
```
lgp171188: @jamestait, nit: we group all the `import *` together followed by the `from * import *`.

instance/models/instance.py
jamestait: I've just pushed that change. :slightly_smiling_face:

instance/tests/models/test_openedx_theme_mixins.py
jamestait: Ah, yes, sorry about this - I'll fix that.

registration/templates/registration/terms.html
jamestait: OK, I'll get that updated. Thanks for the corrected links. :slightly_smiling_face:

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -71,7 +74,10 @@ def test_colors_and_images_applied(self):
         OpenEdXInstanceFactory(name='Integration - test_colors_applied', deploy_simpletheme=True)
         instance = OpenEdXInstance.objects.get()
         user = get_user_model().objects.create_user('betatestuser', 'betatest@example.com')
-        self.make_test_application(instance, user)
+        with freeze_time('2019-01-02 09:30:00') as frozen_time:
```
lgp171188: @jamestait, wouldn't it be easier to create a class variable named say `FROZEN_TIME` with the time to use for the tests and also have the string-formatted version of it in the class so that this code doesn't have to be repeated in multiple places?

instance/models/instance.py
lgp171188: > do we want to keep this editable, or perhaps editable if empty, so we can add this date manually when we get customers' acceptance

This looks like the best way to do this.

registration/templates/registration/terms.html
jamestait: Pushed the updated ToS.

instance/tests/models/test_openedx_theme_mixins.py
jamestait: To summarise the conversation we had in-person about this, we need `accepted_privacy_policy` to be nullable, as we have existing instances where the user hasn't agreed to the privacy policy. In instances where the field is null, it needs to be editable in the admin, so it can be set on behalf of users. In instances where the field is not null, it needs to be read-only in the admin.

instance/tests/models/test_openedx_theme_mixins.py
jamestait: This import wasn't necessary (see conversation below), so it will be removed.

instance/tests/models/test_openedx_theme_mixins.py
jamestait: > Edit: It looks like this code is out of place in this test and can we find out ways to get rid of it?

Right - it doesn't matter to us here what the timestamp is for the privacy policy acceptance, so we can just update `make_test_application` to provide the current time.

PR #409, opened 2019-03-19
[BB-200] add validators for color fields in the registration form

dupainaulevain:
Followup of https://github.com/open-craft/opencraft/pull/402

### Implementation notes

The regex validation only happens when the form is submitted for the same reason [subdomain needs a special patch|https://github.com/open-craft/opencraft/blob/914b759abd4e99c48f3514177b34e0436898dc47/registration/forms.py#L180]. This is not a problem however since this validation is only a safeguard against the **very** rare case where someone (or a program) fills the registration form without the help of a color picker.

### Instructions to test manually:

* make run.dev
* firefox http://localhost:5000/registration
* remove **type="color"** from one color field (right click + inspect)
* enter **#1234** in the field revealed
* enter all required values in the form
* click the submit button
* check there is an error message under the field looking like this:

z


dupainaulevain:
@viadanna I like files with readable names too :-) Is [this ok](https://github.com/open-craft/opencraft/compare/4972d5649568c99f7b52317bb76314c617c38c0a..5460033be9823c51fe251a58ca6ef4ede309aea5)?

PR #412, opened 2019-03-21
Upgraded Swampdragon, Django and django-angular

DevanR:
[Work in Progress]

This PR contains work in an attempt to upgrade Django to 1.11, see SE-710, SE-712

### DIFF
--- a/cleanup_utils/requirements.txt
+++ b/cleanup_utils/requirements.txt
@@ -1,3 +1,4 @@
 boto3==1.9.61
 mysqlclient==1.3.9
 python-novaclient==6.0.0
+PyYAML==3.10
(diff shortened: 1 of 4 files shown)


clemente:
@DevanR how this does relate to https://github.com/open-craft/opencraft/pull/407, and should we close one of those?

DevanR:
> @DevanR how this does relate to #407, and should we close one of those?

It's not really related. The way I see it, there are two hurdles moving forward:

1. Updating PyYAML and Requests is blocked by the dependency on the current version of Jasmine.
2. Replacing Swampdragon is blocked by Django==1.11 which is blocked by Angular.

I think these can be two separate tickets.

DevanR:
This has been closed in favour of https://github.com/open-craft/opencraft/pull/424/ (SE-948).

PR #417, opened 2019-03-25
BB-839 update integration tests to use Hawthorn

xirdneh:
This PR updates the integration tests to use hawthorn edx and configuration branches.
While doing the update there was an issue with migrations taking too long. The reason was because the centralized MySQL server used to run integration tests had too many databses and users created. There's a cleanup script that runs on every CirlceCI pipeline that deletes databases older than 8 hours.
The cleanup script was updated. A new method was added to delete every database and user used in integration tests in case we eventually need to delete everything.
Also, the cleanup script (specifically the MySQL part) is using `information_schema.tables` (and there's no other way to do it). The issue is that if a database is created and no tables are added to the database then it will not appear in `information_schema.tables`. This PR also adds a dummy table named `opencraft` to every database created. This will force every database to appear in `information_schema.tables` table. In case creating an app server fails and no tables are created this will allow us to still keep track of old databases. If the instance created is not an integration instance this will not introduce any issues.
This PR also uses the same random prefix to create databases and users. This way if something goes wrong on a CircleCI run we can later check for old databases and delete the databases as well as the users related to those databases because the random prefix used is the same.

### DIFF
--- a/cleanup_utils/integration_cleanup.py
+++ b/cleanup_utils/integration_cleanup.py
@@ -106,6 +106,7 @@ def run_integration_cleanup(dry_run=False):
         age_limit=DEFAULT_AGE_LIMIT,
         url=os.environ['DEFAULT_INSTANCE_MYSQL_URL'],
         domain=os.environ['DEFAULT_INSTANCE_BASE_DOMAIN'],
+        drop_dbs_and_users=os.environ.get('DROP_INTEGRATION_DBS_AND_USERS', 'False').lower() == 'true',
         dry_run=dry_run
     )
     mysql_cleanup.run_cleanup()
--- a/instance/tests/integration/base.py
+++ b/instance/tests/integration/base.py
@@ -59,7 +59,7 @@ class IntegrationTestCase(TestCase):
         for load_balancer in LoadBalancingServer.objects.iterator():
             load_balancer.delete()
         for instance in OpenEdXInstance.objects.iterator():
-            instance.delete()
+            instance.delete(ignore_errors=True)
 
         super().tearDown()
 
--- a/instance/tests/integration/factories/instance.py
+++ b/instance/tests/integration/factories/instance.py
@@ -46,8 +46,9 @@ class OpenEdXInstanceFactory(DjangoModelFactory):
         # random value for 'internal_lms_domain' if neither 'sub_domain' nor 'internal_lms_domain' are provided.
         if 'sub_domain' not in kwargs and 'internal_lms_domain' not in kwargs:
             kwargs = kwargs.copy()
-            random_id = str(uuid.uuid4())[:8]
+            random_id = str(uuid.uuid4())[:6]
             sub_domain = '{}.integration'.format(random_id)
+            kwargs['random_prefix'] = random_id
             kwargs['internal_lms_domain'] = generate_internal_lms_domain(sub_domain)
         return super(OpenEdXInstanceFactory, cls).create(*args, **kwargs)
 
@@ -58,8 +59,8 @@ class OpenEdXInstanceFactory(DjangoModelFactory):
     # or a release candidate tag will work.  We point both the edx-platform and the configuration
     # versions to the branch "integration" in our own forks.  These branches are based on the
     # corresponding openedx_release versions from upstream, but can contain custom modifications.
-    openedx_release = 'open-release/ginkgo.1'
+    openedx_release = 'open-release/hawthorn.1'
     configuration_source_repo_url = 'https://github.com/open-craft/configuration.git'
-    configuration_version = 'integration-ginkgo'
+    configuration_version = 'integration-hawthorn'
     edx_platform_repository_url = 'https://github.com/open-craft/edx-platform.git'
-    edx_platform_commit = 'opencraft-release/ginkgo.1'
+    edx_platform_commit = 'opencraft-release/hawthorn.1'
--- a/instance/tests/integration/integration_instance.py
+++ b/instance/tests/integration/integration_instance.py
@@ -235,7 +235,7 @@ class InstanceIntegrationTestCase(IntegrationTestCase):
 
         # The logo is found in a line like this in the HTML:
         #       <img src="/static/simple-theme/images/logo.82fb8d18479f.png" alt="danieltest1b Home Page"/>
-        logo_extractor = re.search(r'<img src="(/static/simple-theme/images/logo.[a-z0-9]+\.png)" ',
+        logo_extractor = re.search(r'<img\s+class="logo"\s+src="(/static/simple-theme/images/logo.[a-z0-9]+\.png)"',
                                    server_html)
         self.assertTrue(logo_extractor)
         logo_url = logo_extractor.group(1)
--- a/circle.yml
+++ b/circle.yml
@@ -84,7 +84,7 @@ jobs:
           command: |
             . venv/bin/activate
             bin/run-circleci-tests
-          no_output_timeout: 47m
+          no_output_timeout: 6h
           environment:
             ANSIBLE_HOST_KEY_CHECKING: False
   cleanup:
--- a/cleanup_utils/aws_cleanup.py
+++ b/cleanup_utils/aws_cleanup.py
@@ -108,7 +108,10 @@ class AwsCleanupInstance:
         """
         if self.dry_run:
             return
-        self.iam_client.delete_user(UserName=username)
+        try:
+            self.iam_client.delete_user(UserName=username)
+        except self.iam_client.exceptions.DeleteConflictException as exc:
+            logger.error("Could not delete user: %s. %s", username, exc)
 
     def get_iam_users(self):
         """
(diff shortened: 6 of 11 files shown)


Agrendalath:
@xirdneh, just a nit: you have a typo in the PR title - 'Hawthorn' (not too important, but the title is pulled while squashing commits) ;)

cleanup_utils/integration_cleanup.py
```diff
@@ -109,6 +109,8 @@ def run_integration_cleanup(dry_run=False):
         dry_run=dry_run
     )
     mysql_cleanup.run_cleanup()
+    if os.environ.get('DROP_INTEGRATION_DBS_AND_USERS', 'False').lower() == 'true':
```
giovannicimolin: Should we set this variable on CircleCI Environment?

instance/tests/integration/base.py
```diff
@@ -59,7 +59,7 @@ def tearDown(self):
         for load_balancer in LoadBalancingServer.objects.iterator():
             load_balancer.delete()
         for instance in OpenEdXInstance.objects.iterator():
-            instance.delete()
+            instance.delete(ignore_errors=True)
```
giovannicimolin: Why are you passing ignore errors to the function?

instance/tests/integration/integration_instance.py
```diff
@@ -235,8 +235,9 @@ def assert_theme_provisioned(self, instance, appserver, application):
 
         # The logo is found in a line like this in the HTML:
         #       <img src="/static/simple-theme/images/logo.82fb8d18479f.png" alt="danieltest1b Home Page"/>
-        logo_extractor = re.search(r'<img src="(/static/simple-theme/images/logo.[a-z0-9]+\.png)" ',
-                                   server_html)
+        logo_extractor = re.search(
```
giovannicimolin: Revert this formatting change.

instance/tests/integration/factories/instance.py
```diff
@@ -46,8 +46,9 @@ def create(cls, *args, **kwargs):
         # random value for 'internal_lms_domain' if neither 'sub_domain' nor 'internal_lms_domain' are provided.
         if 'sub_domain' not in kwargs and 'internal_lms_domain' not in kwargs:
             kwargs = kwargs.copy()
-            random_id = str(uuid.uuid4())[:8]
```
giovannicimolin: Why these changes were required?

cleanup_utils/integration_cleanup.py
xirdneh: The idea is to use it every once in a while when we want to delete every database and user used in the integration tests.
For the very day runs we probably don't want to delete everything or tests running in parallel will not work properly.
If things are getting a bit slow we could just have a user with admin access to set this variable in CirlceCI and then run the cleanup step.

instance/tests/integration/base.py
xirdneh: The `ignore_errors` attribute was introduced recently.
I thought it would be a good idea since we're deleting an instance and if there's an error with deleting we won't go all the way through the deletion. For instance, if there's an error while deleting a rabbitMQ user the entire deletion workflow will stop there and throw the error.
This will log those errors and continue with deletion/deprovisioning the rest of the stuff.

instance/tests/integration/factories/instance.py
xirdneh: This is referring to the last point in the PR description.
The thing is that we are adding a random string prefix to each database and username to avoid name clashing. Currently Ocim uses an [8 char] (https://github.com/open-craft/opencraft/blob/master/instance/tests/integration/factories/instance.py#L49) random [prefix](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/domain_names.py#L105) for [databases](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/openedx_database.py#L170) and a [6 char](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/database.py#L194) random [prefix](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/openedx_database.py#L204) for users and both of these are different. There is also a cleanup script that will delete every database that's 8 hours or older in case the Ocim integration test failed to delete the databases created. The issue is that the cleanup script can't delete the users bound to a specific database because there's no way to know it (well, I guess we could check permissions but that can get a bit tricky). With this what we do is that we use the same 6 char prefix to create the database and the [users](https://github.com/open-craft/opencraft/blob/josuebc/BB-839/instance/models/openedx_instance.py#L124). By sharing the same prefix we can later get all the old databases and use the same prefix to [query](https://github.com/open-craft/opencraft/blob/josuebc/BB-839/cleanup_utils/mysql_cleanup.py#L270) the username that are bound to that database. Basically creating a namespace.
The other update that I added was to create an empty table called `opencraft` when [creating a Database](https://github.com/open-craft/opencraft/blob/josuebc/BB-839/instance/models/mixins/database.py#L95). That's because we are using `information_schema.tables` to get when databases were last updated but if the integration test didn't create any tables then the database will not show up there. By creating an empty table every time we create a database we make sure every database can be queried through `information_schema.table`

giovannicimolin: Perfect! Nice way of making the cleanup easier. :grin:

instance/tests/integration/base.py
giovannicimolin: Nice :+1:

cleanup_utils/integration_cleanup.py
giovannicimolin: Seems ok to leave this here.

instance/tests/integration/integration_instance.py
```diff
@@ -235,7 +235,7 @@ def assert_theme_provisioned(self, instance, appserver, application):
 
         # The logo is found in a line like this in the HTML:
         #       <img src="/static/simple-theme/images/logo.82fb8d18479f.png" alt="danieltest1b Home Page"/>
-        logo_extractor = re.search(r'<img src="(/static/simple-theme/images/logo.[a-z0-9]+\.png)" ',
+        logo_extractor = re.search(r'<img\s+class="logo"\s+src="(/static/simple-theme/images/logo.[a-z0-9]+\.png)"',
```
giovannicimolin: Nit: I didn't noticed that you changed the regex here, so if you want, you can change the formatting back.

cleanup_utils/integration_cleanup.py
```diff
@@ -109,6 +109,8 @@ def run_integration_cleanup(dry_run=False):
...
+        mysql_cleanup.drop_integration_dbs_and_users()
```
giovannicimolin: Uh-oh, somehow I missed this, but instead of using an IF here, could you pass a variable to mysql_cleanup to conditionally clean the users?
```
mysql_cleanup = MySqlCleanupInstance(
    age_limit=DEFAULT_AGE_LIMIT,
    url=os.environ['DEFAULT_INSTANCE_MYSQL_URL'],	        
    domain=os.environ['DEFAULT_INSTANCE_BASE_DOMAIN'],	
    drop_users=os.environ.get('DROP_INTEGRATION_DBS_AND_USERS', False),
    dry_run=dry_run
)
```

xirdneh: Ah, good idea. Will do. Thanks.

instance/tests/integration/integration_instance.py
xirdneh: That's ok, this is the same format used in the rest of the module.

instance/models/mixins/database.py
```diff
@@ -92,6 +92,7 @@ def _create_database(cursor, database):
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
         cursor.execute('CREATE DATABASE IF NOT EXISTS `{db}` DEFAULT CHARACTER SET utf8'.format(db=database))
+        cursor.execute('USE `{db}`; CREATE TABLE IF NOT EXISTS `opencraft` (col INT)'.format(db=database))
```
lgp171188: @xirdneh, @giovannicimolin, should we be adding code that is only needed for the integration tests here?

xirdneh: @john2x @giovannicimolin I believe there's only two changes outside of tests. The first one adds the ability to specify the prefix used in MySQL usernames. It is only used in tests and if that parameter is not specified then creating an instance will work in the same way 
The second one is about adding an empty table, which I think it'll help us outside of integration tests. Because Ocim creates empty databases in the beginning, so if anything goes wrong it's difficult to craft a query which will return this unused databases.
We could also remove these changes but they won't break any current behavior. This was tested by spinning up a new app server in stage, too.

lgp171188: @xirdneh, can't we create this database in the integration test itself at a later stage?

xirdneh: @lgp171188 That would defeat the purpose of creating that database because something else could go wrong in between and we'd still have an empty database which are harder to delete.
But, we could probably add it to the integration playbook although that's outside of the scope of this ticket.
Would you like me to delete this and create a ticket to add it to the integration playbook or find another solution?

lgp171188: >That would defeat the purpose of creating that database because something else could go wrong in between and we'd still have an empty database which are harder to delete.
But, we could probably add it to the integration playbook although that's outside of the scope of this ticket.
Would you like me to delete this and create a ticket to add it to the integration playbook or find another solution?

@xirdneh, it would be better to do it in the code/playbook related to the integration tests. You can create a new ticket for doing that and leave this change out of this PR.

PR #418, opened 2019-03-26
[BB-1105] circleci: switch to ubuntu:xenial

dupainaulevain:
Because CircleCI jessie based images are no longer working and Xenial
is what production uses, and it makes sense to test on the same
distribution.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,14 +2,14 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:3.5.4-browsers
+      - image: ubuntu:xenial
         environment:
           DEBUG: 'true'
           DEFAULT_FORK: 'open-craft/edx-platform'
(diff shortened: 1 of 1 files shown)


antoviaque:
@dupainaulevain Was this the PR you mentioned that wasn't necessary because of the future move to terraform? Wouldn't this still be a good thing to merge in the meantime?

dupainaulevain:
I don't think it's worth the effort, but feel free to re-open if you have a different opinion.

antoviaque:
@dupainaulevain How much more work would be needed to complete it? Not sure if this is WIP or close to review status - it's always good to see work end up actually merged :)

dupainaulevain:
The associated issue BB-1105 was fixed differently.

antoviaque:
@dupainaulevain That doesn't really answer my question - but ok, if you really think it's not worth reusing this work, then let's let it go. It did sound like something useful though.

PR #421, opened 2019-03-27
[BB-986] Provide users with sample privacy policy

jamestait:
Adds a field to the registration form to allow users to provide a URL to a Privacy Policy when signing up. The Privacy Policy URL is optional - a default can be specified with the `DEFAULT_PRIVACY_POLICY_URL` Django setting or environment variable. Existing instances without a Privacy Policy already set may be edited without needing to provide a Privacy Policy URL. When the Privacy Policy URL is provided, the `EDXAPP_LMS_ENV_EXTRA` has `MKTG_URL_OVERRIDES` added and populated with the provided value in the `PRIVACY` element.

## Testing Instructions

* Merge this branch and apply migrations.
* Start the dev server, specifying a valid URL in the `DEFAULT_PRIVACY_POLICY_URL` environment variable: `DEFAULT_PRIVACY_POLICY_URL=http://localhost/privacy make run.dev`
* Observe the `Privacy Policy URL` field in the registration form.
* Attempt to submit the application with an invalid URL in the field.
  - The registration attempt should be rejected.
* Provide a valid URL for the field and submit the application.
  - The registration attempt should succeed.
* Provision an appserver from the newly-created instance.
  - Observe in the appserver's `Combined ansible vars` that:
    + `EDXAPP_LMS_ENV_EXTRA` contains `MKTG_URL_OVERRIDES`, which has your provided URL as the `PRIVACY` element.


instance/migrations/0115_auto_20190322_1036.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
giovannicimolin: Give a meaningful name to the migration (ref. https://docs.djangoproject.com/en/2.1/topics/migrations/#workflow)

instance/models/mixins/openedx_config.py
```diff
@@ -390,6 +391,21 @@ def _get_configuration_variables(self):
...
 
+        if self.privacy_policy_url:
+            # Custom Privacy Policy
+            url_components = urlparse(self.privacy_policy_url)
+            root_url = '{}://{}'.format(
+                url_components.scheme, url_components.netloc)
+            template.update({
+                "EDXAPP_MKTG_URLS": {
+                    "ROOT": root_url,
+                    "PRIVACY": self.privacy_policy_url,
```
giovannicimolin: Add a field for TOS here too.

instance/models/openedx_appserver.py
```diff
@@ -66,6 +66,11 @@ class Meta:
         'The default contact email for this instance; also used as the from address for emails '
         'sent by the server.'
     ))
+    privacy_policy_url = models.URLField(
```
giovannicimolin: Add a fields for TOS too.

registration/forms.py
```diff
@@ -238,6 +246,8 @@ def __init__(self, *args, **kwargs):
                 for name, field in self.fields.items():
                     if name not in self.can_be_modified:
                         field.widget.attrs['readonly'] = True
+            else:
+                self.initial['privacy_policy_url'] = settings.DEFAULT_PRIVACY_POLICY_URL
```
giovannicimolin: Why are you setting the default value here?

registration/forms.py
```diff
@@ -320,6 +330,18 @@ def clean_password_confirmation(self):
             )
         return password
 
+    def clean_privacy_policy_url(self):
```
giovannicimolin: The user should be able to submit the form with blank privacy and TOS fields, so that they take the default values.

registration/migrations/0010_betatestapplication_privacy_policy_url.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
giovannicimolin: Give the migration a meaningful name.

registration/tests/test_views.py
```diff
@@ -556,6 +591,44 @@ def test_modifying_design_fields_sends_email(self):
             self._register(modified)
             self.assertEqual(len(mail.outbox), original_emails + 1)
 
+    def test_preserve_empty_privacy_url(self):
```
giovannicimolin: We can make the users leave empty values for all instances, not only the existing ones. This will reduce overall complexity.

registration/models.py
```diff
@@ -134,6 +134,12 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
                    'your expectations?'),
         blank=True, default=''
     )
+    privacy_policy_url = models.URLField(
+        verbose_name='URL to Privacy Policy',
+        help_text=('URL to the privacy policy.'),
+        null=True, # to ease migrations
```
giovannicimolin: Remove this `null=True` and use `default="", blank=True`.

instance/models/openedx_appserver.py
```diff
@@ -66,6 +66,11 @@ class Meta:
...
+        verbose_name='URL to Privacy Policy',
+        help_text=('URL to the privacy policy.'),
+        blank=True,
```
giovannicimolin: Use `default="", blank=True`.

instance/models/mixins/openedx_config.py
```diff
@@ -390,6 +391,21 @@ def _get_configuration_variables(self):
                 ],
             })
 
+        if self.privacy_policy_url:
```
giovannicimolin: Since we're adding TOS and privacy, you'll need to update this.
BTW, kudos on making the configuration only set these variables when the fields are not blank. :+1:

registration/forms.py
jamestait: The rationale is that for new instances, we want to require a privacy policy, so I set the initial value here to provide the hint - the user is free to change it if they require. For existing instances that don't already have a privacy policy set (all instances to begin with), we want the registration to be editable without additionally adding a default Privacy Policy that the user hasn't reviewed.

instance/models/openedx_appserver.py
jamestait: Is ToS required here as well? The ticket and discovery exclusively mention Privacy Policy, whereas BB-985 explicitly mentions ToS in addition to Privacy Policy.

registration/forms.py
jamestait: The approach I've taken allows the user to edit the mutable fields without requiring them to also add a privacy policy, or automatically adding one in the background that they haven't had an opportunity to review. It's a little more complex, but it keeps things visible and means existing instance owners can add the policy URL when they're ready. If we want to force them to have a privacy policy, we should notify them, give them a timeline and apply our default if they don't respond.

registration/tests/test_views.py
jamestait: It will, but it will also add a privacy policy to existing instances without the owner's knowledge, which doesn't feel like a good experience - they should at least have the opportunity to review the privacy policy and provide their own if they don't like the one provided by default.

registration/forms.py
jamestait: I've removed this, and instead allowed empty values for both new and existing applications.

registration/models.py
```diff
@@ -137,8 +137,8 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
     privacy_policy_url = models.URLField(
         verbose_name='URL to Privacy Policy',
         help_text=('URL to the privacy policy.'),
-        null=True, # to ease migrations
```
giovannicimolin: Just a note here, the reason I asked for this change was to avoid having 2 possible values for *empty* stored on the database (`''` and `null`). :grin:

registration/tests/test_views.py
```diff
@@ -598,15 +598,15 @@ def test_preserve_empty_privacy_url(self):
         """
         self._register(self.form_data)
         application = BetaTestApplication.objects.get()
-        application.privacy_policy_url = None
+        application.privacy_policy_url = ''
         application.save()
         application.refresh_from_db()
-        assert application.privacy_policy_url is None
+        assert application.privacy_policy_url == ''
```
giovannicimolin: Change this to self.assertEqual (ref https://docs.python.org/3/library/unittest.html#assert-methods).

jamestait: The use of `assert` rather than `self.assertEqual` in this instance is intentional: we're asserting that the test setup has been performed correctly, rather than making assertions about the behaviour of the code under test. Happy to switch it, but I've found this distinction useful when things don't go according to plan.

giovannicimolin: We don't use `assert` like this in any other tests, so it's better for us to use the same pattern. Please change the assert to `self.assertEqual`.

giovannicimolin:
@jamestait Can you rebase, check if the changes and migrations are ok and ping me again when the tests are passing.

PR #422, opened 2019-03-28
BB-1081: Disable Studio SSO login

giovannicimolin:
This PR disables the Studio login via SSO that was introduced by Ironwood but keeps the subdomain functionality.

### Testing instructions:
1. Go to studio.
2. Check if the Sign In button lead to a login page inside Studio.
3. Check that the login on Studio works.
4. Check that the LMS login works.
5. Do the same steps for the external and internal domain names

### Sandboxes:
**hawthorn.1**: https://stage.console.opencraft.com/instance/3550/
**ironwood.1**: https://stage.console.opencraft.com/instance/3560/
**master**: https://stage.console.opencraft.com/instance/3570/

### Reviewers:
- [ ] @lgp171188

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -208,6 +208,10 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                 "USE_MICROSITES": False,
                 "PREVENT_CONCURRENT_LOGINS": False,
                 "ENABLE_ACCOUNT_DELETION": True,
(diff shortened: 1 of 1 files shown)


giovannicimolin:
@jamestait This is expected behaviour, as the LMS and Studio accounts are the same. The behaviour we are disabling is the SSO login on Studio, which redirects "Sign in" on Studio to LMS and makes it work as a SSO provider.

PR #423, opened 2019-04-05
[SE-943] Consul transactions

SSPJ:
We currently use [consul locks](https://github.com/open-craft/opencraft/blob/master/instance/models/utils.py#L468) to prevent race conditions between Ocim, cert-manager and other consul clients. However, the `ConsulAgent` class [here](https://github.com/open-craft/opencraft/blob/master/instance/models/utils.py#L447) is the only piece of code that writes instance/appserver configuration to Consul and all the other clients only read it. Due to the locking and the delays associated with locking and releasing the locks when there are multiple clients trying to acquire the lock, we often run into the `ConsulLockAcquisitionException` error due to one of the Ocim processes being unable to acquire the lock due to some reason - too much contention for the lock is a prime example.

By using [Consul transactions](https://www.consul.io/api/txn.html), which are atomic, we do not need locks and all the clients reading the information will only read the consistent values.

### DIFF
--- a/instance/management/commands/update_metadata.py
+++ b/instance/management/commands/update_metadata.py
@@ -84,12 +84,12 @@ class Command(BaseCommand):
         metadata from Consul.
         """
         instances_ids = self.get_archived_instances()
-        with ConsulAgent() as agent:
-            self.stdout.write('Cleaning metadata for {} archived instances...'.format(len(instances_ids)))
-            for instances_id in instances_ids:
-                prefix = settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=instances_id)
-                agent.delete(prefix, recurse=True)
-            self.stdout.write(self.style.SUCCESS('Successfully cleaned archived instances\' metadata'))
+        agent = ConsulAgent()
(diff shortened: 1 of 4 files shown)


viadanna:
@SSPJ @lgp171188 @giovannicimolin Sorry for hijacking the PR as SF but it's blocking a few other tasks currently.

I found out that the cause of the JSONDecodeError is that before we conditionally JSON encoded [here](https://github.com/open-craft/opencraft/blob/master/instance/models/utils.py#L530) while here we expect everything to be encoded and decoded.

I've pushed a commit to JSON decode if at all possible but not crash on strings that aren't encoded. What do you think?

lgp171188:
@viadanna, thanks for the fix. Since the existing code which writes the fields does not serialize strings into JSON, the changes you have made are necessary.

@giovannicimolin, I still have to write some unit tests for testing the retry logic. I don't think I can write integration tests unless I want to test the transactions functionality of consul

giovannicimolin:
@giovannicimolin No need to write integration tests on this ticket. As long as we unblock people working on Ocim it's fine to create a new ticket to add the integration tests later.

PR #425, opened 2019-04-08
Log but don't fail on AWS user cleanup

viadanna:
TBD


viadanna:
Current behavior failing but allowing merge is much better and safer than this.

PR #426, opened 2019-04-12
BB-1030 Collect code coverage from every test 

xirdneh:
Code coverage is not being calculated for integration tests nor for front end tests.
The solution is to run `codecov` with `--parallel-mode` that way every run will have a unique number attached at the end of the output file. Then, after every test has been run we run `codecov combine` which combines every code coverage file. Finally, we run `codecov report` to print the report and check for a coverage threshold.
There's multiple updates to the way CircleCI runs the workflow as well as how we run tests:

- Taking advantage of some [new things in CirlceCI](https://discuss.circleci.com/t/circleci-2-1-config-overview/26057) the workflow has been re-written to run in this way:
Screen Shot 2019-04-12 at 9 54 11 AM
The main gains from this type of workflow is to make sure everything is correct before running integration tests (since those are quite heavy) and to be able to re-run any step without having to re-run the entire workflow.

- First the quality test is run together with the new migrations checks. **If either test fails the rest of the test will not run**.
- Then, the workflow runs every unit test (JS jasmine tests, django unit tests and "browser" tests). Although "browsers" tests are not exactly unit test they fall under this category because they are more lightweight than integration tests and don't call any external services. **If either of these tests fail then the integration tests will not run**.
- There was a change on how we run jasmine tests. We used to run them using [python jasmine](https://jasmine.github.io/setup/python.html) which is really simple to setup for python projects but it's really hard to get code coverage since most stable tools for JS code coverage are expecting to be run with JS test runners. The JS tests are now run with Karma + Jasmine which allow us to use [istanbul](https://istanbul.js.org/) to calculate code coverage.
- By using Karma + Jasmine we can now use karma's browser launchers, which are easier to use. The downside here is that we now have to deal with `npm`.
- After the unit tests are run the workflow will run the integration tests separating them into three parallel jobs. The custom [test runner](https://github.com/open-craft/opencraft/blob/master/opencraft/tests/utils.py#L51) is now deprecated since the environment variable used to see in which "cirlce ci node" the test were running is not usable with the new workflow. Instead, a `test_group` variable is specified on each parallel job. The `test_group` variable is then passed into the job as an environment variable and [`unittest`'s `skipIf`](https://docs.python.org/3/library/unittest.html#unittest.skipIf) is used to check which tests should run on each parallel job. This makes the entire integration tests running easier to understand.
- After quality, unit and integration tests are run successfully a job to combine every code coverage is run. Every previously ran job persist its coverage file to a workspace which the `coverage` job then mounts and combine.
- Finally, the cleanup job is now [scheduled](https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow) to run every day at 1 am instead of being run with every CirlceCI run. The cleanup job will also run only for the `master` branch.s


Makefile
```diff
@@ -61,6 +63,21 @@ install_system_db_dependencies: apt_get_update ## Install system-level DB depend
 
 install_system_dependencies: apt_get_update ## Install system-level dependencies from `debian_packages.lst`. Ignores comments.
 	sudo -E apt-get install -y `grep -v '^#' debian_packages.lst | tr -d '\r'`
+	if [ -z $$CI ] ; then \
+		echo "Installing Firefox because we're not in a CI."; \
+		sudo apt-get install -y libgtk3.0-cil-dev libasound2 libasound2 libdbus-glib-1-2 libdbus-1-3; \
```
viadanna: I've created a patch for the deployment playbook `ocim-vagrant` to install those but I think we might as well keep this here. I just recommend adding `no-install-recommends` to prevent installed unnecessary packages.

Makefile
```diff
@@ -61,6 +63,21 @@ install_system_db_dependencies: apt_get_update ## Install system-level DB depend
...
+		sudo dpkg -i /tmp/firefox.deb || sudo apt-get -f install; \
+		sudo rm -rf /tmp/firefox.deb; \
+	else \
+		echo "Not installing Firefox either because this is a CI."; \
+	fi
+
+install_js_dependencies: ## Install dependencies for JS code.
+	curl -sL https://deb.nodesource.com/setup_11.x | sudo -E bash -
+	sudo apt-get install -y nodejs
+	npm install
```
viadanna: This is currently failing for me in the vagrant devstack with a
```
npm ERR! enoent ENOENT: no such file or directory, open '/home/vagrant/opencraft/node_modules/nyc/node_modules/handlebars/node_modules/source-map/package.json.3468720659'
```

Any ideas?

circle.yml
```diff
@@ -18,35 +17,29 @@ jobs:
...
       - checkout
       - restore_cache:
           key: dependencies-{{ checksum "requirements.txt" }}
       - run:
           name: Wait for Redis
-          command: |
-            while ! nc -z localhost 6379; do
-              sleep 0.1
-            done
+          command: dockerize -wait tcp://localhost:6379 -timeout 10s
```
viadanna: :+1:

circle.yml
```diff
@@ -66,6 +59,8 @@ jobs:
             sudo apt-get install postgresql-client
             sudo apt-get install unzip
             sudo wget -P /tmp https://releases.hashicorp.com/consul/1.2.1/consul_1.2.1_linux_amd64.zip
+            make install_system_dependencies
```
viadanna: This was not being run before, was it? We might not need it in CI if that's the case.

xirdneh: We might not need it. I thought maybe it would be better to have it that way the CI's environment is closer to what we have locally.
I can remove it, tho.

Makefile
xirdneh: @viadanna This might just be npm. Try removing `package-lock.json` and `node_modules` and try again.

xirdneh: @viadanna Also, try it again. We are not actually using `nyc` so I deleted it from `package.json`

circle.yml
viadanna: Fair enough, let's keep it.

Makefile
viadanna: Ok, could get it installed but now I'm seeing:
```
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
```
Can you update the necessary packages?

xirdneh: @viadanna Done

circle.yml
```diff
@@ -83,10 +78,57 @@ jobs:
...
+            npm install
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          command: |
+              . venv/bin/activate
+              coverage combine /tmp/workspace/coverage/$CIRCLE_SHA1/
+              make cov.html
+              if [ -e /tmp/workspace/coverage/$CIRCLE_SHA1/text/coverage.txt ]; then cat /tmp/workspace/coverage/$CIRCLE_SHA1/text/coverage.txt; fi
```
viadanna: Just a question: the coverage for JavaScript is just being reported but isn't taken into account for the minimum, is that right?

circle.yml
```diff
@@ -115,5 +157,68 @@ workflows:
...
+            - js_tests
+      - coverage:
+          name: coverage
+          requires:
+            - integration_tests_group_1
+            - integration_tests_group_2
+            - integration_tests_group_3
+              #- python_unit_tests
+              #- browser_tests
+              #- js_tests
```
viadanna: In the end we'll need to restrict merging on this job instead of the current build, right?

circle.yml
```diff
@@ -115,5 +157,68 @@ workflows:
...
+  cleanup-workflow:
+    triggers:
+      - schedule:
+          cron: "0 1 * * *"
```
viadanna: Do we get any sort of logs for this cron job?

circle.yml
xirdneh: There's two minimums. One is calculated by `coverage.py` and another one is calculated by `istanbul`. The codecoverage.py minimum is set in the [coverage threshold](https://github.com/open-craft/opencraft/pull/426/files#diff-b67911656ef5d18c4ae36cb6741b7965R24) and the istanbul minimum is setup in the [karma.config file](https://github.com/open-craft/opencraft/pull/426/files#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfbR122).
There's no simple way to combine both of this (afaik). Services like codecov or coveralls do combine multiple reports in different formats.

circle.yml
xirdneh: @viadanna What do you mean restrict merging on this job?

circle.yml
xirdneh: @viadanna It will run as a regular job so it'll show up in the job list. It'll get [scheduled](https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow) when there's a push to master because of the filter.

viadanna: :+1:

circle.yml
viadanna: When a job fails, so we can prevent merging. If this one or a previous one fails, do we need to restrict merging on all jobs or just this one?

circle.yml
viadanna: :+1:

circle.yml
xirdneh: @viadanna Ah, I see. Well this step only prints out the report it doesn't check for thresholds so it shouldn't fail.
But yes, we probably shouldn't restrict merging on this job since it's purely informational.
The idea is to eventually use this step to send the reports to codecov/coveralls.

xirdneh:
@viadanna We still need to change the circleci integration configuration so it doesn't stay on yellow.
I've left the details in the ticket. Let me know if you can't find them, thanks.

viadanna:
@xirdneh We'll have to remove the expectation of "Build" but what should we use instead? That was my question above about relying on the Coverage job. Can we have it fail if coverage is below the thresholds? What happens when earlier tests fail and it isn't run?

xirdneh:
@viadanna Ah, I see.
Sorry, I got confused.
No, this last step does fails if the threshold is less than 90%.
And this last step is not going to run if any of the last steps fail.
So, we can only keep the expectancy on this last step.

xirdneh:
@viadanna The check runs when we run `make cov.html`

viadanna:
@giovannicimolin Can you review my changes as SF?

circle.yml
```diff
@@ -82,10 +78,57 @@ jobs:
...
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install --upgrade pip
+            pip install --upgrade virtualenv
+            # The jessie-updates and jessie-backports areas are now removed from the repositories. So
+            # the lines corresponding to those areas have to be removed to fix the resulting errors.
+            # This can be removed after upgrading to the image based on Debian Stretch.
+            sudo sed -i '/jessie-updates/d' /etc/apt/sources.list
+            sudo sed -i '/jessie-backports/d' /etc/apt/sources.list
```
viadanna: Is this still relevant after changing the image to a `stretch` one?

xirdneh: @viadanna I don't think so. I'm rebasing with master and taking care of this. Thanks.

circle.yml
```diff
@@ -82,10 +78,57 @@ jobs:
             make install_system_dependencies
             sudo ln -s /usr/lib/x86_64-linux-gnu/libmysqlclient.so /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18
             bin/install-supported-firefox
```
lgp171188: @viadanna, this could be used to replace the Firefox installation code above or can be removed if it is duplicate work.

circle.yml
```diff
@@ -114,5 +157,65 @@ workflows:
...
+            - integration_tests_group_2
+            - integration_tests_group_3
+  cleanup-workflow:
+    triggers:
+      - schedule:
+          cron: "0 1 * * *"
+          filters:
+            branches:
+              only:
+                - master
```
lgp171188: @viadanna, why is the cleanup job running only on the master branch? It should run in all the CI runs to clean up resources older than 8 hours.

viadanna: It's now running on a cron, everyday at 1am, so there's no need to run on all branches given the script will cleanup all resources, @lgp171188

circle.yml
```diff
@@ -82,10 +78,57 @@ jobs:
...
+          no_output_timeout: 2h
           environment:
             ANSIBLE_HOST_KEY_CHECKING: False
+      - persist_to_workspace:
+          root: /tmp
+          paths:
+              - coverage
+  coverage:
+    docker:
+      - image: circleci/python:3.5.4-node
```
lgp171188: @viadanna, this and the following lines (and any other part not using Debian Stretch) should be updated to include the changes made in a733d982ddd7a6435f9b5771f55099ee4d142e28 to update the docker image and code to install the dependencies.

circle.yml
lgp171188: @viadanna, so there is a separate CI job? Nice! :+1:

circle.yml
```diff
@@ -114,5 +157,65 @@ workflows:
...
+          requires:
+            - quality_migrations_tests
+      - build:
+          name: js_tests
+          tests_group: 0
+          tests_type: clean test.js
+          requires:
+            - quality_migrations_tests
+      - build:
+          name: integration_tests_group_1
```
lgp171188: @viadanna, afaik, we can have 4 shards in a run. So it might be a good idea to move the integration tests that don't spawn a full-fledged appserver to a separate shard. This also makes debugging SSL issues and other easier because all the other shards will have just one test that spawns a full-fledged appserver.

circle.yml
viadanna: Alright, I see only the build job had the updated image. I've updated the other ones and will push for another test.

circle.yml
viadanna: :+1:

circle.yml
viadanna: :+1:

debian_packages.lst
```diff
@@ -1,6 +1,5 @@
 git
 libffi-dev
-libmysqlclient-dev
```
swalladge: @viadanna is there a reason this was removed from the package list? It breaks devstack provisioning for me, because one of the python requirements requires it.

PR #428, opened 2019-05-03
Update usage of _create_bucket to work with the new boto3 code

jamestait:
This branch updates the `migrate_swift_to_s3` management command's usage of `OpenEdXInstance._create_bucket()` to use `max_retries` in the kwargs instead of `attempts`. This change was introduced as part of the upgrade to boto3.


jamestait:
As per Jira:

Paulo Viadanna wrote:
> given we are over budget already I prefer to close the PR and reopen it if necessary when the time to migrate those comes.

PR #429, opened 2019-05-03
[BB-1191] Update the default release to Ironwood

lgp171188:
This is done in preparation for mass-upgrading the self-hosted Ocim
instances to Ironwood


viadanna:
@lgp171188 The change looks good but the tests are failing because the integration mongodb is not found. It will be back in an hour or so.

lgp171188:
@viadanna, after the CI checks pass, I will deploy this and the other PR on stage and spawn a new instance to verify that everything works fine and share the details here.

PR #431, opened 2019-05-09
Fix a bug causing the load balancer configuration fragments to not get cleaned up and a datetime warning

lgp171188:
**Testing instructions**
* Verify that no integration test load balancer configuration fragments are skipped when the deletion task is run in the cleanup CI task.
* Verify that there are now no warnings about using a native datetime when timezone support is enabled.


giovannicimolin:
@lgp171188 Just reviewed it ^

PR #433, opened 2019-06-04
Remove swampdragon

swalladge:
This PR strips out all swampdragon and the functionality that it provided.

**Testing instructions**:

1. set up environment ready to create instances
2. make sure there is at least one instance reference to work with
3. select an instance and click "Launch new AppServer"
4. wait, refresh page, observe that new appserver has been created
5. check "Log" tab to verify that logs load
6. if appserver created successfully, select in and expand Log accordian. Verify that appserver logs load.
7. refresh page and expand Log accordian again. Verify that new log entries appear since last load.


instance/static/js/src/instance.js
```diff
@@ -221,27 +203,8 @@ app.controller("Details", ['$scope', '$state', '$stateParams', 'OpenCraftAPI',
 
             OpenCraftAPI.all("openedx_appserver").post({instance_id: $stateParams.instanceId});
             // The API call above is an asynchronous task so it will return a 200 status immediately.
-            // When the new app server is successfully created, it will send an openedx_appserver_update
-            // notification, which will trigger $scope.refresh(), which will reset 'is_spawning_appserver'
```
clemente: I tested this.
- With the old behaviour: after clicking the „Launch new AppServer“ button, it becomes disabled for some seconds (and includes a spinner), then becomes enabled again
- With the new behaviour: after clicking the „Launch new AppServer“ button, it becomes disabled for some seconds… and for ever, and nothing else changes. It looks like there was some error (like a JS error), and I feel it won't be very clear to the user that things are fine and we need to wait & refresh. Note that we have a few non-OpenCraft Ocim users

I thought of adding an `alert(…)` to say that it has been correctly spawned and it will be visible after refreshing. We don't use any other `alert`, and it can be intrusive if we use them a lot… (I would still prefer it but I guess others won't). What we have used to show notification is something [like this](https://github.com/open-craft/opencraft/blob/65e2b87a7e3796044b8c90aa928cf081f8312892/instance/static/js/src/openedx_appserver.js#L89), though the notification isn't very prominent (you can actually test this by going to a terminated server, force-enabling the terminate button through your browser, and then clicking the button).
So could you add some message to the user as mentioned above?

Sorry for not foreseeing the eternal spinner in the launch button.

By the way, any other more user-friendly approach can be good too (like: changing the text of the button to something like „launching. Please refresh after some seconds to see the new server“)

clemente:
:+1:

  - [X] I tested this: what I mentioned above, and in addition the button changes text after being clicked
  - [X] I read through the code
  - [X] Includes documentation: updates it

I'll merge+deploy soon in around half an hour (right now there are servers being deployed).

PR #434, opened 2019-06-08
[BB-1054] Wait for registration validation in browser tests instead of sleeping unconditionally

jamestait:
Description
=========

Adds to the Registration browser tests a custom WebDriver wait based on the Angular validation classes applied to server-side validated fields, and uses that with a longer timeout instead of an unconditional sleep.

Testing
----------
1. Browser tests ran by CircleCI must pass :)

Bonus: In a local devstack, run the browser tests multiple times to make sure there's no flakiness any longer. (I did this already)


registration/tests/utils.py
```diff
@@ -35,6 +35,25 @@
...
+
+    ng_validated_classes = ('ng-valid', 'ng-invalid')
+
+    def __init__(self, locator):
+        self.locator = locator
+
+    def __call__(self, driver):
+        element = driver.find_element(*self.locator)
+        classes = element.get_attribute('class').split(' ')
+        if 'ng-dirty' in classes and any(ng_class in classes for ng_class in self.ng_validated_classes):
```
Agrendalath: Could you please add some more explanation here or in the class docstring?
You have explained this workaround nicely in the comment, but it would be good to have it persistent in the code.

registration/tests/utils.py
```diff
@@ -65,10 +84,12 @@ def form(self):
         """
         return self.client.find_element_by_tag_name('form')
 
-    def fill_form(self, form_data):
+    def fill_form(self, form_data, validate_fields=None):
         """
         Fill in the form with the given data.
         """
+        if validate_fields is None:
+            validate_fields = tuple()
```
Agrendalath: Nit:
```python
validate_fields = validate_fields or ()
```

I personally find this notation much more concise.

registration/tests/utils.py
```diff
@@ -85,12 +106,18 @@ def fill_form(self, form_data):
                 self.client.execute_script("document.getElementById('{}').type='text'".format(id_elem))
 
             if not element.get_attribute('readonly') and not element.get_attribute('type') == 'hidden':
+                element.click()
                 element.clear()
                 if value:
+                    time.sleep(.25)
```
Agrendalath: Could you please add a comment here, explaining why we're doing this?

registration/tests/utils.py
viadanna: :+1:

registration/tests/utils.py
viadanna: :+1:

registration/tests/utils.py
viadanna: :+1:

PR #436, opened 2019-06-11
[SE-708] Upgrade all packages to the latest versions. Move to Django 2

clemente:
This PR upgrades all packages in `requirements.txt` to their latest versions.
In particular it migrates to the latest Django version 2.2.
It also updates many core packages like huey, honcho, ipython, … and also the cleanup tests. No functionality is lost.
It also introduces `pip-tools` to manage our depencies file (see below).

This is the work of SE-711 (upgrading packages) plus SE-713 (making tests work again). Also SE-709 (pip-tools), SE-1256 (fix code quality) and SE-1415 (deployment).

**Details**

***OpenStack***
- the parts about OpenStack had to be a bit rewritten, both the connection setup and some parameters. Also the part about `num_retries` (retrying in case on error). Some more work could help to simplify the code (e.g. we're using 2 OpenStack libraries with a bit different behaviour)

***django-angular***
- migrating to the latest django-angular is unbearable for too many reasons: it forces us to rewrite the way we handle files (logo/favicon) and the code that handles it isn't working (`render() got an unexpected keyword argument 'renderer'`) and it requires more packages and even Node. Disabling  django-angular is troublesome and requires rewriting many tests. So we're not upgrading django-angular, and instead I fixed the version we're using to make it work in Django 2, it. It's in [this branch](https://github.com/open-craft/django-angular/tree/v1-with-django2-support)

***quality tests (pylint)***
- upgrading pylint surfaced hundreds of new code quality errors: no documentation, line too long, redefined function with different methods, too many branches, broad exception, unused import, unneeded return, and more. This PR fixes them. It involves refactoring some functions

***minor changes***
- some functions were changed to use `str` instead of `bytes`, since the new version of some package started passing `b"strings like this one"` to templates and to DB fields, and they ended up in haproxy config files and in DB fields exactly like that: with a lowercase `b`, a quotation mark etc. (3 extra bytes)

***requirements.txt is now autogenerated by pip-compile***
- earlier: we hand-edit `requirements.txt` or use `pip freeze >requirements.txt`. It includes all
- now: we hand-edit `requirements.in` to list the software we care about (our direct dependencies). Then we run `pip-compile`. This creates `requirements.txt` with the full list. We commit both.

***not upgraded to the latest version***
- pylint was a mess; we can't upgrade to the latest versions of packages because there are many dependency conflicts and in addition many open bugs. The status of each package is: 
  - `pep8-naming`: newest version 0.8.2, but latest `prospector` (1.1.6.4) limits us to 0.4.1
  - `pycodestyle`: newest version 2.5.0, but latest `prospector` (1.1.6.4) limits us to 2.4.0. Actually, 2.3.1 due to `flake8`
  - `pyflakes`: newest version 2.1.1, but latest `prospector` (1.1.6.4) limits us to 1.6.0
  - `flake8`: newest version 3.7.7, but the mentioned `pyflakes` downgrade limits us to 3.5.0
  - `astroid`: newest version 2.2.5, but [a bug](https://github.com/PyCQA/pylint-plugin-utils/issues/12) with an old version of pylint/prospector forces us to downgrade to 2.1.0.  By the way the maintainer of some of those tools, Carl, worked in OpenCraft (hello if you read this)
  - `pylint`: newest version 2.3.1 but that same bug affects many components, and downgrading `pylint` to 2.2.3 avoids it
  - `pylint-django`: newest version 2.0.9, but it has [another bug](https://github.com/PyCQA/pylint-django/issues/232) (open!) that I saw in my Ocim, and I had to downgrade to 2.0.6 to fix it
  - that means that: *prospector* is holding many of our packages by a few years (we could try to try it with recent versions and upstream it), but in addition some of the *pylint*-related bugs (including open bugs) would make an upgrade to the latest versions a bit longer (we'd have to upstream some fixes)
- in addition `PyYAML`: newest version 5.1.1, but latest `jasmine` (3.4.0) limits us to 3.13
- selenium isn't upgraded to version 3 since we found new problems in the test cases that would require a non-trivial rewrite. See SE-1414

***Playbooks***
They don't require any update. `.env` files don't require updates either.

**Testing instructions**

- test recreating vagrant
- or `pip install -r requirements.txt` from a current Ocim
- run Ocim and verify that it runs and can deploy. Tto do real deployments you'll need a good `.env`. Ask someone else for theirs or take Ocim stage's .env 
- integration tests should pass
- we'll test this in *stage* first


instance/tests/models/test_openedx_appserver.py
```diff
@@ -134,6 +134,7 @@ def test_provision_unhandled_exception(self, mocks):
         self.assertFalse(result)
         mocks.mock_provision_failed_email.assert_called_once_with("AppServer deploy failed: unhandled exception")
 
+    # pylint: disable=too-many-locals
```
SSPJ: The threshold for `too-many-locals` [is 15](http://pylint-messages.wikidot.com/messages:r0914). By my count, there are 16 in this function. Getting rid of `instance` and `organization` on lines 178, 179 would fix the linter.

The purpose is to discourage making functions that do too much. This function borderline does too much, so if we can get rid of the `#pylint: disable` we can force the next person who adds to this function to also refactor it.

registration/auth_backends.py
```diff
@@ -50,5 +50,6 @@ def authenticate(self, username=None, password=None):
             # difference between an existing and a non-existing user (#20760).
             UserModel().set_password(password)
         else:
-            if user.check_password(password):
+            if user.check_password(password) and self.user_can_authenticate(user):
```
SSPJ: What prompted this change?

registration/forms.py
```diff
@@ -34,7 +34,7 @@
 from django.utils import timezone
 from django.utils.text import capfirst
 from django.template.loader import get_template
-from djng.forms import NgDeclarativeFieldsMetaclass, NgFormValidationMixin, NgModelForm, NgModelFormMixin
+# from djng.forms import NgDeclarativeFieldsMetaclass, NgFormValidationMixin, NgModelForm, NgModelFormMixin
```
SSPJ: Ditto.

registration/forms.py
```diff
@@ -105,7 +105,10 @@ class BetaTestApplicationForm(forms.ModelForm):
     """
     class Meta:
         model = BetaTestApplication
-        exclude = ('user', 'status', 'instance', 'accepted_privacy_policy')
```
SSPJ: Is this related to the failing tests? All of the tests with user registration are passing the wrong default for `accept_privacy_policy` and `accept_terms`.

registration/auth_backends.py
swalladge: This was updated to reflect django 2.2.2's base [authenticate](https://github.com/django/django/blob/9400c96b209855d7b97e3223e4b8ba7751c357cc/django/contrib/auth/backends.py#L16-L27) method. I'm not sure how ocim manages disabling users, but I thought it best to update it and discuss in the review here.

registration/forms.py
swalladge: Yes I think you're right. Perhaps in this case it would be better to keep using the exclude list?

swalladge: Hmm even when I add all the missing fields to the explicit fields list, it still fails those tests. I'm going to switch back to the exclude list method and disable pylint on it for now.

swalladge: Interestingly, those registration tests still fail with the exclude list set back to how it was originally. :thinking:

instance/tests/test_tasks.py
```diff
@@ -369,8 +369,8 @@ def test_shut_down_obsolete_pr_sandboxes(self, data, mock_archive, mock_logger):
             closed_at = None
 
         with patch(
-            'pr_watch.github.get_pr_info_by_number',
-            return_value={'state': pr_state, 'closed_at': closed_at},
+                'pr_watch.github.get_pr_info_by_number',
```
toxinu: Is it `pylint` which ask for double spaced tabs under `with` statement?

pylintrc
```diff
@@ -100,7 +100,8 @@ disable=
     too-many-public-methods,
     protected-access,
     unidiomatic-typecheck,
-    unused-argument
+    unused-argument,
+    no-else-return
```
toxinu: :+1:

registration/models.py
```diff
@@ -243,6 +243,7 @@ def first_activated(self):
         """
         if self.instance:
             return self.instance.first_activated
+        return None
```
toxinu: Is it `pylint` which want to use `return None` instead of just `return` to be more explicit?

swalladge: Yes

instance/tests/test_tasks.py
swalladge: Yes

registration/forms.py
SSPJ: @swalladge Sorry for sending you up a wrong tree. I'm not sure what the issue is. There are two commits: 6f60aa1ef23b0f34522db570dd628896ed58e885 and 727d8752ae48b60350b260c9b2e8f430ce24b034 , which add a default of "checked" for those two boxes during registration. You have both on this branch.

SSPJ:
@swalladge I am comfortable giving this a :+1: for the scope of SE-1256.

- [x] I tested this: checked that the failing CI builds do not appear to be related to any pylint errors
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ n/a
- [ ] ~~Includes documentation~~ n/a

Per @clemente in the description, it does not need to be merged or deployed.

swalladge:
FYI the reason why a bunch of the registration form tests are currently failing is because the error message scraping function is designed to work with django angular forms, which are currently disabled. https://github.com/open-craft/opencraft/blob/cb23f223236e2b45b177920741cbd7f739b9658c/registration/tests/test_views.py#L507-L524

registration/forms.py
swalladge: @SSPJ I believe I discovered the issue - it was related to the form changes and how the value was being parsed. I made a [note of the other related issues](https://github.com/open-craft/opencraft/pull/436#issuecomment-510716910), but didn't attempt to fix them because this is because of the temporary removal of django angular forms. This issue I was able to fix because it was a general incorrectness in how the value was parsed.

SSPJ: Good work sleuthing that out, @swalladge !

swalladge:
`test_os_server_nova_error` may be failing partly because `management_url` has been deprecated - mocking this may longer be having the correct effect:

https://github.com/open-craft/opencraft/blob/clemente/django2-and-other-upgrades/instance/tests/models/test_server.py#L138

swalladge:
@clemente note that I needed to revert https://github.com/open-craft/opencraft/pull/426/files#r311844651 to provision my devstack.

swalladge:
@clemente If some of the nova client methods raise different errors because keystoneauth1, we may need to update areas like: https://github.com/open-craft/opencraft/blob/9039af308f209f3eb41b5540f33ab22e9c1ad4c7/instance/models/server.py#L347-L364

to check for new exceptions. For example, nova client.servers.get('name') appears to raise `keystoneauth1.exceptions.discovery.DiscoveryFailure` instead of the original requests ConnectionError

swalladge:
The browser tests were failing on my devstack and in CI - I managed to fix it by tweaking the selenium driver capabilities: https://github.com/open-craft/opencraft/pull/436/commits/d63522662125b74509d395d2fdc532b7548a3bd4 (fix discovered from https://stackoverflow.com/questions/47782650/selenium-common-exceptions-sessionnotcreatedexception-message-unable-to-find-a)

However now it's failing on CI with a different error. I suspect it's related to the different installation instructions on CI:

https://github.com/open-craft/opencraft/blob/d63522662125b74509d395d2fdc532b7548a3bd4/Makefile#L64-L75

swalladge:
@clemente I think I've tracked the issues down to the update from selenium 2 to 3. Investigating in https://github.com/open-craft/opencraft/pull/452. It's difficult because the CI environment is different to the local devstack.

Also is there a reason we use an old firefox version for testing and ci?

clemente:
> Also is there a reason we use an old firefox version for testing and ci?

No reason, it's probably that we were using the latest version of firefox accepted by our old selenium. We're migrating selenium==2.53.6	→ selenium==3.141.0 so we'll need to check the right version. I tried figuring out the supported firefox version for 3.141.0. [I found out](https://www.seleniumhq.org/about/platforms.jsp) `Support for Firefox is the latest release, the previous release, the latest ESR release and the previous ESR release.` but we'll still need to find the right version: not too high, not too low, not too different from what was working…

swalladge:
@clemente Ah ok, makes sense. I still had heaps of issues with running the tests with selenium 3 in #452 . Perhaps it would be a good idea to upgrade to selenium 3 in a new ticket later?

clemente:
@swalladge Yes we probably need to switch to `keystoneauth1.exceptions.discovery.DiscoveryFailure`… if it makes sense. We'll check that. It may be that ConnectionError is still thrown in other cases.

Thanks for the debugging and fixes!
I see that the 3 integration tests are failing for different reasons, some related to the infrastructure (build 6817).

And I'll create a task to migrate selenium 2 → 3.

registration/auth_backends.py
swalladge: @clemente are you aware if this may cause issues? Is there ever a case where this function may return false but we still want the user to be able to login?

requirements.txt
```diff
@@ -1,156 +1,201 @@
...
+ddt==1.2.1
+debtcollector==1.21.0
+decorator==4.4.0
+Django==2.2.2
+# with Django 1 we were using: -e git+https://github.com/jrief/django-angular.git@3a4eb9e716b0f03e4abe5ec60d6978a52da100f8#egg=django_angular	
+# with Django 2, clean option:
+# django-angular==2.2
+# but django-angular 2.2 requires a heavy rewrite of many parts of the form. Disabling django-angular does too. So a compromise is to patch our django-angular version to work with Django 2:
+# TODO: use opencraft's branch
+-e git+https://github.com/open-craft/django-angular.git@clemente/migrate-our-django-angular-to-django-2#egg=django_angular
```
swalladge: This TODO must be resolved before merging

requirements.txt
```diff
@@ -1,156 +1,201 @@
...
+django-storage-swift==1.2.19
 django-zurb-foundation==5.5.0
-djangorestframework==3.5.3
+djangorestframework==3.9.4
 docutils==0.14
 dodgy==0.1.9
-factory-boy==2.7.0
-fake-factory==0.7.2
-freezegun==0.3.8
+dogpile.cache==0.7.1
```
swalladge: There appear to be some dependencies here that aren't in the original requirements.txt. For example, dogpile is new, and I can't find it directly used anywhere in the codebase. If they are dependencies of dependencies, then perhaps they shouldn't be listed here.

cleanup_utils/requirements.txt
```diff
@@ -1,6 +1,7 @@
...
-PyYAML==3.10
-Django==1.10.7
-
+boto3==1.9.171
+mysqlclient==1.4.2.post1
+python-novaclient==14.1.0
+PyYAML==3.13
+Django==2.2.2
+django-environ==0.4.5
+virtualenv==16.6.1
```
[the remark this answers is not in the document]
clemente: Yes. I got CircleCI tasks whose cleanup phase specifically failed because of not having the virtualenv package (and I think the django-environ package too).

instance/models/mixins/secret_keys.py
```diff
@@ -216,7 +216,7 @@ def http_auth_pass(self):
 
     def http_auth_info_base64(self):
         """
-        Return the HTTP auth information in the format required by Authorization HTTP header.
+        Return the HTTP auth information in the format required by Authorization HTTP header, as a string.
         """
         user_pass = '{}:{}'.format(self.http_auth_user, self.http_auth_pass)
-        return b64encode(user_pass.encode('latin1'))
+        return b64encode(user_pass.encode('latin1')).decode("utf-8")
```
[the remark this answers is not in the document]
clemente: What we're encoding is user/pass which will be ASCII anyway, this is more about `str` to `bytes`.
Ok, I changed it to `utf-8` since if we're ever going to accept any non-ASCII character like `á` we'd also want to accept others like `ĉ` (non-latin1).

registration/auth_backends.py
clemente: No problem here. [user_can_authenticate](https://github.com/django/django/blob/4f7328ce8a35160d155c41d362c3d674f8ef4d2d/django/contrib/auth/backends.py#L54) mainly checks if `user.is_active`. If a user isn't active, they shouldn't log in.

requirements.txt
clemente: I fixed it and I really moved it to a branch without my username, https://github.com/open-craft/django-angular/tree/v1-with-django2-support

requirements.txt
clemente: They are indeed dependencies of dependencies. I created [this dependency tree](https://gist.github.com/clemente/4d0a3dfd0337e4259a12fda6633978ae) with pipdeptree, and there we see that some are several levels deep.

I think they should be listed, mainly to allow us to do a simple `pip freeze > requirements.txt` and not have to manually edit a lot of information afterwards. And 2nd, to pin them to a version that works; this is to avoid auto-upgrading and to solve possible conflicts. Dependencies of dependencies can still create conflicts (A requires P v1, B requires P v2).

I initially deleted everything that we weren't directly using, and in the end I gave up since it was much more complex to maintain than using `pip freeze`.

swalladge: I'm not a huge fan because it makes it difficult to see at a glance which are our pinned dependencies, and which are random sub dependencies. Also blindly using the output of pip freeze can result in unwanted packages making it into the deps list because of previously installing temporary packages for whatever reason. For example, you have added `pipdeptree`, and `pip-autoremove` which would have been in your local environment but I'm pretty sure aren't used anywhere in ocim.

clemente: I know and that's how this file has been maintained until now. We had a task to simplify this. I answered in SE-713.
I found it wouldn't be too hard to migrate to a cleaner system so I just did it, to pip-tools (i.e. pip-compile), and now we have requirements.in with the basic dependencies and it autogenerates requirements.txt with all of them. I also removed the packages that we seem we're not using (there were lots more).

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -336,6 +342,9 @@ def test_email_address_deletion_on_disabling_monitoring(self, mock_newrelic):
         NewRelicEmailNotificationChannel.objects.get(email='admin@opencraft.com')
 
     def test_deleting_shared_notification_email_address(self):
+        """
+        Test that a shared email notification channel can't be deleted.
```
giovannicimolin: @clemente Shared notification channels can't be deleted because they are shared between many instances.
For example: `ops@opencraft.com` will be a shared notification channel that will be used for NewRelic alerts from all instances. `urgent@opencraft.com` is also another alert that is shared by all maintenance instances.  We can't delete them because if we do, all other instances that depend on that notification channel will also be deleted.

This was made this way because how NewRelic alerts works, and we can't add two of the same email address to NewRelic Alerts.

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -126,8 +125,8 @@ def test_enable_monitoring_for_pre_new_relic_alerts_instances(self, mock_newreli
         """
         instance = OpenEdXInstanceFactory()
         monitor_ids = [str(uuid4()) for i in range(len(instance._urls_to_monitor))]
-        for index, monitored_url in enumerate(instance._urls_to_monitor.keys()):
-            NewRelicAvailabilityMonitor.objects.create(instance=instance, pk=monitor_ids[index])
+        for monitor_id in monitor_ids:
```
giovannicimolin: Does this have the same effect as the previous code?
It seems like I let this pass on the review.

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -150,20 +149,27 @@ def test_enable_monitoring_for_pre_new_relic_alerts_instances(self, mock_newreli
         )
 
     @patch('instance.models.mixins.openedx_monitoring.newrelic')
-    def test_enable_monitoring_does_not_skip_any_new_relic_alerts_code_on_retry_after_error(self, mock_newrelic):
+    def test_enable_monitoring_does_not_skip_alerts_on_retry_after_error(self, mock_newrelic):
```
giovannicimolin: I'm not sure about this one too, reading both titles leave me a little confused, but the new one seems cleaner.
I you can change it :+1:

requirements.txt
swalladge: Nice! At the very least this provides a way to tell the difference between packages we care about and sub dependencies. :+1:

instance/tests/models/test_openedx_monitoring_mixins.py
clemente: Thanks, I added this comment to the test, for more context.

instance/tests/models/test_openedx_monitoring_mixins.py
clemente: I think yes (also, the test passes): `monitored_url` wasn't used, only the index is used, and `index` is just 0 1 2 3 4… to get the IDs. Better to directly iterate over the IDs.

clemente:
This was already reviewed (mainly @swalladge) and tested in the stage server and also in the devstack. CircleCI are passing too. There's some flakiness in the CircleCI tests but it's affecting all other servers equally (I verified that the specific errors we got, like the 503 error, were happening on other servers too). Similarly, some deployed servers showed errors in stage (e.g. pymongo: `AutoReconnect: timed out`), but I also found the same errors in recent servers spawned before I put the code there.
I also arranged the commits to remove FIXMEs, to remove very temporary approaches, to group them by topic, to make each one solve what it says (more or less). I didn't squash them still more since they are related to dozens of different issues.

So with all, this is ready to merge and deploy!

PR #437, opened 2019-06-14
BB-184 Update Ocim default values.

xirdneh:
This PR changes the default `EDXAPP_WORKERS` and `MAX_REQ` values.
After a lot of testing we started using 14 threads for a specific client with a lot of traffic and it worked great. This is the reason why we're suggesting a change in the default thread number.
The max request has also been lowered to 1000 for both cms and lms to avoid memory leaks.


instance/models/mixins/openedx_config.py
```diff
@@ -225,8 +225,8 @@ def _get_configuration_variables(self):
             # By default, the number of workers is num_cores*4 for LMS, num_cores*2 for CMS,
             # which turns out to be too much.
             "EDXAPP_WORKERS": {
-                "lms": 3,
-                "cms": 2,
+                "lms": 14,
+                "cms": 14,
```
kaizoku: Do we want CMS workers to be the same as LMS?
I thought they were usually around half that of the LMS.

instance/models/mixins/openedx_config.py
```diff
@@ -235,7 +235,7 @@ def _get_configuration_variables(self):
             # e.g if the instance receives an average of around 20000 requests per day,
             # with 10% for static assets, restarting after 10000 requests means
             # restarting each of the 3 LMS workers about every .8 days.
-            "EDXAPP_LMS_MAX_REQ": 5000,
+            "EDXAPP_LMS_MAX_REQ": 1000,
```
kaizoku: This is to account for the increased number of workers?
Seems like a good idea

xirdneh: @kaizoku yes.

instance/models/mixins/openedx_config.py
xirdneh: @kaizoku Well since CMS has less traffic we could probably lower it.
I'll update the PR, thanks.

PR #440, opened 2019-06-25
[BB-960] Don't fail when Github user does not exist

viadanna:
Description
=========

This PR implements a check for the Github users before provisioning the servers to prevent failures when a user isn't found.

Testing
----------

1. Check out this branch in Stage and restart the server.
2. Replace a user's Github handle with one that does not exist.
3. Start provisioning a new App Server.
4. Make sure it does not fail when adding users, as the user that does not map to a Github user does not show the `github` key.

```
2019-06-25 10:29:07-0300 INFO
changed: [54.38.70.171] => (item={u'type': u'admin', u'name': u'viadannap'})
2019-06-25 10:29:07-0300 INFO
changed: [54.38.70.171] => (item={u'github': True, u'type': u'admin', u'name': u'viadanna'})
```

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -9,7 +9,7 @@ from random import randint
 from django.conf import settings
 
 from instance.models.mixins.common_config import ConfigMixinBase
-from instance.models.utils import retry_session
+from instance.models.utils import check_github_users
 
 
 # Classes #####################################################################
@@ -384,11 +384,6 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             })
 
         if self.admin_users:
-            session = retry_session()
-            users = [
-                username for username in self.admin_users
-                if session.get('https://github.com/{}.keys'.format(username)).status_code == 200
-            ]
             template.update({
                 "COMMON_USER_INFO": [
                     {
@@ -396,7 +391,7 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                         "github": True,
                         "type": "admin"
                     }
-                    for github_username in users
+                    for github_username in check_github_users(self.admin_users)
                 ],
             })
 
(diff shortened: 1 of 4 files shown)


instance/models/mixins/openedx_config.py
```diff
@@ -383,12 +384,20 @@ def _get_configuration_variables(self):
...
+                username for username in self.admin_users
+                if session.get('https://github.com/{}.keys'.format(username)).status_code == 200
+            ]
             template.update({
                 "COMMON_USER_INFO": [
                     {
                         "name": github_username,
                         "github": True,
                         "type": "admin"
+                    } if github_username in users else {
```
lgp171188: @viadanna, do we have to include the users with missing keys as they will not have a way to login?

instance/models/utils.py
```diff
@@ -651,3 +653,18 @@ def _is_json_serializable(obj):
...
+    session = session or requests.Session()
+    retry = Retry(
+        total=retries,
+        read=retries,
+        connect=retries,
+        backoff_factor=backoff_factor,
+        status_forcelist=status_forcelist,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount('http://', adapter)
```
lgp171188: @viadanna, is there a reason we need HTTP?

instance/models/mixins/openedx_config.py
```diff
@@ -9,6 +9,7 @@
 from django.conf import settings
```
lgp171188: @viadanna, can you add unit tests for the changes made in this PR?

viadanna: Sure :+1:

instance/models/mixins/openedx_config.py
viadanna: I thought it was better than just ignoring the user. I can remove this instead.

instance/models/utils.py
viadanna: Not for Github, I can remove it but it'll make the feature less generic.

PR #441, opened 2019-06-28
[BB-1009] Update how we save configuration to use the prefix as key and a json …

xirdneh:
Update how we save configuration to use the prefix as key and a json object as value.
Currently the configuration values are saved in multiple key/value pairs, e.g.:
```json
[{
    "Key": "prefix/key1",
    "Value": "value1",
}, {
    "Key": "prefix/key2",
    "Value": "value2"
}
]
```
This PR will update the way the config values are saved to use the instance's prefix as key and the value will be a json object with all the configuration values.
There are different methods implemented in the 'ConsulAgent' utility class to do basic CRUD opreations on the new formatted config objects.

# Rationale

We are updating the way Consul saves instance's metadata but there are a lot of instances already running with a different metadata format. If we flip the switch and allow Ocim only handle one type of metadata format then we are going to have some down time while we update the other instance's metadata format. The code in this PR is backwards compatible and will only UPDATE old style metadata. For any new instance it will CREATE and UPDATE the new style of metadata.
This works because consul KV store is built in a way that uses prefixed. So, if there's three KV pairs with the prefix `ocmi/instances/121/` (e.g. `ocim/instances/121/slug`, `ocim/instance/121/domains`, `ocim/instances/121/key`) then Consul will only retrieve these values if `kv.get` is called with `recurse=True` if not consul will not return anything for `kv.get("ocim/instances/121")`.

# Testing instructions
1. (Before pulling this PR) In a Django shell, create an instance:
```python

    name='Ironwood sandbox',
    sub_domain='ironwood',
    # The rest of the parameters are all optional:
    email='myname@opencraft.com',
    openedx_release='named-release/ironwood',
    configuration_version='named-release/ironwood',
    configuration_extra_settings='',
)
```
2. Make sure the metadata was saved:
```python
from instance.models.utils import ConsulAgent
ca = ConsulAgent(prefix='ocim/instances/<instasnce_id>')
ca._client.kv.get(ca.prefix, recurse=True)
```
3. Pull this branch and update the created instance metadata
```python
instsance.update_consul_metadata()
```
4. Make sure nothing changed in the consul stored data .This is ensuring the code is backwards compatible.
```python
from instance.models.utils import ConsulAgent
ca = ConsulAgent(prefix='ocim/instances/<instasnce_id>')
ca._client.kv.get(ca.prefix, recurse=True)
```
5. Create a new instance and make sure the metadata was saved in the new format:
```python

    name='New sandbox',
    sub_domain='newsandbox',
    # The rest of the parameters are all optional:
    email='myname@opencraft.com',
    openedx_release='named-release/ironwood',
    configuration_version='named-release/ironwood',
    configuration_extra_settings='',
)
```
6. Make sure the metadata was save in the new format:
```python
from instance.models.utils import ConsulAgent
ca = ConsulAgent(prefix='ocim/instances/<instasnce_id>')
ca._client.kv.get(ca.prefix)
```

### DIFF
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -495,9 +495,9 @@ class ConsulAgent(object):
             }
         }
 
-    def _get_put_data(self, updates):
+    def _get_keys_as_dict(self):
         """
-        Prepare and return the data to be used in the transaction put request
+        Retrieves a list of keys based on a prefix and returns a dictionary.
         """
         request = [
             {
@@ -517,24 +517,32 @@ class ConsulAgent(object):
         for entry in result['Results']:
             entry["KV"]["Value"] = self._tnx_decode(entry["KV"]["Value"])
             existing[entry["KV"]["Key"].split('/')[-1]] = entry["KV"]
+        return existing
 
+    def _get_put_data(self, updates):
+        """
+        Prepare and return the data to be used in the transaction put request
+        """
+        existing = self._get_keys_as_dict()
         # store existing keys which will be updated
         # as a list of dictionaries containing JSON encoded values
         put = []
         for key in set(updates.keys()) & set(existing.keys()):
             if updates[key] != existing[key]["Value"]:
-                put.append(self._tnx_update(
-                    self.prefix + key, updates[key], existing[key]["ModifyIndex"]))
+                put.append(
+                    self._tnx_update(
+                        self.prefix + key,
+                        updates[key],
+                        existing[key]["ModifyIndex"]
+                    )
+                )
 
         # store all new keys in the list
         for key in set(updates.keys()) - set(existing.keys()):
             put.append(self._tnx_set(self.prefix + key, updates[key]))
 
         # get or create the version number
-        if 'version' in existing:
-            version = existing['version']["Value"]
-        else:
-            version = 0
+        version = existing.get("version", {"Value": 0})["Value"]
 
         # increment the version number and use `txn` to update inside a transaction
         if put:
@@ -568,7 +576,7 @@ class ConsulAgent(object):
                 except consul.base.ClientError:
                     if attempt == num_retries:
                         raise
-                    time.sleep(5)
+                    time.sleep(5 * (attempt + 1))
         return version, bool(put)
 
     def get(self, key, index=False, **kwargs):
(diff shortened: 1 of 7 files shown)


instance/models/utils.py
```diff
@@ -517,24 +517,32 @@ def _get_put_data(self, updates):
...
         # store all new keys in the list
         for key in set(updates.keys()) - set(existing.keys()):
             put.append(self._tnx_set(self.prefix + key, updates[key]))
 
         # get or create the version number
-        if 'version' in existing:
-            version = existing['version']["Value"]
-        else:
-            version = 0
+        version = existing.get('version', {"Value": 0})["Value"]
```
viadanna: nit: this is mixing single and double quotes (seems it has been for a while)

instance/models/utils.py
```diff
@@ -608,6 +616,99 @@ def put(self, key, value, **kwargs):
...
+            payload = {
+                k: stored[k] for k in stored
+                if k not in value and
+                k != 'version'
+            }
+            updated = bool(updates)
+            payload['version'] = stored['version']
+            payload.update(value)
+        else:
+            payload = value
```
viadanna: Shouldn't `txn_data` be updated here as well in case of a PATCH is applied to a prefix that hasn't been updated?

instance/models/utils.py
```diff
@@ -608,6 +616,99 @@ def put(self, key, value, **kwargs):
...
+        stored['version'] = stored['version'] + 1
+        return self._client.kv.put(self.prefix, json.dumps(stored).encode('utf-8'))
+
+    def remove_dict(self):
+        """
+        Remove a config dict using this instance's prefix.
+        """
+        self._client.kv.delete(self.prefix)
+
+    def consolidate_values_to_dict(self):
```
viadanna: I expected this to be used to move from the old style to the new style, but found no references to this. Is that expected?

instance/models/utils.py
xirdneh: `txn_data` will only be available if (and only if) the instance's metadata is saved in the old style (e.g. `ocim/instances/<id>/slug`, `ocmi/instances/<id>/domains`, etc...).
If that's the case then we don't create the new style of metadata and we just use `self.txn_update` to use the old code and update the metadata accordingly.
If there's no `txn_data` and there's no `get_data` then this is a new instance and the metadata is created in the new style. Which means just passing the dictionary.
In case this is a PATCH to a metadata object in the new style, that would mean that there will be some value in `get_data` and no value in `txn_data` and we only update the updated keys on the dictionary.

instance/models/utils.py
xirdneh: @viadanna Yes, since the code is backwards compatible then we don't really need to run this.
This is just here as a utility function and meant to be used manually if needed.
Or depends on how we plan to move every instance to the new style.

instance/models/utils.py
viadanna: > If that's the case then we don't create the new style of metadata and we just use self.txn_update to use the old code and update the metadata accordingly.

But when we have the old style metadata, this will have get both txn_data and get_data, right? Won't it be updated here?

xirdneh: @viadanna If there's an old-style configuration in the metadata then only `txn_data` will get populated.
That's because `consul.kv.get` will match the key given exactly and in the old-style the instance's prefix is not stored in the kv store, e.g. `ocim/instsances/1/key1`, `ocim/instsances/1/key2`, etc...
So, if you have this in Ocim:
```json
[{
    "Key": "ocim/instances/123/key1",
    "Value": "value1",
}, {
    "Key": "ocim/instances/123/key2",
    "Value": "value2"
},{
    "Key": "ocim/instances/1234",
    "Value": "{\"key1\": \"value1\", \"key2\": \"value2\"}"
}
]```
And you run `consul.kv.get('ocim/instances/123')` then you will get an empty response because the exact key `ocim/instances/123` do not exists. If you run `consul.kv.get('ocim/instances/1234')` then the result will be:
```json
{
    "Key": "ocim/instances/1234",
    "Value": "{\"key1\": \"value1\", \"key2\": \"value2\"}"
}```
And if you do `consul.kv.get('ocim/instances/123', recurse=True)` then the response will be:
```json
[{
    "Key": "ocim/instances/123/key1",
    "Value": "value1",
}, {
    "Key": "ocim/instances/123/key2",
    "Value": "value2"
}]```

This is because consul's KV is built around the idea of prefixes in keys (when using `recurse=True`).

viadanna: > when using recurse=True

Oh, somehow I had in my mind both were fetched using that :man_facepalming: 

All's good.

xirdneh:
@viadanna I've added the custom domains to an integration test. Is that what you had in mind?

viadanna:
> @viadanna I've added the custom domains to an integration test. Is that what you had in mind?

Does this refers to https://github.com/open-craft/opencraft/pull/442#pullrequestreview-259220323 ?

PR #442, opened 2019-07-05
(BB-1402) Add support for setting up arbitrary subdomain.

xirdneh:
When creating an instance there's a few [default subdomains](https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/domain_names.py#L170) that are configured by default. In this context *configured* means the correct [configuration](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/load-balancer-v2/templates/backend.map.ctmpl) [line](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl#L105) in [haproxy](https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#4-use_backend)'s [config file](https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#4.2-http-request%20do-resolve) as well as having [cert-manager](https://github.com/open-craft/cert-manager/blob/master/manage_certs.py#L49) use let's encrypt to create the domain's certificate.
The way this configuration works for default subdomains is that Ocim creates a consul k/v pair with the necessary information (instance and domain list) and there are two different `consul watch` processes running that configure OC's haproxy and run `certbot` when there's a change on consul's side.
This PR leverages that by adding a text area in the instance's admin interface. In this text area a user can add multiple sub-domains (one in each line). Each subdomain is verified to be a subdomain of the main `lms` domain and added to the instance's consul configuration. The rest of the configuration (haproxy and cert-manager) works as expected.

# Testing instructions

1. Checkout this branch and run Ocim and consul locally.
2. Create an instance and query consul. You should see a domain config key like this one:
Screen Shot 2019-07-05 at 5 57 35 PM
3. Go to Django's admin interface and add multiple sub-domains:
Screen Shot 2019-07-05 at 5 57 03 PM
4. Save the updated instance.
5. Check consul for the updated domains list. Should look like this (note the extra sub-domains):
Screen Shot 2019-07-05 at 6 16 30 PM


instance/models/mixins/domain_names.py
```diff
@@ -179,6 +182,9 @@ def get_load_balanced_domains(self):
             self.internal_discovery_domain,
             self.internal_ecommerce_domain,
         ]
+        custom_domains = self.extra_custom_domains.split("\r\n")
+        print(custom_domains)
+        domain_names += [domain for domain in custom_domains if domain.endswith(self.internal_lms_domain)]
```
viadanna: Do you think it would be a good idea storing just the subdomains and concatenating the lms domain here instead?

instance/models/mixins/domain_names.py
```diff
@@ -179,6 +182,9 @@ def get_load_balanced_domains(self):
             self.internal_discovery_domain,
             self.internal_ecommerce_domain,
         ]
+        custom_domains = self.extra_custom_domains.split("\r\n")
+        print(custom_domains)
```
viadanna: Debugging leftovers?

xirdneh: Indeed, sorry about that.

instance/models/mixins/domain_names.py
xirdneh: @viadanna We could do that, but then we should probably change how we save the rest of the subdomains, i.e. `self.internal_*_domain` and `self.external_*_domain`.
Or else some of the domains pertaining to an instance will be the complete name and some others would be the name minus the lms domain.

viadanna: Alright, makes sense. On second thought, this gives more flexibility as well.

xirdneh:
@viadanna Yes, the added [integration test](https://github.com/open-craft/opencraft/pull/442/files#diff-ea586dca3f707d2466b9230bc4c84f0dR273) was for this review. Sorry about this.

instance/migrations/0117_auto_20190416_2130.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2019-04-16 21:30
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
viadanna: Can you merge this migration along the previous one and keep a reasonable name for it?

instance/tests/integration/integration_instance.py
```diff
@@ -317,6 +349,7 @@ def test_spawn_appserver(self):
             self.assert_secret_keys(instance, appserver)
             self.assert_lms_users_provisioned(user, appserver)
             self.assert_theme_provisioned(instance, appserver, application)
+        self.assert_load_balanced_domains(instance)
```
viadanna: :+1:

PR #448, opened 2019-07-29
[BB-1481] Fix failing unit test

lgp171188:
The unit test started failing after the `PRELIMINARY_PAGE_HOSTNAME` was
set in the CircleCI environment and the test expected it to not be
set. The fix explicitly makes the value `None` so that the test doesn't fail.


lgp171188:
This is related to https://github.com/open-craft/opencraft/pull/446

PR #449, opened 2019-08-08
BB-1030 Reorganize system dependencies

viadanna:
This PR reorganizes Debian/Ubuntu dependencies to get both CircleCI and Vagrant devstack to install all requirements.

This was broken in #426

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -43,23 +43,21 @@ jobs:
       - run:
           name: Install Dependencies
           command: |
+            sed -i 's/libmysqlclient/default-libmysqlclient/' debian_packages.lst
+            make install_system_dependencies
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libmysqlclient.so /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18
+            bin/install-supported-firefox
             python3 -m venv venv
             . venv/bin/activate
             pip install --upgrade pip
             pip install --upgrade virtualenv
-            sudo apt-get update; sudo apt-get install python2.7-dev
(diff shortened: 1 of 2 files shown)


viadanna:
@swalladge This fixes the devstack, can you review it?

circle.yml
```diff
@@ -43,23 +43,21 @@ jobs:
       - run:
           name: Install Dependencies
           command: |
+            sed -i 's/libmysqlclient/default-libmysqlclient/' debian_packages.lst
+            make install_system_dependencies
+            sudo ln -s /usr/lib/x86_64-linux-gnu/libmysqlclient.so /usr/lib/x86_64-linux-gnu/libmysqlclient.so.18
+            bin/install-supported-firefox
```
swalladge: It would be nice if the logic for installing firefox on local devstack was also moved to install-supported-firefox sometime. It's currently fragmented between this script and the makefile.

viadanna: I couldn't agree more @swalladge 

I moved all debian/ubuntu packages installation to the Makefile with that in mind, let's see if I end up having time this week to move Firefox installation as well.

swalladge: Maybe this can be done later in another ticket in the interest of unbreaking master sooner?

debian_packages.lst
```diff
@@ -1,10 +1,14 @@
...
+postgresql-client
 python-dev
+python2.7-dev
 python3-pip
 redis-server
 subversion
-xvfb
-postgresql-client
+unzip
+xvfb
```
lgp171188: @viadanna, it is a good idea to add a newline at the end of the file..

PR #451, opened 2019-08-08
[BB-1537] Migrate to New Relict Alerts for monitoring alert notifications

lgp171188:
With the legacy New Relic Synthetics notification reaching EOL in September 2019 and us having issues with not getting notified for the failing heartbeat checks, a migration to the New Relic Alerts is required and this PR implements the changes needed for it.

**Testing instructions**:
Though the production environment is the ideal place to test this, it is possible to test the changes in the stage environment with some configuration and tweaks.
* Create a new `.env.<suffix>` file in the stage server by copying the existing `.env` file and add the variable configuring the New Relic API admin key. It's okay to use the production key but ensure that all the alert policies, notification channels, alert conditions and monitors created for testing purposes are deleted after testing.
* Run `honcho -e <the created .env.* file> run python3 manage.py shell_plus` command to start the Django shell.
* Play around with the new functions added in `instance/newrelic.py` for creating/deleting New Relic Alert policies, email notification channels, adding them to policies, creating alert conditions for existing monitors etc. and verify that everything works as expected.
* Apply the migrations and test the following scenarios through Ocim.
  * Open the Django shell using the ``honcho -e <the created .env.* file> run python3 manage.py shell_plus` command.
  * Add `NewRelicEmailNotificationChannel` instances for the `settings.ADMINS` email addresses with the `shared` flag set to `True`. Note that this is likely to send a few emails to these addresses.
  * Run the `enable_monitoring()` method on an existing, successfully provisioned instance and verify that everything works okay and the monitoring is enabled properly.
  * Add some additional monitoring emails and repeat the previous step and verify the result.
  * Run the `disable_monitoring()` method and verify that all the resources created by `enable_monitoring()` are deleted. Verify that the shared email addresses are not removed from New Relic or deleted from the database.
  * Try a series of `enable_monitoring()` and `disable_monitoring()` calls in random order and verify that nothing unexpected happens. Ensure that at the end the `disable_monitoring()` method is called to clean up any test resources created.
  * Update the `.env` file with the New Relic admin API key (get it from production or use `.env.newrelic` on stage). It is a good idea to inform the team not to create new sandboxes or spawn new appservers while testing this as they will end up creating more resources (including the monitors which may cost money)  to be cleaned up.
  * Restart the app by killing the `make run` process and restarting it.
  * Restart the Django shell as well.
  * Create a new sandbox instance, spawn a new appserver and verify that every thing works okay. Confirm and verify that the monitoring has been set up properly.
  * Disable monitoring on the new sandbox and verify that it works okay. Confirm and verify that all the monitoring resources are deleted.
  * Archive the instance and verify that the cleanup is performed and nothing related to that instance is left undeleted.
  * Also verify that the `NewRelicEmailNotificationChannel` instances with the `shared` flag set to `True` are not deleted when disabling the monitoring and archiving instances.

* All the new models use the integer type for the primary key to match the type returned by te New Relic API. This could require changes if New Relic changes the data types.
* This might require additional testing scenarios not listed in the testing instructions above.


instance/models/mixins/openedx_monitoring.py
```diff
@@ -80,12 +140,25 @@ def disable_monitoring(self):
         for monitor in self.new_relic_availability_monitors.all():
             monitor.delete()
 
+        try:
+            alert_policy = NewRelicAlertPolicy.objects.get(instance=self)
+            for channel in alert_policy.email_notification_channels.filter(shared=False):
```
giovannicimolin: Why are we deleting all `email_notification_channels`? Won't this fail with `ProtectedError` if we try to delete a shared notification channel (like ops@opencraft.com)?

giovannicimolin: The `_add_emails` function above reuses existing notification channels. Shouldn't we try to delete, but skip if the channel is shared?

instance/models/mixins/openedx_monitoring.py
```diff
@@ -44,33 +45,92 @@ def enable_monitoring(self):
...
         already_monitored_urls = set()
 
         for monitor in self.new_relic_availability_monitors.all():
             url = newrelic.get_synthetics_monitor(monitor.pk)['uri']
             if url in urls_to_monitor:
                 already_monitored_urls.add(url)
+                # Check if the monitor has an associated alert condition
+                # and create one if it doesn't. This helps when the alert condition
+                # wasn't created in a previous invocation due to some issue.
+                if not monitor.new_relic_alert_conditions.exists():
```
giovannicimolin: This seems to take care of previously provisioned instances that used the old Synthetics. :+1:

instance/models/mixins/openedx_monitoring.py
lgp171188: @giovannicimolin, the queryset filters on `shared=False` and that should take care of this.

instance/models/mixins/openedx_monitoring.py
lgp171188: @giovannicimolin, this also takes care of the scenario where the alert policy creation failed in a run and we retry.

lgp171188: @giovannicimolin, I have added a unit test to verify that this works for the instances that already have the monitoring enabled using the old code.

lgp171188:
@giovannicimolin, I have added a couple of unit tests to verify that these changes work okay for existing instances with monitoring enabled using the existing code (before the changes in this PR) and also for verifying that if the `enable_monitoring()` method fails due to some New Relic API errors in the new API calls added, subsequent retries will not ignore the creation of any resources.

lgp171188:
@giovannicimolin, I have created a follow-up ticket for documenting this setup. Will merge and deploy the changes now.

instance/tests/models/test_openedx_monitoring_mixins.py
```diff
@@ -93,11 +104,86 @@ def test_enable_monitoring(self, additional_monitoring_emails, expected_monitor_
...
+        """
+        Check that the enable_monitoring doesn't skip the creation of monitoring resources when a previous attempt
+        to enable monitoring failed with an error.
+        """
+        instance = OpenEdXInstanceFactory()
+
+        class CustomException(Exception):
+            pass
+
+        def check(instance):
```
clemente: @giovannicimolin this function isn't very clear to me. It seems that the first invocation will also throw `CustomException` (it seems that all invocations will do), so why are we using a broad „catch Exception“ as if we didn't know whether it will succeed or not? And if we expect 1+10 failures, why not directly loop 11 times? (or 10, to round it).
I also added a doctest [in my branch](https://github.com/open-craft/opencraft/pull/436/commits/892c199fba57e5a918de6b2e52218a881a5ca98d) (still untested). I'll write some questions in the JIRA task.

lgp171188: @clemente, the idea behind this test is - if some step in enabling the monitoring, particularly in the new changes added in this PR, for an instance fails, retrying the same should not skip the failed step. The code uses only the data in the models for checking if a New Relic resource (alert policy, alert condition, notification channel) exists or not. While testing the code changes in this PR, I found that I had introduced a bug where the model instance was getting created before the actual New Relic resource was created and when the New Relic API call failed due to some reason, the retry skipped the code to try creating it again because the model instance exists.

Since that code was confusing, I added this test to ensure that if a New Relic API call fails due to some reason, a retry will still traverse through the same code path on every retry till it succeeds.

PR #452, opened 2019-08-09
SE-713 temporary PR testing browser tests

swalladge:
This is a temporary PR for testing browser tests in CI.

Do not merge.


swalladge:
Determined that the selenium 3 and geckodriver isn't working for our tests currently. Have reverted to selenium 2 on https://github.com/open-craft/opencraft/pull/436 and will go from there.

PR #454, opened 2019-08-13
BB-1576: Add variables required to gate playbook reconfiguration

giovannicimolin:
This PR adds a flag to gate load balancer reconfiguration on Ocim.

**Testing instructions**:
1. Checkout this branch on your Ocim devstack
2. Run tests related to this change and check that they pass.
```
make test.one instance.tests.models.test_openedx_appserver.OpenEdXAppServerTestCase.test_make_active_no_load_balancer_reconfiguration
make test.one instance.tests.models.test_openedx_appserver.OpenEdXAppServerTestCase.test_make_active
make test.one instance.tests.models.test_load_balanced_mixin
```
3. Configure Ocim with the `.env` used for development and add `DISABLE_LOAD_BALANCER_CONFIGURATION = True`.
4. Run Ocim Shell and check that the setting is correctly set.
5. Create a new instance:
```
production_instance_factory(name='123', sub_domain='bb1576-test')
```
6. Run Ocim with `make run.dev` and spawn a new appserver. Check that the following log appears on the instance log:
```
Direct load balancer reconfiguration disabled. Skipping 123 (bb1576-test.plebia.net) configuration...
```

Make sure you clean up the instance afterwards.

**Reviewer:**
- [x] @Agrendalath 

FYI @SSPJ

### DIFF
--- a/instance/tests/models/test_openedx_appserver.py
+++ b/instance/tests/models/test_openedx_appserver.py
@@ -28,7 +28,7 @@ import novaclient
 import requests
 import responses
 import yaml
-from ddt import ddt, data, unpack
+from ddt import ddt, data
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core import mail as django_mail
--- a/instance/tests/models/test_openedx_appserver.py
+++ b/instance/tests/models/test_openedx_appserver.py
@@ -28,7 +28,7 @@ import novaclient
 import requests
 import responses
 import yaml
-from ddt import ddt, data
+from ddt import ddt, data, unpack
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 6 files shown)


instance/tests/models/test_openedx_appserver.py
```diff
@@ -28,7 +28,7 @@
 import requests
 import responses
 import yaml
-from ddt import ddt, data
+from ddt import ddt, data, unpack
```
Agrendalath: Just a nit: are we actually using `unpack` anywhere?

giovannicimolin: I tried to :stuck_out_tongue:, in the end I needed a setting override and created a new test for it.
I will update the PR and remove this.

PR #457, opened 2019-08-23
Upgrade PyYAML and remove some unused dependencies

clemente:
This is a follow-up of https://github.com/open-craft/opencraft/pull/436. It upgrades some packages with vulnerabilities (according to github):
- it upgrades PyYAML, which wasn't possible before due to version conflicts, but now it is
- it removes ansible and its dependencies, since it wasn't initially in requirements.txt and it was added by mistake
- it does some minor upgrades to stay up to date (`pip-upgrade -U`)

Testing:
- similar to the cited PR
- CI tests should pass
- deploy to stage and test
- deploy to production


swalladge:
:+1:

- [x] I tested this:
  - provisioned a local devstack (with the fixes from #449) and verified that it completed successfully
  - I _didn't_ do a smoke test for yaml loading; that can happen on stage. hopefully since unit tests pass, everything will be ok.
- [x] I read through the code

clemente:
Tested in production. Thanks!

PR #458, opened 2019-08-27
[BB-1604] Add delete_archived management command

viadanna:
Description
=========

This pull requests adds a `delete_archived` management command.

The command will fetch all archived instances older than the given number of months, and after user confirmation, delete these instances.

Testing
----------

1. Create a couple instances in your devstack and archive them:
```
instance = instance_factory(name="Sandbox instance", sub_domain="sandbox")
instance.ref.is_archived = True
instance.ref.save()
```
2. Modify the `modified` attribute of one of the instance references:
```
from datetime import timedelta as td
instance.ref.modified -= td(days=365)
instance.ref.save(update_modified=False)
```
3. Run the management command and check it finds the correct instances
```
$ honcho run python manage.py delete_archived 0
$ honcho run python manage.py delete_archived 1
$ honcho run python manage.py delete_archived 24
(etc)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -858,6 +858,15 @@ before the redeployment command. For example:
 
     HUEY_QUEUE_NAME=opencraft_low_priority make manage "instance_redeploy ..."
 
+**`delete_archived`**: Deletes archived instances older than the given number
+of months. For example, to delete instances archived more than one year ago:
+
+    make manage "delete_archived 12"
+
+This command will actually check when the InstanceReference model was last
+modified, so it might miss archived instances that were modified after being
+archived.
+
 Databases
 ---------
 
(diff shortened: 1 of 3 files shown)


viadanna:
Thanks for the review @giovannicimolin 

I'll update the README :+1:

viadanna:
@giovannicimolin Added a section to the README and now it will also log the instance name to STDERR in case of failure when running in automated mode `-y`

[This is the diff](https://github.com/open-craft/opencraft/pull/458/files/5c3e404e358fba85ef13fd9096cc7146eeacc85e..134b5e5cf94831c1b60e474ac9d57f7a67db4c21)

PR #459, opened 2019-08-29
SE-1463 revert changes to EDXAPP_WORKERS

swalladge:
14 and 7 were way too high to use as safe defaults, resulting in small
 instances locking up.

The changes that we're reverting were introduced in https://github.com/open-craft/opencraft/pull/437

**Reviewer**:

- [ ] @viadanna


swalladge:
@lgp171188 @viadanna I also reverted the changes to the max_requests setting (back to 5000), now that we are defaulting back to a lower number of workers.

swalladge:
Hmm github didn't let me merge to master. Will need to wait until the CI finishes.

PR #463, opened 2019-09-10
SE-1463 Update default extended /heartbeat checks

swalladge:
- forum heartbeat path changed in  https://github.com/edx/edx-platform/commit/2839851bde247a3d33db1133048e0442341204f2
- need to add default celery heartbeat check here too since we're
  overriding it from the default HEARTBEAT_EXTENDED_CHECKS.

**JIRA tickets**: 

**Dependencies**: None

**Sandbox URL**: TBD

**Merge deadline**: None

**Testing instructions**:

1. run ocim from this branch
2. launch a new open edx instance
   - default extra settings
   - (openedx revision newer than https://github.com/edx/edx-platform/commit/2839851bde247a3d33db1133048e0442341204f2 )
   - configuration branch https://github.com/edx/configuration/pull/5400
3. visit /heartbeat?extended=true on the new instance and verify that both the
   forum and celery heartbeat checks are present and successful.

- ~this will break new sandboxes running edx-platform master prior to  https://github.com/edx/edx-platform/commit/2839851bde247a3d33db1133048e0442341204f2 (commit in april 2019), that successfully override the default EDXAPP_LMS_ENV_EXTRA (anything prior to https://github.com/edx/edx-platform/commit/2108b69ff15cca7c9f44f08c77a9f33d1ca2cad1 in march), and don't manually override HEARTBEAT_EXTENDED_CHECKS (from extra settings).~ conditional checks added so this doesn't happen

**Reviewers**
- [ ] @viadanna

**Settings**
```yaml
```

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -51,14 +51,24 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             # edxapp
             "EDXAPP_PLATFORM_NAME": self.instance.name,
             "EDXAPP_SITE_NAME": self.instance.domain,
+
+            # Available as ENV_TOKENS in the django setting files.
+            # Note: use EDXAPP_CMS_ENV_EXTRA and/or EDXAPP_LMS_ENV_EXTRA; don't use
+            # EDXAPP_ENV_EXTRA as the other two overwrite this.
+            "EDXAPP_CMS_ENV_EXTRA": {
+                "ADDL_INSTALLED_APPS": [
+                    "openedx.core.djangoapps.heartbeat",
+                ],
+            },
             "EDXAPP_LMS_ENV_EXTRA": {
                 "ADDL_INSTALLED_APPS": [
                     "openedx.core.djangoapps.heartbeat",
                 ],
                 "HEARTBEAT_EXTENDED_CHECKS": [
-                    "lms.lib.comment_client.utils.check_forum_heartbeat",
+                    "openedx.core.djangoapps.heartbeat.default_checks.check_celery",
                 ],
             },
+
             "EDXAPP_LMS_NGINX_PORT": 80,
             "EDXAPP_LMS_SSL_NGINX_PORT": 443,
             "EDXAPP_LMS_BASE_SCHEME": 'https',
@@ -178,11 +188,6 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             # Misc
             "EDXAPP_LANG": 'en_US.UTF-8',
             "EDXAPP_TIME_ZONE": 'UTC',
-
-            # Available as ENV_TOKENS in the django setting files.
-            "EDXAPP_ENV_EXTRA": {
-                "LANGUAGE_CODE": 'en',
-            },
 
             # Features
             "EDXAPP_FEATURES": {
@@ -401,6 +406,14 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"] = {
                 "PRIVACY": self.privacy_policy_url,
             }
+
+        # use the new forum heartbeat path if on master. Note: will need to
+        # update this after the release after ironwood has landed.
+        if self.openedx_release == 'master':
+            forum_hb_path = "openedx.core.djangoapps.django_comment_common.comment_client.utils.check_forum_heartbeat"
+        else:
+            forum_hb_path = "lms.lib.comment_client.utils.check_forum_heartbeat"
+        template["EDXAPP_LMS_ENV_EXTRA"]["HEARTBEAT_EXTENDED_CHECKS"].append(forum_hb_path)
 
         return template
 


instance/models/mixins/openedx_config.py
```diff
@@ -56,7 +56,8 @@ def _get_configuration_variables(self):
                     "openedx.core.djangoapps.heartbeat",
                 ],
                 "HEARTBEAT_EXTENDED_CHECKS": [
-                    "lms.lib.comment_client.utils.check_forum_heartbeat",
```
lgp171188: @swalladge, since the path was changed only after the Ironwood release, removing this will break all the instances running Ironwood or an older version. The way we usually deal with it is to add a conditional checking for all the releases which use this and set the old value if it is true or if not, set the new value.

instance/models/mixins/openedx_config.py
```diff
@@ -56,7 +56,8 @@ def _get_configuration_variables(self):
...
+                    "openedx.core.djangoapps.django_comment_common.comment_client.utils.check_forum_heartbeat",
+                    "openedx.core.djangoapps.heartbeat.default_checks.check_celery",
```
lgp171188: @swalladge, the `check_celery` task was intentionally removed when we added this feature because there was an import error which was causing the task and hence the heartbeat checks to fail. (details [here](https://github.com/open-craft/opencraft/pull/340#issuecomment-412379171)). Can you check and confirm if that issue is now resolved before enabling this? If it is resolved, you will have to check in which release/branch this was resolved and add conditionals if there are any branches that we run in production which don't have the fix.

swalladge: @lgp171188 That import error happens when `openedx.core.djangoapps.heartbeat` is missing from the django installed apps. That looks like it was already fixed as part of https://github.com/open-craft/edx-platform/pull/135. However it is broken since ironwood because LMS_ENV_EXTRA doesn't get passed to lms.yml, where ironwood reads from. This is the exact error that is causing half the sandboxes extended heartbeat checks to be failing right now.

swalladge: I guess what I'm trying to say above is that it was resolved at the time, but has been broken again since ironwood for different reasons. https://github.com/edx/configuration/pull/5400 should fix that again.

instance/models/mixins/openedx_config.py
swalladge: Please see the latest changes - I added conditional checking (untested). It's a bit messy because it appears the path changed between ironwood.1 and ironwood.2. Is this the kind of conditional checking you'd like to see?

I guess a neat alternative would be to disable forum heartbeat checking by default, and we can enable it manually for instances where we need it?

swalladge:
I moved the default settings that were set on EDXAPP_ENV_EXTRA to both EDXAPP_CMS_ENV_EXTRA and EDXAPP_LMS_ENV_EXTRA, because EDXAPP_ENV_EXTRA is overridden/lost for lms or studio when setting LMS_ENV_EXTRA or CMS_ENV_EXTRA respectively.

instance/models/mixins/openedx_config.py
lgp171188: >  However it is broken since ironwood because LMS_ENV_EXTRA doesn't get passed to lms.yml, where ironwood reads from. This is the exact error that is causing half the sandboxes extended heartbeat checks to be failing right now.

@swalladge, since the issue with the `LMS_ENV_EXTRA` and `CMS_ENV_EXTRA` variables is only in the `master` branch of the upstream after Ironwood branches were cut (I assumed that `since Ironwood` means this) and the upstream PR branches are usually based on it, I don't think there is anything that we can do till it is fixed upstream.

instance/models/mixins/openedx_config.py
```diff
@@ -403,6 +402,13 @@ def _get_configuration_variables(self):
                 "PRIVACY": self.privacy_policy_url,
             }
 
+        # use the new forum heartbeat path if newer than ironwood.1
+        if 'ironwood' in self.openedx_release and 'ironwood.1' not in self.openedx_release:
```
lgp171188: Continuing the discussion in https://github.com/open-craft/opencraft/pull/463#discussion_r322669445 here.

>  It's a bit messy because it appears the path changed between ironwood.1 and ironwood.2. Is this the kind of conditional checking you'd like to see?

@swalladge, I checked the files and the commit history in the upstream `open-release/ironwood.master` branch and also our `opencraft-release/ironwood.2` branch and could not find the change to move the `check_forum_heartbeat` function from `lms/lib/comment_client/utils.py` to `openedx/core/djangoapps/django_comment_common/comment_client/utils.py`. So it looks like the path change is needed only for the master branch and the PRs based on it as the change was not backported to Ironwood for obvious reasons. Did I miss something?

swalladge: Ah true. I had just assumed that because that change was in master before ironwood.2 was released, that it would be part of ironwood.2. That's not the case though. The change to only read from the yml files instead of json is only in master too.

instance/models/mixins/openedx_config.py
swalladge: Yep, hopefully edx/configuration#5400 will be merged quickly and then we won't need to worry. :)

swalladge:
@lgp171188 I updated the conditional based on what you discovered, so now it only uses the new forum heartbeat path if openedx_release is set to master.

lgp171188:
> @lgp171188 sandboxes created for testing (on ocim stage, with this PR's branch deployed):

@swalladge, I tested the Ironwood sandbox and can see that configuration is set properly and that the extended heartbeat checks are passing.

However, since the provisioning of the other sandbox for testing the `master` branch is failing due to some unrelated issue, I am unable to verify the fix on it. I have spawned a new appserver now to check if the issue has been resolved now.

swalladge:
@lgp171188 Argh that's the issue from [SE-1553](https://tasks.opencraft.com/browse/SE-1553)... I'll manually fix the migrations and hopefully that will help. :)

lgp171188:
@swalladge, the latest appserver was provisioned successfully and I was able to verify that the changes are working. That said, I did get a `False` for `celery` when I loaded the extended heartbeat URL for the first time using the Appserver's IP address. After that it worked fine. This is something to keep an eye on after we deploy this to production.

lgp171188:
@swalladge, this is ready for merge now that the CI checks have passed.

swalladge:
Thanks @lgp171188 :) 

> That said, I did get a False for celery when I loaded the extended heartbeat URL for the first time using the Appserver's IP address. 

Yes I have observed the same thing - I believe it's because the celery workers take a while to start up and become operational.

swalladge:
@lgp171188 Just a heads up that I haven't merged this yet because the flakey integration tests haven't passed yet - they started rebuilding since I squashed and force-pushed.

lgp171188:
> @lgp171188 Just a heads up that I haven't merged this yet because the flakey integration tests haven't passed yet - they started rebuilding since I squashed and force-pushed.

@swalladge, ah no! I meant using the `Squash and merge` button in the PR to avoid re-running CI checks. Let's wait now for the CI checks to pass: :crossed_fingers:

swalladge:
Yay at last! :)

> ah no! I meant using the Squash and merge button in the PR

@lgp171188 I always prefer doing operations locally so I know exactly what is happening, and so that the resulting commit is still signed by my pgp key.

PR #466, opened 2019-09-24
BB-1682: Deprecate swift features

giovannicimolin:
This PR adds some deprecation warnings, comments and removes mentions about `swift` from the user interface.

**Testing instructions:**
1. Check that the tests run correctly.
2. Provision an instance with S3 storage and check that the Instance Configuration page correctly states when S3 is provisioned and deprovisioned.

**Reviewer:**
- [x] @xitij2000


backup_swift/tasks.py
```diff
@@ -47,6 +47,7 @@ def do_backup_swift():
     """
     Perform full swift backup sequence.
     """
+    logger.warning('SWIFT storage is not maintained and will be dropped in a future version.')
```
xitij2000: nit: I think there is a specific system for this in python:
``
import warnings
warnings.warn(message, DeprecationWarning)
```

File: instance/management/commands/migrate_swift_to_s3.py [hunk 2]
```diff
@@ -334,6 +334,7 @@ def handle(self, *args, **options):
         """
         Finds all instances and migrates them.
         """
+        self.logger.warning('SWIFT storage is not maintained and will be dropped in a future version.')
```
xitij2000: nit: Since this command was specifically designed to migrate from a deprecated system to a non-depricated system I don't think the command itself should be deprecated. 

Although... it will be removed, so maybe?

backup_swift/tasks.py
giovannicimolin: This warning doesn't use the module's logging system, so I decided to stick with a warning.

instance/management/commands/migrate_swift_to_s3.py
giovannicimolin: It doesn't work without the django models, so it's deprecated too. :grin:

instance/models/mixins/storage.py
```diff
@@ -179,6 +181,7 @@ class Meta:
             ' When set empty, the bucket is created in the default region us-east-1.'
         )
     )
+    s3_provisioned = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
giovannicimolin: I ended up taking this out since properties aren't being rendered on templates.

PR #467, opened 2019-09-26
BB-1714 Prevent Consul collision on integration tests

viadanna:
Intro
===

This PR adds a random ID to the OpenEdxInstance created during integration tests. This is necessary as the instance ID is used as a prefix to Consul data and multiple instances are created in parallel.

Testing
---

1. Checkout this branch and start the devstack.
2. Run the steps used by CircleCI to run Consul locally:
```bash
sudo wget -P /tmp https://releases.hashicorp.com/consul/1.2.1/consul_1.2.1_linux_amd64.zip
sudo unzip /tmp/consul_1.2.1_linux_amd64.zip -d /usr/local/bin
consul agent -dev
```
3. Fire up the shell and use the modified factory method to create an instance:
```python
from instance.tests.integration.factories.instance import OpenEdXInstanceFactory                                                
i = OpenEdXInstanceFactory()
print(i.id)
```
4. Check the Consul log for the prefix containing the random ID.
```
    2019/09/26 21:09:00 [DEBUG] http: Request GET /v1/kv/ocim/instances/1488912035 (106.941µs) from=127.0.0.1:42116
```
5. Stop the Consul process and try to get an id, making sure it works:
```
from instance.tests.integration.factories import instance
print(instance._get_unused_id())
```

### DIFF
--- a/instance/tests/integration/factories/instance.py
+++ b/instance/tests/integration/factories/instance.py
@@ -39,9 +39,15 @@ def _get_unused_id():
     Find an unused ID for Consul
     """
     while True:
-        i = randint(2 ** 30, 2 ** 31)
-        if not ConsulAgent(prefix=settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=i)).get(''):
-            return i
+        try:
+            uid = randint(2 ** 30, 2 ** 31)
+            current = ConsulAgent(
+                prefix=settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=uid)
+            ).get('')
+            assert current is not None
+        except (ConnectionError, AssertionError):
+            # Returns when current is None or Consul unreachable
+            return uid
 
 
 # Classes #####################################################################
--- a/instance/tests/integration/factories/instance.py
+++ b/instance/tests/integration/factories/instance.py
@@ -45,7 +45,7 @@ def _get_unused_id():
                 prefix=settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=uid)
             ).get('')
             assert current is not None
-        except Exception:
+        except (ConnectionError, AssertionError):
             # Returns when current is None or Consul unreachable
             return uid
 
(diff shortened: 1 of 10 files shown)


instance/tests/integration/factories/instance.py
```diff
@@ -23,12 +23,25 @@
...
+from instance.models.utils import ConsulAgent
+
+
+def _get_unused_id():
+    """
+    Find an unused ID for Consul
+    """
+    while True:
+        i = randint(2 ** 30, 2 ** 31)
+        if not ConsulAgent(prefix=settings.CONSUL_PREFIX.format(ocim=settings.OCIM_ID, instance=i)).get(''):
```
giovannicimolin: If  `Consul` isn't available, this errors out with 
```
HTTPConnectionPool(host='127.0.0.1', port=8500): Max retries exceeded with url: /v1/kv/ocim/instances/1183429413 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f135e757208>: Failed to establish a new connection: [Errno 111] Connection refused',))
```

Can you catch this exception and return the first generated number when this happens?

viadanna: Sure! I didn't expect it to be used without Consul but yeah, better handle that gracefully.

instance/tests/integration/factories/instance.py
```diff
@@ -23,12 +23,31 @@
...
+from instance.models.utils import ConsulAgent
+
+
+def _get_unused_id():
+    """
+    Find an unused ID for Consul
+    """
+    while True:
+        try:
+            uid = randint(2 ** 30, 2 ** 31)
```
giovannicimolin: This line should be placed outside the `try-except` statement as context isn't passed to the exception handler when an error happends, yelding:
```
UnboundLocalError: local variable 'uid' referenced before assignment
```

viadanna: That's weird, I couldn't hit that error, could you explain how you triggered that? E.g:
```
In [1]: try: 
   ...:     uid = 1000 
   ...:     assert False 
   ...: except: 
   ...:     print(uid) 
   ...:                                                                         
1000
```

PR #470, opened 2019-10-01
Use Loader=SafeLoader for PyYAML

SSPJ:
See https://msg.pyyaml.org/load for why this change is happening.

This is a drive-by PR not related to any ticket.

### DIFF
--- a/instance/ansible.py
+++ b/instance/ansible.py
@@ -50,8 +50,8 @@ def load_yaml(string):
     """
     if string.startswith('@'):
         with open(string[1:]) as f:
-            return yaml.load(f)
-    return yaml.load(string)
+            return yaml.load(f, Loader=yaml.SafeLoader)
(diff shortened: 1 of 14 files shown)


clemente:
:+1: If tests pass we can merge this. No need to deploy it separately, I think, since it's a trivial change that doesn't change the behaviour.

  - [X] I tested this: I only tested this in local. The code execution mentioned at https://msg.pyyaml.org/load is possible with python2 but doesn't work in python3 (`expected a class, but found <class 'builtin_function_or_method'>`), so Ocim isn't vulnerable. In python3, it's using SafeLoader by default (if I explicitly ask to use UnsafeLoader, the code execution works in python3 too)
  - [X] I read through the code. This change doesn't change the behaviour, it just removes the warning by explicitly mentioning the default behaviour. This is now the recommended way to parse YAML. We don't need to abstract it further. I checked that all uses were replaced
  - [X] Includes documentation: none needed

PR #473, opened 2019-10-03
BB-1726 Simplify integration tests

viadanna:
Description
========

This PR merges all integration tests that spawn an AppServer into a single test. Other integration tests that don't require an AppServer are run on a separate container.

The objective here is simplifying and reducing time wasted trying to get a green build.

~As a test, the integration tests aren't required for merging, let's see how it goes.~

Testing
-------

1. Just watch CircleCI 🙂


circle.yml
```diff
@@ -182,22 +177,7 @@ workflows:
...
-          requires:
-            - python_unit_tests
-            - browser_tests
-            - js_tests
-      - build:
-          name: integration_tests_group_4
-          tests_group: 4
-          tests_type: test.integration
-          requires:
+            - quality_migrations_tests
```
giovannicimolin: Shouldn't we gate the integration tests with `python_unit_tests`? They will catch any issues before an integration test starts running.

instance/tests/integration/integration_instance.py
```diff
@@ -439,7 +436,7 @@ def test_ansible_failure(self, git_checkout, git_working_dir):
         self.assertEqual(appserver.status, AppServerStatus.ConfigurationFailed)
         self.assertEqual(appserver.server.status, ServerStatus.Ready)
 
-    @skipIf(TEST_GROUP is not None and TEST_GROUP != '4', "Test not in test group.")
+    @skipIf(TEST_GROUP is not None and TEST_GROUP != '2', "Test not in test group.")
```
giovannicimolin: With a single integration test running, do we need these decorators? If not, can you remove them?

circle.yml
viadanna: Not sure I follow. The integration tests do required python unit tests.

instance/tests/integration/integration_instance.py
viadanna: I'm keeping two test groups. One for the test that spawn an appserver and another for those that don't. This has the potential to make retrying the workflow faster if the 1st integration test group passes.

circle.yml
giovannicimolin: Oh yeah, the diff on github was misleading.

PR #474, opened 2019-10-04
[SE-1603] Mock Consul for unit tests

SSPJ:
**Testing Instructions**

Use this in `.env.test`:

```
CONSUL_SERVERS=fake-consul-dns.name
DEFAULT_LOAD_BALANCING_SERVER=ubuntu@fake-load-balancer
CONSUL_ENABLED=true
DISABLE_LOAD_BALANCER_CONFIGURATION=true
```

And run `make test.unit` inside the `opencraft` directory in a vagrant machine.

**Author's notes and concerns**

`pr_watch.tests.test_tasks.TasksTestCase` has failing tests for me locally, even on master, I suspect because I have something misconfigured.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -20,3 +20,5 @@ DEFAULT_LOAD_BALANCING_SERVER='ubuntu@haproxy-test.fake.domain'
 LOAD_BALANCER_FRAGMENT_NAME_PREFIX='opencraft-'
 DEFAULT_RABBITMQ_API_URL='https://admin:password@rabbitmq.example.com:15671'
 DEFAULT_INSTANCE_RABBITMQ_URL='amqps://rabbitmq.example.com:5671'
+CONSUL_ENABLED: false
+DISABLE_LOAD_BALANCER_CONFIGURATION: false
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -156,7 +156,7 @@ class OpenEdXInstance(
                 'Direct load balancer reconfiguration disabled. No haproxy '
                 'configuration fragment and backend map will be generated.'
             )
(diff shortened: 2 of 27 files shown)


clemente:
@SSPJ Great!
I was going to post an exhaustive review. But then I found only this test which fails due to the new changes:

```
honcho -e .env.test run ./manage.py test instance.tests.management.test_instance_redeploy.InstanceRedeployTestCase

ERROR: test_default_arguments (instance.tests.management.test_instance_redeploy.InstanceRedeployTestCase)
  File "/home/vagrant/.virtualenvs/opencraft/lib/python3.5/site-packages/testfixtures/utils.py", line 83, in patched
    return func(*args, **keywargs)
TypeError: test_default_arguments() missing 1 required positional argument: 'mock_consul'

```

Could you fix that one (maybe a `with` block, if nothing helps?) and then I'll post the full review? (The rest looks fine).

SSPJ:
@clemente It passes for me now. It seems the `LogCapture()` decorator doesn't play nice with other decorators, but a with block worked fine.

Makefile
```diff
@@ -132,6 +132,7 @@ ifneq ($(wildcard .env.integration),)
 	$(HONCHO_COVERAGE_INTEGRATION) --pattern=integration_*.py
 else ifdef OPENSTACK_USER
 	echo -e "\nRunning integration tests with credentials from environment variables..."
+	echo -e "\n and by the way CONSUL_SERVERS is: --> $$CONSUL_SERVERS <--"
```
clemente: @SSPJ This was for debugging, I don't think it's needed, could you remove it before merging?
If it's useful, maybe place it somewhere else or with another message.

PR #475, opened 2019-10-06
[BB-1375] Re-add real-time updates to Ocim (with websockets through Django Channels)

lgp171188:
Adds real-time updates to Ocim using Django channels.

**Testing instructions**:

The instructions given below are for testing in the vagrant environment and will require some changes to test in other environments.

* The `channels_redis` library used as the channel layer for `channels` requires Python 3.6 or above, but the version of Python available on Ubuntu 16.04 (used in production and on vagrant) only supports 3.5.x. So to test this PR, install `pyenv` and `pyenv-virtualenv` plugin. Install Python 3.6.9 (the latest Python 3.6 release), create a virtualenv using `pyenv-virtualenv`, activate it and install all the requirements in `requirements.txt.`,
* Setup the configuration in `.env` using the file in the private documentation repository. It is missing some configuration related to AWS - `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. So add those variables with appropriate values.
* Run the unit test suite by running `make test.unit` from the top-level directory of this repository inside the vagrant VM to ensure that the environment is setup okay.
* Create a local superuser account if it doesn't exist already.
* Start the local Ocim instance by running `make run.dev`.
* Navigate to http://localhost:8000 and log in to the site using the super user credentials.
* Check the browser console to confirm that there are no errors.
* Create an instance using one of the factory methods.
* Verify that the instance automatically shows up in the Ocim interface.
* Click on the instance and launch an appserver. Immediately click on the `Logs` tab to check if the logs are updated in real time there.
* Also verify that there is a green circle near the instance name on the left side bar indicating that the instance is being provisioned. Later, check that the icon turns to red in real-time if the appserver fails to provision.
* Verify that after a few seconds, a new appserver shows up automatically under the instance.
* Click on the logs section of the appserver and verify that the logs are updated in real-time.
* Verify that there no JS errors in the browser console.
* Open the local Ocim website on a separate, isolated browser session where the user is unauthenticated. Open the browser JavaScript console and try creating a WebSocket client using the following code in the console.
  ```javascript
  var ws = new WebSocket('ws://localhost:8000/ws/');
  ```
  Verify that an error is thrown and the connection is rejected and that the websocket connections are allowed only for authenticated users.
* Check and test any other things of your choice related to real-time updates and verify that those work.

* All authenticated users are allowed to connect to the WebSocket endpoint `/ws/` and will receive all the websocket messages. The users with `manage_own` permission will also see the effects of the notifications for their instances alone. This mirrors the functionality in the removed swampdragon code. As a workaround, it might be better to allow only the superusers and the users with `manage_own` permissions to connect to the websocket endpoint. Any further changes to fix this issue will probably have to be done in a separate PR as a part of a separate ticket.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,3 @@
-web: python3 manage.py runserver_plus
+web: python3 manage.py runserver 0.0.0.0:8000
 worker: python3 manage.py run_huey --no-periodic
 worker_low_priority: HUEY_QUEUE_NAME=opencraft_low_priority python3 manage.py run_huey --no-periodic
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -64,12 +64,22 @@ server at http://localhost:5000/ using your web browser.
 
 If you prefer not to use Vagrant, you can install OpenCraft Instance Manager manually.
 Refer to the [Ansible playbooks](https://github.com/open-craft/ansible-opencraft) used
-by Vagrant for an example. Instructions based on Ubuntu 16.04.
+by Vagrant for an example. Ocim requires Python 3.6 or a newer version. The instructions
+here are based on Ubuntu 16.04. Since Ubuntu 16.04 ships with Python 3.5, we will use `pyenv`
+to install a newer, supported version on it. This is not required when a supported version of
+Python is available. All the `pyenv` commands can be then ignored and the built-in `venv` module
+can be used to create a virtualenv environment.
 
-Install the system package dependencies & virtualenv:
+Install [pyenv](https://github.com/pyenv/pyenv) by following the [pyenv documentation](https://github.com/pyenv/pyenv#installation).
+Also install [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv#installation) by following its [documentation](https://github.com/pyenv/pyenv-virtualenv#installation).
+
+Install a supported version of Python and create a virtualenv environment. We will be using Python 3.6.9 as an example.
+
+    pyenv install 3.6.9
+
+Install the system package dependencies:
 
     make install_system_dependencies
-    pip3 install --user virtualenv
 
 You might also need to install PostgreSQL, MySQL and MongoDB:
 
@@ -78,10 +88,10 @@ You might also need to install PostgreSQL, MySQL and MongoDB:
 Note that the tests expect to be able to access MySQL on localhost using the
 default port, connecting as the root user without a password.
 
-Create a virtualenv, source it, and install the python requirements:
+Create a virtualenv, activate it, and install the python requirements:
 
-    virtualenv -p python3 venv
-    source venv/bin/activate
+    pyenv virtualenv 3.6.9 opencraft
+    pyenv activate opencraft
     pip install -r requirements.txt
 
 You will need to create a database user to run the tests:
@@ -133,14 +143,14 @@ server:
 Process description
 -------------------
 
-This runs three processes via Honcho, which reads `Procfile` or `Procfile.dev`
+This runs multiple processes via Honcho, which reads `Procfile` or `Procfile.dev`
 and loads the environment from the `.env` file:
 
-* *web*: the main HTTP server (Django + Werkzeug debugger in dev, gunicorn in prod)
-* *worker*: runs asynchronous jobs (Huey)
+* *web*: runs an ASGI server supporting HTTP and Websockets in dev and a WSGI HTTP server using gunicorn in productino.
+* *websocket*: runs an ASGI websocket server in production.
+* *worker\**: runs asynchronous jobs (Huey).
+* *periodic*: runs periodic, asynchronous jobs in production (Huey).
 
-Important: the Werkzeug debugger started by the development server allows remote
-execution of Python commands. It should *not* be run in production.
 
 Static assets collection
 ------------------------
--- a/instance/logging.py
+++ b/instance/logging.py
@@ -29,6 +29,8 @@ import traceback
 from django.apps import apps
 from django.db import connection, models, ProgrammingError
 
+from instance.serializers.logentry import LogEntrySerializer
+from instance.utils import publish_data
 
 # Logging #####################################################################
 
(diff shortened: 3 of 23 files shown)


instance/static/js/src/instance.js
```diff
@@ -69,11 +69,14 @@ app.factory('OpenCraftAPI', function(Restangular) {
     });
 });
 
+app.factory('WebSocketClient', function() {
+    return new WebSocket('ws://' + window.location.host + '/ws/');
```
lgp171188: This is a problem in production where we will be using the `wss` protocol. I have to come up with a way to select the correct protocol in production and development environments.

Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: python3 manage.py runserver_plus
+web: python3 manage.py runserver 0.0.0.0:8000
```
viadanna: Is this change necessary? RunServerPlus is a nice extension for debugging during development.

instance/logging.py
```diff
@@ -72,14 +74,27 @@ def emit(self, record):
...
+        # in debug mode, like "GET /static/img/favicon/favicon-96x96.png":
+        if content_type:
+            log_event = {
+                'type': 'object_log_line',
+                'log_entry': LogEntrySerializer(log_entry).data
+            }
+            if hasattr(obj, 'event_context'):
+                log_event.update(obj.event_context)
+            # TODO: Filter out log entries for which the user doesn't have view rights
+            # TODO: More targetted events - only emit events for what the user is looking at
```
viadanna: This means the logged events will be broadcast to everyone online in OCIM, irrespective of which org the user belongs to, right? I'll have to check some logs, but I'm looking for any security/privacy risks here. What do you think?

instance/static/js/src/openedx_appserver.js
```diff
@@ -111,6 +111,25 @@ app.controller("OpenEdXAppServerDetails", ['$scope', '$state', '$stateParams', '
                               'alert');
             });
         };
+        $scope.$on("websocket:object_log_line", function(event, data){
+            if (!$scope.appserverLogs) {
+                return;
+            }
+
+            if (data.appserver_id == $scope.appserver.id || ($scope.appserver.server && data.server_id == $scope.appserver.server.id)) {
+                if (data.log_entry.level == 'ERROR' || data.log_entry.level == 'CRITICAL') {
```
viadanna: If we keep these filters, can we move the conditional to the backend? This would reduce the traffic and privacy/security risks mentioned earlier.

instance/models/instance.py
```diff
@@ -90,6 +91,16 @@ def delete(self, **kwargs):  # pylint: disable=arguments-differ
...
         super().delete(**kwargs)
 
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        """
+        Save this InstanceReference.
+        """
+        super().save(*args, **kwargs)
+        publish_data({
+            'type': 'instance_update',
+            'instance_id': self.pk,
```
viadanna: I couldn't find any use for the `instance_id`, can we remove it?

instance/models/openedx_appserver.py
```diff
@@ -533,3 +534,8 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
         if not self.pk:
             self.configuration_settings = self.create_configuration_settings()
         super().save(*args, **kwargs)
+        publish_data({
+            'type': 'openedx_appserver_update',
+            'appserver_id': self.pk,
+            'instance_id': self.owner.pk,
```
viadanna: Again it seems the instance id isn't used.

instance/static/js/src/instance.js
viadanna: Excellent!

Procfile.dev
lgp171188: @viadanna, `runserver_plus` doesn't support ASGI. So yes this has to be changed. :(

instance/logging.py
lgp171188: @viadanna, this code along with the comments has been copied from the reverted code. This issue has been present in the previous implementation as well and since it will require more time to fix it in this ticket, it might be better to create a follow-up task for fixing this.

instance/models/instance.py
lgp171188: @viadanna, it is not used in the existing code and I still think that it will be useful to know in the client about the instance which triggered this notification. All the websocket notifications have an identifier to represent the entity that triggered it and retaining this will conform to that. What do you say?

instance/models/openedx_appserver.py
lgp171188: @viadanna, the instance ID is used for this notification type in the JS code below. (see `instance/static/js/src/instance.js` line 236).

instance/static/js/src/openedx_appserver.js
lgp171188: @viadanna, I am not sure I understand your comment here. The filtering here is for segregating the error and non-error messages in separate arrays and I am not sure how it is related to the issue of not restricting access to the websocket notifications.

instance/logging.py
viadanna: Sure :+1:

instance/static/js/src/openedx_appserver.js
viadanna: Nevermind, I misunderstood the indentation and blocks here, all's good.

instance/models/instance.py
viadanna: I was worried of leaking information due to the broadcast nature of the websockets implementation. If that's going to change, no need to touch this.

documentation/installation.md
```diff
@@ -64,12 +64,22 @@ server at http://localhost:5000/ using your web browser.
...
+Python is available. All the `pyenv` commands can be then ignored and the built-in `venv` module
+can be used to create a virtualenv environment.
 
-Install the system package dependencies & virtualenv:
+Install [pyenv](https://github.com/pyenv/pyenv) by following the [pyenv documentation](https://github.com/pyenv/pyenv#installation).
+Also install [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv#installation) by following its [documentation](https://github.com/pyenv/pyenv-virtualenv#installation).
+
+Install a supported version of Python and create a virtualenv environment. We will be using Python 3.6.9 as an example.
+
+    pyenv install 3.6.9
```
viadanna: Thanks for the detailed documentation :+1:

PR #476, opened 2019-10-10
SE-1606: Periodic builds

swalladge:
This PR adds support for periodically building sandboxes.

**Dependencies**: None

**Testing instructions**:

1. deploy ocim (test on stage?)
1. enable periodic builds on an instance and set a reasonable interval for testing (~4h)
1. wait
1. verify that new appservers are being built and activated every `interval` time periods
1. probably more playing around to search for edge cases. (?)
1. view the settings table in the frontend for instances that do and don't have periodic builds enabled. Verify that the information displayed about periodic builds for said instances are correct.

**Screenshots**:

Frontend changes showing periodic builds status to user when disabled:

enabled:

- now is the time to agree on terminology. Please thoroughly check the names of the new table columns and help_text.

**Reviewers**

- [x] @pkulkark 
- [x] @toxinu


periodic_builds/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+    Automatically deploy new servers for all Open edX instances configured for periodic builds.
+    """
+
+    # get all instances configured for periodic builds
+    instances = OpenEdXInstance.objects.filter(periodic_builds_enabled=True)
+
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    for instance in instances:
+        # if the instance is due for a re-build, then we spawn a new appserver
+        latest_appserver = instance.appserver_set.latest("created")
```
swalladge: Try to resolve appserver_set in tests when I'm testing locally just results in the entire function silently returning. Don't know why.

swalladge: fixed: turns out the huey decorator suppresses errors. removed the decorator for debugging, and then re-added it.

instance/models/mixins/openedx_periodic_builds.py
```diff
@@ -0,0 +1,52 @@
...
+    Mixin to provide support for periodic builds.
+    """
+
+    class Meta:
+        abstract = True
+
+    build_interval = models.DurationField(
+        default=datetime.timedelta(hours=24),
+        help_text=(
+            "Time interval between periodic builds. Note that this will not be precise; "
```
pkulkark: @swalladge It would probably help if you specified the time units to be used here i.e. "Time interval (in hours) between periodic builds...."

toxinu:
:+1: Very nice work @swalladge!
 
- [x] Testing instructions seems to be reasonable to cover basic usage. I don't see any corner cases for now
 - [x] I read through the code
 - [ ] ~~I checked for accessibility issues n/a~~
 - [x] Includes documentation: model fields are well described

instance/models/mixins/openedx_periodic_builds.py
swalladge: @pkulkark Ah good point. I will add that to the help text :+1:

periodic_builds/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+                        Status.WaitingForServer,
+                        Status.ConfiguringServer,
+                ):
+                    break
+            else:
+                spawn_appserver(
+                    instance.ref.pk,
+                    num_attempts=instance.periodic_builds_retries + 1,
+                    mark_active_on_success=True,
+                )
```
swalladge: It just occurred to me - this isn't async, so if we have multiple instances with periodic builds enabled, this will block for a long time. We should change this to launching an async task. I will research and update accordingly.

@pkulkark CC @toxinu

swalladge: ... or is it async? :thinking:

swalladge: Wait, I got mixed up. I'm pretty certain now that because the spawn_appserver is a huey `db_task`, it will be run async and return immediately. :sweat_smile:

pkulkark:
@swalladge Checked the deployed instance on stage with the periodic build enabled. Seems to be working fine. LGTM :+1:

PR #477, opened 2019-10-15
SE-1682 deactivate old appservers periodic builds

swalladge:
This deactivates old appservers on successful provisioning of a new appserver when launched from the periodic builds task.

**Testing instructions**:

1. Set up a sandbox for periodic builds
2. wait for at least two appservers to be launched by the periodic builder
3. verify that previous appservers are deactivated and the latest appserver is activated always

Note that this can currently be verified with the [master test](https://stage.console.opencraft.com/instance/4330/) instance on stage.

**Reviewers**
- [ ] @pkulkark

### DIFF
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -69,6 +69,10 @@ def spawn_appserver(
     logger.info('Retrieving instance: ID=%s', instance_ref_id)
     instance = OpenEdXInstance.objects.get(ref_set__pk=instance_ref_id)
 
+    # NOTE: this is not async; blocks up to an hour.
+    # The actual appserver model is created fairly quickly though (after
+    # instance-wide provisioning things happen (mysql, dns records, mongo, s3,
+    # rabbitmq, etc.) but before appserver provision happens)
     appserver = instance.spawn_appserver(
         num_attempts=num_attempts,
         success_tag=success_tag,
--- a/periodic_builds/tasks.py
+++ b/periodic_builds/tasks.py
@@ -26,7 +26,7 @@ import logging
 import datetime
 
 from huey.api import crontab
-from huey.contrib.djhuey import db_periodic_task
+from huey.contrib.djhuey import db_periodic_task, lock_task
 
 from instance.models.openedx_instance import OpenEdXInstance
 from instance.models.appserver import Status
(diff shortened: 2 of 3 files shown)


pkulkark:
@swalladge This would only allow one AppServer to be active at any time. What happens when we need more than 1 server active (like in the case of Cloudera)? Do we just not enable periodic builds in such cases?

swalladge:
@pkulkark Yes the idea is to not enable periodic builds for these cases. See [discussion on discovery doc](https://docs.google.com/a/opencraft.com/document/d/16u3iQXGomLZeHJSW5hKmw71QL-kZtWJVDs9lJ-H55Sc/edit?disco=AAAADikhSHQ) for more context, but basically we decided there weren't any currently known cases where we would want to automatically build and mark new appservers active, but not deactivate old servers. If we find we really want them in the future, it should be a small fix (add a database field to configure the behaviour).

pkulkark:
@swalladge Ah okay. Thanks for the clarification. Looks like the provisioning is failing on Master test instance. Would you be able to fix it so this can be tested?

swalladge:
@pkulkark The master test instance is building correctly now, but it doesn't appear to be deactivating old appservers. Not sure why. Ideas? CC @giovannicimolin

swalladge:
@pkulkark  CC @giovannicimolin I've done some tracing for the issue. Can't verify anything yet until stage is free to play on.

There are two observed issues:

1. old appservers are not deactivated when a new appserver is launched, even though we have `deactivate_old_appservers=True` set.
2. sometimes 2 new appservers are launched simultaneously. eg appserver [42](https://stage.console.opencraft.com/instance/4330/edx-appserver/1469/) and [43](https://stage.console.opencraft.com/instance/4330/edx-appserver/1470/).

For 1., I can't find a reason. It may be possible that orm caching is interfering - unit tests aren't going to pick that up. Eg.

https://github.com/open-craft/opencraft/blob/f7ac4492019ae0cb3d58b66a4a8368f7e0d5d6ae/instance/tasks.py#L104

This seems a bit far-fetched though...

For 2. a possibility is that the huey launch_periodic_builds task was somehow registered twice. If that function was called to quickly in succession, there would be a race condition where the appserver in progress wouldn't be registered yet, so another appserver could be spawned from the function. I don't know how that can happen though. 

There could be an issue perhaps if the task isn't imported by default correctly (may need to add imports to __init__.py files to ensure that). However, it should raise some exception in that case.

There could be race conditions because huey tasks are being called from the periodic task. Eg. we could end up with several appserver spawn tasks in the queue. Now that I say that, it seems the most plausible. We need a way to take a lock on an instance when we launch an appserver, and release it once the appserver has been launched. Ideally, we move the appserver creation to a blocking method we can call from `launch_periodic_builds`, and the rest of the provisioning and setup to a task. Currently, the appserver object isn't created until in `instance.spawn_appserver`, which is too late:

https://github.com/open-craft/opencraft/blob/f7ac4492019ae0cb3d58b66a4a8368f7e0d5d6ae/instance/models/openedx_instance.py#L305

swalladge:
@pkulkark please see discussion on the ticket. This is ready for final review pass. :)

pkulkark:
@swalladge I don't see a new appserver being provisioned for [master test](https://stage.console.opencraft.com/instance/4330/edx-appserver/1597/). The configuration says the periodic builds is set to 4 hours and the last appserver was created more than 4 hours ago. Am I missing something?

swalladge:
@pkulkark It won't be exactly every 4 hours - it could be up to 2 hours past the time depending on when the periodic build launcher runs (and this time it missed out by about 30 seconds :P ). It's been about 5 hours now, so I'd expect a new one to be launched in about an hour.

PR #478, opened 2019-10-15
SE-1607 extra emails on build failure

swalladge:
This PR adds the extra monitoring email addresses as recipients to the email send when an appserver fails to provision. This allows us to specify extra addresses to receive alerts for the periodic builds. It is also general to allow using this for example to ping a PR author if the sandbox fails to provision.

This also includes the patch from #301 (link to instance from admin) and the complementary link from instance page to admin for convenience. As a result, this closes [SE-1409](https://tasks.opencraft.com/browse/SE-1409).

**Screenshots**:

**Testing instructions**:

1. Add an email address you can monitor to an instance's `additional_monitoring_emails` field.
2. set the instance up to fail (eg. choose an invalid git revision for edx-platform, set extra settings that are known to fail, etc.)
3. launch an appserver
4. verify that the email entered in step 1 is emailed with details of the appserver failure when it moves to configuration failed state.
5. Navigate to the settings tab of an instance, find and click the edit icon next to 'Current configuration', verify that it navigates to the admin page for that instance  (see screenshots for how buttons should look)
6. on the admin page for the instance, find the 'view on site' button on the top right, and verify that clicking that navigates to the frontend page for that instance.

**Reviewers**
- [ ] @pkulkark


pkulkark:
@swalladge I tried this on [Master test](https://stage.console.opencraft.com/instance/4330/edx-appserver/1428/) instance on stage but did not receive the email notification when the appserver failed to provision. Not sure if i'm missing something?

swalladge:
@pkulkark Did you add your email to the `additional monitoring emails` field in [admin](https://stage.console.opencraft.com/admin/instance/openedxinstance/456/change/)?

pkulkark:
@swalladge Yup I did. I reverted it back after testing but you can see it was set in the `Configuration` section of AppServer 9.

swalladge:
@pkulkark That's weird... I've been receiving emails alright. I added your email address back in the monitoring emails field (along with mine) and so I guess we just wait and see what happens. ¯\\\_(ツ)\_/¯

pkulkark:
@swalladge My bad! The email ended up in my spam folder for some reason! It's working fine.
Another thing I noticed was, clicking on the link to the django admin page from the instance page redirects to it on the same tab, whereas clicking on the `View on site` link in the django admin page opens a new tab for the instance page. Is that the expected behaviour?

swalladge:
@pkulkark Ah cool! :) Re the link opening, yep that's expected - it's just the default behaviour of the admin app. What behaviour do you reckon it should have?

pkulkark:
@swalladge Ah I see. I thought it would need to redirect on the same tab. But if that's the expected behaviour then I'm fine leaving it as is :slightly_smiling_face: 
The rest of the changes looks good to me :+1:

pomegranited:
@swalladge  OMG Both of the changes you made here are so useful!  Thank you so much for addressing our tech debt by adding that link to the instance's admin page.  😃

PR #481, opened 2019-10-23
SE-1764 Add field for emails on appserver failure

swalladge:
This is a followup to #478 . It has since been determined that re-using the additional_monitoring_emails field for the purpose of extra alert emails about appserver provision failure is bad, because this field's semantics is usually urgent alerts only.

This adds a new field specifically for extra emails to receive alerts on appserver provision fail.

**Merge deadline**: ASAP this is causing a lot of false positive alerts and blocking other tasks

**Testing instructions**:

1. add an email address to an instance's `provision_failed_emails` and another address to additional monitoring emails.
2. ensure the instance's config is such that the next appserver to launch fails (eg. set `AWS_GATHER_FACTS: true` because that currently fails)
3. launch a new appserver
4. wait until it reaches configuration failed state
5. check the email address' inbox and verify that it received the alert email. also verify that the email in additional_monitoring_emails does *not* receive an alert
6. view the settings for an instance and the configuration for an appserver (from the frontend) and verify the provision failed emails field is correctly displayed
7. launch a new *production* instance (eg. using the prod instance factory from the shell).
8. view the new instance in admin and verify that the emails set in the `PROD_APPSERVER_FAIL_EMAILS` (urgent@ currently on stage) is entered in the `provision_failed_emails` field.

For convenience, this branch is checked out on stage.

**Reviewers**
- [x] @lgp171188


instance/models/mixins/utilities.py
```diff
@@ -77,7 +77,7 @@ def provision_failed_email(self, reason, log=None):
             self.EmailBody.PROVISION_FAILED.format(name=self.name, instance_name=self.instance.name, reason=reason),
             self._get_exc_info(default=None),
             attachments=attachments,
-            extra_recipients=self.instance.additional_monitoring_emails,
+            extra_recipients=self.instance.provision_failed_emails,
```
lgp171188: @swalladge, the docstring of this method has to be updated. 

Also, does it make sense to send the logs (which can be large and not easier to check than doing the same via the Ocim interface) in an attachment? I know this was already reviewed and merged. So this is just a question.

instance/static/html/instance/appserver.html
```diff
@@ -98,6 +98,10 @@ <h4>{{ appserver.name }}</h4>
           <td>Additional monitoring email addresses</td>
           <td>{{ appserver.additional_monitoring_emails }}</td>
         </tr>
+        <tr>
+          <td>Email addresses monitoring AppServer provision fail</td>
```
lgp171188:
```suggestion
          <td>Email addresses monitoring AppServer provision failures</td>
```

instance/static/html/instance/details.html
```diff
@@ -99,6 +99,7 @@ <h4>Current configuration <a href="{{ instance.admin_url }}"><i class="fa fa-pen
         <tr><td>configuration_theme_settings</td><td><pre>{{ instance.configuration_theme_settings }}</pre></td></tr>
         <tr><td>Additional security groups</td><td><pre>{{ instance.additional_security_groups }}</pre></td></tr>
         <tr><td>Additional monitoring email addresses</td><td>{{ instance.additional_monitoring_emails }}</td></tr>
+        <tr><td>Email addresses monitoring AppServer provision fail</td><td>{{ instance.provision_failed_emails }}</td></tr>
```
lgp171188:
```suggestion
        <tr><td>Email addresses monitoring AppServer provision failures</td><td>{{ instance.provision_failed_emails }}</td></tr>
```

instance/factories.py
```diff
@@ -151,6 +151,7 @@ def production_instance_factory(**kwargs):
         # Allow production instances to use a different OpenStack instance flavor by default.
         # This allows using a larger instance flavor for production instances.
         openstack_server_flavor=settings.OPENSTACK_PRODUCTION_INSTANCE_FLAVOR,
+        provision_failed_emails=settings.PROD_APPSERVER_FAIL_EMAILS,
```
lgp171188: @swalladge, it will be a good idea to update the test(s) in `instance/tests/test_factories.py` to check this.

instance/models/openedx_appserver.py
```diff
@@ -169,7 +169,16 @@ class Meta:
         help_text=(
             "Optional: A list of additional email addresses other than settings.ADMINS "
             "who should receive alerts from New Relic Synthetics Monitors when this instance "
-            "becomes unavailable, and alerts on AppServer provision failure."
+            "becomes unavailable."
+        )
+    )
+    provision_failed_emails = ArrayField(
```
lgp171188: @swalladge, I am not a native English speaker, but isn't there ambiguity on whether this is for `provision "failed emails"` or `"provision failed" emails`? Would it be better to make it `provisioning_failed_emails` or something like `provisioning_failure_notifcation_emails`? This is just a suggestion and I will leave it to you to decide. :)

opencraft/settings.py
```diff
@@ -543,6 +543,8 @@
 # Email confirmation
 SIMPLE_EMAIL_CONFIRMATION_AUTO_ADD = False
 
+# alert emails on production instance appserver provision fail
```
lgp171188: @swalladge, nit: is this suggestion better?

```suggestion
# Email addresses to alert for production instance appserver provisioning failures.
```

instance/models/openedx_appserver.py
swalladge: Yep makes sense. I wasn't happy with the wording either :+1:

instance/models/mixins/utilities.py
swalladge: Good catch!

> does it make sense to send the logs

probably not, but this was a decision made long ago and not in scope for changing in this ticket. :)

swalladge:
@lgp171188 Thanks! All comments addressed and ready for another pass. :smile:

PR #485, opened 2019-10-30
BB-1837: Start/stop services when starting/stopping vms.

giovannicimolin:
This PR implements methods to start/stop services on VMs that are being started/stopped.

**Testing instructions:**
1. Checkout this branch on Stage.
2. Pick an instance with two running appservers (such as [this one](https://stage.console.opencraft.com/instance/4210/).
3. Activate one appserver, check on the logs that the playbook to enable services was correctly run.
4. Deactivate one appserver, check tat the appserver was correctly run, SSH into the VM and check that the services were stopped.
5. Activate the same appserver, and check that the services were started.
6. Rename `/edx/bin/supervisord` to something else, and deactivate the appserver, it should work.
7. Try to activate the broken appserver, the appserver should not be activated.
8. Repeat this a few times.

**Reviewers:**
- [ ] @viadanna 

CC @pomegranited and @toxinu


instance/models/openedx_appserver.py
```diff
@@ -240,8 +243,17 @@ def make_active(self, active=True):
...
-                         self.name, "active" if active else "inactive", self.instance.name)
+        self.logger.info(
+            'Making %s %s for instance %s...',
+            self.name,
+            "active" if active else "inactive",
+            self.instance.name
+        )
+        # Try to start/stop services on VM
+        # If this fails, don't block activation, but log output
+        if not self.manage_instance_services(active=active):
```
giovannicimolin: I've changed this to avoid activating an appserver if this fails.

viadanna: Not sure I follow, it will continue and activate the instance, no?

giovannicimolin: I've updated the code to prevent activating appservers when the playbook to activate services fail. This is to avoid adding a broken server to the load balancer pool.

instance/tests/models/test_openedx_appserver.py
```diff
@@ -544,6 +546,32 @@ def test_make_active(self, mocks):
...
+        """
+        mocks.mock_run_appserver_playbooks.return_value = ('', 2)
+        instance = OpenEdXInstanceFactory(internal_lms_domain='test.activate.opencraft.co.uk')
+        appserver_id = instance.spawn_appserver()
+        self.assertEqual(mocks.mock_load_balancer_run_playbook.call_count, 1)
+        appserver = instance.appserver_set.get(pk=appserver_id)
+
+        self.assertEqual(instance.appserver_set.get().last_activated, None)
+
+        appserver.make_active()
```
viadanna: Is this supposed to fail?

giovannicimolin: No, it should simply return without modifying the appserver.
Do you think its better if we raise an exception?

viadanna: It would be good to have some indication if it failed. Maybe return True/False?

giovannicimolin: The other methods below don't return anything, so the behavior won't be consistent if I return `true/false`.

instance/models/openedx_appserver.py
```diff
@@ -249,8 +252,29 @@ def make_active(self, active=True):
...
+            self.instance.name
+        )
+        # Try to start/stop services on VM.
+        # If this fails, block instance activation/deactivation
+        if not self.manage_instance_services(active=active):
+            self.logger.error(
+                'Failed to start/stop Open edX services on instance %s...',
+                self.instance.name
+            )
+            # Allow deactivating failing appservers but block activating
```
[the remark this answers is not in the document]
viadanna: Thanks for the comments explaining the behavior [here](https://github.com/open-craft/opencraft/pull/485/files#diff-22ca602ba3ef86e0e14f219a74caa12eR262).

instance/tests/integration/ansible/playbooks/failignore.yml
```diff
@@ -12,6 +12,17 @@
 
 - name: Fail and ignore on purpose
   hosts: all
+  become: true
   tasks:
+    - file:
```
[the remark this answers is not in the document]
viadanna: Both the ansible_failure and ansible_failignore tests failures in running playbooks and how we deal with those in code. We actually provision a new test instance and appserver for each, so there should be no conflicts, unless I misunderstood the comments. I don't know why they run `apt` but I guess it's simply testing the VMs are working fine and we can leave that.

Due to the unexpected complexity mocking this, our constant improvements to the integration tests and the fact this has been tested working, I'd say we keep this behavior.

instance/tests/integration/ansible/playbooks/failignore.yml
```diff
@@ -12,6 +12,17 @@
...
+        path: /edx/bin/
+        state: directory
+        mode: '0755'
+    - copy:
```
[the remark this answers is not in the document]
viadanna: This was added because there's no real `/edx/bin/supervisorctl` in the test instance, as each of the four tests in this test group create an independent instance to prevent conflicts.

instance/tests/integration/ansible/playbooks/failignore.yml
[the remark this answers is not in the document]
giovannicimolin: > I don't know why it updates packages through apt. I don't fully understand the script so I might be wrong.
> I don't know why they run apt but I guess it's simply testing the VMs are working fine and we can leave that.

`python` is required so that ansible can run any commands besides apt on the hosts. So the script runs `apt update` and then installs `python-minimal`. :)

PR #486, opened 2019-11-01
Wire integration tests to their own load balancer

SSPJ:
A small change (which took a lot of work to get, thanks @clemente !) to not run the consul agent in dev mode but instead to connect it to the Consul server running on `haproxy-integration`.

This means that when integration tests run, they will write instance data to that consul, which will then request certs for them via the cert-manager.

To avoid test failures during DNS propagation, there is a wildcard in Gandi for `*.integration` which points to `haproxy-integration`.

The other change required is setting a few variables in CircleCI config (and updating the entry in Vault).

```
DEFAULT_LOAD_BALANCING_SERVER: ubuntu@haproxy-integration.net.opencraft.hosting
CONSUL_SERVERS: haproxy-integration.net.opencraft.hosting
CONSUL_ENABLED: true
DISABLE_LOAD_BALANCER_CONFIGURATION: true
```

**Testing Instructions**

None.

Note to the reviewer. The vars in `circleci.yml` were set to make the tests pass. I will remove them before merging this.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -21,4 +21,5 @@ LOAD_BALANCER_FRAGMENT_NAME_PREFIX='opencraft-'
 DEFAULT_RABBITMQ_API_URL='https://admin:password@rabbitmq.example.com:15671'
 DEFAULT_INSTANCE_RABBITMQ_URL='amqps://rabbitmq.example.com:5671'
 CONSUL_ENABLED: false
+CONSUL_SERVERS: haproxy-integration.net.opencraft.hosting
 DISABLE_LOAD_BALANCER_CONFIGURATION: false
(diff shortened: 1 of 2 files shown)


clemente:
:+1: This was already tested, and the tests pass now. If you move the variables to CircleCI, then it's ready. We can continue deploying, running CI and tweaking the CircleCI if needed after it's merged. Though the current variables seem to work. I didn't check why we need  each one of these variables, but they seem good. But are all of them needed? Maybe they already have good values in CircleCI.

  - [X] I already tested it in SE-1376: the shared consul is populated with values when the integration tests run
  - [X] I read through the code
  - [] Includes documentation: none needed

PR #489, opened 2019-11-10
SE-1832 Move auto provisioning failure email alerts to registration

swalladge:
This moves the auto-setting of proviosing failure notification emails from production instance factory to the provision instance function that is called when creating an instance for an end user. Doing this because we only want to auto set PROD_APPSERVER_FAIL_EMAILS for end user instances.

**Test instructions**:

- deploy this branch to stage (no migrations and only small changes; easy to switch between this and master)
- register as an end user
- verify that the resulting instance has the settings.PROD_APPSERVER_FAIL_EMAILS address(es) set on provisioning_failure_notification_emails
- launch a new instance via production_instance_factory
- verify that no provisioning_failure_notification_emails are set

**Reviewer**:

- [x] @gr4yscale

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -151,7 +151,6 @@ def production_instance_factory(**kwargs):
         # Allow production instances to use a different OpenStack instance flavor by default.
         # This allows using a larger instance flavor for production instances.
         openstack_server_flavor=settings.OPENSTACK_PRODUCTION_INSTANCE_FLAVOR,
-        provisioning_failure_notification_emails=settings.PROD_APPSERVER_FAIL_EMAILS,
     )
     instance_kwargs.update(kwargs)
 
--- a/instance/tests/test_factories.py
+++ b/instance/tests/test_factories.py
@@ -122,7 +122,8 @@ class FactoriesTestCase(TestCase):
         instance = production_instance_factory(sub_domain=sub_domain)
         instance = OpenEdXInstance.objects.get(pk=instance.pk)
         self._assert_field_values(instance, sub_domain, **self.PRODUCTION_DEFAULTS)
-        self.assertEqual(instance.provisioning_failure_notification_emails, ['appserverfail@localhost'])
+        # standard production instances should not have this set
+        self.assertEqual(instance.provisioning_failure_notification_emails, [])
 
         # Create instance with custom field values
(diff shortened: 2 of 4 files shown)


gr4yscale:
@swalladge I'll take a look at this in a few hours while you're probably sleeping. :)

gr4yscale:
@swalladge I'm not seeing the `provisioning_failure_notification_emails` set after doing the first 2 steps: 

> verify that the resulting instance has the settings.PROD_APPSERVER_FAIL_EMAILS address(es) set on provisioning_failure_notification_emails

After waiting a bit of time, I've checked and the array for `provisioning_failure_notification_emails` is empty, where we should expect to see "urgent@opencraft.com":

```
In [49]: instance = OpenEdXInstance.objects.latest('id')

In [50]: instance
Out[50]: <OpenEdXInstance: test-se-1832 (test-se-1832.stage.opencraft.hosting)>

In [51]: instance.provisioning_failure_notification_emails
Out[51]: []

In [52]: from django.conf import settings                                                           

In [53]: settings.PROD_APPSERVER_FAIL_EMAILS                                                        
Out[53]: ['urgent@opencraft.com']
```


This could entirely be my fault as I've had very little experience working with OCIM thus far.

1) I've attached to the screen session on `stage.console.opencraft.com`, switched to your branch, and restarted the services (killed the web and worker processes, `make run`),

2) I've gone to https://stage.console.opencraft.com/registration/ and registered a new account and the domain `test-se-1832.opencraft.hosting`, and clicked the link in verification email that I received.

3) Back to the screen session, after restarting the django shell for good measure, I ran the above ^^

Am I missing something?

swalladge:
@gr4yscale No you're not missing anything. Thanks for being so thorough! Back to debugging I guess. :)

swalladge:
@gr4yscale I did some debugging and lol the logic was working, but the instance wasn't being saved. [this instance](https://stage.console.opencraft.com/instance/4460/) was created on the latest revision following the same steps you did. This time it correctly sets the alert emails!

By the way, this field can be viewed directly from the frontend in the instance settings tab - look for 'Email addresses monitoring AppServer provision failures'.

Ready for final review. :)

gr4yscale:
@swalladge :+1:  Tested this and it is working now. Good to merge. :)

PR #490, opened 2019-11-11
BB-1857 Respecting DISABLE_LOAD_BALANCER_CONFIGURATION during cleanup

viadanna:
## Testing
1. In your devstack, run `make test.integration_cleanup` and see the tasks for manipulating load balancers being run. (might need to comment the previous cleanup steps).
1. Then run `DISABLE_LOAD_BALANCER_CONFIGURATION=true make test.integration_cleanup` and see that the task for manipulating load balancers is ignored.


instance/tests/integration/integration_instance.py
```diff
@@ -379,13 +380,7 @@ def test_spawn_appserver(self):
 
         if settings.DEFAULT_INSTANCE_MYSQL_URL and settings.DEFAULT_INSTANCE_MONGO_URL:
             self.assert_swift_container_provisioned(instance)
-            self.assert_server_ready(instance)
-            self.assert_instance_up(instance)
-            self.assert_appserver_firewalled(instance)
-            self.assertTrue(instance.successfully_provisioned)
```
kaizoku: Do we no longer need to check that all of these steps succeeded?

viadanna: These assertions were duplicated when merging integration tests. Those are running 10 lines above.

cleanup_utils/load_balancer_cleanup.py
```diff
@@ -47,8 +51,15 @@ def __init__(self, load_balancer_address, fragment_prefix, age_limit, dry_run):
 
     def run_cleanup(self):
         """Run the actual cleanup"""
+        logger.info("\n --- Starting Consul cleanup ---")
+        self._clean_consul()
```
giovannicimolin: This should come after the self.dry_run` check below. 
No keys should be deleted when running with `dry_run`, but the log output should work as usual.

cleanup_utils/load_balancer_cleanup.py
```diff
@@ -24,6 +24,10 @@
 import logging
 import os
 import pathlib
+from operator import itemgetter
+
+from consul import Consul
```
giovannicimolin: I'm not sure this requirement is being installed (the cleanup requirements are on a separate file).

cleanup_utils/load_balancer_cleanup.py
viadanna: I was just working on that and pushed some changes to it.

cleanup_utils/load_balancer_cleanup.py
viadanna: The cleanup jobs is working, but I'll double-check it.

circle.yml
```diff
@@ -149,6 +158,8 @@ workflows:
   version: 2
   build-test-and-deploy:
     jobs:
+      - cleanup:
```
giovannicimolin: Remove this before merging. We already have a schedule cleanup tasks below.

instance/tests/integration/integration_instance.py
```diff
@@ -379,13 +380,7 @@ def test_spawn_appserver(self):
 
         if settings.DEFAULT_INSTANCE_MYSQL_URL and settings.DEFAULT_INSTANCE_MONGO_URL:
             self.assert_swift_container_provisioned(instance)
```
giovannicimolin: Is this check really needed now that swift support is deprecated? Can you remove it too?

circle.yml
viadanna: Sure, I was just keeping this until the approval to make sure none of the changes broke the cleanup job and there's no flakiness. It's quite difficult to test otherwise.

instance/tests/integration/integration_instance.py
viadanna: I intended to create a task to remove all deprecated code related to Swift in one go.

Given I'll have to get another CI run for the other change, I'll include removing this.

PR #494, opened 2019-11-19
BB-1876: Add frontend automatic build and deployment WIP

giovannicimolin:
This PR set's up automatic deployment of the new Ocim static frontend. It does this:
1. Runs tests.
2. Builds static files.
3. Waits for user approval to run deployment.
4. Deploys to S3 and invalidates CloudFront.

**Testing instructions:**
1. Rerun the CI checks
2. Wait until the `allow-deployment` stage then click on the stage and **Approve** the deployment:
click
approve
3. Watch the redeployment, check that the files were updated and a CloudFront invalidation was triggered. The AWS account is: [Ocim Stage](https://vault.opencraft.com:8200/ui/vault/secrets/secret/show/core/OpenCraft%20Tools%20:%20Resources%20:%20AWS%20:%20AWS%20S3%20Stage).
4. Check the live app: https://app.stage.console.opencraft.com/

**Reviewers:**
- [ ] @Agrendalath

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,14 +12,14 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps--{{ .Branch }}-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Run frontend tests
           command: npm test -- --reporters=default --reporters=jest-junit --ci --runInBand
@@ -34,14 +34,14 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Run frontend quality tests
           command: npm run lint
@@ -56,14 +56,14 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ checksum "package-lock.json" }}
+          key: frontend-deps-{{ .Branch }}
       - run:
           name: Render environment variables to .env file
           command: |
--- a/circle.yml
+++ b/circle.yml
@@ -65,6 +65,10 @@ jobs:
             - node_modules
           key: frontend-deps-{{ checksum "package-lock.json" }}
       - run:
+          name: Render environment variables to .env file
+          command: |
+            echo $(env) >> .env
+      - run:
           name: Build React frontend
           command: npm run build
       - persist_to_workspace:
--- a/circle.yml
+++ b/circle.yml
@@ -12,14 +12,17 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps-{{ .Branch }}
+          keys:
+            - frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - frontend-deps-{{ .Branch }}-
+            - frontend-deps-
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ .Branch }}
+          key: frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
       - run:
           name: Run frontend tests
           command: npm test -- --reporters=default --reporters=jest-junit --ci --runInBand
@@ -34,14 +37,17 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps-{{ .Branch }}
+          keys:
+            - frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - frontend-deps-{{ .Branch }}-
+            - frontend-deps-
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ .Branch }}
+          key: frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
       - run:
           name: Run frontend quality tests
           command: npm run lint
@@ -56,14 +62,17 @@ jobs:
       - checkout:
           path: ~/project
       - restore_cache:
-          key: frontend-deps-{{ .Branch }}
+          keys:
+            - frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - frontend-deps-{{ .Branch }}-
+            - frontend-deps-
       - run:
           name: Installing frontend dependencies
           command: npm install
       - save_cache:
           paths:
             - node_modules
-          key: frontend-deps-{{ .Branch }}
+          key: frontend-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
       - run:
           name: Render environment variables to .env file
           command: |
--- a/circle.yml
+++ b/circle.yml
@@ -50,7 +50,7 @@ jobs:
           paths:
             - .
 
-  deploy-frontend:
+  deploy-frontend-stage:
     docker:
       - image: circleci/python:3.8
     steps:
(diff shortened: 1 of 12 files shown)


circle.yml
```diff
@@ -208,9 +262,25 @@ workflows:
...
+          type: approval
+          requires:
+            - build-frontend
+      - deploy-frontend-stage:
+          requires:
+            - allow-deployment
+          filters:
+            branches:
+              only:
+                - giovanni/add-frontend-ci-deployment
```
giovannicimolin: This needs to be renamed before merging, and renamed again before merging to `master`

circle.yml
```diff
@@ -20,6 +25,52 @@ jobs:
...
+      - run:
+          name: Installing frontend dependencies
+          command: npm install
+      - save_cache:
+          paths:
+            - node_modules
+          key: dependencies-{{ checksum "package-lock.json" }}
+      - run:
+          name: Build React frontend
+          command: npm run build
```
Agrendalath: Don't we need to generate any `.env` file for the frontend deployment?

circle.yml
```diff
@@ -1,11 +1,16 @@
 version: 2.1
+
+orbs:
+  aws-s3: circleci/aws-s3@1.0.11
+
 jobs:
-  build-frontend:
+  test-frontend:
```
Agrendalath: I see a couple console errors in the [frontend tests](https://circleci.com/gh/open-craft/opencraft/9606l). Should we create a follow-up to fix them?

circle.yml
```diff
@@ -208,9 +262,30 @@ workflows:
...
+          type: approval
+          requires:
+            - build-frontend
+          filters:
+            branches:
+              only:
+                - giovanni/add-frontend-ci-deployment
+      - deploy-frontend-stage:
+          requires:
+            - allow-deployment
```
Agrendalath: Do we really want to restrict the stage deployments with a click? It might be worth doing this for the prod deployment, but do we actually need this for stage?

circle.yml
```diff
@@ -20,6 +25,52 @@ jobs:
...
+          key: dependencies-{{ checksum "package-lock.json" }}
+      - run:
+          name: Build React frontend
+          command: npm run build
+      - persist_to_workspace:
+          root: build
+          paths:
+            - .
+
+  deploy-frontend-stage:
```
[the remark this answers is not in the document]
giovannicimolin: Yes, the production deployment will be added in a follow up, once we have the registration form working.

circle.yml
giovannicimolin: No, the environment variables are taken directly from the CircleCI configuration and used in the build.
You can see the variables [here](https://circleci.com/gh/open-craft/opencraft/edit#env-vars), prefixed with `FRONTEND_STAGE_`.

circle.yml
giovannicimolin: They were merged like that because the JS tests/checks were not properly set up. I will create a follow up task to address those issues.

circle.yml
```diff
@@ -20,6 +25,52 @@ jobs:
...
+          arguments: |
+            --acl public-read \
+            --profile=default
+          aws-access-key-id: FRONTEND_STAGE_AWS_ACCESS_KEY_ID
+          aws-secret-access-key: FRONTEND_STAGE_AWS_SECRET_ACCESS_KEY
+          aws-region: FRONTEND_STAGE_AWS_REGION
+      - run:
+          name: Invalidate CloudFront cache
+          command: |
+            aws cloudfront create-invalidation --distribution-id $FRONTEND_STAGE_AWS_CLOUDFRONT_DISTRIBUTION_ID --paths '/*' --profile=default
```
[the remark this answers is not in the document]
giovannicimolin: Yeah, its required because otherwise the `awscli` orb picks up the credentials from the environment variables (that are the ones for the integration environment).

circle.yml
giovannicimolin: Isn't it better to wait the stage backend deployment before frontend changes?

giovannicimolin: Anyway, it's not a big deal for stage, so I've removed it.

circle.yml
```diff
@@ -20,6 +25,52 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: build-frontend
+      - aws-s3/sync:
+          from: build-frontend/
+          to: "s3://$FRONTEND_STAGE_AWS_S3_BUCKET/frontend"
+          overwrite: true
+          arguments: |
+            --acl public-read \
+            --profile=default
```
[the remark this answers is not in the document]
giovannicimolin: Done.

circle.yml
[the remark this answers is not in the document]
Agrendalath: What I meant is that we already have it defined [here](https://github.com/open-craft/opencraft/pull/494/files#diff-29944324a3cbf9f4bd0162dfe3975d88R91). Should it still be in the [current line](https://github.com/open-craft/opencraft/pull/494/files#diff-29944324a3cbf9f4bd0162dfe3975d88R98)? Sorry for not making this clear.

circle.yml
Agrendalath: Oh, nice! Just curious - how does it work? Do we have some custom build scripts for supporting this naming convention?

circle.yml
Agrendalath: > Isn't it better to wait the stage backend deployment before frontend changes?

Technically yes, but it doesn't matter for stage. My understanding of the CD for stage is that I merge the PR and I have the results automatically deployed. Adding a manual step that doesn't ping you when it's ready to be triggered (I checked their docs community forums, but it doesn't seem to be possible to trigger custom notification) makes it very easy to forget about it. At least I'm pretty sure I would forget to click it from time to time :D

circle.yml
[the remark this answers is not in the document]
giovannicimolin: Oh yeah, the same bug happens here too. If this argument isn't passed, they incorrect credentials are used. I think that this happens because `awscli` prioritizes environment variables over the `~/.aws/credentials` file, and the only way to circumvent this and leave the variables for the integration tests untouched is to pass the argument on every line run.

circle.yml
giovannicimolin: Ah, I see what you mean. Yeah, I'm not passing the variables correctly to the react app, just for the AWS deployment script. 
Let me copy paste a few lines from the sprints app and fix this up :stuck_out_tongue:

circle.yml
```diff
@@ -26,7 +26,29 @@ jobs:
...
-  build-frontend:
+  frontend-quality:
+    docker:
+      - image: circleci/node:lts
+    working_directory: ~/project/frontend
+    steps:
+      - checkout:
+          path: ~/project
+      - restore_cache:
+          key: dependencies-{{ checksum "package-lock.json" }}
```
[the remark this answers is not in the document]
giovannicimolin: @Agrendalath I tried it, and it doesn't work. Following the documentation instructions the CI throws an error:
```
# |   |   2. [#/jobs/frontend-build/steps/1/restore_cache] only 1 subschema matches out of 2
# |   |   |   1. [#/jobs/frontend-build/steps/1/restore_cache/key] expected type: String, found: Sequence
# |   |   |   |   SCHEMA:
# |   |   |   |     type: string
# |   |   |   |   INPUT:
# |   |   |   |     null
```
And the error happens for cache restores that are using a list of arguments instead of a string. :disappointed: 
Do we have any repo with this working so that I can check/copy?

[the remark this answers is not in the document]
giovannicimolin: See [this build](https://circleci.com/gh/open-craft/opencraft/9659?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

circle.yml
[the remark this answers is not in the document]
giovannicimolin: Removed these workarounds in favor of using contexts

circle.yml
```diff
@@ -1,25 +1,99 @@
...
+      - restore_cache:
+          key: frontend-deps-{{ checksum "package-lock.json" }}
+      - run:
+          name: Installing frontend dependencies
+          command: npm install
+      - save_cache:
+          paths:
+            - node_modules
+          key: frontend-deps-{{ checksum "package-lock.json" }}
+      - run:
```
giovannicimolin: @Agrendalath Here is the `.env` file rendering. It'll only render variables starting with `REACT_APP`, and this allows us to add new variables without editing the file.
React takes care of importing this using the `dotenv` package.

Agrendalath: Oh, this is cool! I haven't thought about adding them this way. I'll gladly reuse it in Sprints CI :)

Btw. dotenv supports also other files, like `.env.local` and `.env.production`. Just wanted to point it out.

circle.yml
Agrendalath: Haha, sure :P

circle.yml
Agrendalath: Oh, this is cool! I haven't thought about adding them this way. I'll gladly reuse it in Sprints CI :)

Btw. dotenv supports also other files, like `.env.local` and `.env.production`. Just wanted to point it out.

circle.yml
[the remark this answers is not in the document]
Agrendalath: I checked your commits and it looks like you were using `key` instead of `keys` for restoring multiple cache options ;)

I pushed the commit to the separate branch, [verified it on the CI](https://circleci.com/workflow-run/dc915f85-1bbf-4806-8baf-d3256f1426c2) and cherry-picked it to this branch.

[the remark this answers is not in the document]
giovannicimolin: Woah, thanks for fixing that for me! I didn't even notice that there was a difference in the usages.

circle.yml
giovannicimolin: That's good to know! I'm going with this implementation for now so that I can reuse the build and deployment scripts and just change the context.

PR #495, opened 2019-11-19
[BB-1909] Disable strict host key checking

Agrendalath:
This removes the key checking, which could interfere if we got an IP address that we've already used.

### DIFF
--- a/instance/models/mixins/ansible.py
+++ b/instance/models/mixins/ansible.py
@@ -54,6 +54,8 @@ class AnsibleAppServerMixin(models.Model):
         help_text='YAML variables for commonly needed services.')
 
     INVENTORY_GROUP = 'generic'
+    # It is possible to get an IP address that has already appeared, so we want to disable strict host key checking.
+    HOST_ARGS = 'ansible_ssh_common_args="-o StrictHostKeyChecking=no"'
 
     class Meta:
         abstract = True
@@ -96,9 +98,9 @@ class AnsibleAppServerMixin(models.Model):
             raise RuntimeError("Cannot prepare to run playbooks when server has no public IP.")
         return (
             '[{group}]\n'
-            '{server_ip}\n'
+            '{server_ip} {host_args}\n'
             '[app:children]\n'
-            '{group}'.format(group=self.INVENTORY_GROUP, server_ip=public_ip)
+            '{group}'.format(group=self.INVENTORY_GROUP, server_ip=public_ip, host_args=self.HOST_ARGS)
         )
 
     def _run_playbook(self, working_dir, playbook):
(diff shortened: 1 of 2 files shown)


instance/models/mixins/ansible.py
```diff
@@ -54,6 +54,8 @@ class AnsibleAppServerMixin(models.Model):
         help_text='YAML variables for commonly needed services.')
 
     INVENTORY_GROUP = 'generic'
+    # It is possible to get an IP address that has already appeared, so we want to disable strict host key checking.
```
swalladge: Drive by comment: can this comment include a warning that this is a special case and disabling strict host key checking is not something one should do?

Agrendalath:
@itsjeyd, thank you for the detailed response. I was thinking about this approach (removing the key during terminating an instance) before implementing this, but I didn't see any good point in doing this, because currently the only place we could get this is during the deployment of the new AppServer. As this is one-time operation for the **new** VM, we wouldn't accidentally reuse an existing one anywhere. But removing these keys is future proof, so it sounds better to do it this way. If we decided to run playbooks on existing instances, such check would become very handy.

What should we do about the current content of `known_hosts`? Should we just erase it or write some more sophisticated management command for removing the hosts that are not related to currently running AppServers?

If nothing serious comes up, I should have some time for taking this the next week. Otherwise I'll follow-up on this in the next sprint.

**Btw.** are you thinking of any specific vectors of the MiTM attack for this? As we're using the actual IP addresses here, the only vulnerability I could think of would be spoofing a VM provider to return an IP address used by already existing instance to replace it with custom configuration.
cc: @swalladge

swalladge:
> Btw. are you thinking of any specific vectors of the MiTM attack for this?

Nothing in particular. In this case, disabling strict host checking may in fact be the best option. My only point was to leave a comment to dissuade people from copying it for other purposes where it wouldn't be a good idea.

itsjeyd:
@Agrendalath Thanks for the details :)

> What should we do about the current content of `known_hosts`? Should we just erase it or write some more sophisticated management command for removing the hosts that are not related to currently running AppServers?

That's a good point! If we make removing the entry for a given VM from `known_hosts` part of the process of terminating that VM, I don't think we'll need a more sophisticated management command. Cleaning up the current contents of `known_hosts` can just be a one-off operation (even if we end up scripting it).

> If nothing serious comes up, I should have some time for taking this the next week. Otherwise I'll follow-up on this in the next sprint.

I think we'll need to shoot for next sprint: edX's request about making some of our XBlocks Python 3-compatible is likely higher priority at this point (cf. [SE-1901](https://tasks.opencraft.com/browse/SE-1901)).

So let's create a follow-up task for implementing the approach we've been discussing and close out [BB-1909](https://tasks.opencraft.com/browse/BB-1909).

I'll let you decide if you want to reuse this PR for the follow-up ticket, or close it and create a new one :)

Agrendalath:
This will be handled in [BB-1913](https://tasks.opencraft.com/browse/BB-1913).

PR #499, opened 2019-12-03
[BB-1873] Add v2 of account and instance config APIs

xitij2000:
This PR adds new API endpoints to power the new React-based registration flow. It also adds support Swagger API documentation via `drf_yasg` to enable automatic generation of an API Client. It also moves all existing API from before this PR into the v1 namespace. 

Under the v2 namespace it adds:
- /v2/accounts/ Accounts API: This API enables public registration, account validation, and allows fetching and updating account data.
- /v2/instances/openedx_config OpenedX instance configuration API: This allows new users to register their first instance and to update their instances' configuration and to provision a new server for it after changes are complete.


registration/api/v2/views.py
```diff
@@ -0,0 +1,219 @@
...
+    @action(detail=True, methods=['post'])
+    def commit_changes(self, request, pk=None):
+        """
+        Commit configuration changes to instance and launch new AppServer.
+
+        This API call will copy over any changes made to the instance config to
+        the actual instance used to launch AppServers and launch a new AppServer
+        with the applied changes.
+
+        It checks to see if an AppServer is already being provisioned and in that
```
Agrendalath: Nit: redundant `to see`.

registration/api/v2/views.py
```diff
@@ -0,0 +1,219 @@
...
+            raise ValidationError("Instance launch already in progress", code="in-progress")
+
+        if not EmailAddress.objects.get(email=instance_config.public_contact_email).is_confirmed:
+            raise ValidationError("Updated public email needs to be confirmed.", code="email-unverified")
+
+        instance.theme_config = instance_config.draft_theme_config
+        instance.name = instance_config.instance_name
+        instance.privacy_policy_url = instance_config.privacy_policy_url
+        instance.email = instance_config.public_contact_email
+        instance.save()
```
Agrendalath: Wouldn't it be better to extract this to the serializer (e.g. subclassing `OpenEdXInstanceConfigSerializer`)?

registration/api/v2/views.py
```diff
@@ -0,0 +1,219 @@
...
+
+        spawn_appserver(instance.ref.pk, mark_active_on_success=True, num_attempts=2)
```
Agrendalath: I believe it might be nice to extract `num_attempts` to settings and env variables instead of hardcoding it.

registration/api/v2/serializers.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    user = serializers.HiddenField(default=serializers.CurrentUserDefault())
+
+    def validate_user(self, value):
+        """
+        Prevent user from creating more than the allowed number of instances.
+
+        Currently each user is allowed one Open edX instance.
+        """
+        if BetaTestApplication.objects.filter(user=value).exists():
```
Agrendalath: We could change `exists` to `count` with and overriddable value for forward compatibility.
We may also just leave it as it is for now and change it in the future, if needed.

Makefile
```diff
@@ -174,7 +174,7 @@ test.one: clean
 # Files #######################################################################
 
 static_external:
-	$(MAKE) -C static/external
+	$(MAKE) -C static/external -j4
```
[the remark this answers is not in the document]
xitij2000: The reason for limiting here is mainly to respect the bandwidth of the sites we are pulling from. Ideally we would just avoid fetching this again if it hadn't changed but making that kind if change was hard and out of scope here.

api/urls.py
```diff
@@ -24,15 +24,32 @@
...
 
 # URL Patterns ################################################################
 
 app_name = 'api'
+
+# pylint: disable=invalid-name
+schema_view = get_schema_view(
+    openapi.Info(
+        title="OpenCraft Instance Manager",
+        default_version="v1",
```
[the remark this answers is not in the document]
xitij2000: I think we can eventually just get rid of and remove the v1 APIs when all of them have replacement.

registration/api/v2/views.py
xitij2000: This method simply doesn't involve serialisation so it didn't feel right. Perhaps it would be better placed on the model itself?

registration/api/v2/serializers.py
xitij2000: My understanding is that `exists` is much better for performance if applicable.

registration/api/v2/views.py
```diff
@@ -0,0 +1,219 @@
...
+from registration.models import BetaTestApplication
+from registration.utils import verify_user_emails
+from userprofile.models import UserProfile
+
+logger = logging.getLogger(__name__)
+
+
+@viewset_swagger_helper(
+    create="Create new user registration",
+    list="Get current user registration data",
```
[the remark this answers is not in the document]
xitij2000: I think it's sufficient that the list method returns only the current user's profile. That can be used by the frontend without any need for a /current/ model since these APIs are only for regular users, not admins.

xitij2000:
@Agrendalath  There are no specific scenarios other than those covered by tests anyway. 

It mainly just adds an API for mostly existing data.

registration/api/v2/views.py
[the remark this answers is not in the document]
Agrendalath: Sure, that makes sense too.

registration/api/v2/views.py
Agrendalath: Moving it partially to the model could also be confusing. Let's just leave it here.

registration/api/v2/serializers.py
Agrendalath: Yep, it is. Let's use `exists` then, as we currently don't care about multiple instances.

registration/api/v2/views.py
Agrendalath: Ah, I see you moved them. That's fine :+1:

PR #500, opened 2019-12-03
BB-1919: Add registration flow

giovannicimolin:
This PR introduces the new Ocim registration pages with the basic registration flow.

**Testing instructions:**
1. `npm i && npm start`
2. Go to http://localhost:3000/registration/domain.
3. Insert `existing` and click "Check availability".
4. Check that an error message is displayed.
5. Write something and check that the error message immediately vanishes after editing the field.
6. Test that that registration form is working (next/back buttons work, data is store on the state).
7. Check that the styles are similar to the ones in the mockups.

Since the validators aren't implemented yet, change a few values in the `registrationReducer` to show when fields have feedback error messages. Example: show error on password field:
```
    case RegistrationActions.Types.ROOT_STATE_UPDATE:
      // Merge state without erasing previous values
      return {
        loading: state.loading,
        registrationData: {
          ...state.registrationData,
          ...action.data.registrationData
        },
        registrationFeedback: {
          ...state.registrationFeedback,
          ...action.data.registrationFeedback,
          // ADD THIS LINE
          password: "This password already exists!"
        }
      };
```

**Reviewer:**
- [ ] @Agrendalath

### DIFF
--- a/frontend/src/registration/actions.ts
+++ b/frontend/src/registration/actions.ts
@@ -2,7 +2,7 @@ import { push } from 'connected-react-router';
 import { OcimThunkAction } from 'global/types';
 import { Action } from 'redux';
 import { Types as UIActionTypes } from 'ui/actions';
-import { RegistrationModel, InstanceInfoModel } from './models';
+import { RegistrationModel, InstanceInfoModel, DomainInfoModel } from './models';
 
 export enum Types {
   REGISTRATION_SUBMIT = 'REGISTRATION_SUBMIT',
@@ -30,6 +30,19 @@ export type ActionTypes =
   | RegistrationSuccess
   | RegistrationFailure;
 
+
+export const updateDomainInfoState = (
+  data: DomainInfoModel,
+): OcimThunkAction<void> => async dispatch => {
+  dispatch({ type: Types.REGISTRATION_SUCCESS, data });
+};
+
+export const updateInstanceInfoState = (
+  data: InstanceInfoModel,
+): OcimThunkAction<void> => async dispatch => {
+  dispatch({ type: Types.REGISTRATION_SUCCESS, data });
+};
+
 export const submitRegistration = (
   data: RegistrationModel,
   nextStep?: string
@@ -39,26 +52,28 @@ export const submitRegistration = (
     data
   });
 
-  try {
-    // try submitting form data
-    // TODO
-    if (data.domain === 'existing') {
-      throw Error('Test error');
+  setTimeout(() => {
+    try {
+      // try submitting form data
+      // TODO
+      if (data.domain === 'existing') {
+        throw Error('Test error');
+      }
+      dispatch({ type: Types.REGISTRATION_SUCCESS, data });
+      if (nextStep) {
+        dispatch(push(nextStep));
+      }
+      dispatch({ type: UIActionTypes.NAVIGATE_NEXT_PAGE });
+    } catch (error) {
+      dispatch({
+        type: Types.REGISTRATION_FAILURE,
+        error
+      });
     }
-    dispatch({ type: Types.REGISTRATION_SUCCESS, data });
-    if (nextStep) {
-      dispatch(push(nextStep));
-    }
-    dispatch({ type: UIActionTypes.NAVIGATE_NEXT_PAGE });
-  } catch (error) {
-    dispatch({
-      type: Types.REGISTRATION_FAILURE,
-      error
-    });
-  }
+  }, 800);
 };
 
-export const submitInstanceInfo = (
+export const validateInstanceInfo = (
   data: InstanceInfoModel,
   nextStep?: string
 ): OcimThunkAction<void> => async dispatch => {
@@ -66,22 +81,23 @@ export const submitInstanceInfo = (
     type: Types.REGISTRATION_SUBMIT,
     data
   });
-
-  try {
-    // try submitting form data
-    // TODO
-    if (data.publicContactEmail === 'existing') {
-      throw Error('Test error');
+  setTimeout(() => {
+    try {
+      // try submitting form data
+      // TODO
+      if (data.publicContactEmail === 'existing') {
+        throw Error('Test error');
+      }
+      dispatch({ type: Types.REGISTRATION_SUCCESS, data });
+      if (nextStep) {
+        dispatch(push(nextStep));
+      }
+      dispatch({ type: UIActionTypes.NAVIGATE_NEXT_PAGE });
+    } catch (error) {
+      dispatch({
+        type: Types.REGISTRATION_FAILURE,
+        error
+      });
     }
-    dispatch({ type: Types.REGISTRATION_SUCCESS, data });
-    if (nextStep) {
-      dispatch(push(nextStep));
-    }
-    dispatch({ type: UIActionTypes.NAVIGATE_NEXT_PAGE });
-  } catch (error) {
-    dispatch({
-      type: Types.REGISTRATION_FAILURE,
-      error
-    });
-  }
+  }, 800);
 };
--- a/frontend/src/registration/actions.ts
+++ b/frontend/src/registration/actions.ts
@@ -9,8 +9,8 @@ import {
 } from './models';
 
 export enum Types {
-  // Support action to update root state
-  ROOT_STATE_UPDATE = 'ROOT_STATE_UPDATE',
+  // Support action to update root state and clean error messages when users change fields
+  CLEAR_ERROR_MESSAGE = 'CLEAR_ERROR_MESSAGE',
   // Data validation actions
   REGISTRATION_VALIDATION = 'REGISTRATION_VALIDATION',
   REGISTRATION_VALIDATION_SUCCESS = 'REGISTRATION_VALIDATION_SUCCESS',
@@ -21,9 +21,9 @@ export enum Types {
   REGISTRATION_FAILURE = 'REGISTRATION_FAILURE'
 }
 
-export interface RootStateUpdate extends Action {
-  readonly type: Types.ROOT_STATE_UPDATE;
-  readonly data: RegistrationStateModel;
+export interface FeedbackMessageChange extends Action {
+  readonly type: Types.CLEAR_ERROR_MESSAGE;
+  readonly data: RegistrationFeedbackModel;
 }
 
 export interface RegistrationValidation extends Action {
@@ -32,7 +32,7 @@ export interface RegistrationValidation extends Action {
 
 export interface RegistrationValidationSuccess extends Action {
   readonly type: Types.REGISTRATION_VALIDATION_SUCCESS;
-  readonly data: RegistrationStateModel;
+  readonly data: RegistrationModel;
 }
 
 export interface RegistrationValidationFailure extends Action {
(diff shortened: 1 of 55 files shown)


frontend/src/ui/components/DomainInput/DomainInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <FormControl
+          value={props.domainName}
+          onChange={(event: any) => {
+            props.handleDomainChange(event.target.value);
+          }}
+          isInvalid={isInvalid}
+        />
+        {props.internalDomain === true && (
+          <InputGroup.Append>
+            <InputGroup.Text>.opencraft.hosting</InputGroup.Text>
```
Agrendalath: Could we move this to the configs instead of hardcoding it? It reduces the risk of typos.

frontend/src/ui/components/DomainSuccessJumbotron/DomainSuccessJumbotron.tsx
```diff
@@ -28,7 +28,10 @@ export const DomainSuccessJumbotron: React.FC<DomainProps> = (
         <WrappedMessage id={domainStatusText} messages={messages} />
       </h2>
       <div className="domain-name">
-        <p>{props.domain}</p>
+        <p>
+          {props.domain}
+          {props.domainIsExternal === false && <span>.opencraft.hosting</span>}
```
Agrendalath: Nit: isn't ` === false` redundant? `domainIsExternal` is already a `boolean`.

Could we move this suffix to the configs instead of hardcoding it? It reduces the risk of typos.

frontend/src/registration/components/DomainInputPage/DomainInputPage.tsx
```diff
@@ -1,51 +1,58 @@
...
     });
   };
 
-  private submitForm = () => {
+  private submitDomain = () => {
     this.props.submitRegistration(
-      { domain: this.state.domainName },
+      { domain: this.props.domain },
       ROUTES.Registration.INSTANCE
     );
```
Agrendalath: Are we planning to do some domain validation on the frontend?
Something like `^[a-zA-Z0-9][a-zA-Z0-9-]{1,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$` should work.

frontend/src/registration/components/DomainInputPage/displayMessages.ts
```diff
@@ -1,12 +1,4 @@
 const messages = {
-  typeDomainNameBelow: {
-    defaultMessage: 'Type your domain name below',
-    description: 'Message above domain input text box.'
-  },
-  checkAvailability: {
-    defaultMessage: 'Check Availability',
-    description: 'Label for button that check availability of Domain'
-  },
```
Agrendalath: Won't we use it later for own domains?

frontend/src/registration/components/InstanceSetupPage/InstanceSetupPage.tsx
```diff
@@ -58,12 +95,25 @@ export class InstanceSetupPage extends React.PureComponent<Props> {
...
+              type="email"
+              name="publicContactEmail"
+              value={this.props.publicContactEmail}
+              onChange={this.onChange}
+            />
             <p>
               <WrappedMessage id="publicContactEmailHelp" messages={messages} />
             </p>
           </FormGroup>
         </Form>
```
Agrendalath: We should add some validation to this form - instance name shouldn't be empty and email should match something like `/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)\|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])\|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/`.

frontend/src/registration/components/InstanceSetupPage/InstanceSetupPage.tsx
```diff
@@ -4,33 +4,66 @@ import { Form, FormControl, FormGroup, FormLabel } from 'react-bootstrap';
...
+    loading: getRegistrationData(state, 'loading')
   }),
   {
-    submitRegistration
+    validateInstanceInfo,
+    updateInstanceInfoState
   }
 )
-export class InstanceSetupPage extends React.PureComponent<Props> {
+export class InstanceSetupPage extends React.PureComponent<Props, StateProps> {
```
Agrendalath: The props of this page (and all other ones in the registration flow) will be lost after refresh. We could make it persistent, e.g. with [redux-storage](https://github.com/react-stack/redux-storage).

giovannicimolin: I think this is the intended behavior of the current implementation. If the user want's to change anything, he can go back using the navigation buttons or browser back button.
If the page is refreshed it should start from scratch.

frontend/src/registration/components/DomainInputPage/displayMessages.ts
giovannicimolin: This was moved to `ui/components`, the file is still there. :)

Agrendalath: @giovannicimolin, oh, sorry, I missed this :)

frontend/src/registration/components/InstanceSetupPage/InstanceSetupPage.tsx
Agrendalath: domain
I don't think it's a user-friendly behavior ;)

> If the page is refreshed it should start from scratch.

From the user perspective I would expect data to be retained for the completed steps when I refresh the page, especially as these steps have different routes.

giovannicimolin: Oh, now I got what you meant. Yeah, this was actually a bug in the reducers that erased and replaced the global state. 
It's fixed now (I tested by going back and forth multiple times).

giovannicimolin: Ok, this was partially fixed :stuck_out_tongue: 
I will add `redux-storage` to fix this.

frontend/src/registration/components/InstanceSetupPage/InstanceSetupPage.tsx
xitij2000: I think we can just rely on the server side for this. Otherwise we will need to keep the validation in sync.

frontend/src/registration/actions.ts
```diff
@@ -2,58 +2,131 @@ import { push } from 'connected-react-router';
...
-import { RegistrationModel } from './models';
+import {
+  RegistrationModel,
+  RegistrationStateModel,
+  RegistrationFeedbackModel
+} from './models';
 
 export enum Types {
+  // Support action to update root state
+  ROOT_STATE_UPDATE = 'ROOT_STATE_UPDATE',
```
xitij2000: Instead of this call it 'REGISTRATION_UPDATE'

frontend/src/registration/actions.ts
```diff
@@ -2,58 +2,131 @@ import { push } from 'connected-react-router';
...
 export type ActionTypes =
+  | RootStateUpdate
+  | RegistrationValidation
+  | RegistrationValidationSuccess
+  | RegistrationValidationFailure
   | SubmitRegistration
   | RegistrationSuccess
   | RegistrationFailure;
 
+export const updateRootState = (
```
xitij2000: rename to `updateRegistraiton`

frontend/src/registration/actions.ts
```diff
@@ -2,58 +2,131 @@ import { push } from 'connected-react-router';
...
+  data: RegistrationStateModel
+): OcimThunkAction<void> => async dispatch => {
+  dispatch({ type: Types.ROOT_STATE_UPDATE, data });
+};
```
xitij2000: This need not be a ThunkAction since it's not calling any APIs. 
A simplified system can be used like the following:
```javascript
import { createAction } from 'redux-actions';
export const updateRegistration = createAction(Types.ROOT_STATE_UPDATE);
```
Here though it would be better to have:
```javascript
export const updateRegistration = (field: keyof RegistrationStateModel, value) => async dispatch => 
dispatch({ type: Types.REGISTRATION_UPDATE, field, value });
```

frontend/src/registration/components/AccountSetupPage/AccountSetupPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+@connect<StateProps, ActionProps, {}, Props, RootState>(
+  (state: RootState) => ({
+    loading: state.registration.loading,
+    registrationData: state.registration.registrationData,
+    registrationFeedback: state.registration.registrationFeedback
+  }),
+  {
+    submitRegistration,
+    updateRootState
```
xitij2000: -> updateRegistration

frontend/src/registration/components/AccountSetupPage/AccountSetupPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const { value } = e.target;
+
+    this.props.updateRootState({
+      registrationData: {
+        [field]: value
+      },
+      registrationFeedback: {
+        [field]: ''
+      }
+    });
```
xitij2000:
```javascript
    this.props.updateRegistration({[field]: value});
```

frontend/src/registration/reducers.ts
```diff
@@ -1,19 +1,63 @@
...
-export const initialState: Readonly<RegistrationModel> = blankRegistration;
+export const initialState: Readonly<RegistrationStateModel> = blankRegistrationState;
 
 export function registrationReducer(
   state = initialState,
   action: RegistrationActions.ActionTypes
-): RegistrationModel {
+): RegistrationStateModel {
   switch (action.type) {
+    case RegistrationActions.Types.ROOT_STATE_UPDATE:
```
xitij2000:
```suggestion
    case RegistrationActions.Types.REGISTRATION_UPDATE:
```

frontend/src/registration/reducers.ts
```diff
@@ -1,19 +1,63 @@
...
+        loading: state.loading,
+        registrationData: {
+          ...state.registrationData,
+          ...action.data.registrationData
+        },
+        registrationFeedback: {
+          ...state.registrationFeedback,
+          ...action.data.registrationFeedback
+        }
+      };
```
xitij2000: This might be a good time to include [immutability helper](https://github.com/kolodny/immutability-helper) which will reduce this to:
```javascript
return update(state, {
    registrationData: {[field]: {$set: value}},
    registrationFeedback: {[field]: {$set: '}},
});
```

Agrendalath:
@giovannicimolin, I'm getting these errors in the console for the second step (`registration/instance`), but you don't need to address this in the current ticket.
err

giovannicimolin:
@Agrendalath Thanks for the review. I've fixed the console issue, as well as updated the snapshot tests and added a few new tests.

PR #503, opened 2019-12-09
[BB-1874] Add new email templates 

xitij2000:
This PR adds new email templates and a new email and setup for sending HTML emails. A new email is also added in the email process after instance creation but before the site is fully provisioned. 

Testing:
- Setup email setting in the dev environment 
- Go through the registration process
- Ensure you get all three emails at the correct time with the correct details. 

Optionally, you can temporary add the email templates as templateviews to view them in a browser. The code can then be sent using a tool like https://www.pilotmail.io/ to see if it work in an email client

Sample code to test welcome emails. 
```python
def send_welcome_email_test(email: str) -> None:
    """
    Send a test welcome email.
    """
    context = dict(
        user_name="John Doe",
        lms_link="http://my.edx.site",
        studio_link="http://studio.my.edx.site",
        # TODO: add in once implemented
        customise_link="http://ocim.site/customise/",
    )
    html_email_helper(
        template_base_name='emails/welcome_email',
        context=context,
        subject=settings.WELCOME_EMAIL_SUBJECT,
        recipient_list=(email,)
    )
```


opencraft/urls.py
```diff
@@ -42,6 +42,38 @@
...
+            theme="Dark",
+            primary_color="#000000",
+            secondary_color="#dddddd",
+            logo_url="http://logo",
+            logo_file="logo.png",
+            header_url="http://header",
+            header_file="header.png",
+        )),
+        name="email-test-info"
+        ),
```
xitij2000: @Agrendalath These are just for testing the emails with some data. I will delete them before merge.

opencraft/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    text_message = text_template.render(email_context)
+
+    html_template = get_template(f'{template_base_name}.html')
+    html_message = html_template.render(email_context)
+
+    send_mail(
+        subject=subject,
+        message=text_message,
+        html_message=html_message,
+        from_email=settings.DEFAULT_FROM_EMAIL if from_email is None else from_email,
```
Agrendalath: Nit: do we have a use-case for "falsy, but not None" `from_email` variable here? This one-liner with reversed check is a little hard to read. IMO `from_email or settings.DEFAULT_FROM_EMAIL` could be a little cleaner.

opencraft/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        scheme = "http"
+    else:
+        scheme = "https"
+    return f"{scheme}://{domain}{relative_path}"
+
+
+def build_email_context(context=None):
+    """
+    Builds a context object with common settings for emails.
+    """
```
Agrendalath: Please add a docstring about `context` and information that it overrides generated values.

templates/emails/welcome_email.html
```diff
@@ -0,0 +1,78 @@
...
+                </a>
+            </td>
+            <td>
+                <a href="{{ studio_link }}">
+                    Studio
+                </a>
+            </td>
+            <td>
+                <a href="{{ customise_link }}">
+                    Customise
```
Agrendalath: Nit: on mockups it was "Customize".

registration/utils.py
```diff
@@ -35,4 +37,77 @@ def verify_user_emails(user: User, request: HttpRequest, *email_addresses: str):
...
+        studio_link="http://studio.my.edx.site",
+        # TODO: add in once implemented
+        customise_link="http://ocim.site/customise/",
+    )
+    html_email_helper(
+        template_base_name='emails/welcome_email',
+        context=context,
+        subject=settings.WELCOME_EMAIL_SUBJECT,
+        recipient_list=(email,)
+    )
```
Agrendalath: Is it something that should be removed (as well as the URLs) before merging?

registration/provision.py
```diff
@@ -77,4 +77,7 @@ def _provision_instance(sender, **kwargs):
         )
         application.instance.lms_users.add(user)
         application.save()
+        # At this point we know the user has confirmed their email and set up an instance.
+        # So we can go ahead and send the account info email.
+        transaction.on_commit(lambda: send_account_info_email(application))
```
Agrendalath: Neat! :)

templates/emails/welcome_email.html
```diff
@@ -0,0 +1,78 @@
...
+                </a>
+            </td>
+        </tr>
+    </table>
+    <div class="block">
+        <h2>Open edX documentation</h2>
+        <p>
+            All product documentation for Open edX can be found at
+            <a href="https://docs.edx.org/.">
```
Agrendalath: Nit: should this dot be here?

static/css/emails.css
```diff
@@ -0,0 +1,261 @@
...
+    color: #0f1f24;
+    background-color: #e2ebef;
+    text-align: left;
+    padding: 16px;
+    margin: 16px 0;
+}
+
+.block h2 {
+    margin: 0;
+    font: 20px bold Times, serif;
```
Agrendalath: The `welcome_email` looks like on the mockups without this `font` line. Should we actually use it there?

templates/emails/email_base.html
```diff
@@ -0,0 +1,42 @@
...
+    </style>
+</head>
+{% inlinecss "css/emails.css" %}
+    <body>
+    <div class="main">
+        <img class="opencraft-logo" src="{{ base_url }}{% static 'img/svg/logo-for-email.svg' %}"
+             alt="Opencraft Logo" aria-hidden="true">
+        <div class="main">
+            <img src="{{ base_url }}{% static 'img/jpg/header-background.jpg' %}" alt="Email Header" aria-hidden="true">
+            <div class="content">
```
Agrendalath: layout

@xitij2000, there is a weird padding between header and content and the header is not scaled correctly.

email_verification/__init__.py
```diff
@@ -35,27 +34,22 @@
...
                                      'Please verify this email address')
 EMAIL_VERIFICATION_TEMPLATE = getattr(settings, 'EMAIL_VERIFICATION_TEMPLATE',
-                                      'email_verification/email.txt')
+                                      'emails/verify_email')
 
 
 # Functions ###################################################################
 
-def send_email_verification(email, request):
+def send_email_verification(email: str):
```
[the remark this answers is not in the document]
xitij2000: @Agrendalath Sorry yeah, I refactored this earlier to take the email address directly instead of the model but I guess I undid that. I now remember why I did so. Adding a typehint here breaks things. It imports a model before Django is ready, so I will just remove the hint, or add it in documentation.

static/css/emails.css
```diff
@@ -0,0 +1,261 @@
...
+    color: white;
+    line-height: 1.31;
+}
+
+/*
+CSS for verification email
+*/
+
+.content h1 {
+    font: 24px 'Times New Roman', serif;
```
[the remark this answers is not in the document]
xitij2000: @Agrendalath Ideally yes, but we don't have too much control over that since we're running an inlining script. It is because of this error that I changed to single quotes in CSS, where I usually use double quotes. But the inliner always produces this.

templates/emails/email_base.html
```diff
@@ -0,0 +1,42 @@
+{% load static %}
+{% load inlinecss %}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="Content-Type" content="text/html charset=UTF-8"/>
+    <title>Confirm your email address</title>
```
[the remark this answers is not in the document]
xitij2000: I added it mainly for testing in the browser. I can definitley remove this.

templates/emails/email_base.html
```diff
@@ -0,0 +1,42 @@
...
+            <div class="addendum">
+                <p>
+                    And if you need anything, or have any questions
+                    or remarks, don't hesitate to reply to this email.
+                </p>
+            </div>
+        </div>
+        <div class="footer">
+            <p>
+                <a href="{{ unsubscribe_url }}">Unsubscribe</a> | edX and Open edX are registered trademarks of edX Inc.
```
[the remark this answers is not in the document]
xitij2000: I think this is more of a TODO item for now. Perhaps I should add an explicit TODO here. 
Similar to the instance customisation link. If we have a link to the profile edit page this can point to that since we'll have an option to disable update emails. 

Or we can remove this entirely since the emails right now are triggered by user actions and are not a subscription.

registration/utils.py
xitij2000: I was wondering if it's OK to keep it around as a way to test the email setup. I can remove it also.

templates/emails/email_base.html
xitij2000: This is odd. I was not getting this while testing, but then again I moved on from verify to other emails so perhaps I broke the verify emails at some point.

email_verification/__init__.py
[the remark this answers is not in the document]
xitij2000: > Also, should we move it to `registration/utils.py` to group all methods for sending these emails in one module?

So I think it's best to keep it where it is since it's closer to email verification then registration, and can happen outside the context of registraiton.

templates/emails/email_base.html
xitij2000: Found the issue! Linking to an asset from localhost doesn't work on gmail so I had uploaded these assets to an s3 bucket of mine and linked from there temporarily. When switching back to the regular image I forgot to copy over the class I'd applied.

templates/emails/email_base.html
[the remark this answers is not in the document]
xitij2000: It seems that having the subject as the title could be useful since some email apps use it for notifications and such. So I've make it something you can specify in the template.

email_verification/__init__.py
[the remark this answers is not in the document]
Agrendalath: > So I think it's best to keep it where it is since it's closer to email verification then registration, and can happen outside the context of registraiton.

Sure, sounds reasonable.

templates/emails/email_base.html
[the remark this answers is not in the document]
Agrendalath: > Or we can remove this entirely since the emails right now are triggered by user actions and are not a subscription.

It's fine to leave it with a TODO here :+1: 
We might need to add it later to some emails anyway.

registration/utils.py
Agrendalath: It's fine to keep this commented out here with a note that you can uncomment it for testing.
Or we can add this part of code to testing instructions in case we wanted to test it locally. I'll leave it up to you :)

templates/emails/email_base.html
Agrendalath: Ah, got it. Looks great now!

registration/utils.py
xitij2000: @Agrendalath commented out code quickly goes out of date since it isn't ever quality checked. 

I'll move it to the testing instructions isntead.

PR #505, opened 2019-12-17
BB-1970: Connect to backend

giovannicimolin:
This PR connects the new Ocim frontend to the v2 backend APIs and adds a few missing features to the registration form.

**Testing instructions:**
1. Run the backend and frontend from this branch.
2. Go through the registration form and complete a registration.
3. Check that the user was created.
4. Click on the email validation links (they are printed on the backend console output).
5. Check that the instance is created and starts provisioning.
6. Check that the field validation is working.
6. Try to break the form: start on a wrong page (such as `/registration/instance`).

**Reviewer:**
- [ ] @xitij2000


api/auth.py
```diff
@@ -0,0 +1,22 @@
...
+
+class TokenSerializer(serializers.Serializer):
+    refresh = serializers.CharField()
+    access = serializers.CharField()
+
+class TokenErrorSerializer(serializers.Serializer):
+    details = serializers.CharField()
+
+# Override method to fix swagger schema
+class JWTAuthToken(TokenObtainPairView):
```
giovannicimolin: @xitij2000 I had to override the method and provide correct response so that the auto generated client could work.

frontend/package.json
```diff
@@ -54,7 +55,8 @@
     "eject": "react-scripts eject",
     "lint": "eslint src/ --ext .ts,.tsx",
     "lint-fix": "eslint src/ --ext .ts,.tsx --fix",
-    "prettier": "prettier --write src/**/*.ts src/**/*.tsx"
+    "prettier": "prettier --write src/**/*.ts src/**/*.tsx",
+    "update-api-client": "./scripts/update-api-client.sh"
```
giovannicimolin: @xitij2000 Handy to auto update API client.

frontend/src/registration/actions.ts
```diff
@@ -72,96 +75,102 @@ export const clearErrorMessage = (
...
+  } else {
+    registrationFeedback.passwordConfirm = "The password confirmation should match the password from the field above."
+  }
+
+  if (Object.entries(registrationFeedback).length === 0) {
+    V2Api.accountsCreate({
+      data: { ...registrationData }
+    }).then(() => {
+      // Perform authentication and create new instance
+      let x = dispatch(performLogin({
```
giovannicimolin: @xitij2000 I need help here.

I created an user account using `accountsCreate` above, then when it succeeds, I call `performLogin` and make the request to return a JWT token. The thing is that I'm not sure how to link these tasks, I need the response from `performLogin` (or at least an updated root state) to be able to call the instance creation endpoint.

I see that the request and API response are correct, but I don't know how to chain these tasks:
1. Create account (already here)
2. Login (almost done, need to save tokens to state)
3. Create instance (not here yet)

Should I chain `.then(` and nest the logic? If not, what is the best way of doing this?

frontend/package.json
xitij2000: @giovannicimolin  Could we have a step in the CI try to auto-generate the api-client and if that modifies files we can fail the deployment? This will make sure the API client is always up to date.

frontend/src/registration/actions.ts
xitij2000: Yup, I think it makes sense to chain the events. So it should go as follows:
1. Call the create account endpoint.
2. If successful call the login endpoint
3. If successful save the token by dispatching a LOGIN_SUCCESS event.

Now in the next steps you can resume the process.

api/auth.py
xitij2000: We might want to look into doing this through the customisation support provided by swagger-codegen.

frontend/src/auth/actions.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export type ActionTypes = SubmitLogin | LoginSuccess | LoginFailure | Logout;
+
+export const performLogin = (
+  data: LoginFormModel
+): OcimThunkAction<void> => async dispatch => {
+  await V2Api.authTokenCreate({
+    data: { ...data }
+  })
+    .then(response => {
```
xitij2000: TypeScript is complaining about this variable not having a type. I think you can set it to `Token` and `import { Token } from 'ocim-client';`

frontend/src/registration/actions.ts
```diff
@@ -72,96 +74,118 @@ export const clearErrorMessage = (
...
+    })
+    .catch((e: any) => {
+      e.json().then((feedback: any) => {
+        // If validation fails, return error to form through state
+        const error = { ...feedback };
+        // Loop at each error message and join them.
+        // Also convert keys from snake_case to camelCase
+        Object.keys(error).forEach(key => {
+          error[toCamelCase(key)] = error[key].join();
+        });
```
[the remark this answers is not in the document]
giovannicimolin: @xitij2000 The mechanism I use to link the feedback messages to the fields requires them to have the same name. 
It'd be good if we could make the autogenerated client to convert the values automatically, but since that would take a lot of time I just converted here.

frontend/src/global/api.ts
```diff
@@ -0,0 +1,31 @@
...
+  // middleware: [
+  //     {
+  //         post: async (context) => {
+  //             if (context.response.status === 401) {
+  //                 redirectToLogin();
+  //             }
+  //             return context.response;
+  //         },
+  //     },
+  // ],
```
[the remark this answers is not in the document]
giovannicimolin: I was planning to use this to add a token refresh and redirect if logged out, but since it's out of the scope of this ticket I'll remove it.

opencraft/settings.py
```diff
@@ -144,6 +146,8 @@
 
 ROOT_URLCONF = 'opencraft.urls'
 
+CORS_ORIGIN_ALLOW_ALL = True
```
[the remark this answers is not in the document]
giovannicimolin: This needs to be properly set up on production. This setting just makes testing easier on the devstack.

frontend/src/auth/actions.ts
```diff
@@ -0,0 +1,67 @@
...
+      // Perform authentication and create new instance
+      dispatch({ type: Types.LOGIN_SUCCESS, data: { ...response } });
+      // Save auth data to localStorage so the API client picks it up
+      window.localStorage.setItem('token_access', response.access);
+      window.localStorage.setItem('token_refresh', response.refresh);
+    })
+    .catch((e: any) => {
+      e.json().then((feedback: any) => {
```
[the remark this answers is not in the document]
giovannicimolin: @xitij2000 Can I handle this with an Error boundary pointing to an error page?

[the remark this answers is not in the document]
giovannicimolin: Nvm, the error boundary doesn't catch these types of things.

frontend/src/registration/actions.ts
```diff
@@ -72,96 +76,124 @@ export const clearErrorMessage = (
...
+      })
+      .catch((e: any) => {
+        e.json().then((feedback: any) => {
+          // If validation fails, return error to form through state
+          const error = { ...feedback };
+          // Loop at each error message and join them.
+          // Also convert keys from snake_case to camelCase
+          Object.keys(error).forEach(key => {
+            error[toCamelCase(key)] = error[key].join();
+          });
```
xitij2000: This code has appeared twice, and will likely show up again, so perhaps convert it into a utility function. Then this can become:

`const error = toCamelCaseKeys(feedback)`

frontend/src/registration/actions.ts
```diff
@@ -72,96 +76,124 @@ export const clearErrorMessage = (
...
+        });
       });
-    }
-  }, 800);
+  } else {
+    // Failing local validation
+    dispatch({
+      type: Types.REGISTRATION_VALIDATION_FAILURE,
+      error: { ...registrationFeedback }
```
xitij2000:
```suggestion
      error: registrationFeedback 
```

api/urls.py
```diff
@@ -40,11 +42,18 @@
...
     url(r'^$', RedirectView.as_view(url='v1/', permanent=False), name='index'),
+    # v1 urls
     url(r'^v1/', include((v1_router.urls, 'api_v1'), namespace='v1')),
-    url(r'^v2/', include((v2_router.urls, 'api_v2'), namespace='v2')),
     url(r'^v1/auth/', include('rest_framework.urls', namespace='rest_framework')),
+    # v2 urls
+    url(r'^v2/', include((v2_router.urls, 'api_v2'), namespace='v2')),
+    url(r'^v2/auth/token/', JWTAuthToken.as_view(), name='token_obtain_pair'),
+    # url(r'^v2/auth/refresh/', TokenRefreshView.as_view(), name='token_refresh'),
+    # url(r'^v2/auth/verify/', TokenVerifyView.as_view(), name='token_verify'),
```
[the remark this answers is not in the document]
giovannicimolin: These aren't used yet, and I haven't created serializers to fix the endpoints documentation, so I'll remove them for now.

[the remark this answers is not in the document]
xitij2000: It's OK to leave them around, with a comment about the same.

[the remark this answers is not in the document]
giovannicimolin: Done!

PR #507, opened 2020-01-03
[BB-2003] Remove the 'clean' target dependency for the 'test.integration_cleanup' target

lgp171188:
It runs a lot of stuff (including `coverage`) which are not needed in the cleanup job.

**Testing instructions**:
* After the PR is merged (the job runs only on the master branch), verify that the cleanup job doesn't run the commands in the `clean` target and that the error about the missing `coverage` command is not thrown. Since the cleanup job runs as a scheduled task, we will have to wait till the next scheduled run.
* Alternatively, use the [CircleCI local CLI tool](https://circleci.com/docs/2.0/local-cli/) to run the cleanup job locally after [converting the configuration file](https://circleci.com/docs/2.0/local-cli/#running-a-job) and verifying that the execution proceeds past the step that is currently failing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ test.migrations_missing: clean ## Check if migrations are missing.
 test.browser: clean static_external ## Run browser-specific tests.
 	@echo -e "\nRunning browser tests..."
 	xvfb-run --auto-servernum $(HONCHO_COVERAGE_TEST) --pattern=browser_*.py
-	
+
 test.integration: clean ## Run integration tests.
 ifneq ($(wildcard .env.integration),)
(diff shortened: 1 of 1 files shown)


lgp171188:
The integration tests are failing because of the Gandi v5 migration issue (unrelated to this PR) and require the corresponding changes to be merged in order to pass.

lgp171188:
@xitij2000, without any code changes, the cleanup job has been running okay for the past few days. I am guessing that these errors happen when the cleanup job is run on a different container/environment which doesn't already have the virtualenv environment from the regular CI runs. This fix is still relevant and will prevent any errors in the future.

lgp171188:
Merged as a part of PR #508

PR #508, opened 2020-01-10
[BB-2008] Migrate to Gandi v5 LiveDNS API

lgp171188:
This PR adds the code changes required to use the Gandi v5 LiveDNS API to manage the DNS records of the Ocim domains.

**Update**:

This also contains the fixes to:
* fix the failing integration cleanup job.
* fix the `xqueue` service error breaking provisioning.

**Testing instructions**:
* Ensure that the CI checks pass and there are no errors in the integration tests related to managing the DNS records.
* Check out the source branch of this PR and run Ocim.
* Create a new instance and spawn a new appserver for it.
* Check and verify that the DNS records for the instance are created during provisioning and the added records are present in the zone file for the domain in the Gandi admin interface. After a short while, ensure that the domains in the added records resolve properly.
* Once the instance has completed provisioning, verify that the DNS record for the active appserver VM is added without any errors.
* Archive the instance and verify that the DNS records are removed successfully.

**Testing instructions for other fixes**:
* Verify that the integration cleanup job doesn't fail due to an error about not being able to find the `coverage` command.
* Verify that the instance provisioning completes successfully and the right version of `xqueue` is installed.

* The API key for the account with access to the integration test base domain, doesn't have the permission to list the domains in the account. So the value  of the `DEFAULT_INSTANCE_BASE_DOMAIN` environment variable is added as a fallback. We have to ensure that the API key used in production has all the required permissions.
* The code for the existing XML-RPC API is left intact and can be restored easily with very minimal changes. This can be removed after the migration in production is complete and verified to work without any issues.
* The `cleanup_utils` code changes should work, butt the actual removal of the records couldn't be tested locally due to there being no data left to cleanup.
* This PR has to be merged very close to the deployment on production, after all the preparation for the domain migration is done and a maintenance window for Ocim is announced.


instance/tests/utils.py
```diff
@@ -40,7 +39,11 @@ def patch_gandi(func=None):
     """
     Decorator to temporarily replace the Gandi API by the fake implementation.
     """
-    patcher = patch('instance.gandi.api', gandi.GandiAPI(client=FakeGandiClient()))
+    patcher = patch('instance.gandi.api', FakeGandiV5APIClient())
+    # mocked_set_dns_record = patch.object(gandi.GandiV5API, 'set_dns_record', MagicMock())
```
giovannicimolin: Are these comments still needed? 
If the mocks are required, can you add a comments explaining, if not, can you clean them up?

instance/tests/fake_gandi_client.py
```diff
@@ -162,6 +162,7 @@ def _domain_list(self, api_key):
 
     def _domain_info(self, api_key, domain):
         """
+
```
giovannicimolin: Remove blank line.

instance/gandi.py
```diff
@@ -189,4 +192,157 @@ def remove_dns_record_callback(zone_id, zone_version):
...
+        return result
+
+    def set_dns_record(self, domain, **record):
+        """
+        Set a DNS record. This method takes the mandatory 'domain' parameter to be able to support multiple domains
+        handled by the same Gandi account
+        """
+        if 'ttl' not in record.keys():
+            record['ttl'] = 1200
+            assert 'name' not in record, 'The name gets extracted from the FQDN passed in `domain`'
```
giovannicimolin: Why is this assert required, what happens if it fails?

instance/gandi.py
```diff
@@ -189,4 +192,157 @@ def remove_dns_record_callback(zone_id, zone_version):
...
 
-api = GandiAPI()
+class GandiV5API:
+    """
+    Gandi v5 LiveDNS API client
+    """
+    def __init__(self, api_key, api_base='https://dns.api.gandi.net/api/v5'):
+        self.api_base = api_base
+        self.api_key = api_key
+        self._domain_cache = None
```
giovannicimolin: Shouldn't we populate the domain cache when instancing the class? 
Can you add a call to `_populate_domain_cache` here?

instance/gandi.py
```diff
@@ -189,4 +192,157 @@ def remove_dns_record_callback(zone_id, zone_version):
...
+
+    def _populate_domain_cache(self):
+        """
+        Populate the internal domain cache with all domains in the current Gandi account.
+        """
+        self._domain_cache = []
+        domains_api_url = '{}/domains'.format(self.api_base)
+
+        try:
```
giovannicimolin: Do you think it's worth trying to upstream this method of retrieving domains in account to Lexicon? If so, and they already have some sort of support for it on other providers, can you create an upstream ticket?

instance/gandi.py
```diff
@@ -189,4 +192,157 @@ def remove_dns_record_callback(zone_id, zone_version):
...
+        lexicon_config['action'] = 'create'
+        lexicon_config['type'] = record['type']
+        lexicon_config['name'] = record['name']
+        lexicon_config['content'] = record['value']
+        lexicon_config['ttl'] = record['ttl']
+        config = ConfigResolver()
+        config.with_dict(dict_object=lexicon_config)
+        client = Client(config)
+        result = client.execute()
+        return result
```
giovannicimolin: Can we do some basic error handling internally? 
We should handle the result internally and throw a known error when an issue happens. 
Example: 
- **If adding a DNS record succeeds:** no need to return anything
- **If it fails:** return a custom exception class with error (`SetDnsFailed` or something like that that that allows error handling and logging on modules using this class).

instance/gandi.py
lgp171188: @giovannicimolin, the domain cache is needed only to split the given hostname into the registered domain name and its subdomain in the `_split_domain_name()` method and that's why it is called there before performing the split operation.

instance/gandi.py
lgp171188: @giovannicimolin, the lexicon code only returns an `Exception` instance with a custom error message. The only improvement that can be done here is to catch all exceptions, log an error message and re-raise the exception since we cannot proceed further without setting the DNS record. There is no error handling in the XML-RPC API client code as well.

instance/gandi.py
lgp171188: @giovannicimolin, when the assertion fails, it raises an `AssertionError` with the given message. I borrowed this assert statement from the `GandiAPI` client code.

instance/tests/utils.py
lgp171188: @giovannicimolin, I had added these while testing the changes. Will remove.

instance/gandi.py
lgp171188: @giovannicimolin, the domain cache is needed only to split the given hostname into the registered domain name and its subdomain in the `_split_domain_name()` method and that's why it is called there before performing the split operation.

instance/gandi.py
lgp171188: > Do you think it's worth trying to upstream this method of retrieving domains in account to Lexicon? If so, and they already have some sort of support for it on other providers,

@giovannicimolin, `lexicon` explicitly mentions that it is for manipulating DNS records. From what I can see, all the lexicon operations require a domain name. I checked the [base argparse configuration](https://github.com/AnalogJ/lexicon/blob/master/lexicon/parser.py#L9) for the CLI and also some of the main providers and couldn't find this sort of operation supported anywhere. 

Hence imho this feels out of scope for upstreaming, It is just a simple `GET` API call and can be done in our code itself.

giovannicimolin: Makes sense.

cleanup_utils/dns_cleanup.py
```diff
@@ -173,3 +175,28 @@ def run_cleanup(self, hashes_to_clean):
                         zone_id=self.zone_id,
                         zone_version_id=new_zone_version
                     )
+
+
+class DNSCleanupInstance:
```
giovannicimolin: Nit: `DNSCleanupInstance` -> `DnsCleanupInstance`.
You can keep the old one as `OldDnsCleanupInstance`.

cleanup_utils/integration_cleanup.py
```diff
@@ -31,7 +31,7 @@
 from pytz import UTC
 
 from cleanup_utils.aws_cleanup import AwsCleanupInstance
-from cleanup_utils.dns_cleanup import DnsCleanupInstance
+from cleanup_utils.dns_cleanup import DNSCleanupInstance
```
giovannicimolin: Nit: `CamelCaseNaming` should always be 1 uppercase + lowercase characters (

cleanup_utils/integration_cleanup.py
```diff
@@ -112,8 +112,7 @@ def run_integration_cleanup(dry_run=False):
     mysql_cleanup.run_cleanup()
 
     # Run DNS cleanup
-    dns_cleanup = DnsCleanupInstance(
-        zone_id=int(os.environ['GANDI_ZONE_ID']),
+    dns_cleanup = DNSCleanupInstance(
```
giovannicimolin: Nit: here too, `DnsCleanupInstance`

cleanup_utils/integration_cleanup.py
lgp171188: > Nit: CamelCaseNaming should always be 1 uppercase + lowercase characters 

@giovannicimolin, not when a word in the name is an abbreviation like DNS :stuck_out_tongue: For example, we have `OpenEdXInstance`, `OpenEdXAppServer`, `DBHandler`, `MySQLServer`, and even `GandiAPI` for the v4 API client. So I would like to keep the naming as-is. What do you say?

giovannicimolin: Keep it as is, it's not really a blocker, just a minor nit. :)

PR #509, opened 2020-01-11
[BB-1913] Delete SSH key upon terminating the AppServer

Agrendalath:
This PR is a follow-up to #495. It removes SSH key from `~/.ssh/known_hosts` before we terminate the sever. This way we won't encounter SSH key conflict when we get the server with the same IP we've used before.

## Testing instructions.
1. Log into stage.
1. Create new AppServer for any existing instance.
1. Log to stage via SSH and switch to the proper user.
1. Run `ssh-keygen -F $IP` to check that the key has been added to `~/.ssh/known_hosts`. If the response is empty, try again in a minute.
1. Use `Terminate` button in Ocim.
1. Run `ssh-keygen -F $IP` to confirm that the key has been deleted.

Note: integration tests are failing, but they're not related to this change.

### DIFF
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -493,4 +493,3 @@ class OpenStackServer(Server):
         self.logger.info('Deleting SSH key of "%s" host.', ip)
         command = f'ssh-keygen -R {ip}'
         subprocess.Popen(command, shell=True)
-
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -23,6 +23,7 @@ Instance app models - Server
 # Imports #####################################################################
 
 import logging
+import subprocess
 import time
 
 from django.conf import settings
@@ -441,6 +442,8 @@ class OpenStackServer(Server):
             self._status_to_terminated()
             return
 
+        ip = str(self.public_ip)  # We should retrieve this before terminating the server.
+
         try:
             os_server = self.os_server
             server_closed = self._shutdown(os_server)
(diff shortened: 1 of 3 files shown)


instance/models/server.py
```diff
@@ -477,3 +482,14 @@ def _shutdown(self, os_server, poll_interval=10, max_wait=None):
...
+    def _delete_ssh_key(self) -> None:
+        """
+        Delete SSH key from `~/.ssh/known_hosts`.
+
+        We can safely ignore the command's return code, because we just need to be sure that the key has been removed
+        for non-existing server - we don't care about non-existing keys.
+        """
+        self.logger.info('Deleting SSH key of "%s" host.', self.public_ip)
+        command = f'ssh-keygen -R {self.public_ip}'
+        subprocess.Popen(command, shell=True)
```
lgp171188: @Agrendalath, can you add some error handling code here?

instance/models/server.py
```diff
@@ -477,3 +482,17 @@ def _shutdown(self, os_server, poll_interval=10, max_wait=None):
...
+        """
+        Delete SSH key from `~/.ssh/known_hosts`.
+
+        We can safely ignore the command's return code, because we just need to be sure that the key has been removed
+        for non-existing server - we don't care about non-existing keys.
+        """
+        self.logger.info('Deleting SSH key of "%s" host.', self.public_ip)
+        command = f'ssh-keygen -R {self.public_ip}'
+        try:
+            subprocess.Popen(command, shell=True)
```
lgp171188: @Agrendalath, nit: This logs the output of the command to the Ocim logs/console output. I think we should suppress this.

Agrendalath: @lgp171188, done.

Agrendalath: @lgp171188, done.

PR #510, opened 2020-01-12
WIP [SE-1968] Send deployment failure e-mail after all N attempts failed instead of after each attempt

clemente:
This code replaces the deployment failure e-mails, so that instead of being 1 per deployment attempt, they become 1 after all deployment attempts have been used.

Closed in favour of https://github.com/open-craft/opencraft/pull/515, which adds a new e-mail type instead of replacing the existing one.


clemente:
Closed in favour of https://github.com/open-craft/opencraft/pull/515, which adds a new e-mail type instead of replacing the existing one.

PR #517, opened 2020-01-28
BB-2083: Add deployment toolbar + modals, fix tests and lint code

giovannicimolin:
This PR implements and connects the redeployment toolbar to Ocim backend.

**Testing instructions:**
1. Checkout this branch. Run Ocim backend.
2. On another console `cd frontend/` and run the new UI `npm run start`.
3. Create a new betatest instance using the form.
4. Click login and enter your credentials.
5. Wait for the instance to provision (or force the appserver to an active state).
6. Check that the redeployment toolbar is working as expected.
7. Change the instance name and email.
8. Check that the deploy button was enabled.
9. Click deploy.
10. Check that the deploy button got disabled and that now you have an option to cancel the redeployment.
11. Cancel the redeployment.
12. Deploy again, and wait until it's provisioned.
13. Check code quality & implementation.

This is currently running on stage, so feel free to test there too.

**Reviewer:**
- [ ] @kaizoku


frontend/src/console/actions.ts
```diff
@@ -70,3 +111,57 @@ export const listUserInstances = (): OcimThunkAction<void> => async dispatch =>
...
+        [fieldName]: value
+      }
+    });
+    // dispatch({
+    //   type: Types.UPDATE_INSTANCE_INFO_FAILURE,
+    //   data: {
+    //     [fieldName]: value
+    //   }
+    // });
+  }, 2000);
```
bradenmacdonald: You can also write this using `await` which lets you include the error handling code uncommented:

```typescript
try {
    await new Promise((resolve) => setTimeout(resolve, 2000)); // sleep for 2 seconds
    dispatch({
        type: Types.UPDATE_INSTANCE_INFO_SUCCESS,
        data: {
          [fieldName]: value
        }
    });
} catch {
    dispatch({
      type: Types.UPDATE_INSTANCE_INFO_FAILURE,
      data: {
        [fieldName]: value
      }
    });
}
```

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -31,6 +34,19 @@ export class InstanceSettingsComponent extends React.PureComponent<
...
   }
 
+  public componentDidUpdate(prevProps: Props) {
+    // Fill fields after finishing loading data
+    if (
+      prevProps.activeInstance.data === null &&
+      this.props.activeInstance.data
+    ) {
+      this.setState({
+        instanceName: this.props.activeInstance.data.instanceName,
```
bradenmacdonald: Whenever you have similar data in `State` and in `Props`, where the state is the possibly new value that the user is entering via a form, and the props is the existing value, I recommend naming the state variable something like `newInstanceName`. Then it's much less confusing to know why it sometimes uses `this.props. ... .instanceName` and sometimes `this.state.instanceName`

frontend/src/console/actions.ts
```diff
@@ -70,3 +111,57 @@ export const listUserInstances = (): OcimThunkAction<void> => async dispatch =>
...
+  dispatch({
+    type: Types.UPDATE_INSTANCE_INFO,
+    fieldName
+  });
+
+  setTimeout(function() {
+    dispatch({
+      type: Types.UPDATE_INSTANCE_INFO_SUCCESS,
+      data: {
+        [fieldName]: value
```
bradenmacdonald: This action will get dispatched after some delay (while waiting for the API response), and the `activeInstance` may have changed in the meantime. So you should check and make sure that `getState().activeInstance.id === instanceId` before you dispatch `UPDATE_INSTANCE_INFO_SUCCESS`.

Otherwise if the user submit an update then changes the active instance, then the API response comes in, the user will see the new (title or whatever) on the wrong instance in the UI.

frontend/src/global/constants.ts
```diff
@@ -47,3 +47,11 @@ export const REGISTRATION_STEPS = [
   ROUTES.Registration.ACCOUNT,
   ROUTES.Registration.CONGRATS
 ];
+
+export enum RedeploymentStatus {
```
bradenmacdonald: There should be a way to have this enum auto-generated in the API client?

giovannicimolin: It will be, this is just the frontend implementation with some dummy values until the APIs land (I expected to land earlier this sprint, but there were some delays).

frontend/src/console/actions.ts
giovannicimolin: Thanks for catching that! There's probably a few places that'll need something like this before dispatching the signal.

frontend/src/global/constants.ts
giovannicimolin: Done. The enum is generated from the API spec now.

bradenmacdonald: Nice :)

kaizoku:
@giovannicimolin, deployment works, but after canceling, it shows "(0 Updates)", and I can't retrigger the deployment without making another UI change.

The code otherwise looks good.

kaizoku:
Looks like that bug is actually due to the backend, and not this code, so it won't be handled in this PR.

:+1:

- [✓] I tested this: (followed test steps)
- [✓] I read through the code
- [✓] I checked for accessibility issues

giovannicimolin:
@kaizoku The bug is in the endpoint that manage the changes, so it's not in the scope of this ticket. I'll create a ticket to fix this in the epic. 

Thanks for the review.

PR #518, opened 2020-01-29
SE-2121 Don't filter custom domains for load balancer

samuelallan72:
This will allow adding extra external domains to be handled by the load
balancer. We don't remove filtering for other areas, because we don't
want to manage dns, etc. for external domains.

**Test instructions**:

- deploy to stage
- choose an openedx instance to test against
- set up dns (cname or A record) for an external domain to point to that instance 
- add the external domain to the extra custom domains list for the instance
- verify that this extra custom domain is added to the load balancer and points to the instance

**Reviewer**:

- [x] @symbolist

### DIFF
--- a/instance/models/mixins/domain_names.py
+++ b/instance/models/mixins/domain_names.py
@@ -166,14 +166,6 @@ class DomainNameInstance(models.Model):
         prefixes = ['studio', 'preview', 'discovery', 'ecommerce']
         return ['{}-{}'.format(prefix, self.internal_lms_domain) for prefix in prefixes]
 
-    def get_custom_domains(self):
-        """
-        Returns a list of custom (internal) sub-domains configured for the instance.
-        """
-        return [
-            domain for domain in self.extra_custom_domains.split("\r\n") if domain.endswith(self.internal_lms_domain)
-        ]
-
     def get_load_balanced_domains(self):
         """
         Return an iterable of domains that should be handled by the load balancer.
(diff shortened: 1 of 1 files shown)


symbolist:
@swalladge The code change seems reasonable to me. Have you already deployed this as part of setting up the instance you are working on? (so we don't have to set up other instances to test with)

swalladge:
@symbolist 

> Have you already deployed this as part of setting up the instance you are working on? 

No, this change has not been deployed anywhere yet.

instance/models/mixins/domain_names.py
```diff
@@ -168,7 +168,7 @@ def get_prefix_domain_names(self):
 
     def get_custom_domains(self):
         """
-        Returns a list of custom sub-domains configured for the instance.
+        Returns a list of custom (internal) sub-domains configured for the instance.
         """
         return [
             domain for domain in self.extra_custom_domains.split("\r\n") if domain.endswith(self.internal_lms_domain)
```
xitij2000: Wouldn't ``.splitlines`` be a better option here.

swalladge: Yes.

swalladge: Just pushed some changes to hopefully make it clearer exactly when and why we're filtering extra_custom_domains.

symbolist:
I tested this by adding a custom domain to an OCIM instance; deploying a new app server, pointing the custom domain to the LB ip in `/etc/hosts` and was able to access the OCIM instance via the custom domain. 👍

symbolist (approved):

swalladge:
Thanks @symbolist! Deployed to stage and prod.

PR #520, opened 2020-02-04
BB-2027: New API endpoints for frontend

giovannicimolin:
This PR implements the following features:
- Adds new `deployments` endpoint, that allows retrieving status of deployment and cancelling latest deployments.
- Fixes Swagger generated documentation
- Toggles a flag that deactivates old appservers for self-deployed instances.

I went with a simple approach and just added the minimally required endpoints/functionality.

**Testing instructions:**
1. Checkout this branch.
2. Run Ocim.
3. Go to the swagger documentation (http://localhost:5000/api/swagger/)
4. Check the new `deployment` endpoints.
5. Create a new betatest instance.
9. Check the status endpoint.
6. Try to cancel the deployment of the first appserver, it should fail.
9. Check the status endpoint.
7. Create a mock `active` appserver.
8. Spawn a new appserver and try stopping the provisioning by making a `DELETE` request to the deployments endpoint.
9. Check the status endpoint.

**Reviewer:**
- [ ] @xitij2000 

**Notes:**
We don't currently have any way of separating user-triggered deployments from OpenCraft deployments, so the user will be able to stop any provisioning for now. Any ideas of how we can implement this easily?


registration/api/v2/views.py
```diff
@@ -235,3 +238,92 @@ def get_queryset(self):
...
+
+            # Get number of undeployed changes
+            num_changes = 0
+            if instance:
+                num_changes += 1 if application.instance_name != instance.name else 0
+                num_changes += 1 if application.privacy_policy_url != instance.privacy_policy_url else 0
+                num_changes += 1 if application.public_contact_email != instance.email else 0
+                if application.use_advanced_theme:
+                    # TODO: Improve counting differences of theming changes
+                    num_changes += 1 if application.draft_theme_config != instance.theme_config else 0
```
xitij2000: I think this structure is a bit confusing. I understand it avoids multiple levels of ifs, however the else 0 looks a bit messy. Could you just switch these to standard ifs without an else? For example:
```suggestion
                    if application.draft_theme_config != instance.theme_config: num_changes += 1 
```

xitij2000: Another approach is:

```python
num_changes = sum((
    application.instance_name != instance.name,
    application.privacy_policy_url != instance.privacy_policy_url,
    application.public_contact_email != instance.email,
    application.use_advanced_theme & application.draft_theme_config != instance.theme_config,
))
```

registration/api/v2/views.py
```diff
@@ -235,3 +238,92 @@ def get_queryset(self):
             return BetaTestApplication.objects.filter()
         else:
             return BetaTestApplication.objects.filter(user=self.request.user)
+
+
+class OpenEdxInstanceDeploymentViewSet(RetrieveDestroyAPIView, GenericViewSet):
```
xitij2000: Shouldn't this endpoint should also have a create option? It could take over the current commit changes API.

giovannicimolin: The frontend already needs to wait to perform a query on this, it waits until the `selectedInstance` is available to make the query, and only pulls the required data.

I prefer to avoid a list endpoint because it'll result in a huge number of queries due to get `instance` -> `appserver` for each betatest application and it's not really required.

registration/api/v2/views.py
```diff
@@ -258,17 +213,83 @@ def get_queryset(self):
...
+            retry_attempts=settings.SELF_SERVICE_SPAWN_RETRY_ATTEMPTS,
+        )
+
+        return Response(status=status.HTTP_200_OK)
+
+    def list(self, request, *args, **kwargs):
+        """
+        List method not allowed.
+        """
+        return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
```
xitij2000: I think this is the default behaviour right? 
In any case, doesn't hurt to be explicit.

giovannicimolin: The default behavior isn't implemented because I used only the exact views required (Create, Destroy, Retrieve). That meant that the abstract method `list` wasn't implemented, and this broke the browsable API. Adding this fixed it :)

PR #521, opened 2020-02-07
SE-2108 Fix appservers/vms dns records clean up

pkulkark:
This PR removes the DNS records for the appservers of an instance that is being archived.

**Testing Instructions:**
1. Create a new instance and spawn new appservers.
2. Activate the new appservers.
3. Verify the corresponding vm dns records are shown in the dns console.
4. Archive the created instance with active appservers.
5. Verify that the vm dns records are removed.

**Reviewers:**

- [ ] @lgp171188


instance/models/openedx_instance.py
```diff
@@ -223,7 +223,15 @@ def set_active_vm_dns_records(self):
...
                 domain = "vm{index}.{base_domain}".format(index=i, base_domain=self.internal_lms_domain)
                 gandi.api.set_dns_record(domain, type="A", value=ip_addr)
-        # TODO: implement cleaning up DNS addresses that are no longer needed.
+
+    def clean_up_appservers_dns_records(self):
+        """
+        Removes the DNS records for app servers
+        """
+        self.logger.info("Cleaning up DNS records for app servers...")
+        for i, _ in enumerate(self.get_active_appservers(), 1):
```
lgp171188: @pkulkark, there is a flaw in the logic used here (same as in the `set_active__vm_dns_records` method. Let's say I create a new instance, spawn an appserver and activate it. The appserver VM will get a DNS record set for `vm1.<instance domain>` on activation. Now there is a security update for which the instance gets redeployed and hence a new appserver is spawned. Before deactivating the 1st appserver which will cause some downtime, one would activate the 2nd appserver (which will then get the `vm2.<instance domain>` record set) and then deactivate the old one after the former is active. This will trigger a call to the `set_active_vm_dns_records` method which will re-assign the `vm1.<instance>` domain name to the 2nd appserver but won't do anything about the `vm2.<instance domain>` record previously assigned to the 2nd appserver.

The current implementation of the cleanup method will only remove the `vm1.<instance domain>` record and leave the `vm2.<instance domain>` record intact. This is not what we want. In scenarios where we have multiple active appservers at any point in time, this problem gets worse. So the logic in this method and the `set_active_vm_dns_records` method (and possibly, its callers) has to be updated to cleanup the records properly.

pkulkark: @lgp171188 I changed the logic for `set_active_vm_dns_records` so that it only sets a DNS record while activating and takes the particular appserver number i.e. when we activate appserver 1, it will set `vm1.<instance domain>`. And calling the clean up method when we deactivate an appserver. Does that approach make more sense?

instance/models/openedx_appserver.py
```diff
@@ -278,9 +278,12 @@ def make_active(self, active=True):
         self.is_active = active
         self.save()
         self.instance.reconfigure_load_balancer()
+        appserver_num = self.name[-1]
```
lgp171188: @pkulkark, is this logic correct? For example, an instance could have an active appserver with name 'AppServer 17` and this will return `7`.

pkulkark: Oops! sorry looks like I pushed the wrong change.

instance/models/openedx_instance.py
```diff
@@ -213,17 +213,22 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
         return backend_map, backend_conf
 
-    def set_active_vm_dns_records(self):
+    def set_active_vm_dns_record(self, appserver_num=1, ip_addr=None):
         """
-        Set DNS A records for all active app servers.
+        Set DNS A record for the active app server.
         """
-        self.logger.info("Setting DNS records for active app servers...")
-        for i, appserver in enumerate(self.get_active_appservers(), 1):
```
lgp171188: @pkulkark, wouldn't it be a good and correct idea to use the existing logic and just remove the `vm<N+1>.<instance domain>` record? The removal method doesn't throw an error when trying to delete a non-existing record.

pkulkark: Do you mean like just loop over all appservers for an instance and remove the `vm<i>.<instance domain>` record? Wouldn't that add unnecessary overhead? Say for example, if we archive an instance that has a large number of appservers but only a couple that were activated?

instance/models/openedx_appserver.py
```diff
@@ -278,9 +278,12 @@ def make_active(self, active=True):
         self.is_active = active
         self.save()
         self.instance.reconfigure_load_balancer()
+        appserver_num = self.name.split()[-1]
```
lgp171188: @pkulkark, I am not sure you understood the logic here right. The number to be used in the `vmX.<instance domain>` DNS record has nothing to do with the appserver number shown in the name. It is based on the number of active appservers - if there N active appservers, DNS records `vm<1-N>.<instance domain>` will be set for the active appservers from oldest to the newest. So for example, if the appservers 11, 12 are the active ones, the `vm1` record will point to the IP address of the appserver 11 and `vm2` record will point to the IP address of the appserver 12. Then when 2 new appservers 13, 14 are spawned to replaces the existing pair and are activated, they will initially get assigned `vm4`, `vm5` and then once the appservers 11, 12 are deactivated, the `vm1`, `vm2` records will get assigned to the appservers 13, 14 and the `vm3`, `vm4` records will get deleted.

pkulkark: @lgp171188 Based on your [previous comment](https://github.com/open-craft/opencraft/pull/521#discussion_r376969949), I was under the impression that the logic for `set_active_vms_dns_records` is flawed and needs fixing. Sorry if I misunderstood. 
This was one approach I thought that might make it simpler - i.e. use the appserver number for the dns record and delete them when deactivating. 
I think what you're suggesting is to keep the same logic for setting the dns record, and add a new step to it to delete the non-used dns records i.e. if appserver 11 and 12 are active initially, `vm1` and `vm2` dns records will be associated with their corresponding ip address, and when we spawn and activate a new appserver 13, then that gets `vm3` dns record. Now if we deactivate appserver 11, then `vm1` and `vm2` get assigned to appserver 12 and 13, and the new step needs to delete `vm3`. Is that correct? If so, my question is how do we get the no-longer-used dns records? do we just iterate through the remaining appservers and delete `vm<i>` dns records, where, number_of_active_appservers < i < total_number_of_spawned_appservers?
Would that not add unnecessary overhead if we have a lot of inactive appservers and this would get called everytime we activate an appserver?

lgp171188: @pkulkark, see [this comment](https://github.com/open-craft/opencraft/pull/521/files#r380140691) of mine. I meant deleting the `vm<N+1>.<instance domain>` where N is the number of active appservers. The DNS client API will not throw an error when deleting a non-existent record. So if the record exists, it gets deleted. If it doesn't, nothing happens.

This will require some testing in the case of instances which should always have at least 2 active appservers, the reason being, once the 2 new appservers are active, it is very common to deactivate both older appservers in quick succession. So if the 2nd deactivation starts running before the first one finishes the reassignment, there will be an undeleted, dangling record `vm3.<instance domain>` left after the process.

instance/models/openedx_instance.py
```diff
@@ -220,10 +220,24 @@ def set_active_vm_dns_records(self):
...
             ip_addr = appserver.server.public_ip
+
             if ip_addr:
                 domain = "vm{index}.{base_domain}".format(index=i, base_domain=self.internal_lms_domain)
                 gandi.api.set_dns_record(domain, type="A", value=ip_addr)
-        # TODO: implement cleaning up DNS addresses that are no longer needed.
+
+        active_appservers = self.get_active_appservers()
+        unused_dns_index = active_appservers.count() + 1
+        domain = "vm{index}.{base_domain}".format(index=unused_dns_index, base_domain=self.internal_lms_domain)
```
lgp171188: @pkulkark, this `set_active_vm_dns_records()` method is called from the `make_active()` method [here](https://github.com/open-craft/opencraft/blob/master/instance/models/openedx_appserver.py#L283) which already knows whether the appserver is getting deactivated or not. So if that information is passed to this method via variable named, say `deactivate_an_appserver`, the DNS record removal here can be done only if that is the case. This will help avoid making an additional deletion call every time irrespective of whether the DNS record exists or not.

instance/models/openedx_instance.py
```diff
@@ -220,10 +220,24 @@ def set_active_vm_dns_records(self):
         self.logger.info("Setting DNS records for active app servers...")
```
lgp171188: @pkulkark, the code in this method that loops over the active appservers and sets the DNS records and the code that conditionally removes a DNS record should all be wrapped in a `cache.lock` block, with a lock named something like `set_active_vm_dns_records_<unique identifier for the instance>` so that only the calls to the same method for the same instance are serialized. The unique identifier could be the instance `id`. Unless we use a lock, we cannot guarantee that multiple `set_active_vm_dns_records()` calls are not executed at the same time and that can cause issues.

For the same reason, the same lock should also be used in the `clean_up_appserver_dns_records()` method.

instance/tests/models/test_openedx_instance.py
```diff
@@ -1251,4 +1253,33 @@ def test_active_vm_dns_records(self, mocks, mock_consul):
         self._verify_vm_dns_records('test.com', [appserver1.server.public_ip])
         appserver2.make_active()
         self._verify_vm_dns_records('test.com', [appserver1.server.public_ip, appserver2.server.public_ip])
-        # We haven't implemented cleaning up VM DNS records yet, so we can't test it.
+
+    @patch('instance.models.mixins.load_balanced.LoadBalancedInstance.remove_dns_records')
+    @patch('instance.models.mixins.openedx_monitoring.OpenEdXMonitoringMixin.disable_monitoring')
+    @patch('instance.models.load_balancer.LoadBalancingServer.reconfigure')
+    @patch_services
+    def test_clean_up_dns_records(self, mocks, *mock):
```
lgp171188: @pkulkark, this test can be simplified to only check whether all the DNS records for the active appservers of an instance are removed during archival.

The checks to the active VM DNS records on various sequences of activation, deactivation, can be moved to the test above. Also, please test the `instance must have 2 active appservers at all the times` scenario in which there will 4 active VMs temporarily before the 2 older ones are deactivated.

instance/models/openedx_instance.py
pkulkark: Makes sense. Added a variable.

instance/models/openedx_instance.py
pkulkark: Ah I see. Added a lock. Thanks!

instance/tests/models/test_openedx_instance.py
pkulkark: Modified it. Let me know if that's better?

instance/models/openedx_instance.py
```diff
@@ -213,17 +214,34 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
-    def set_active_vm_dns_records(self):
+    def set_active_vm_dns_records(self, deactivate_appserver=False):
         """
         Set DNS A records for all active app servers.
         """
         self.logger.info("Setting DNS records for active app servers...")
-        for i, appserver in enumerate(self.get_active_appservers(), 1):
-            ip_addr = appserver.server.public_ip
-            if ip_addr:
+        with cache.lock('set_active_vm_dns_records_' + str(self.ref.instance_id)):
```
lgp171188: @pkulkark, nit: better to use `str.format` instead of `+`.

instance/models/openedx_instance.py
```diff
@@ -213,17 +214,34 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+                unused_dns_index = active_appservers.count() + 1
+                domain = "vm{index}.{base_domain}".format(index=unused_dns_index, base_domain=self.internal_lms_domain)
+                gandi.api.remove_dns_record(domain, type="A")
+
+    def clean_up_appserver_dns_records(self):
+        """
+        Removes the DNS records for the app servers.
+        """
+        self.logger.info("Cleaning up DNS records for app servers...")
+        with cache.lock('clean_vm_dns_records_' + str(self.ref.instance_id)):
```
lgp171188: @pkulkark, nit: better to use `str.format` instead of `+`.

lgp171188: @pkulkark, the locking is useful here only when using the same lock as in the `set_active_vm_dns_records` method. There is no risk in two concurrent archive operations running at the same time as that will not leave any dangling DNS records. However, if a new appserver is activated at the same time as the instance is archived (I know it is highly unlikely, but still possible), there is a chance that one DNS record maybe left undeleted.

Since the lock is used in 2 places, it might be better to name it something like `appserver_dns_record_update` instead.

instance/tests/models/test_openedx_instance.py
```diff
@@ -1220,15 +1225,12 @@ def _verify_vm_dns_records(self, domain, active_vm_ips):
...
         dns_ips = set()
-        vm_indices = set()
         for record in gandi.api.list_records(domain):
             match = re.match(r'vm(\d+)\.', record['name'])
             if match:
                 self.assertEqual(record['type'], 'A')
                 dns_ips.add(record['value'])
-                vm_indices.add(int(match.group(1)))
         self.assertEqual(dns_ips, set(active_vm_ips))
-        self.assertEqual(vm_indices, set(range(1, len(vm_indices) + 1)))
```
lgp171188: @pkulkark, is there a reason for which you removed these lines? These are just testing the indices used in the active appserver records and are useful.

instance/tests/models/test_openedx_instance.py
```diff
@@ -1251,4 +1253,45 @@ def test_active_vm_dns_records(self, mocks, mock_consul):
...
+        appserver4 = instance.appserver_set.first()
+        appserver4.make_active()
+
+        # deactivate first two appservers
+        appserver1.make_active(active=False)
+        appserver2.make_active(active=False)
+        self._verify_vm_dns_records('test.com', [appserver3.server.public_ip, appserver4.server.public_ip])
+
+        # archive instance and verify no dns records
+        instance.archive()
```
lgp171188: @pkulkark, nit: testing the removal when archiving an instance should be a separate test.

instance/models/openedx_instance.py
pkulkark: Updated it.

instance/models/openedx_instance.py
pkulkark: Sure. I updated it.

instance/tests/models/test_openedx_instance.py
pkulkark: This was causing the test case to fail earlier (mainly because of the way they were checked) but works now after I changed the logic. Reverted the change.

instance/tests/models/test_openedx_instance.py
pkulkark: I figured since archive calls the `clean_up_appserver_dns_records` method, this would be a good place to add it. Does that make sense?

instance/models/openedx_instance.py
```diff
@@ -213,17 +214,34 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+        with cache.lock('appserver_dns_record_update_{}'.format(self.ref.instance_id)):
+            for i, appserver in enumerate(self.get_active_appservers(), 1):
+                ip_addr = appserver.server.public_ip
+
+                if ip_addr:
+                    domain = "vm{index}.{base_domain}".format(index=i, base_domain=self.internal_lms_domain)
+                    gandi.api.set_dns_record(domain, type="A", value=ip_addr)
+
+            if deactivate_appserver:
+                active_appservers = self.get_active_appservers()
```
lgp171188: @pkulkark, nit: if the result of the previous `self.get_active_appservers()` call was stored in a variable, this duplicate call can be avoided.

pkulkark: Done :+1:

instance/models/openedx_instance.py
```diff
@@ -213,17 +214,34 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
-    def set_active_vm_dns_records(self):
+    def set_active_vm_dns_records(self, deactivate_appserver=False):
         """
         Set DNS A records for all active app servers.
         """
         self.logger.info("Setting DNS records for active app servers...")
-        for i, appserver in enumerate(self.get_active_appservers(), 1):
-            ip_addr = appserver.server.public_ip
-            if ip_addr:
+        with cache.lock('appserver_dns_record_update_{}'.format(self.ref.instance_id)):
```
lgp171188: @pkulkark, `self.ref.instance_id` points to `self.id`.

```suggestion
        with cache.lock('appserver_dns_record_update_{}'.format(self.id)):
```

instance/models/openedx_instance.py
```diff
@@ -213,17 +214,34 @@ def get_load_balancer_configuration(self, triggered_by_instance=False):
...
+                unused_dns_index = active_appservers.count() + 1
+                domain = "vm{index}.{base_domain}".format(index=unused_dns_index, base_domain=self.internal_lms_domain)
+                gandi.api.remove_dns_record(domain, type="A")
+
+    def clean_up_appserver_dns_records(self):
+        """
+        Removes the DNS records for the app servers.
+        """
+        self.logger.info("Cleaning up DNS records for app servers...")
+        with cache.lock('appserver_dns_record_update_{}'.format(self.ref.instance_id)):
```
lgp171188: @pkulkark, `self.ref.instance_id` points to `self.id`.
```suggestion
        with cache.lock('appserver_dns_record_update_{}'.format(self.id)):
```

lgp171188:
@pkulkark, once the CI checks pass, make sure you squash your changes into a single commit and merge. This can be done using the `Squash and merge` option in the `Merge pull request` dropdown and it allows you to edit the commit message of the squashed commit.

PR #522, opened 2020-02-10
[BB-2057] Add management command to quickly create sandbox for any edx-platform PR

xitij2000:
This PR adds a management command that will quickly create a sandbox for any PR targeting edx-platform.


pr_watch/management/commands/setup_pr_sandbox.py
```diff
@@ -0,0 +1,75 @@
...
+        for release_name, release_config in RELEASE_BRANCH_MAP.items():
+            if release_name in pr.target_branch:
+                for key, val in release_config.items():
+                    setattr(instance, key, val)
+        instance.save()
+        if created:
+            spawn_appserver(instance.ref.pk, mark_active_on_success=True, num_attempts=2)
+        else:
+            self.stderr.write(self.style.ERROR(
+                f"Failed to create watched pull request for {pr_url.geturl()}"
```
kaizoku: Could we add some more descriptive error handling for this case?
Not sure how we would do so without updating `WatchedPullRequest.objects.get_or_create_from_pr()` though.

I tested with the PR you'd used before and it only gives a very general error.

xitij2000: @kaizoku I thought about it a bit and I see two general scenarios for logging errors:
- A sandbox already exists
- Other error during provisioning/configuration etc. 

I think the first case shouldn't really be an error even though I was treating it as such. We can just update the existing PR instance, and launch a new appserver. I've updated the code to do just this. 

For any other kind of error, it will come from an exception. Since this is a developer command I think just throwing that exception will give enough info. 

Does that seems reasonable?

kaizoku: Yep, that sounds good by me. Thanks for updating it.

PR #526, opened 2020-02-12
BB-2085: Add custom domain support

giovannicimolin:
This adds a custom domain workflow into the new registration form & APIs.

**Testing instructions:**
1. Checkout this branch locally.
2. Start Ocim (inside vagrant `make dev.run`) 
3. Install the updated API driver: from `frontend/`, run `./scripts/build-api-client.sh`.
4. Run the frontend `npm start`.
5. Make a normal betatest application with an internal domain, click on the activation links and check that the instance domain are correctly provisioned.
6. Clear your local storage to avoid stored token issues (specifically, delete `access_token` and `refresh_token`).
7. Start a new beta test registration, but click on the "I want to use my own domain" link in the first page.
8. Click on the activation links and check that the instance has the correct domain names set.

**Reviewer:**
- [x] @xitij2000 

**Notes:** 
I currently fill the subdomain as a slugified domain + some random characters in the frontend to avoid clashes (if the domain is `test.com`, the subdomain will be `testcom-23m4`). Let me know if you want me to move that to the backend (probably into the serializer validator).


registration/models.py
```diff
@@ -121,6 +121,37 @@ class BetaTestApplication(ValidateModelMixin, TimeStampedModel):
             'blacklisted': 'This domain name is not publicly available.',
         },
     )
+    external_domain = models.CharField(
+        max_length=255,
+        unique=True,
+        verbose_name='external domain name',
+        blank=True,
+        null=True,
```
giovannicimolin: Needs `null=True` because of the `unique` keyword.

frontend/src/registration/components/CustomDomainSetupPage/CustomDomainSetupPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+            </Col>
+          </Row>
+          {extra && (
+            <Row>
+              <Col className="extra">{extra}</Col>
+            </Row>
+          )}
+        </div>
+      );
+    };
```
xitij2000: This could just be implemented as another component that in internal to this file. It would clarify the code a little bit.

frontend/src/registration/components/DomainInputPage/DomainInputPage.tsx
```diff
@@ -53,16 +58,45 @@ export class DomainInputPage extends React.PureComponent<Props, State> {
...
+      );
+    } else {
+      this.props.performValidationAndStore(
+        {
+          subdomain: this.state.subdomain
+        },
+        RegistrationSteps.INSTANCE
+      );
+    }
   };
```
xitij2000: I'm wondering if it would be better to just ask the user for an internal domain as well. It is used in a lot of places, for instance to generate the database names etc so might be best to give users an option.

giovannicimolin: I think we should avoid this for now. It won't make much sense for a user that's registering for a external domain to be asked for an internal one too.

giovannicimolin: We're already discussing removing the instance extra email from the registration form because users get confused with having to activate 2 emails.

frontend/src/registration/components/CustomDomainSetupPage/CustomDomainSetupPage.tsx
giovannicimolin: Done!

giovannicimolin:
Merging with red checks because failures are not related to these changes.

PR #529, opened 2020-02-19
[BB-2082] Deploy the simple theme using the theme configuration

lgp171188:
Adds the changes to deploy the `simple-theme` using the theme configuration stored in the `theme_config` field. Also defines schema for the v0, v1 configuration, adds validators and unit tests. Also adds two new `SIMPLE_THEME_SKELETON_THEME_LEGACY*` variables for using a different branch for the v0 theming implementation.

**Testing instructions**:
This branch is already deployed on stage and there is a sandbox to test and verify the functionality (details in the task).
* Configure the values for the `SIMPLE_THEME_SKELETON_THEME_*` variables.
* Create a new beta-test instance from the registration form.
* Verify that the simple theme configuration for the initial appserver is correct.
* Once the first appserver is provisioned, activate it and test the LMS theming and verify that the theme configuration has been applied correctly.
* Add valid theme configuration in the `theme_config` field of the instance. A valid configuration based on the variable mappings listed in the ticket is already available for usage or as reference.
* Check the generated simple theme configuration in the instance settings.
* Launch a new appserver for the instance.
* Once that is provisioned succesfully, check its LMS and verify that the theming is applied as configured in the `theme_config` field.

* The schema constraints are very lax and in the v1 schema all the theme variables are optional at the moment. This can be changed once we get a better idea of what those need to be. 
* The CI checks are failing for unrelated issues.


instance/models/mixins/openedx_theme.py
```diff
@@ -55,29 +55,40 @@ class OpenEdXThemeMixin(models.Model):
...
-            # and simple_theme won't be used (unless manually requested through other settings).
-            return ""
+    def get_common_settings(self, application):
+        """
+        Returns a dict with the simple-theme settings common to the v0 and v1 implementation.
+        """
+        return {
+            "SIMPLETHEME_ENABLE_DEPLOY": True,
+            "EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO": settings.SIMPLE_THEME_SKELETON_THEME_REPO,
+            "EDXAPP_COMPREHENSIVE_THEME_VERSION": settings.SIMPLE_THEME_SKELETON_THEME_VERSION,
```
giovannicimolin: @lgp171188 Is there a way we can keep supporting the old theme for a while? 
We can add a temporary `SIMPLE_THEME_SKELETON_THEME_LEGACY_VERSION` for a while.

I want to deploy this production without breaking the current production deployments.

instance/schemas/theming.py
```diff
@@ -19,74 +19,122 @@
...
 """
-from django.core.exceptions import ValidationError
-from schema import Schema, Regex, Optional, Or, SchemaError
+from jsonschema import validate
 
-color = Regex(r'^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$')
+color_regex = '^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$'
 
 
-def nullable(schema):
```
giovannicimolin: Can we keep allowing the frontend to set a null value?
The registration form doesn't have any theming fields and doesn't send the key when the user is registering.

instance/schemas/theming.py
```diff
@@ -19,74 +19,122 @@
...
+            "properties": {
+                "color": ref("color"),
+                "border-bottom": ref("color"),
+            }
         }
-    }
-})
+    },
+    "type": "object",
+    "properties": {
```
giovannicimolin: @lgp171188 How does the mapping of unassigned variables work here?
Let's say the user only set `main_color`, `accent_color` and `light` in the theme options, is the validation going to fail?

instance/models/mixins/openedx_theme.py
```diff
@@ -55,29 +55,40 @@ class OpenEdXThemeMixin(models.Model):
     class Meta:
```
giovannicimolin: Remember to update the frontend API client after implementing the changes by going to `frontend/` and running `./scripts/update-api-client.sh`. The frontend should keep working after these changes are merged.

instance/schemas/theming.py
lgp171188: @giovannicimolin, as discussed in the meeting we had yesterday, whatever values are stored in the database will be rendered. Now, what the client has to send and what has to be done in the API handler before save is still TBD.

instance/schemas/theming.py
lgp171188: @giovannicimolin, this is now implemented.

instance/models/mixins/openedx_theme.py
lgp171188: @giovannicimolin, I have not made any changes to the model at all. So will this still be required?

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -184,3 +187,54 @@ def test_get_contrasting_font_color(self, test_colors, mock_consul):
...
         )
+
+    # pylint: disable=too-many-locals
+    @patch(
+        'instance.tests.models.factories.openedx_instance.OpenEdXInstance._write_metadata_to_consul',
+        return_value=(1, True)
+    )
+    def test_v1_config_chosen_when_theme_config_provided(self, mock_consul):
+        """
+        Test that when theme_config is provided, the v1 version of the theme configuratino is generated.
```
giovannicimolin: Nit: `configuration`
```suggestion
        Test that when theme_config is provided, the v1 version of the theme configuration is generated.
```

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -184,3 +187,54 @@ def test_get_contrasting_font_color(self, test_colors, mock_consul):
...
+                'EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO': settings.SIMPLE_THEME_SKELETON_THEME_REPO,
+                'EDXAPP_COMPREHENSIVE_THEME_VERSION': settings.SIMPLE_THEME_SKELETON_THEME_VERSION,
+            }
+            for ansible_var, value in expected_settings.items():
+                self.assertEqual(value, parsed_vars[ansible_var])
+
+            self.assertNotIn('SIMPLETHEME_EXTRA_SASS', parsed_vars)
+
+            # We check that the files are in URLs
+            # If this fails in local tests it can be because you don't have SWIFT upload enabled
```
giovannicimolin: We left SWIFT behind some time ago. This should upload to S3 instead of SWIFT.

instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -90,6 +91,8 @@ def test_colors_and_images_applied(self):
             parsed_vars = yaml.load(variables, Loader=yaml.SafeLoader) or {}
             expected_settings = {
                 'SIMPLETHEME_ENABLE_DEPLOY': True,
+                'EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO': settings.SIMPLE_THEME_SKELETON_THEME_LEGACY_REPO,
```
giovannicimolin: Do we need to add these values to the CircleCi variables or the defaults will do?

instance/schemas/theming.py
```diff
@@ -20,73 +20,160 @@
...
+theme_schema_v1 = {
+    "id": "theme_schema_v1",
+    "definitions": {
+        "color": color_schema,
+        "flag": flag_schema,
+    },
+    "type": "object",
+    "properties": {
+        "version": {"type": "number", "enum": [1, ]},
+        "theme": {"anyOf": [{"type": "string", "enum": ["light", "dark"]}, ref("color")]},
```
giovannicimolin: Can you remove this `light/dark` option as we've discussed in the 121?

instance/models/mixins/openedx_theme.py
giovannicimolin: @lgp171188  No need, I just tested and checked that the API client didn't break with these new changes.

instance/tests/models/test_openedx_theme_mixins.py
lgp171188: @giovannicimolin, yes, we will have to add these variables to CircleCI. But the defaults will also work because the integration tests just look for the presence of the color value in the CSS file.

PR #530, opened 2020-02-20
Make OCIM tests independent of environment locale

alanoe:
This patch affects the following tests:
    - test_string_to_file_path_ok
     - test_get_pr_by_number
     - test_get_pr_info_by_number
    
These tests currently break if 2 conditions occur simultaneously:
    
  - they are run in Python 3.6-, which always uses the locally configured locale by default
  - the environment locale is the (default) legacy C locale, which uses ASCII encoding
    
On the other hand, in Python 3.7+, this problem does not happen. If the environment is
using a legacy C locale, Python coerces it to a UTF-8 based locale and avoids the issue
by implementing https://www.python.org/dev/peps/pep-0538/ , as documented in
https://docs.python.org/3/whatsnew/3.7.html.
    
Unfortunately, the OCIM tests currently run in a Vagrant VM with Ubuntu Xenial and
Python 3.6.9, so they are subject to the problem above. In order to avoid this possibility,
these tests now explicitly set encoding to UTF-8 before reading/writing to a file.


pomegranited:
Hi @alanoe , thanks for fixing these tests!  Could you put the ticket ID in the title so we can track it (and/or create a ticket for it if there isn't already one there?)

lgp171188:
> Problem: File reading/writing methods in Python 3.6 and below assume
ASCII encoding. A file with different encodings requires that the
encoding is explicitly provided when opening the file.

@alanoe, can you provide more details on why this is needed, particularly since Ocim still runs on Python 3.6 and these unit tests have been running without any errors for a long time now?

CC @giovannicimolin

alanoe:
@pomegranited I have done this using my onboarding time, so there is no ticket open for these improvements. @giovannicimolin opened one for tracking review work: https://tasks.opencraft.com/browse/BB-2208 , but it does not make sense to use that one in the PR title.

alanoe:
@lgp171188 I think I was not clear enough in the original commit msg. First, these tests only break if 
- they are run in Python 3.6, which uses the locally configured locale by default
- if the locally configured locale is not set and uses the legacy C locale, which uses ASCII encoding

In case the environment is using a legacy C locale, Python 3.7+ coerces it to a UTF-8 based locale and avoids the problem by implementing https://www.python.org/dev/peps/pep-0538/ , as documented in https://docs.python.org/3/whatsnew/3.7.html.

From your previous message, I assume that these tests only fail in my workstation. Is that true? If so, I *think* that this may happen because my LC_CTYPE env var is set to pt_BR.UTF-8 and for some reason Vagrant doesn't like it. I get this message when I vagrant ssh:

warning: setlocale: LC_CTYPE: cannot change locale (pt_BR.UTF-8)

So possibly the tests fail in my workstation due to my local environment, not the test code. Anyway, my patch forces UTF8-encoding in those file read/write operations and make the tests more robust.

lgp171188:
@alanoe,

>From your previous message, I assume that these tests only fail in my workstation. Is that true? If so, I think that this may happen because my LC_CTYPE env var is set to pt_BR.UTF-8 and for some reason Vagrant doesn't like it. I get this message when I run "vagrant ssh":

The `ubuntu/xenial64` vagrant base box that we use has SSH configured to allow clients to pass locale environment variables (`AcceptEnv LANG LC_*`) and I think the SSH configuration on your host is configured to pass it (`SendEnv LANG LC_*`) - I can see this as the default behaviour on Debian and Debian-based distros.

So your fix is good and solves an uncommon problem :+1:

lgp171188:
@alanoe, please note that we usually create PRs from branches in the same repo and do not use forks.

alanoe:
I have improved the commit msg, rebased the dev branch on top of master and resubmitted the PR. Another test is failing now, but it is not caused by this PR. I think the problem was introduced by a recently merged PR.

PR #531, opened 2020-02-26
[SE-2273] Update keystone calls (OpenStack Identity API) from v2 to v3

clemente:
Update keystone (OpenStack Identity Service) API v2.0 to v3. See [deprecation notice](http://travaux.ovh.net/?do=details&id=42179)

The API change affects the way Ocim authenticates with OpenStack.
I don't think it affects other cases (e.g. the SWIFT usage done from an edxapp instance), though I didn't particularly test it since we've deprecated it. Likewise for the `migrate_swift_to_s3` command: it passes some parameters to `rclone` to create a configuration. This call may need some update if we ever use the script again and it doesn't work.

# Notes

- This PR changes the authorization part. It doesn't change other OpenStack APIs. For instance, this mustn't change from 2 to 3: `def get_nova_client(region_name, api_version=2):`. It's a different setting, it's the client version, not the „auth_version“. If I change it, I get: `novaclient.exceptions.UnsupportedVersion: Invalid client version '3.0'. Major part should be '2'`.

About this part:
```
+            user_domain_id="default",
+            user_domain_name="Default",
```
The names `default` and `Default` are extracted from a settings file downloaded from OVH, which contains:
```
export OS_USER_DOMAIN_NAME="Default"
if [ -z "$OS_USER_DOMAIN_NAME" ]; then unset OS_USER_DOMAIN_NAME; fi
export OS_PROJECT_DOMAIN_ID="default"
if [ -z "$OS_PROJECT_DOMAIN_ID" ]; then unset OS_PROJECT_DOMAIN_ID; fi
```

We don't require customizing them and we can keep them at `default`.

Note that e.g. *domain ID* appears twice: user domain ID, and project domain ID. This is because [the expected v3 request](https://storage.p19.cloud.ovh.net/v1/AUTH_47f7eb0153a7447799377451f59357ce/public/keystone-v3-request.txt) contains this information twice too: the `"domain": { "id": "default" }` appears twice. The code didn't work when the request only set the user domain ID but not the project domain ID.  

As for user_domain_name/project_domain_name, note that it's not the same as the „project name“ (aka „tenant“) which is set in another line, `project_name=settings.OPENSTACK_TENANT`, to a private numeric ID.

Finally, the authorization code is repeated in two places because we use two different variables: `openstacksdk` (in `get_openstack_connection`) and `python-novaclient` (in `get_nova_client`). Luckily, the authorization works in the same way (I think both of them use keystone).

# Testing protocol

- Phase 1: testing in CI
  - the tests should pass. The flaky tests should flake
  - you should check that the change covers APIs mentioned in the deprecation notice 
  - This PR is the final one. But [in this other branch](https://github.com/open-craft/opencraft/compare/clemente/keystone-v3-api...clemente/keystone-v3-api-testable) I'm using a temporary variable, `OPENSTACK_AUTH_URL_V3`, to avoid changing the real one `OPENSTACK_AUTH_URL` while I test. I added `OPENSTACK_AUTH_URL_V3` to CircleCI. The code in that branch uses `OPENSTACK_AUTH_URL_V3`
- Phase 2: in Ocim stage
  - remove the `OPENSTACK_AUTH_URL_V3` in the code (update: done in this PR), use `OPENSTACK_AUTH_URL`. CircleCI will not pass (we didn't change `OPENSTACK_AUTH_URL` yet), but it's not important in this phase
  - the PR should be the final one
  - put it in stage
  - try to deploy an instance
  - in particular, check that it's created in the right place: right account, right security group, right SSH key (i.e. you can access), etc. Try to archive it too, just in case
  - if the deployment works, and the CircleCI tests from phase 1 passed, it's ready to be deployed
- Phase 3: in Ocim production
  - update `OPENSTACK_AUTH_URL` in CircleCI to `https://auth.cloud.ovh.net/v3`. Remove `OPENSTACK_AUTH_URL_V3`
  - (update it in Vault too)
  - re-run the CircleCI tests from github. They should be reading `OPENSTACK_AUTH_URL` now, and they should pass now
  - (note that up to now, CircleCI is broken for everyone else who isn't using this branch)
  - merge the PR
  - deploy to Ocim production. This shouldn't change any behaviour. No `.env` change is required
  - ask people to rebase on master

### DIFF
--- a/instance/tests/test_openstack_utils.py
+++ b/instance/tests/test_openstack_utils.py
@@ -484,7 +484,7 @@ class ServicePassesAuthTestCase(TestCase):
 
         service.assert_called_once_with(
             options={
-                'auth_version': '2',
+                'auth_version': '3',
                 'os_username': 'user',
                 'os_password': 'password',
                 'os_tenant_name':  'tenant',
--- a/cleanup_utils/openstack_cleanup.py
+++ b/cleanup_utils/openstack_cleanup.py
@@ -55,7 +55,11 @@ class OpenStackCleanupInstance:
                 username=openstack_settings['username'],
                 password=openstack_settings['password'],
                 project_name=openstack_settings['project_name'],
-                region_name=openstack_settings['region_name']
+                region_name=openstack_settings['region_name'],
+                user_domain_id="default",
+                user_domain_name="Default",
+                project_domain_id="default",
+                project_domain_name="Default",
             )
         except Exception as e:  # pylint: disable=broad-except
             logger.error("ERROR: %s", e)
--- a/instance/management/commands/migrate_swift_to_s3.py
+++ b/instance/management/commands/migrate_swift_to_s3.py
@@ -110,7 +110,7 @@ class Command(BaseCommand):  # pragma: no cover
             key=instance.swift_openstack_password,
             authurl=instance.swift_openstack_auth_url,
             tenant_name=instance.swift_openstack_tenant,
-            auth_version='2',
+            auth_version='3',
             os_options={'region_name': instance.swift_openstack_region}
         )
 
@@ -136,7 +136,7 @@ class Command(BaseCommand):  # pragma: no cover
         """
         rclone_config = (
             "rclone config create ocim-swift swift "
-            "user '%s' key '%s' auth '%s' tenant '%s' auth_version '2' region '%s'" % (
+            "user '%s' key '%s' auth '%s' tenant '%s' auth_version '3' region '%s'" % (
                 instance.swift_openstack_user,
                 instance.swift_openstack_password,
                 instance.swift_openstack_auth_url,
--- a/instance/models/mixins/openedx_storage.py
+++ b/instance/models/mixins/openedx_storage.py
@@ -131,7 +131,7 @@ class OpenEdXStorageMixin(StorageContainer, SwiftContainerInstanceMixin, S3Bucke
 
             "EDXAPP_DEFAULT_FILE_STORAGE": "swift.storage.SwiftStorage",
             "EDXAPP_FILE_UPLOAD_STORAGE_BUCKET_NAME": self.swift_container_name,
-            "EDXAPP_SWIFT_AUTH_VERSION": '2',
+            "EDXAPP_SWIFT_AUTH_VERSION": '3',
             "EDXAPP_SWIFT_USERNAME": self.swift_openstack_user,
             "EDXAPP_SWIFT_KEY": self.swift_openstack_password,
             "EDXAPP_SWIFT_TENANT_NAME": self.swift_openstack_tenant,
--- a/instance/openstack_utils.py
+++ b/instance/openstack_utils.py
@@ -84,6 +84,10 @@ def get_openstack_connection(region_name):
             username=settings.OPENSTACK_USER,
             project_name=settings.OPENSTACK_TENANT,
             password=settings.OPENSTACK_PASSWORD,
+            user_domain_id="default",
+            user_domain_name="Default",
+            project_domain_id="default",
+            project_domain_name="Default",
         ))
     conn = connection.from_config(cloud_config=cloud_region)
     conn.session.user_agent = "opencraft-im"
@@ -142,6 +146,10 @@ def get_nova_client(region_name, api_version=2):
         project_name=settings.OPENSTACK_TENANT,
         auth_url=settings.OPENSTACK_AUTH_URL,
         region_name=region_name,
+        user_domain_id="default",
+        user_domain_name="Default",
+        project_domain_id="default",
+        project_domain_name="Default",
     )
 
     # API queries via the nova client occasionally get connection errors from the OpenStack provider.
@@ -208,7 +216,7 @@ def swift_service(
     """
 
     return SwiftService(options=dict(
-        auth_version='2',
+        auth_version='3',
         os_username=user,
         os_password=password,
         os_tenant_name=tenant,


instance/tests/test_openstack_utils.py
```diff
@@ -292,23 +290,25 @@ def test_get_server_public_address(self, addresses, expected_address, version=4)
...
-        """
-        Connection error during a request from the nova client
-        Ensure requests are retried before giving up, with a backoff sleep between attempts
-        """
-        def getresponse_call(*args, **kwargs):
-            """ Invoked by the nova client when making a HTTP request (via requests/urllib3) """
-            raise ConnectionResetError('[Errno 104] Connection reset by peer')
-        mock_getresponse.side_effect = getresponse_call
-        nova = openstack_utils.get_nova_client(settings.OPENSTACK_REGION)
-        with self.assertRaises(keystoneauth1.exceptions.discovery.DiscoveryFailure):
```
Agrendalath: @clemente, it seems that for this API version we can expect `keystoneauth1.exceptions.connection.ConnectFailure` here.
cf. [session.py](https://github.com/openstack/keystoneauth/blob/master/keystoneauth1/session.py#L1012-L1020) and [test_session.py](https://github.com/openstack/keystoneauth/blob/master/keystoneauth1/tests/unit/test_session.py#L1518-L1532).

Agrendalath:
@clemente, should we replace all `tenant_id` keys with `project_id` for our fixtures, as described [here](https://specs.openstack.org/openstack/neutron-specs/specs/newton/moving-to-keystone-v3.html)?
Currently we have these, but I'm not sure whether they're being used:
```bash
ocim -> ag tenant_id
instance/tests/fixtures/openstack/api_server_1_building.json
58:        "tenant_id": "1c26352f78f842e2b38ff9cec3643541",
105:    "tenant_id": "1c26352f78f842e2b38ff9cec3643541",

instance/tests/fixtures/openstack/api_server_2_active.json
72:        "tenant_id": "1c26352f78f842e2b38ff9cec3643541",
133:    "tenant_id": "1c26352f78f842e2b38ff9cec3643541",

instance/tests/fixtures/openstack/api_server_3_active.json
72:        "tenant_id": "1c26352f78f842e2b38ff9cec3643541",
133:    "tenant_id": "1c26352f78f842e2b38ff9cec3643541",
```

instance/tests/test_openstack_utils.py
clemente: @Agrendalath This is a more complex issue. Now I understand it, after much debugging.
- The main reason for this test failure is that I forgot to add this line to `.env.test`: `OPENSTACK_AUTH_URL_V3='http://example.com/auth/url`. It's a long story from there.
- The variable is one I introduced temporarily (I'll remove it later and we'll use the default one, `OPENSTACK_AUTH_URL`)
- The long story is: because I didn't define the _V3 in .env.test, it took the value from other place (environment, I think), and therefore `OPENSTACK_AUTH_URL_V3` ended up with the value `'https://auth.cloud.ovh.net/v3'`.
- The test doesn't run any external connection to any Internet server (I verified this), so it doesn't matter whether it says `auth.cloud.ovh.net` or `example.com`, both are fake. However, the `/v3` in `'https://auth.cloud.ovh.net/v3'` matters a lot, because keystone [is looking for it](https://github.com/openstack/keystoneauth/blob/1fd67089f7714fcaf0df1e3db845b93272b95b15/keystoneauth1/identity/generic/base.py#L155). In our original test we used `'http://example.com/auth/url'`, which doesn't contain `/v3`, and therefore it ran the last branch of the `if` and raised `DiscoveryFailure`. Because of the `/v3`, it ran another branch and caused another exception (`ConnectFailure`, it seems)
- since the test we're discussing ([test_nova_client_connection_error](https://github.com/open-craft/opencraft/blob/2b53d2049088ec7accfc408c88a605ac262141a3/instance/tests/test_openstack_utils.py#L298]) isn't testing any particular version (v2 nor v3), but just the side effects, I think it's better to keep it API-version-agnostic, and avoid mentioning any `/v2.0` or `/v3`, and leave it with `OPENSTACK_AUTH_URL='http://example.com/auth/url'` as it was before
- therefore: after I delete the `OPENSTACK_AUTH_URL_V3` and we use `OPENSTACK_AUTH_URL`, which is `'http://example.com/auth/url'`, this test won't fail. Tested in CircleCI through SSH

clemente:
> @clemente, should we replace all `tenant_id` keys with `project_id` for our fixtures, as described [here](https://specs.openstack.org/openstack/neutron-specs/specs/newton/moving-to-keystone-v3.html)?

@Agrendalath 

That link is to [neutronclient (Networking services)](https://docs.openstack.org/python-neutronclient/train/), which we're not using (it's to create networks etc.); we're using [novaclient](https://docs.openstack.org/python-novaclient/train/reference/api/index.html) (it's for VMs etc). Anyway, in those libraries the initialization code is very similar or the same. We already use the right parameters in the code (openstack_utils.py).

As for the project/tenant included in these fixtures: I went through all the usages of those 3 fixtures files (which are: `tests/models/test_server.py`, `tests/models/test_openedx_instance.py`, `tests/models/test_ openedx_ansible_mixins.py`) and found that the tests don't check any openstack details or openstack-related information; they mainly checke the status of the server. I guess the whole JSON was included by copy/pasting it, but most of the information isn't actually used. The 3rd fixture file isn't used at all.
So we don't need to update any data.

In addition, the change in the PR is about v2→v3 in the Identity API. The fixtures speak about a v2 API (e.g. `"manager": "<novaclient.v2.servers.ServerManager object…`) which isn't the Identity one, and we're not migrating away from it yet. The fixtures don't include authorization data.

clemente:
By the way, [all CircleCI passed](https://circleci.com/workflow-run/91f5a4b2-5fe1-4393-ba8a-9049d29af0e9) with the PR that was using `OPENSTACK_AUTH_URL_V3`. Now I renamed it back to `OPENSTACK_AUTH_URL` and changed it in CircleCI and I'm rerunning them.

clemente:
I will merge this because the latest tests already pass in CircleCI: [integration+unit](https://app.circleci.com/pipelines/github/open-craft/opencraft/1233/workflows/15f1784f-db9c-4130-9e12-97b9acebedea), [frontend](https://app.circleci.com/pipelines/github/open-craft/opencraft/1233/workflows/0a16aa7d-01c8-447a-92e7-380ec0d6d574). The status hasn't been updated in GitHub (and they still show as running since more than 10 hours ago), maybe it's because I had to hit „rerun“ some times in CircleCI.
Tests are flaky due to other reasons that I couldn't identify (some timeout, [e.g.](https://app.circleci.com/pipelines/github/open-craft/opencraft/1231/workflows/44de654b-3cdf-4b68-9aa5-5a32b79b2dd6/jobs/13604)). Some tests are taking longer to run (more than 3 h 30 m).

This has been tested in Ocim stage, with deployment + LB activation + VM termination, all working.
The code has been reviewed by @giovannicimolin in the task.

PR #534, opened 2020-02-27
SE-2017 Add new ssl monitor

pkulkark:
This PR adds a new monitor to check the SSL certificate expiration of an endpoint. The monitor sends an alert if the ceritificate expires within 2 days.

**Testing Instructions:** TBD

**Reviewers:**

- [ ] @toxinu


instance/scripts/check_ssl_expiration.js
```diff
@@ -0,0 +1,52 @@
+var request = require('request');  
+
+/* 
+************************************************** 
+What is the URL that you want to check? 
+************************************************** 
+*/ 
+var urlToMonitor = 'https://example.com';
```
toxinu: Can you add a comment here to explain that this url is replaced programmatically?

instance/newrelic.py
```diff
@@ -84,11 +85,53 @@ def create_synthetics_monitor(uri, name=None, monitor_type='SIMPLE',
...
+    """
+    Update the ssl monitor script
+    """
+    url = '{0}/monitors/{1}/script'.format(SYNTHETICS_API_URL, monitor_id)
+    logger.info('PUT %s', url)
+    headers = _request_headers()
+    headers['Content-Type'] = 'application/json'
+    current_directory = os.path.dirname(os.path.realpath(__file__))
+    script_file = os.path.join(current_directory, 'scripts', 'check_ssl_expiration.js')
+    with open(script_file, 'r') as f:
```
toxinu: Maybe it could be cleaner to use the Django template engine? Could be overkill too, feel free to ignore my comment. :)

instance/models/mixins/openedx_monitoring.py
```diff
@@ -198,6 +223,26 @@ def delete(self, *args, **kwargs):  # pylint: disable=arguments-differ
         super().delete(*args, **kwargs)
 
 
+class NewRelicSSLMonitor(models.Model):
+    """
+    A New Relic Synthetics ssl monitor for an instance.
+    """
+    id = models.CharField(max_length=256, primary_key=True)
```
toxinu: Just out of curiosity, why do we need primary key as `CharField` here?

toxinu: Oh because we are using NewRelic generated ids which are string, right?

instance/scripts/check_ssl_expiration.js
pkulkark: :+1:

instance/models/mixins/openedx_monitoring.py
pkulkark: Yes, I based it on the existing NewRelicAvailabilityMonitor.

pkulkark:
@toxinu It looks like the simplest way to test this is to deploy it to prod and only enable the ssl monitor for a test instance. So I plan to merge this PR once the CI tests pass and test it out.

lgp171188:
@pkulkark, @toxinu, was all this needed, when just adding a flag in [this API call](https://docs.newrelic.com/docs/apis/synthetics-rest-api/monitor-examples/manage-synthetics-monitors-rest-api#create-monitor) would've sufficed?

PR #535, opened 2020-02-28
BB-2131: Implement theming customization API and frontend models

giovannicimolin:
This PR introduces a new theming customization API with a custom PATCH endpoint to allow editing single fields from each `draft_theme_config`.
This also changes the behavior of some endpoints, adding a automatically generated serializer from the theme schema, so that the frontend properly parses these values.

**Setup instructions:**
1. Checkout this branch.
2. Run vagrant `vagrant up && vagrant ssh`.
3. Check if all migrations are applied and run Ocim `make run.dev`.
4. Open another terminal window and navigate to `frontend/`.
5. Set up the frontend with `npm install` and them set up the API client running `./scripts/build-api-client.sh`.

**Testing instructions:**
1. Create a new account using the new Ocim UX frontend (http://localhost:3000/registration/domain)
2. Go to http://localhost:5000/admin/registration/betatestapplication/ and check if the betatest application was correctly created, and the `Draft Theme Configuration JSON` field is filled.
3. Go to http://localhost:5000/api/swagger/ and test the `
​/v2​/instances​/openedx_config​/{id}​/theme_config​/` API endpoint.
* Change a single variable and check that it is correctly returned and saved:
```
{"footer-bg": "#fafafa"}
```
* Erase a default variable (such as `main-color`) and check that it goes back to the theme default:
```
{"main-color": ""}
```
* Set one of the optional theme variables from [here|https://github.com/open-craft/opencraft/blob/new-console/instance/schemas/theming.py#L76] and then erase it:
```
{"main-nav-color": "#fefefe"}
```
Then:
```
{"main-nav-color": ""}
```
4. Log in the new console using the account you just created (http://localhost:3000/login).
5. Click on `Theme`, then `Preview and Colors` and click on the color fields, pick a color and lose focus on the component. You should see a network request patching that exact fields. All theming fields should be greyed out while the request is being processed.

**Notes:**

1. No need to review files under `frontend/packages/api_client` as they are auto-generated by swagger codegen.

**Reviewers:**
- [ ] @alanoe 
- [ ] @lgp171188


instance/schemas/theming.py
```diff
@@ -167,6 +166,27 @@ def ref(definition_name):
     "oneOf": [nullable_schema, theme_schema_v0, theme_schema_v1]
 }
 
+DEFAULT_THEME = {
+    "version": 1,
+    "main-color": "#126F9A",
+    "link-color": "#126F9A",
+    "header-bg": "#FFFFFF",
+    "footer-bg": "#FFFFFF",
```
alanoe: IMHO it's clearer to have "color" in the end of the name of all attributes which are colors (eg. header-bg-color, footer-bg-color).

registration/api/v2/serializers.py
```diff
@@ -164,12 +165,47 @@ class Meta:
...
+            field_type = None
+            if key == 'version':
+                field_type = serializers.IntegerField()
+            elif value == ref('flag'):
+                field_type = serializers.BooleanField(required=False)
+            else:
+                field_type = serializers.CharField(
+                    max_length=7,
+                    required=False
+                    # TODO: Add a color validator here
```
alanoe: All attributes except version and flag will always be colors? I suggest that we only assume that the attribute is a color if its name ends with "color" and only validate it in this case.

instance/schemas/theming.py
giovannicimolin: I agree with you, but there's not much we can do in the scope of this epic unless we want to do some variable name remapping (I'd avoid that).
The main reason for the variables names being like this is that they need to match the ones described on this repo: https://github.com/open-craft/edx-simple-theme/tree/guruprasad/wip-high-level-theme (we're using this simple theme).

registration/api/v2/serializers.py
giovannicimolin: >I noticed that currently all attributes except version and flag are colors, but couldn't that change in the future? I'd prefer to avoid that assumption that all these attributes are colors.

Yes, that's a good point, but I have no of what the fields will look like in the future. I can change the `else` statement to `elif value == ref('color')` if that makes the solution more acceptable.

instance/schemas/theming.py
lgp171188: @alanoe, in the simple-theme that Giovanni has linked to, I have used variable names like `footer-bg`, `footer-color` etc. where the `bg` (the short form for "background") suffix refers to the background color and the `-color` suffix refers to the foreground/text color. This was done to create a distinction between the two and at the same time avoid the variable names from becoming too long.

registration/api/v2/serializers.py
```diff
@@ -180,8 +180,10 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
 
+        # Just using the new v1 schema for now, adding the v0 schema breaks
+        # code generation because one variable is named main_color and other is
+        # main-color and both get converted to mainColor in TS
```
alanoe: if we must support v0 and v1 and this is a temp solution, shouldn't you add a TODO here?

registration/api/v2/serializers.py
alanoe: > I can change the else statement to elif value == ref('color') if that makes the solution more acceptable.

+1

instance/schemas/theming.py
alanoe: @lgp171188 @giovannicimolin Thanks for explaining the origin of those attributes names. IMHO bg-color and fg-color are more explicit and clearer than bg and color. Anyway, I guess we won't make such renamings improvements like this at this point, so never mind.

registration/api/v2/serializers.py
giovannicimolin: >Yes, that's the code duplicating part of the data definition done in the theme_schema_v1 JSON Schema in instance/schemas/theming.py. I'm still getting familiar with the code and right now cannot point out a better solution, but this duplication does not look good. I'd at least add a TODO tag.

This serializer actually avoids duplication schema definition code and allows the frontend API client to be automatically generated without needing to add serializer fields one by one.

registration/api/v2/serializers.py
```diff
@@ -164,12 +165,47 @@ class Meta:
...
+            self.fields[key] = field_type
+
+
 class OpenEdXInstanceConfigSerializer(serializers.ModelSerializer):
     """
     Serializer with configuration details about the user's Open edX instance.
     """
 
     user = serializers.HiddenField(default=serializers.CurrentUserDefault())
+    draft_theme_config = ThemeSchemaSerializerGenerator(required=False)
```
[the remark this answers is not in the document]
giovannicimolin: The `draft_theme_config` is where we store the theming variables before they are actually deployed using the `Deploy` button in the new console. This is just a temporary storage for changes made by the user.

frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -92,6 +95,11 @@ export interface InstancesOpenedxConfigReadRequest {
     id: string;
 }
 
+export interface InstancesOpenedxConfigThemeConfigRequest {
```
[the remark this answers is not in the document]
giovannicimolin: We have control, but at the expense of adding override decorators to everything. 
Since it would make the backend side way less flexible we're not using it.

registration/api/v2/serializers.py
giovannicimolin: We actually plan to remove all v0 related code after this is deployed to production, so it doesn't make much sense to implement it.

frontend/src/console/models.ts
```diff
@@ -1,12 +1,16 @@
-import { OpenEdXInstanceDeploymentStatusStatusEnum } from 'ocim-client';
+import {
+  OpenEdXInstanceDeploymentStatusStatusEnum,
+  ThemeSchemaSerializerGenerator
+} from 'ocim-client';
 
 export interface InstanceSettingsModel {
-  [key: string]: string | number;
+  [key: string]: string | number | undefined | ThemeSchemaSerializerGenerator;
```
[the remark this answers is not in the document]
giovannicimolin: So, this is Typescript, and to define interfaces we need to set explicit types to each variable. For example, if I want to add a variable that will only hold a string I can use:
```
interface InstanceSettingsModel {
  variableName: string;
}
```
In this case, this line is more generic: it means that any variable inside that interface, needs to have a `string` typed key, and only contain one of the values defined after it. For example:
```
interface InstanceSettingsModel {
  [key: string]: string | number;
  // If I try to add a new variable that is not from one of the types above, the compilation fails
  wrongDefinition: ThemeSchemaSerializerGenerator;
  // If I add a variable that is allowed, then the compilation succeeds
  correctDefinition: string;
```
In any case, this line of code is only required so that I can define that all keys should be strings, so I can replace it with the code below to have the same effect:
```
  [key: string]: any | undefined;
```

frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -33,6 +33,9 @@ import {
     OpenEdXInstanceDeploymentStatus,
     OpenEdXInstanceDeploymentStatusFromJSON,
     OpenEdXInstanceDeploymentStatusToJSON,
+    ThemeSchemaSerializerGenerator,
+    ThemeSchemaSerializerGeneratorFromJSON,
+    ThemeSchemaSerializerGeneratorToJSON,
```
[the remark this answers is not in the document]
giovannicimolin: I'll try to rename it using overrides, makes sense to have a better name for this.

giovannicimolin:
@alanoe All commits are going to be squashed when the PR is merged, so I'm not paying much attention to the current commit messages. I'll write a proper one when I merge the PR.

registration/api/v2/views.py
```diff
@@ -185,6 +188,54 @@ def perform_update(self, serializer):
...
+            }.items() if value != ""
+        }
+        # To make sure the required values are always available, merge dict with
+        # default values dictionary.
+        # This API only changes values for the v1 schema, so it's hardcoded here
+        safe_merged_theme = {
+            'version': 1,
+            **DEFAULT_THEME,
+            **merged_theme
+        }
```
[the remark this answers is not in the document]
giovannicimolin: >Could you explain why this is needed?
This code is needed because the code above erases default values and removes the keys if the user empties a field. Since this would break the provisioning of an instance I overlay the default theme with the changes sent by the user, so if a required variable is erased, it gets set to the default value.

>Won't application.draft_theme_config always contain all attributes in DEFAULT_THEME and 'version' == 1?
No for all `DEFAULT_THEME` variables for the reason explained above. But yes., I can remove the `'version' == 1` since it's already present in the default theme.

frontend/src/console/models.ts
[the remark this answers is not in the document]
alanoe: Thanks for the explanation! I prefer definition with "any" at the end of your comment. Should we allow an undefined value? Isn't there always a theme enabled in an OpenEDX instance?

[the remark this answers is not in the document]
giovannicimolin: Not always, we need to leave the `undefined` there because some clients won't have any theming setup by the time this lands, and if the backend returns an undefined value while we're expecting a defined one the frontend will crash.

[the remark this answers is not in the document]
giovannicimolin: Changed to `[key: string]: any | undefined;`

alanoe:
> All commits are going to be squashed when the PR is merged, so I'm not paying much attention to the current commit messages. I'll write a proper one when I merge the PR.

@giovannicimolin OK. I commented because it was a little cumbersome to identify which patches I should review from the current long list of patches in this PR. I'm just ignoring the ones which have "fix" now. 

By the way, it'd be easier to review if the API and the models were implemented in two distinct PRs. No need to split now, but IMHO we should avoid such huge PRs.

registration/api/v2/serializers.py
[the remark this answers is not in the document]
alanoe: Thanks for the explanation. Given that rationale, shouldn't there be two different theme configs (the draft one and the effective one, currently used in the instance) ? If we won't keep the effective theme config in the db, is there a reason to name this as "draft"? If this "draft" theme config is kept in the db after the theme is deployed, it will no longer be a draft and the name will become misleading.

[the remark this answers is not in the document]
lgp171188: @alanoe, both the draft and the final theme configuration values are stored in the database. Just that the latter will have been deployed on an appserver - i.e., the user can keep changing the theme configuration from the frontend and each change will get stored and overwrite the already stored draft configuration in the database. But when the user clicks a button to deploy the changes, the version of draft configuration is copied to the final configuration and the instance is redeployed to apply the changes.

frontend/packages/api_client/src/apis/V2Api.ts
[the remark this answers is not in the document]
lgp171188: @giovannicimolin, nit: if you are going to override the name, can you name it `InstancesOpenEdXConfigThemeConfigRequest` to ensure that camel case is used throughout and `Open edX` is spelled the same as it is done in the backend code? This is not required if you are not going to add the overrides.

frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -575,6 +583,52 @@ export class V2Api extends runtime.BaseAPI {
...
+
+        const headerParameters: runtime.HTTPHeaders = {};
+
+        headerParameters['Content-Type'] = 'application/json';
+
+        if (this.configuration && this.configuration.apiKey) {
+            headerParameters["Authorization"] = this.configuration.apiKey("Authorization"); // api_key authentication
+        }
+
+        if (this.configuration && (this.configuration.username !== undefined || this.configuration.password !== undefined)) {
```
lgp171188: @giovannicimolin, shouldn't these 2 if conditions be mutually exclusive and hence this be in an `else if` block?

frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -575,6 +583,52 @@ export class V2Api extends runtime.BaseAPI {
...
+            headerParameters["Authorization"] = "Basic " + btoa(this.configuration.username + ":" + this.configuration.password);
+        }
+        const response = await this.request({
+            path: `/v2/instances/openedx_config/{id}/theme_config/`.replace(`{${"id"}}`, encodeURIComponent(String(requestParameters.id))),
```
lgp171188: @giovannicimolin, can the URL, at least the base part `/v2/instances/openedx_config/` if not the whole URL with the placeholder, be defined somewhere as a constant outside of the usage so that it is easier to check and change in the future?

frontend/packages/api_client/src/apis/V2Api.ts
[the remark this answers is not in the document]
giovannicimolin: @lgp171188 I ended up not adding a decorator to change the name of this variable. I prefer to not mess around much with the auto-generated code for now.

frontend/packages/api_client/src/apis/V2Api.ts
giovannicimolin: This is auto-generated code, so I didn't even check it. Not sure about these conditions, but I can check.

frontend/packages/api_client/src/apis/V2Api.ts
giovannicimolin: @lgp171188 This is auto-generated code from the Swagger spec, I think that this is pretty much hardcoded into how the generator works, so we'll not going to be able to change it (unless we do a manual patch, but we don't do that in any auto-generated code).

frontend/packages/api_client/src/apis/V2Api.ts
giovannicimolin: @lgp171188 Can you add a little bit more context here? Is this about the line 609?

lgp171188: @giovannicimolin, I was referring to the `if` conditions in the lines 605 and 609.

giovannicimolin: Those lines are like that because there are 2 auth methods defined in our Swagger spec (basic and API key).
I agree with you that they should be mutually exclusive, but I think we don't have control on how that code is generated. 
Since we control the `configuration` variable, and we never set the 2 auths at the same time this should be fine (https://github.com/open-craft/opencraft/blob/new-console/frontend/src/global/api.ts#L3).

Additionally, adding this to the cleanup step of the epic can be done (remove the basic auth from the swagger spec).

alanoe: If we reach a consensus that the generated code here could be better, we could create a backlog item to submit a patch to Swagger in the future.

alanoe:
One of the test instructions says "Erase a default variable (such as main-color) and check that it goes back to the theme default". Is there such a thing as a theme default? I had understood that there is only effective (deployed) theme and possibly a draft (undeployed) theme. Are you referring to the first here? 

If that is true, it seems inconsistent that trying to update a required field with an empty string resets it to its former value and doing the same with an optional field deletes it. IMHO it'd be more consistent that trying to set a required field to "" returned an error and resetting was done with a special more explicit string, eg "reset". That would allow reseting an optional field value to its former value, which is not allowed now.

frontend/packages/api_client/src/apis/V2Api.ts
giovannicimolin: I'm not sure that's a good idea since we don't have any expertise on it (and we also don't have budget). In any case, the current generated code works well and the issue is not really in our usage scope so we're fine to leave this as is for now.

giovannicimolin:
@alanoe Thanks for the review. Regarding some of the points you raised:

>Regarding the new HTTP API endpoint, IMHO it is inconsistent that trying to update a required theme attribute (eg main-color) with an empty string resets it to its default value and doing the same in an optional attribute (eg main-nav-color) deletes it. It'd be more consistent that trying to set a required field to "" returned an error and resetting to the default was done with a special more explicit string, eg "default".

I already scheduled a follow up to handle this. The improvements will be done in a separate task. This needs to be merged to unblock further work on the new UI.

>I'm not sure how this OpenEDX instance preview will work. Will it be a thumbnail of the OpenEDX instance? If so, it is cumbersome that this preview only will show up in the "Preview and Colors" submenu, it'd be nicer if it appeared during the entire theme customization, in all "Theme" submenus.

Yes, it'll be a small thumbnail of the instance showing general colors. Other customization pages have their own previews.
We hired a UX company to design the interface, so I'm just following the mockups provided.

>It'd be better to show only the selected color instead of its hexadecimal code painted in the selected color in Theme > Preview and Colors UI. The user does not care about the color hexadecimal code.

Again, I'm just following the UX design agency designs. I think having the hex field is not a big compromise, specially when the user has a friendly UI to pick the color from after clicks on the field.

>Why can I only select main color and link color in Theme > Preview and Colors UI? Shouldn't we allow that the user customizes all colors in the theme using this UI? Or will header/footer/buttons colors be set in the other Theme options (eg navigation and buttons)? If that is the case, the word "Colors" in the "Preview and Colors" submenu is a little misleading, as it makes the user think that all theme colors could be set here. Maybe something such as "General" instead of "Colors" would be better.

There are other customization pages that won't be added here. The main goal of this work in not to add this UI in, but to lay the groundwork on the APIs and typescript models to enable the frontend to save theming values in the backend.
This screen was designed by the UX agency, and we already finished the discussions around design/theming (see this discussion in the forum).

>In the current implementation, we currently restrict the user choice of colors to a small list of predefined colors. Isn't there a color palette Javascript component we could use instead so that the user could select any color he/she wants?

Since this task is more about implementing the APIs and code and less about UI, I'll leave the UI improvements to the next task (already scheduled for next sprint).

>One of the tests is broken. Was it caused by your PR? If so, could you pls fix it?

This is unrelated to my PR, but it looks like our testing infrastructure is having some issues. I'll investigate.

alanoe:
AE> It'd be better to show only the selected color instead of its hexadecimal code painted in the selected color in Theme > Preview and Colors UI. 

GC> Again, I'm just following the UX design agency designs. I think having the hex field is not a big compromise, specially when the user has a friendly UI to pick the color from after clicks on the field.

It's indeed not a big issue, it's just (IMHO) unnecessary for the kind of user using the tool and makes seeing selected lighter colors harder (as a block of color is more easily identifiable than a string written in that color, especially if the string font is thin). If it is already decided, never mind.

GC> Since this task is more about implementing the APIs and code and less about UI, I'll leave the UI improvements to the next task (already scheduled for next sprint).

That's fine.

AE> One of the tests is broken. Was it caused by your PR? If so, could you pls fix it?
GC> This is unrelated to my PR, but it looks like our testing infrastructure is having some issues. I'll investigate.

Any news about that?

giovannicimolin:
>Any news about that? IMHO we should not approve the PR if all tests don't pass.

The failing test already passed. Just waiting for integration checks now.

PR #536, opened 2020-03-01
[SE-2245] Override instance storage type to s3 for tests

pkulkark:
This PR overrides the instance storage type setting for failing integration tests.

**Testing Instructions:**

- The CI tests should pass

- Verify the [CI variables](https://vault.opencraft.com:8200/ui/vault/secrets/secret/show/core/OpenCraft%20Tools%20:%20Resources%20:%20OpenCraft%20IM%20-%20Integration%20test%20settings) are updated with newly created s3 bucket details

**Reviewers:**

- [ ] @mtyaka


instance/tests/models/test_openedx_theme_mixins.py
```diff
@@ -65,6 +66,7 @@ def make_test_application(instance, user):
         )
         return application
 
+    @override_settings(INSTANCE_STORAGE_TYPE='s3')
```
lgp171188: @pkulkark, wouldn't it be easier and cleaner to just change the default in `.env.test` instead of changing it in multiple places?

pkulkark: @lgp171188 I suppose so. I was following the [suggestion](https://tasks.opencraft.com/browse/SE-2245) made in the ticket :). I'll update it.

pkulkark: @lgp171188 changing `.env.test` resulted in more test failures so I went back to changing it only for the ones that needed it.

lgp171188:
@pkulkark, a quick question - why do we need to set `INSTANCE_STORAGE_TYPE` for the unit tests? That is primarily used for provisioning storage buckets on external systems, something that shouldn't be happening in the unit tests. Since the environment variables are shared between the unit and integration tests, with the latter requiring this and the related settings, it looks like the unit tests also interact with external systems some how.

This also explains why these tests never fail in the devstack where only the values set in the `.env.test` file are used and there are no additional environment variables.

So would it help to unset the problematic environment variables before running the unit tests? That feels like a better solution.

FYI @mtyaka

pkulkark:
@lgp171188 I'm not exactly sure how the configuration is setup but I don't think that the unit tests are actually interacting with the external services. The `INSTANCE_STORAGE_TYPE` is actually defined in `.env.test` only and doesn't exist in the [environment variables](https://vault.opencraft.com:8200/ui/vault/secrets/secret/show/core/OpenCraft%20Tools%20:%20Resources%20:%20OpenCraft%20IM%20-%20Integration%20test%20settings) used for integration tests. My guess is that it's probably used to verify the logic but the actual call is mocked.

mtyaka:
This fixes the unit tests, and the updated CI variables fix the integration tests, so this is good to merge :+1: 

I would like to schedule a follow-up ticket for this though. SWIFT storage is effectively deprecated in Ocim/OpenEdX, so we should remove `INSTANCE_STORAGE_TYPE='swift'` from [.env.test](https://github.com/open-craft/opencraft/blob/196229828a21f39f2662870ebbd3d4f4b5b2eee4/.env.test#L6) to let it default to `s3` just like we do for integration tests. If for any reason some unit tests actually require swift because they are explicitly testing swift related code, we should override `INSTANCE_STORAGE_TYPE='swift'` just for those tests.

- [x] I tested this: I verified that the build passes with these changes.
- [x] I read through the code
- [ ] Includes documentation

registration/tests/test_views.py
```diff
@@ -589,7 +593,7 @@ def test_modify_user(self):
         application = BetaTestApplication.objects.get()
         self._assert_application_matches_form_data(application)
 
-    @override_settings(VARIABLES_NOTIFICATION_EMAIL='notifications@opencraft.com')
+    @override_settings(VARIABLES_NOTIFICATION_EMAIL='notifications@opencraft.com', INSTANCE_STORAGE_TYPE='s3')
```
lgp171188: @pkulkark, though the unit tests on this PR and the merge commit passed, this change is now causing this unit test to fail with the below error.

```
ERROR: test_modifying_design_fields_sends_email (registration.tests.test_views.BetaTestApplicationViewTestCase)
  File "/home/vagrant/.pyenv/versions/opencraft/lib/python3.6/site-packages/django/test/utils.py", line 373, in inner
    return func(*args, **kwargs)
  File "/home/vagrant/opencraft/registration/tests/test_views.py", line 632, in test_modifying_design_fields_sends_email
    EmailAddress.objects.confirm(email.key)
  File "/home/vagrant/.pyenv/versions/opencraft/lib/python3.6/site-packages/simple_email_confirmation/models.py", line 203, in confirm
    email_confirmed.send(sender=address.user, email=address.email)
  File "/home/vagrant/.pyenv/versions/opencraft/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 175, in send
    for receiver in self._live_receivers(sender)
  File "/home/vagrant/.pyenv/versions/opencraft/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 175, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "/home/vagrant/opencraft/registration/provision.py", line 51, in provision_instance
    _provision_instance(sender, **kwargs)
  File "/home/vagrant/opencraft/registration/provision.py", line 79, in _provision_instance
AttributeError: 'NoneType' object has no attribute 'lms_users'

```
Removing the `INSTANCE_STORAGE_TYPE='s3' setting fixes it. Can you check?

lgp171188: The issue is happening because of the check [here](https://github.com/open-craft/opencraft/blob/master/instance/factories.py#L47) causing `None` to be returned by the `production_instance_factory` function instead of an instance. So we should remove this.

pkulkark: @lgp171188 I see. But reverting this would result in the CI failure again right? I can see this test case fails in the [example build](https://app.circleci.com/jobs/github/open-craft/opencraft/11978) that was failing before this change. Is this error seen in the devstack? I don't see it failing on a [recent unit test run](https://circleci.com/gh/open-craft/opencraft/12786?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

lgp171188: @pkulkark, even I am not sure why and how the unit tests for this PR passed. This unit test is failing in the devstack and also on CircleCI (see https://github.com/open-craft/opencraft/pull/530).

pkulkark: @lgp171188 Interesting. I'll raise another PR which reverts the override for this test case and see if the CI still passes.

PR #538, opened 2020-03-05
Create a BulkEmailFlag when first appserver is provisioned [SE-2078].

mtyaka:
This automatically enables bulk emails for new instances. The playbook that enables bulk emails is run when provisioning the first appserver for the instance.

Note that this doesn't modify the default bulk email template, which is an edX-branded template.

**Test instructions**:
1. Checkout this code and create a new instance. Spawn the first appserver.
2. When the first appserver is successfully spawned, go to the django admin on the new appserver, and verify that a `BulkEmailFlag` with `enabled` attribute set to `True` and `required_course_email_auth` set to `False`. You may also want to verify that you can see the Email tab in the Instructor dashboard in the LMS.
3. Spawn another appserver. Verify that after the appserver spawns successfully the `BulkEmailFlag` remains intact and no new `BulkEmailFlag` object is created.

**Reviewers**:

- [ ] @pkulkark


instance/models/mixins/ansible.py
```diff
@@ -35,15 +36,31 @@
 
 # Classes #####################################################################
 
+# Represents and ansible playbook. It can be located in a remote git repository or
```
pkulkark: Nit: spelling check - should be `Represents an ansible playbook`.

mtyaka: Thanks for catching that; I fixed it.

playbooks/enable_bulk_emails/enable_bulk_emails.yml
```diff
@@ -0,0 +1,10 @@
+---
+# A simple one-off playbook that allows starting and stopping instance services.
+# This is used by Ocim to stop celery workers from inactive appservers to pick
+# up tasks.
```
0x29a, line 4: @mtyaka, looks like this comment was copied from other playbook.

mtyaka: You're right :/ I'll open a PR to fix this in the upcoming sprint.

mtyaka: I added a commit to update this text to a PR I'm currently working on: https://github.com/open-craft/opencraft/pull/782

PR #541, opened 2020-03-10
BB-2179: Multiple bug fixes & addressing implementation review comments

giovannicimolin:
This PR fixes multiple bugs and addresses a few review comments.
**To make the review easier, review the changes by commit, and not all at the same time.**

**Setup instructions:**
1. Checkout this branch.
2. Run Ocim as usual.
3. Navigate to `frontend/` in a new console window.
4. Run `npm install`, then `./scripts/build-api-client.sh`.
5. Run `npm start` to run the frontend.

**(REMOVED) Testing instructions:** 
https://github.com/open-craft/opencraft/commit/9255d9b61e6485b655de78410b62019a25057636:
1. Create a betatest account and instance.
2. Manually change the appserver status to provisioned.
3. Make some changes to variables.
4. Click on the `Deploy` button in the user toolbar.
5. Check that the status was update just after the request finished.

This was removed from this PR, too complex for available timebox.

https://github.com/open-craft/opencraft/commit/b948140e6f208e4739f828bd55ca2fcb20526e00:
1. Log in the user console.
2. Delete the access token, check that the page refreshes.
3. Delete the refresh token, and check that you were redirected to the login page after a while.

https://github.com/open-craft/opencraft/commit/9e01fa359e1b569a591aad82f00eb98e40b7b206
1. Create a new betatest instance.
2. Activate emails.
3. Check that `draft_theme_config` was copied over to the instance model.

https://github.com/open-craft/opencraft/commit/ad1ac41a8a9d1104e2dbb400a701d1f1e0afd805:
This is just error treating for some exceptions. A little hard to replicate, but this essentially redirects the user to the error page when certain requests fail.

https://github.com/open-craft/opencraft/commit/6bb11589b84deb32ae6c99a855ca2892e942d24a:
1. Log in the new console, open developer tools.
2. Check that there is a recurring request every few seconds.
3. Log out and check that it stops.

Other commits are just quality fixes and removing placeholder strings, as pointed out by @bradenmacdonald  in his review of the code (link on task).

**Reviewer:**
- [ ] @lgp171188

### DIFF
--- a/frontend/src/auth/actions.ts
+++ b/frontend/src/auth/actions.ts
@@ -70,24 +70,28 @@ export const performLogin = (
       }
     })
     .catch((e: any) => {
-      e.json().then((feedback: any) => {
-        // If validation fails, return error to form through state
-        let error = feedback.detail;
-        // Don't provide individual feedback on each field for login
-        // If feedback.detail isn't set, then it means we got an
-        // response like this:
-        // {
-        //    username: ["This field is required.]
-        // }
-        // So manually set feedback error message.
-        if (!feedback.detail) {
-          error = 'Both the username and password fields are required.';
-        }
-        dispatch({
-          type: Types.LOGIN_FAILURE,
-          error
+      try {
+        e.json().then((feedback: any) => {
+          // If validation fails, return error to form through state
+          let error = feedback.detail;
+          // Don't provide individual feedback on each field for login
+          // If feedback.detail isn't set, then it means we got an
+          // response like this:
+          // {
+          //    username: ["This field is required.]
+          // }
+          // So manually set feedback error message.
+          if (!feedback.detail) {
+            error = 'Both the username and password fields are required.';
+          }
+          dispatch({
+            type: Types.LOGIN_FAILURE,
+            error
+          });
         });
-      });
+      } catch (e) {
+        dispatch(push(ROUTES.Error.UNKNOWN_ERROR));
+      }
     });
 };
 
--- a/frontend/src/auth/actions.ts
+++ b/frontend/src/auth/actions.ts
@@ -89,7 +89,7 @@ export const performLogin = (
             error
           });
         });
-      } catch (e) {
+      } catch (jsonParseError) {
         dispatch(push(ROUTES.Error.UNKNOWN_ERROR));
       }
     });
--- a/frontend/src/console/components/ConsolePage/ConsolePage.tsx
+++ b/frontend/src/console/components/ConsolePage/ConsolePage.tsx
@@ -30,7 +30,7 @@ interface Props extends StateProps, ActionProps {
 }
 
 export class ConsolePageComponent extends React.PureComponent<Props> {
-  refreshInterval: any | undefined;
+  refreshInterval?: NodeJS.Timer;
 
   public componentDidMount() {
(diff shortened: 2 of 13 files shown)


frontend/src/console/components/ConsolePage/ConsolePage.tsx
```diff
@@ -30,17 +30,26 @@ interface Props extends StateProps, ActionProps {
 }
 
 export class ConsolePageComponent extends React.PureComponent<Props> {
+  refreshInterval: any | undefined;
```
bradenmacdonald: The type of this should be

```typescript
refreshInterval?: NodeJS.Timer;
```

giovannicimolin: Done! Thanks for the tip, stronger the typing the better. :)

frontend/src/auth/actions.ts
```diff
@@ -70,24 +70,28 @@ export const performLogin = (
...
-        // }
-        // So manually set feedback error message.
-        if (!feedback.detail) {
-          error = 'Both the username and password fields are required.';
-        }
-        dispatch({
-          type: Types.LOGIN_FAILURE,
-          error
+      try {
+        e.json().then((feedback: any) => {
```
lgp171188: @giovannicimolin, as the code below is treating `feedback` as an object, should we still use `any`?

giovannicimolin: @lgp171188 If I were to set `feedback` to a type I'd need to create an interface to describe the possible fields inside the response feedback.
We know that the return of `.json()` is a `object`, but if I put that in the prop definition I get type errors because the members of this object are not defined.
I could define a generic object like this:
```
interface GenericJsonObject {
    [key: string]: any;
}
```
But I think it's overly complex for just catching and parsing the response here.

lgp171188:
> 9e01fa3
>   1. Create a new betatest instance.
>   1. Activate emails.
>   1. Check that draft_theme_config was copied over to the instance model.

:heavy_check_mark: I tested this and it worked as expected. Verifying the email address was tricky because I didn't realize right away that there were two email addresses to verify. :smile:

lgp171188:
> b948140:
>    1. Log in the user console.
>    1. Delete the access token, check that the page refreshes.
>    1. Delete the refresh token, and check that you were redirected to the login page after a while.

:heavy_check_mark: This worked as expected. It took me a while to figure out that the access and refresh tokens were in local storage (found it by looking at the source code :P)

lgp171188:
> 6bb1158:
>   1. Log in the new console, open developer tools.
>   1. Check that there is a recurring request every few seconds.
>   1. Log out and check that it stops.

:heavy_check_mark: Verified this. @giovannicimolin, instead of repeated polling, wouldn't it be more efficient to use websockets?

lgp171188:
An issue I found - pressing the `Enter` key on the login page doesn't trigger the login action. It requires manually clicking the `Login` button.

lgp171188:
> 9255d9b:
>  1.  Create a betatest account and instance.
>  1.  Manually change the appserver status to provisioned.
>  1.  Make some changes to variables.
>  1.  Click on the Deploy button in the user toolbar.
>  1.  Check that the status was update just after the request finished.

This requires spawning an appserver, so I will test this on stage.

PR #542, opened 2020-03-10
[BB-2225] Fix the validation issue in the combined schema

lgp171188:
This fixes the issue causing the validation to not work properly in the combined schema though it works as expected for the individual schemas.

Also fixes a minor issue with the `.env.test` file which was causing all the environment variables to not be set correctly.

**Testing instructions**:
* Verify that the new unit tests added for verifying that this issue is fixed, pass.
* Call the `theme_schema_validate()` function with invalid values for the v0 and v1 schemas and verify that the validation fails as expected.


instance/tests/test_theme_schemas.py
```diff
@@ -243,3 +243,24 @@ def test_data_in_v0_and_v1_schema_accepted(self):
             'header_bg_color': '#222',
             'footer_bg_color': '#333'
         })
+
+    def test_invalid_v0_schema_data_causes_validation_failure(self):
+        """
+        Test that the schema validation for the v0 schema works even in the combined schema.
+        """
+        with self.assertRaises(JSONValidationError):
+            theme_schema_validate({'version': 0, 'main_color': 'red', 'link_color': '#fff', 'header_bg_color': '#fff', 'footer_bg_color': '#fff'})
```
giovannicimolin: @lgp171188 Quality violation here: line too long.

PR #543, opened 2020-03-10
BB-2177: Add colors and logo editor pages

giovannicimolin:
This PR adds two pages to the new Ocim frontend: 
1. Colors and Preview page
2. Logos page

**Testing instructions:**
1. Checkout this branch.
2. Set `DEBUG=True` in Django to enable media urls.
2. Run Ocim (inside vagrant with `make run.dev`).
3. Go to `frontend/` and build the new API client with `./scripts/build-api-client.sh`
4. Run the new frontend: `npm start`.
5. Create a new test instance and log in the console.
6. Go to the theme preview and colors page and make sure it's working as expected (http://localhost:3000/console/theming/preview-and-colors). Compare it to the mockups.
7. Change colors and check if the values are correctly saved in the backend.
8. Go to the logos pages (http://localhost:3000/console/theming/logos). Compare it to the mockups.
9. Try uploading multiple images, check if django validation is working properly. See if the images are correctly saved in the backend.

**Reviewers:**
- [ ] @alanoe 
- [ ] @lgp171188


frontend/src/console/components/Logos/Logos.spec.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
alanoe: I know you're just following an already set pattern, but IMHO it'd be better to put test files in a separate directory (instead of putting them together with source code).

frontend/src/console/components/Logos/displayMessages.ts
```diff
@@ -0,0 +1,24 @@
+const messages = {
+  logo: {
+    defaultMessage: 'Logo',
+    description: ''
+  },
+  logoDescription: {
+    defaultMessage:
+      "Your branding to be displayed throughout your instance. " +
```
alanoe: instance => OpenEdX instance

frontend/src/routes/console.tsx
```diff
@@ -1,7 +1,11 @@
 import * as React from 'react';
 
 import { Switch, Route, Redirect } from 'react-router';
-import { InstanceSettings, ThemePreviewAndColors } from 'console/components';
+import {
+  InstanceSettings,
+  ThemePreviewAndColors,
+  Logos
```
alanoe: list imports in alphabetical order

frontend/src/ui/components/CollapseEditArea/CollapseEditArea.tsx
```diff
@@ -0,0 +1,33 @@
...
+) => {
+  const [open, setOpen] = React.useState(props.initialExpanded || false);
+
+  return (
+    <div className="collapsible-edit-area">
+      <div className="edit-collapse">
+        <Nav.Link onClick={() => setOpen(!open)}>
+          <WrappedMessage messages={messages} id="edit" />
+          {!open && <i className="fas xs fa-chevron-up" />}
+          {open && <i className="fas xs fa-chevron-down" />}
```
alanoe: IMHO it'd be clearer to use an "if" statement here instead of the && logical operator.

frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -463,6 +467,45 @@ export class V2Api extends runtime.BaseAPI {
...
+
+        const queryParameters: runtime.HTTPQuery = {};
+
+        const headerParameters: runtime.HTTPHeaders = {};
+
+        if (this.configuration && this.configuration.apiKey) {
+            headerParameters["Authorization"] = this.configuration.apiKey("Authorization"); // api_key authentication
+        }
+
+        if (this.configuration && (this.configuration.username !== undefined || this.configuration.password !== undefined)) {
```
alanoe: I'd use instead:

if (this.configuration) {
    if (this.configuration.apiKey) {}
    else if (this.configuration.username !== undefined || this.configuration.password !== undefined) {}
    else { 
        // handle error
    }
}

giovannicimolin: I can't edit these files, they are auto-generated by the DRF YASG.

frontend/packages/api_client/src/models/OpenEdXInstanceConfigUpdate.ts
```diff
@@ -68,6 +68,18 @@ export interface OpenEdXInstanceConfigUpdate {
...
      */
     draftThemeConfig?: ThemeSchema;
+    /**
+     * Your branding to be displayed throughout your instance. It should be 48px tall. If unset, OpenCraft\'s logo will be used.
+     * @type {string}
+     * @memberof OpenEdXInstanceConfigUpdate
+     */
+    readonly logo?: string;
+    /**
+     * This is used as the browser tab icon for your instance\'s pages. If unset, OpenCraft\'s icon will be used.
```
alanoe: I assume there is no size requirement here, unlike the logo image.

frontend/packages/api_client/src/apis/V2Api.ts
alanoe: Never mind then.

frontend/packages/api_client/src/models/OpenEdXInstanceConfigUpdate.ts
giovannicimolin: I'm not sure what you're referring to on this line, but this structure is an auto-generated Typescript `interface` (similar to a model in Django) from the DRF Swagger Schema.
To view the schema, spin up your devstack and go to http://localhost:5000/api/swagger/?format=openapi

alanoe: Sorry if I was not clear. I noticed that in the logo image, you require a 48px tall image and here there is no image dimensions restriction. I wanted to know if there is none or if you had forgotten to add it. Even though this file is auto-generated, I believe you wrote the doc in other file.

opencraft/urls.py
```diff
@@ -51,3 +52,4 @@
     urlpatterns = [
         url(r'^__debug__/', include(debug_toolbar.urls)),
     ] + urlpatterns
+    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
alanoe: Could you explain in the commit msg how exactly this improves the devstack setup?

frontend/packages/api_client/src/models/OpenEdXInstanceConfigUpdate.ts
giovannicimolin: The feedback on this is enforced by the backend (`/image` endpoint).
If you upload an invalid logo image it should error out.

opencraft/urls.py
giovannicimolin: This is just to make testing easier. The devstack doesn't set up the Django media assets path (so all urls from uploaded files are broken). To avoid having to deal with this headache while testing I've added this line.

alanoe:
I have tested the PR.

- [x] Made sure that color change UI works and new color is saved in the backend
- [ ] Made sure that logo/favicon image upload UI works and images are saved in the backend
This test failed, I got an error page when trying to upload the logo image.

UI bugs:
- default cursor changes to text cursor when hovering over "Theme", "Custom Pages" and "Instance Settings" menu options in the left. it should change to the hand cursor, as it happens with "Theme" menu suboptions
- arrow next to edit is supposed to point upwards when edition pane is shown and downwards when edition pane is hidden. The opposite is implemented right now.

- I got an error page when trying to update the image.

usability comments (not related to the patch):
- currently, in order to update the logo or the favicon image, the user must click in "change", then double click in the "choose file" button,  select the file and then click in the "update image" button (5 clicks). It'd be much easier if file browsing window opened when user clicked on "change" and image was updated after selecting a file (3 clicks).  The whole upload logo / upload favicon subwindows seem unnecessary.
 - "Browser tab icon" would be a much clearer subtitle to user than "Favicon", which does not mean anything to him. I noticed it is described right below, but it could be clearer with less words. 
- "Upload Logo" and "Upload Favicon" texts are useless and confusing (I thought they were clickable at first glance)

Feel free to ignore the usability issues if the UI design should not be changed.

giovannicimolin:
@alanoe Thanks for the thorough review!
I've addressed some comments and added other to the backlog of improvements. Overall I agree with all of your suggestions.

>This second test failed the first time and it worked correctly afterwards. I was not able to replicate it, but I recommend you check if any random factor could cause such an error.

This happens because the media files are not properly set up on Ocim devstack, but it works on stage and prod.

>default cursor changes to text cursor when hovering over "Theme", "Custom Pages" and "Instance Settings" menu options in the left. it should change to the hand cursor, as it happens with "Theme" menu suboptions

Fixed!

> arrow next to edit is supposed to point upwards when edition pane is shown and downwards when edition pane is hidden. The opposite is implemented right now.

Fixed!

> I got an error page when trying to update the image.

This is probably happening because of the devstack setup. It didn't happen for me. If you manage to make it work, can you add instructions on how to reproduce this issue in the epic later?

>currently, in order to update the logo or the favicon image, the user must click in "change", then double click in the "choose file" button, select the file and then click in the "update image" button (5 clicks). It'd be much easier if file browsing window opened when user clicked on "change" and image was updated after selecting a file (3 clicks). The whole upload logo / upload favicon subwindows seem unnecessary.

This was me overdoing things. :stuck_out_tongue: 
I added that window to implement a image editor, so the user can select their logo size, crop, and other things using some react component and ran out of timebox.
I'll add this to the list of improvements, so let's merge this as is for now.

>"Upload Logo" and "Upload Favicon" texts are useless and confusing (I thought they were clickable at first glance)

This is from the mockups, and I don't see an issue in keeping them for now. If we get any user feedback mentioning that this is confusing we'll improve it later.

frontend/packages/api_client/src/models/OpenEdXInstanceConfigUpdate.ts
lgp171188: > The feedback on this is enforced by the backend (/image endpoint).
> If you upload an invalid logo image it should error out.

Can we add the validation to check this in the frontend as well so that we don't have to upload the image to the backend to find out issues with the height, if any?

registration/api/v2/serializers.py
```diff
@@ -38,6 +38,46 @@
...
+            "additionalProperties": True,
+        }
+
+
+class LogoFaviconUploadSerializer(DataSerializer):
+    """
+    Serializer for any response that returns a JSON dict, without specifying
+    the fields of that dict in detail.
+    """
+    logo = serializers.FileField(required=False, use_url=True)
```
lgp171188: @giovannicimolin, shouldn't we be using [ImageField](https://www.django-rest-framework.org/api-guide/fields/#imagefield)?

registration/api/v2/views.py
```diff
@@ -265,6 +266,45 @@ def theme_config(self, request, pk=None):
 
         return Response(status=status.HTTP_200_OK, data=application.draft_theme_config)
 
+    @action(
+        detail=True,
+        methods=['post'],
+        parser_classes=(MultiPartParser, ),
+        serializer_class=LogoFaviconUploadSerializer
+    )
+    def image(self, request, pk: str):
```
lgp171188: @giovannicimolin, can you add a test for this action? Also, though the [DRF documentation](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) is ambiguous and sounds like the `GET` method is always allowed, 

> The action decorator will route GET requests by default, but may also accept other HTTP methods by setting the methods argument.

from [the code](https://github.com/encode/django-rest-framework/blob/master/rest_framework/decorators.py#L145), it looks like that `GET` is only allowed as a default and the provided methods take precedence over it. Did you test if the `GET` method works for this action? If not, can you test and fix this action appropriately?

CC @alanoe

giovannicimolin: @lgp171188 Making a `GET` request on the endpoint returns:
```
HTTP 405 Method Not Allowed
Allow: POST, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "Method \"GET\" not allowed."
}
```

lgp171188: @giovannicimolin, thanks for checking and confirming. In this case, do we still need the `if` condition since `POST` is the only allowed method?

lgp171188:
@giovannicimolin, I tested the changes and they look good to me.
:+1: 
- [x] I tested this on the devstack and verified that the logo/favicon images were uploaded and saved. Also verified that the ansible configuration generated was correct.
- [x] I *partially* read through the code and will piggyback on Alan's review.

PR #546, opened 2020-03-17
[BB-2219] Add button customization page

Agrendalath:
This PR adds button customization page to the new Ocim frontend.

**Testing instructions:**
1. Checkout this branch.
1. Create a new test instance and log in the console.
1. Go to the `Theme` -> `Buttons` page and make sure it's working as expected (http://localhost:3000/console/theming/buttons).
1. Change colors and check if the values are correctly saved in the backend.

**Reviewers:**
- [ ] @giovannicimolin


frontend/src/console/components/ThemeButtons/ThemeButtons.tsx
```diff
@@ -0,0 +1,163 @@
...
+                      messages={messages}
+                      loading={instance.loading.includes('draftThemeConfig')}
+                      hideTooltip={true}
+                    />
+                  </Col>
+                </Row>
+              </Container>
+            </CollapseEditArea>
+
+            <CollapseEditArea initialExpanded={false}>
```
giovannicimolin: @Agrendalath You just need 1 `CollapseEditArea` for each set of buttons.

Agrendalath: Right, good catch :+1:

alanoe:
Please include test instructions. For now, I have done ad-hoc tests.

UI bugs:

- Am I supposed to be able to open the color edition UI for several colors (eg button color, button text color and button border color) simultaneously? I can do it if I click in the corresponding color textfields sequentially. Also, if I do it, clicking outside the textfields don't close them. The only way to close them is to click again on color edition UI and then click outside it.

- If I hover the "i" icon in each button header (eg Primary button), the information popup shows up far from the "i" icon,in the far right edge of the screen. It show up next to the "i" icon.

- arrows next to "Edit" strings are inverted. They should point down when the button theme edition UI is hidden and up when it is shown.

- "register button" information says "reregistration". I think that is a typo?


Usability issues: 

- "i" (information) icons would be more easily associated with buttons headers (eg Primary button) if they were in the same line and closer to the header instead of in the next line.
- Each button header (eg Secondary button) and "Edit" strings are very close, making UI elements too crammed. Adding a vertical space between those strings would make reading easier.
- "Logistration button" header (and corresponding information show when hovering the information icon) makes no sense to user. A better name and description should be used.
- button preview should not be in the far right edge of the screen, but rather closer to the button theme UI.
  Maybe in the line after the button header (eg "Primary button")? Closely related UI elements should appear next each other   to make clear to the user they refer to the same thing. 

If the design cannot be changed, feel free to ignore these.

giovannicimolin:
@Agrendalath I think I've found an issue with the current implementation:
If the `customize-sign-in-btn` CSS flag isn't set, the button customizations won't be deployed ([reference in simple-theme](https://github.com/open-craft/edx-simple-theme/blob/simple-theme-enhancements/lms/static/sass/theme-overrides.scss#L71)).

And setting that variable will only work if all button related variables are set otherwise the deployment will break. Can you modify the theme API to check if all variables are set before enabling the flag? You can also put a switch in the frontend and require all variables to be set before letting the user set it.

Agrendalath:
@giovannicimolin, 
> And setting that variable will only work if all button related variables are set otherwise the deployment will break. Can you modify the theme API to check if all variables are set before enabling the flag? You can also put a switch in the frontend and require all variables to be set before letting the user enable it.

It took me a moment to understand this, as this is a bit confusing that we have a different approaches for for primary/secondary buttons and login/registration ones. Therefore I think we should go with the explicit (switch and a message on the button customization page) approach if we want to implement this on the frontend side, because it will also be confusing for the end user.

However I don't fully understand why we're doing it like this. We are setting default values for primary/secondary buttons [here](https://github.com/open-craft/edx-simple-theme/blob/simple-theme-enhancements/lms/static/sass/theme-colors.scss), right? Couldn't we do the same for login/registration buttons as well to avoid making the UX more complex? Am I missing something here?

giovannicimolin:
@lgp171188 Can you give some advice here? How hard would it be to add defaults to our current simple-theme?
If we add those defaults, can we remove the flag dependency? 

@Agrendalath 
>Therefore I think we should go with the explicit (switch and a message on the button customization page) approach if we want to implement this on the frontend side, because it will also be confusing for the end user.

:+1: Yup, agreed. Can you add the switch in the changes?

lgp171188:
> However I don't fully understand why we're doing it like this. We are setting default values for primary/secondary buttons here, right? Couldn't we do the same for login/registration buttons as well to avoid making the UX more complex? Am I missing something here?

@Agrendalath, the reason why the simple-theme skeleton was implemented like this is though we have "primary" and "secondary" button styles defined, there are no guidelines or well-defined ways to determine which style to apply for each button. And in some of the client themes that I used as the basis for developing this skeleton, these login, register and the login/register page buttons had different colors that couldn't be easily fit into the "primary"/"secondary" model that will work for all those themes.

In the LMS, there are 2 distinct styles of buttons that are used. Ones which swap the text/border color with the background color on hover and the others which remain the same on hover and without hover.

Often, these buttons were styled differently from the rest of the buttons - some of the themes had solid, unchanging button styles for these buttons on hover whereas the rest of the buttons behaved as defined in the primary/secondary button styles. Hence the implementation ended up being how it is now. Now that we don't have any budget to fix or improve the theme skeleton, the tweaks to the values have to be done in Ocim either in the frontend or the backend wherever it is more appropriate.

To simplify the UX, I think we could automatically set the hover colors to be the inverse of the default colors when the user selects the background and text colors. After that, the user could still change the hover colors if they want to.

CC @giovannicimolin

Agrendalath:
@lgp171188, thank you for explaining the reasoning behind this.

> To simplify the UX, I think we could automatically set the hover colors to be the inverse of the default colors when the user selects the background and text colors. After that, the user could still change the hover colors if they want to.

I'm not sure if this implicit behavior would be user-friendly, since we [need](https://github.com/open-craft/opencraft/blob/new-console/instance/schemas/theming.py#L116-L141) them to fill all 6 styles anyway.

@giovannicimolin, I added the switch locally, but disabling the customization does not work - when I submit PATCH with `{"customize-sign-in-btn": false}`, I receive error 400 with `{"non_field_errors": "Schema validation failed."}` body. I tried settings this to `false` with Django admin and got the following exception:
```python
Failed validating 'enum' in schema[1]['properties']['version']:
    {'enum': [0], 'type': 'number'}

On instance['version']:
    1
```


Enabling it via API or removing it with Django admin works fine.

lgp171188:
@Agrendalath,

> I added the switch locally, but disabling the customization does not work

We can always have it enabled and just the set the default values to match the values in the default theme if the user hasn't enabled the customization or turned it off (if there is such a thing). Would that help?

Agrendalath:
@lgp171188,
> if the user hasn't enabled the customization or turned it off (if there is such a thing).

AFAIK there is only this `customize-sign-in-btn` flag in schema, which cannot be disabled.

> We can always have it enabled and just the set the default values to match the values in the default theme

Yep, it sounds the most user-friendly to just have all these fields pre-filled. I added this:
```python
    # Sign In button theming
    "customize-sign-in-btn": True,
    "btn-sign-in-bg": "#126F9A",
    "btn-sign-in-color": "#FFFFFF",
    "btn-sign-in-border-color": "#126F9A",
    "btn-sign-in-hover-bg": "#FFFFFF",
    "btn-sign-in-hover-color": "#126F9A",
    "btn-sign-in-hover-border-color": "#126F9A",
    # Register button theming
    "customize-register-btn": True,
    "btn-register-bg": "#126F9A",
    "btn-register-color": "#FFFFFF",
    "btn-register-border-color": "#126F9A",
    "btn-register-hover-bg": "#FFFFFF",
    "btn-register-hover-color": "#126F9A",
    "btn-register-hover-border-color": "#126F9A",
    # Logistration button theming
    "customize-logistration-btn": True,
    "btn-logistration-bg": "#126F9A",
    "btn-logistration-color": "#FFFFFF",
    "btn-logistration-border-color": "#126F9A",
    "btn-logistration-hover-bg": "#FFFFFF",
    "btn-logistration-hover-color": "#126F9A",
    "btn-logistration-hover-border-color": "#126F9A",
```
to the [DEFAULT_THEME](https://github.com/open-craft/opencraft/blob/new-console/instance/schemas/theming.py#L154), but creating new instance with `POST` to `/v2/instances/openedx_config/` gives me:
```python
ValidationError at /api/v2/instances/openedx_config/
1 is not one of [0]
Failed validating 'enum' in schema[1]['properties']['version']:
    {'enum': [0], 'type': 'number'}
On instance['version']:
    1
```
This is pretty confusing.
cc: @giovannicimolin

lgp171188:
> to the DEFAULT_THEME, but creating new instance with POST to /v2/instances/openedx_config/ gives me:

@Agrendalath, this is a known issue with the `jsonschema` library where when two schemas are combined using operators like `anyOf`, `oneOf`, when there is some extra field added in the 2nd or further sub-schemas, it tries to match it against the first schema and then fails with an error like this which is misleading.

Can you locally try to validate it against the `v1` schema to find out what the issue here is? From the error, It looks like some additional non-existing field is added as I remember seeing a similar error when that happened.

Agrendalath:
@lgp171188, thank you for the resources, they helped me with debugging this! It turned out that `flag_schema` supports [only](https://github.com/open-craft/opencraft/blob/new-console/instance/schemas/theming.py#L34) `True` value.

@giovannicimolin, I might not have the full overview of this, but if I understand this correctly, we have 2 options here:
1. Add `False` (or simply `{"type": "boolean"}`) to [flag_schema](https://github.com/open-craft/opencraft/blob/new-console/instance/schemas/theming.py#L34) and add boolean checks for [customize-* variables](https://github.com/open-craft/edx-simple-theme/blob/simple-theme-enhancements/lms/static/sass/theme-overrides.scss#L71).
1. Create new Ocim endpoint for removing the field from theme config.

What do you think?

lgp171188:
> @lgp171188, thank you for the resources, they helped me with debugging this! It turned out that flag_schema supports only True value.

@Agrendalath, this is because it is a literal flag and not a toggle. It is either present or not present and since it needed a value, I chose a truthy value both in the schema and in the theme SCSS.

giovannicimolin:
@Agrendalath The API should have support for erasing the values.
After checking the code, I realized that I didn't consider the possibility of some values being booleans, just empty strings (this is the exact line: https://github.com/open-craft/opencraft/blob/new-console/registration/api/v2/views.py#L239).
By modifying it to:
```
if not value
``` 
Any falsey value will be erased from the final theme config (including empty strings and `false` values).

frontend/src/ui/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,37 @@
...
+    setValue(props.initialValue || false);
+  }, [props.initialValue]);
+
+  const onChange = () => {
+    props.onChange(props.fieldName, !value);
+  };
+
+  return (
+    <label className="switch">
+      <input
```
giovannicimolin, line 28: @Agrendalath Can you use the react boostrap default switch (https://react-bootstrap.github.io/components/forms/#forms-custom-switch)?

It also looks too big and misaligned for the rest of the text (the one from bootstrap handles scaling automatically).

image

giovannicimolin:
@Agrendalath Don't forget to rebase this before hitting `Squash and merge`. The **new-console** branch was rebased.

PR #548, opened 2020-03-18
BB-2179: Registration form and welcome email fixes.

giovannicimolin:
This PR introduces 2 changes:
1. Fixes broken link on welcome email and add support for empty link.
2. Add custom password validators.

**Testing instructions:**
1. Checkout this branch.
2. Set up some email configuration in `.env`.
```
EMAIL_BACKEND='django.core.mail.backends.smtp.EmailBackend'
EMAIL_HOST=smtp.gmail.com
EMAIL_USE_TLS=True
EMAIL_PORT=587
EMAIL_HOST_USER=your_email@opencraft.com
EMAIL_HOST_PASSWORD=
```
3. Run Ocim and create a betatest instance (no need to activate emails).
4. Run `make shell` and then the following commands:
```
from registration.utils import send_welcome_email
from registration.models import BetaTestApplication

# Pick the betatest instance that you created here (the user email must be valid)
x = BetaTestApplication.objects.all()[0] 
send_welcome_email(x)                                                                            
```
5. Open the email and check that the `Customize` link has a link to localhost.
6. Exit shell, change `USER_CONSOLE_FRONTEND_URL` to an empty string (in `.env` file).
7. Run step 4 and check if the email has a "Coming soon" button instead of "Customize".
8. Check if password validators are working properly using:
- Swagger API
- Old registration form
- New registration form
- Django Admin

**Reviewer:** 
- [ ] @lgp171188

### DIFF
--- a/registration/validators.py
+++ b/registration/validators.py
@@ -27,7 +27,7 @@ class UppercaseValidator:
     Uppercase password validator
     """
     def validate(self, password, user=None):  # pylint: disable=missing-docstring
-        if not re.findall('[A-Z]', password):
+        if not re.findall('r[A-Z]', password):
             raise ValidationError(
                 "The password must contain at least 1 uppercase letter, A-Z.",
                 code='password_no_upper',
@@ -42,7 +42,7 @@ class LowercaseValidator:
     Lowercase password validator
     """
     def validate(self, password, user=None):  # pylint: disable=missing-docstring
-        if not re.findall('[a-z]', password):
+        if not re.findall('r[a-z]', password):
             raise ValidationError(
                 "The password must contain at least 1 lowercase letter, a-z.",
                 code='password_no_lower',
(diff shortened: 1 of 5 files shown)


registration/validators.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_help_text(self):  # pylint: disable=missing-docstring
+        return "Your password must contain at least 1 lowercase letter, a-z."
+
+
+class SymbolValidator:
+    """
+    Symbol password validator
+    """
+    def validate(self, password, user=None):  # pylint: disable=missing-docstring
+        if not re.findall('[,.<>/?:;"-+=%$&@#{}()]', password):
```
lgp171188: @giovannicimolin, though `!` is a valid symbol character, this validator doesn't accept it.

giovannicimolin: I've improved the regex and added other missing special characters.

lgp171188: @giovannicimolin, it looks like the fix for this has caused the quality checks to fail. Can you check?

lgp171188:
@giovannicimolin, the unit tests are failing now after your latest update. One of the failing tests is a flaky one and a retry should likely fix it.

PR #550, opened 2020-03-19
[BB-2217] Store static templates customizations and deploy the changes

lgp171188:
This PR implements the changes to store static the templates customizations and deploy them by generating the appropriate ansible variables to make use of the changes in https://github.com/open-craft/configuration/pull/121. It also implements the registration v2 API changes to allow setting and patching these values from the frontend.

**Testing instructions**:
* Apply the migrations.
* Start the Ocim app server (`make run.dev` in the devstack).
* Log in to Ocim as a superuser after creating a superuser account, if there isn't one already.
* In a separate terminal tab/window, start the frontend app.
* Register for a new betatest instance using the frontend and verify the email addresses given.
* Once the instance is created automatically after verification, navigate to the swagger API page `http://localhost:5000/api/swagger`.
* Check the listed API endpoints and verify that the `openedx_config` endpoints include the `draft_static_content_overrides` field.
* Verify that in the response of the `/v2/instances/openedx_config/` endpoint, the instance created from the new betatest application is listed and has the `draft_static_templates_overrides` set to `null`, the default value.
* Verify that there is a separate API endpoint which allows the `PATCH` request to modify individual fields.
* In the Django admin interface, navigate to the betatest instance created before and change the `draft_static_content_overrides` field's value to `{"version": 0}` and save it.
* Try executing `PATCH` requests to add more supported fields and update existing fields.
* Trigger a redeployment by executing the `/v2/instances/openedx_deployment/` endpoint with a POST request.
* Verify that the `draft_static_content_overrrides` value is copied to the `static_content_overrides` field in the `OpenEdXInstance` model instance corresponding to the betatest form.
* If it is possible to launch an appserver, verify that there is a section for the `SiteConfiguration` settings in the appserver configuration section. Also verify that the ansible variables there correspond to the format expected by https://github.com/open-craft/configuration/pull/121.
* If it is not possible to launch an appserver in the devstack, run the `.get_site_configuration_settings()` method on the `OpenEdXInstance` model instance and verify the generated ansible variables.
* Try testing other permutations/combinations.

* The `draft_static_content_overrides` field and the `static_content_overrides` field use `None` as their default value. However, to be able to update the former using `PATCH` request, a non-null value for the field must be set. So the expectation is that, whenever the user enables the customization for the first static template, a `PUT` request to the instance config endpoint will be required to populate the field with a non-null value.
* The `create_or_update_site_configuration` management command does not allow deleting any `SiteConfiguration` value keys. So the keys once added, cannot be deleted. However, setting their value to `''` or `null` will trigger the default value in the template to be used.
* I have not updated the frontend API client in this PR as that will add an additional field to the instance config API models and that may or may not break the frontend tests, fixing which is not in the scope of this task. But when the frontend task is being worked upon, I will fix any reported issues with the generated schema.


registration/tests/test_views.py
```diff
@@ -592,7 +592,7 @@ def test_modify_user(self):
         application = BetaTestApplication.objects.get()
         self._assert_application_matches_form_data(application)
 
-    @override_settings(VARIABLES_NOTIFICATION_EMAIL='notifications@opencraft.com', INSTANCE_STORAGE_TYPE='s3')
+    @override_settings(VARIABLES_NOTIFICATION_EMAIL='notifications@opencraft.com')
```
giovannicimolin: @lgp171188 Any reason for this change?

registration/tests/test_api.py
```diff
@@ -430,6 +430,53 @@ def test_theme_config_api_erase_value(self):
...
+        """
+        self.client.force_login(self.user_with_instance)
+        self._setup_user_instance()
+
+        response = self.client.patch(
+            reverse(
+                f"api:v2:openedx-instance-config-static-templates-overrides-config",
+                args=(self.instance_config.pk, )
+            ),
+            data={'version': 1},
```
giovannicimolin: Isn't this supposed to be version 0? Or are you specifically trying to make the API error out here?

giovannicimolin: Btw, if there's no template overrides set, and I make a request, will it error out?

instance/migrations/0124_openedxinstance_static_templates_overrides.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('instance', '0123_newrelicsslalertcondition_newrelicsslmonitor'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='openedxinstance',
+            name='static_templates_overrides',
+            field=django.contrib.postgres.fields.jsonb.JSONField(blank=True, help_text="The final static templates' override data committed by the user for deployment. This should bepicked up the appservers launched for this instance.", null=True, validators=[instance.schemas.static_templates_overrides.static_templates_overrides_schema_validate], verbose_name='Draft static templates overrides JSON'),
```
giovannicimolin: Nit: Missing space on `This should bepicked up the appservers` ("bepicked").

instance/models/mixins/openedx_custom_static_pages.py
```diff
@@ -0,0 +1,49 @@
...
+    class Meta:
+        abstract = True
+
+    static_templates_overrides = JSONField(
+        verbose_name='Draft static templates overrides JSON',
+        validators=[static_templates_overrides_schema_validate],
+        null=True,
+        blank=True,
+        default=None,
+        help_text=("The final static templates' override data committed by the user for deployment. This should be"
```
giovannicimolin: Nit: Can you use consistent quotation marks and add a line break on this string?

instance/models/mixins/openedx_site_configuration.py
```diff
@@ -0,0 +1,52 @@
...
+            if static_templates_overrides:
+                site_configuration_settings.update(static_templates_overrides)
+
+        if site_configuration_settings:
+            return yaml.dump(
+                {'EDXAPP_SITE_CONFIGURATION': [{'values': site_configuration_settings}]},
+                default_flow_style=False
+            )
+        else:
+            return ''
```
giovannicimolin: Shouldn't we return `{}` instead of an empty string? This might cause issues when merging the dictionaries, no?

instance/tests/integration/integration_instance.py
```diff
@@ -327,6 +337,7 @@ def test_spawn_appserver(self):
         OpenEdXInstanceFactory(
             name='Integration - test_spawn_appserver',
             deploy_simpletheme=True,
+            static_templates_overrides={'version': 0, 'static_template_about_content': 'Hello world!'},
```
giovannicimolin: Awesome! Thanks for adding an integration test for this :)

giovannicimolin:
@lgp171188 

>The draft_static_templates_overrides field and the static_templates_overrides field use None as their default value. However, to be able to update the former using PATCH request, a non-null value for the field must be set. So the expectation is that, whenever the user enables the customization for the first static template, a PUT request to the instance config endpoint will be required to populate the field with a non-null value.

A `PUT` request isn't valid on the `theme_config` endpoint since the templates override field is defined by a read only serializer. You either need to add support to the endpoint to automatically set values without erroring out or creating a second endpoint just for making the initial post request (I prefer if we go with the first solution since it makes handling things in the frontend easier).

instance/migrations/0124_openedxinstance_static_templates_overrides.py
lgp171188: @giovannicimolin, the source of this issue is in the field definition. So I will fix it here and there.

instance/models/mixins/openedx_custom_static_pages.py
lgp171188: @giovannicimolin, I default to using the single quotes for strings and in this case I had to switch to using the double quotes for the string because there is a single quote within the string and I didn't want to escape it.

> add a line break on this string?

I don't see line breaks used anywhere in the help text for the fields. Even the `deploy_simple_theme` field, which has a significantly longer help text and the `theme_config` field which is of a comparable length (and from where I copied the help text format) don't have line breaks. So is this really required?

instance/models/mixins/openedx_site_configuration.py
lgp171188: @giovannicimolin, the return statement inside the `if` block above returns the string containing the rendered variables. So the empty string `''` is consistent with that. If we have to merge the `SiteConfiguration` variables from different sources - for example, there will be one needed for the home page hero text customizastion, we have the `yaml_merge` function [here](https://github.com/open-craft/opencraft/blob/master/instance/ansible.py#L57) that can be used like it has been in a few existing places.

registration/tests/test_views.py
lgp171188: @giovannicimolin, this was a change that I picked up from the `master` branch to fix a failing unit test as that hasn't yet been merged to the `new-console` branch, on which this PR branch is based.

registration/tests/test_api.py
lgp171188: > Isn't this supposed to be version 0? Or are you specifically trying to make the API error out here?

Yes, you are right. This looks like a typo. I will fix it :+1: 

> Btw, if there's no template overrides set, and I make a request, will it error out?

Yes, with the current implementation, it will error out. However, the [RFC 5789](https://tools.ietf.org/html/rfc5789) for the `PATCH` method states 

```
If the Request-URI does not point to an existing resource, the server MAY create a new resource,
depending on the patch document type (whether it can logically modify a null resource)
and permissions, etc.
```
In this case, we could either allow the `PATCH` method to treat this as a missing resource and create it if it doesn't exist or treat the default `null` value as the resource being present and allow patching. Either way the implementation will be the same.

lgp171188:
> A PUT request isn't valid on the theme_config endpoint since the templates override field is defined by a read only serializer. You either need to add support to the endpoint to automatically set values without erroring out or creating a second endpoint just for making the initial post request (I prefer if we go with the first solution since it makes handling things in the frontend easier).

@giovannicimolin, I have replied to this comment [here](https://github.com/open-craft/opencraft/pull/550/files#r396112726) and the changes I am going to make for addressing that review comment is going to resolve this as well.

.prospector/opencraft.yml
```diff
@@ -5,7 +5,7 @@ dodgy:
   enable: []
   options: {}
   run: null
-ignore-paths: ['migrations', 'deploy/', 'venv/']
+ignore-paths: ['migrations', 'deploy/', 'venv/', 'frontend/']
```
lgp171188: I added this because there is no Python code that we have written inside the `frontend` folder,

instance/models/mixins/openedx_custom_static_pages.py
lgp171188: @giovannicimolin, I have updated the PR address all the review comments except this one as you may not have had a chance to see my reply yet.

registration/tests/test_api.py
lgp171188: @giovannicimolin, I have updated the `PATCH` method to automatically update the `None` default value to the data passed in the `PATCH` request and also add the version if it is not present in the request data.

lgp171188:
@giovannicimolin, the integration tests are failing due to the Keystone migration issue `keystoneauth1.exceptions.http.BadRequest: Expecting to find domain in project`, the fix for which was merged in the `master` branch and is not in the `new-console` branch or this PR branch. I'd like to avoid pulling in those changes in this PR.

instance/models/mixins/openedx_custom_static_pages.py
giovannicimolin: These changes are not required at all! 
I was just going over the code and commented on everything a little bit off, but it's okay to keep the implementation here to leave the code consistent.

PR #552, opened 2020-03-20
[BB-2220] Add navigation customization page

Agrendalath:
**Styling of the right side of the navbar needs some tweaks.**

This PR adds button customization page to the new Ocim frontend.

**Testing instructions:**
1. Checkout this branch.
1. Create a new test instance and log in the console.
1. Go to the `Theme` -> `Navigation` page and make sure it's working as expected (http://localhost:3000/console/theming/navigation).
1. Change colors and check if the values are correctly saved in the backend.

**Reviewers:**
- [x] @giovannicimolin


frontend/src/console/components/ThemeNavigation/ThemeNavigation.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <h2>
+            <WrappedMessage messages={messages} id="themeNavigation" />
+          </h2>
+
+          {themeData && themeData.version === 1 && (
+            <div className="theme-navigation-container">
+              <NavigationMenu themeData={themeData} />
+              <NavigationMenu themeData={themeData} loggedIn />
+              <Row>
+                <Col>
```
giovannicimolin: @Agrendalath Can you try using `<Col md={3}>` for each field?

Agrendalath: @giovannicimolin, I already tried this, but the vertical alignment of the input fields looked odd to me due to text wrapping. I just fixed it with `align-items: end`.

This is how it looks with `md={4}`:
md4

This is how it looks with `md={3}`:
md3

For now I left `md={4}`, as it looked more readable to me. What do you think?

opencraft/settings.py
```diff
@@ -192,6 +192,8 @@
 
 STATIC_ROOT = root('build/static')
 STATIC_URL = '/static/'
+MEDIA_ROOT = root('media')
+MEDIA_URL = '/media/'
```
giovannicimolin: Isn't this change missing the `urlpatterns` entry too?

frontend/src/console/components/NavigationMenu/NavigationMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {/*  Sign in */}
+        {/* </CustomizableButton> */}
+      </Col>
+    </Row>
+  );
+
+  return (
+    <Row
+      className="main-navigation-menu navigation-menu"
+      style={{ background: themeData.mainNavColor }}
```
giovannicimolin: The nav bar is showing up as transparent by default for me. The same happens when clicking on the reset button in the background color field.
image

Can you add the LMS default colors if `themeData` doesn't have those set?

frontend/src/console/components/ThemeNavigation/ThemeNavigation.tsx
```diff
@@ -0,0 +1,140 @@
...
+                loggedIn
+              />
+              <Row>
+                <p className="style-name">
+                  <WrappedMessage messages={messages} id="navigationLinks" />
+                </p>
+              </Row>
+              <Row>
+                <Col md={4}>
+                  <ColorInputField
```
giovannicimolin: There's some issue with the `useEffect` usage in this page. When clicking on `reset`, the colors are reset to the default value, but the field contents are not updated. The same thing doesn't happen in the `Theme preview and colors` page. 
Peek 2020-03-27 10-42
Can you check what's causing this?

opencraft/settings.py
Agrendalath: I added `ENABLE_DEBUG_TOOLBAR=true` to my `.env`, so it was automatically added [here](https://github.com/open-craft/opencraft/blob/master/opencraft/urls.py#L52). Not sure if we have a common `.env` file for development devstack. If so, it might be good to add it there.

giovannicimolin: Btw, can you add a conditional here? Only set these values if `DEBUG=True`, otherwise this might mess up the S3 stage and production configuration.

frontend/src/console/components/ThemeNavigation/ThemeNavigation.tsx
giovannicimolin: Alignment looks great now!

opencraft/settings.py
Agrendalath: Sure, done.

frontend/src/console/components/NavigationMenu/NavigationMenu.tsx
Agrendalath: Done, I also fixed the "bumping" text on hover.

frontend/src/console/components/ThemeNavigation/ThemeNavigation.tsx
Agrendalath: We're getting this error:
```
Warning: A component is changing a controlled input of type undefined to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

It seems to be happening only when from/to empty string.

giovannicimolin: That's weird, this might not happen in the other page because the backend returns default values instead of empty ones when resetting.
Can you check if there's a straightforward fix for this?

PR #553, opened 2020-03-27
[BB-2239] Allow customizing the homepage hero section content

lgp171188:
TBD


registration/api/v2/views.py
```diff
@@ -292,6 +292,10 @@ def image(self, request, pk: str):
                 application.logo = request.data['logo']
                 application.save()
 
+            if 'hero_cover_image' in request.data:
```
lgp171188, line 295: Atm, this endpoint only allows uploading and saving an image and doesn't have a way to delete an existing image and reset the value of the `ImageField` to `None`. I will implement the changes for it soon.

lgp171188:
@Agrendalath, @giovannicimolin, almost all the backend changes required for this PR are now in this PR. So a review pass will be helpful.

instance/models/mixins/openedx_theme.py
```diff
@@ -150,6 +158,12 @@ def get_v1_theme_settings(self):
                 'value': value
             })
 
+        if application.hero_cover_image:
+            sass_overrides.append({
+                'variable': 'homepage-bg-image',
```
Agrendalath, line 163: Is this variable used in simple-theme?
I see only `$home-page-hero-title-color` and `$home-page-hero-subtitle-color` in the [README](https://github.com/open-craft/edx-simple-theme/tree/simple-theme-enhancements).

lgp171188: @Agrendalath, this is a built-in SASS variable already in the platform.

frontend/src/console/actions.ts
```diff
@@ -307,6 +329,57 @@ export const updateThemeFieldValue = (
...
+    });
+  }
+};
+
+export const updateHeroText = (
+  instanceId: number,
+  title: string,
+  subtitle: string
+): OcimThunkAction<void> => async (dispatch, getState) => {
+  dispatch({
```
giovannicimolin, line 371: The `ActionType` is being dispatched twice, once here, and once in `updateStaticContentOverridesFieldValue`.

Can you remove this dispatch?

instance/models/mixins/openedx_theme.py
```diff
@@ -71,6 +72,13 @@ def get_simple_theme_static_files_settings(self, application):
             },
         ]
 
+        if application.hero_cover_image:
```
giovannicimolin, line 75: @lgp171188 Did you get a chance to test if this change is properly deployed?

giovannicimolin: I'm worried that the file might not be deployed to the appserver.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,253 @@
...
+      if (matched) {
+        this.state = {
+          title: matched[1],
+          subtitle: matched[2]
+        };
+      }
+    }
+  }
+
+  private doesHomepageOverlayHtmlExist = (props: Props) => {
```
giovannicimolin, line 67: Since this function is bound to the class, you can use `this` inside, so there's no need to pass the props.

instance/models/mixins/openedx_theme.py
lgp171188: @giovannicimolin, this uses the same logic as the favicon and logo images. So I see no reason why this shouldn't work.

lgp171188:
@giovannicimolin, @Agrendalath, due to some issue related to the GitHub outage, I was unable to push changes to this branch. So I deleted the branch and recreated it which caused this PR to close. So I will create a new PR. Sorry!

Agrendalath:
@lgp171188, [FYI](https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c), but a new PR is also fine :)

PR #554, opened 2020-03-29
SE-2413 Fixes formatting on the rendered documentation How To page

pomegranited:
Fixes some formatting issues with the `mkdocs` rendered markdown for the "How To" documentation page, cf https://ocim.opencraft.com/en/latest/howtos/

Changes made:

* established standard line length of 200 characters.
* added yaml/bash code rendering hints to code blocks
* code blocks cannot be indented in mkdocs markdown
* code blocks mess with ordered list numbering, so converted to unordered lists.
* comments in pasted shell commands mess with the backslash line continuation character, so have moved them into the main content.
* changed heading levels, added more headings.
* moved all extra configuration changes to the top instance configuration section.

**Testing instructions**

1. Run `make docs` to build the documentation.
1. Run `make docs-serve` to check the rendered documentation on `http://localhost:5001`.
1. Read through the rendered documentation and check for formatting issues.

**Reviewer**

- [x] @swalladge

### DIFF
--- a/documentation/howtos.md
+++ b/documentation/howtos.md
@@ -1,6 +1,8 @@
 ## Configuring Ecommerce and Course Discovery
 
-By default, instances are not provisioned with either ecommerce or the Course Discovery service. However, support is available to manually enable those services.
+By default, instances are not provisioned with either Ecommerce or the Course Discovery service. However, support is available to manually enable those services.
+
+Running Course Discovery requires a separate, persistent VM to be deployed to host ElasticSearch, for performance and deployment reasons.
 
 ### Ocim instance extra configuration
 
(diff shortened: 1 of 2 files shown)


documentation/howtos.md
```diff
@@ -16,14 +18,6 @@ EDXAPP_ADDITIONAL_CRON_JOBS:
   hour: "*"
   minute: "43"
   day: "*"
-# update_index is required but isn't currently part of the provisioning process. So in the interests of saving time, this is run every 10 minutes through cron (it's not an expensive operation).
```
pomegranited: FYI @swalladge , I moved this command down to a [one-off manual step](https://github.com/open-craft/opencraft/pull/554/commits/9f12462b7d4f7a746f86cd547a4dff21382b5094#diff-e62c375eb49e96b787ac278c5d78e70dR165), since we're now aware that we need to run ES on a separate VM.

swalladge: Looks good. Could you  also clarify that this command is run on the appserver, and it's a once-off thing because the ES server is expected to be on a long-lived dedicated instance?

pomegranited: I don't think we need to clarify that at this line, since we're removing it entirely?  And I think [this line under Manual Steps](https://github.com/open-craft/opencraft/pull/554/commits/9f12462b7d4f7a746f86cd547a4dff21382b5094#diff-e62c375eb49e96b787ac278c5d78e70dR90) indicates that these are one-off steps.. does it need further clarification?

swalladge: Yep I meant the clarifications should go where the manual step is documented. It wasn't really clear to me that this command is to be run on the openedx appserver, and that the ES server is expected to be a dedicated server.

pomegranited: Gotcha, I see why that was confusing.  Is 71c62ce  better?

swalladge: LGMT :+1:

PR #557, opened 2020-04-01
[BB-2288] Upgrade requirements

xitij2000:
Upgrade requirements.


pomegranited:
@xitij2000  I've tested this on stage, and it seems to be working fine, so good to merge 👍 

* The websockets are still working. Changes made at the backend (instance config changes, appserver activation, spawning new appservers) are all reflected without refreshing on the UI.
* No errors reported in the logs.
* Browsing around produces nothing unexpected.

PR #559, opened 2020-04-02
[BB-2239] Hero customization page

lgp171188:
Replaces https://github.com/open-craft/opencraft/pull/553

Implements the changes for the hero customization page in the console and the corresponding changes in the backend.

**Testing instructions**:
* Checkout the PR branch.
* Apply the migrations and start the backend app server.
* Build the API client and start the frontend server.
* Create a new betatest instance using the new registration form.
* Verify the email addresses provided during the registration from the django shell by running the below statement for each distinct email address:
  ```python
  EmailAddress.objects.confirm(EmailAddress.objects.get(email='<email address>').key)
  ```
* Verify that the `draft_static_content_overrides` field has been populated with the below default value.
  ```python
  {
      "version": 0,
      "homepage_overlay_html": "<h1>Welcome to <instance name></h1><p>It works! Powered by  Open edX®</p>"
  }
  ```
* Also verify that the `home-page-hero-title-color` and `home-page-hero-subtitle-color` variables have been set to their default values in the `draft_theme_config` field.
* Log in to the user console from the new frontend.
* You will land on the `Instance settings -> General` section. If the fields in that page are missing, click on each field and click outside to trigger the `onblur` event to force the data to get loaded. 

  **Note**: This is an existing bug somewhere in the component lifecycle which affects multiple components including the components created in this PR. This will be investigated and fixed in a separate task and PR.
* Navigate to the `Theme -> Hero` section. Verify that the hero preview component shows up with the default hero title and subtitle.
* Also verify that the text boxes below for the title and subtitle fields are filled with the same values as shown in the preview.
* For all the steps below that involve editing, uploading, deleting, verify that the changes are stored as expected in the database.
* Type in those text boxes and verify that the preview is updated when the `onblur` event happens.
* Delete all the text in either of the text boxes and click outside. Verify that the default value gets filled in that case.
* Verify that the color input fields corresponding to the title and subtitle colors have the correct value set and that changing those values updates the color in the preview.
* Upload a cover image of appropriate dimensions for the hero section and verify that it is set as the background image for the hero section, replacing the default shade of grey. Also verify that a `remove` link shows up below the upload button once the image is successfully uploaded.
* Click the `remove` link and verify that the background image is removed.
* Try refreshing the page. If the fields become empty. Visit the `Instance settings -> General` section, confirm that the fields there have the values filled and return to the Hero customization page.
* Verify that the preview component shows the current stored state as expected and all the fields have the expected values.
* Perform any other tests of your choice to try and make the component break in unexpected ways.
* No breakages for existing, un-migrated instances.
  * On the `new-console` branch create a new instance from the old registration form.
  * Switch to the PR branch and run the app.
  * Log in using the account created above via the new frontend.
  * Verify that the hero customization page loads without any errors but doesn't have all its functionality exposed.

* The preview of the background image will not be a 100% accurate and match what will be seen on the instance. This is because the hero preview section is a lot smaller in size than the LMS hero section and the background image there is set with the `background-size` property set to `cover`, which will stretch the image to fit the box. In this PR, I have scaled the uploaded image to fit the dimensions of the box that is there in the preview component. There is no way to give a 100% accurate result without having the dimensions of the preview component match that of the LMS.
* The component lifecycle issue affecting the components added in this PR and other existing components hasn't been fixed in this PR.
* The data migration to migrate the existing instances to pre-populate the default values for the `draft_static_content_overrides` field which are needed for the changes in this PR hasn't been created in this PR. This is because the changes in this PR also require migrating the instance theme configuration to the v1 version and that hasn't been done yet.


frontend/packages/api_client/src/runtime.ts
```diff
@@ -144,7 +144,7 @@ export class Configuration {
     constructor(private configuration: ConfigurationParameters = {}) {}
 
     get basePath(): string {
-        return this.configuration.basePath || BASE_PATH;
+        return this.configuration.basePath != null ? this.configuration.basePath : BASE_PATH;
```
giovannicimolin, line 147: Was this auto-generated? If so we probably have a version mismatch between my version and yours.

frontend/src/ui/components/HeroPreview/HeroPreview.spec.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { setupComponentForTesting } from "utils/testing";
+import { HeroPreview } from './HeroPreview';
+
+it('renders without crashing', () => {
```
giovannicimolin, line 5: Can you add a few tests for this component?

registration/api/v2/views.py
```diff
@@ -292,6 +299,10 @@ def image(self, request, pk: str):
                 application.logo = request.data['logo']
                 application.save()
 
+            if 'hero_cover_image' in request.data:
```
giovannicimolin, line 302: Does the current implementation allows erasing the image? Can we do a similar thing to the other endpoints, where sending an empty message erases the stored value?

frontend/packages/api_client/src/runtime.ts
lgp171188: @giovannicimolin, it was auto-generated and I verified that I have the latest version of all the packages installed.

registration/api/v2/views.py
lgp171188: @giovannicimolin, I have implemented the changes to allow deleting the image from the field. However, Django doesn't delete the image from the storage backend and recommends creating a cron job to delete unused images.

Please check the testing instructions and the code changes for the details.

frontend/packages/api_client/src/runtime.ts
lgp171188: @giovannicimolin, in `frontend/packages/api_client/.openapi-generator/VERSION`, the version in my PR is newer than the version that is in the `new-console` branch. This might be due to that. If using a newer version is a problem, we should pin the docker image to a known good version and update it when required.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+export class HeroComponent extends React.PureComponent<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      title: '',
+      subtitle: ''
+    };
+
+    if (this.homePageOverlayHtmlExists()) {
```
Agrendalath, line 55: This is `false` on page refresh. That's why it will work only when coming from another page.
We should add a `FIXME` to this part and address it in a follow-up, as this is a bug.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+                  <WrappedMessage id="hero" messages={messages} />
+                </h2>
+                <p>
+                  <WrappedMessage id="heroDescription" messages={messages} />
+                </p>
+              </Col>
+            </Row>
+            {themeData && themeData.version === 1 && (
+              <Row>
+                <Col md={12}>
```
Agrendalath, line 175: Couldn't we use just `<Col>`?

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+                      />
+                    </Col>,
+                    <Col md={4} key="heroSubtitleColor">
+                      <ColorInputField
+                        fieldName="homePageHeroSubtitleColor"
+                        initialValue={themeData.homePageHeroSubtitleColor}
+                        onChange={this.onChangeColor}
+                        messages={messages}
+                        loading={instance.loading.includes('draftThemeConfig')}
+                      />
```
Agrendalath, line 267: The styling of these fields doesn't look correct:
image

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+  if (props.homePageHeroSubtitleColor) {
+    subtitleStyles = {
+      color: props.homePageHeroSubtitleColor
+    };
+  }
+  if (props.homepageOverlayHtml) {
+    matched = heroTextRegex.exec(props.homepageOverlayHtml as string);
+  }
+  if (props.heroCoverImage) {
```
Agrendalath, line 33: Wouldn't it work without these checks? It seems to be falling back to defined styles, when these variables are `undefined`.

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
```diff
@@ -0,0 +1,51 @@
...
+    containerStyles = {
+      backgroundImage: `url("${props.heroCoverImage}")`,
+      backgroundSize: '731px 258px'
```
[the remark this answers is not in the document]
lgp171188: @Agrendalath, 

> Should we move backgroundSize to the scss file?

The reason I added it here because the `background-size` property is needed only when the `background-image` property is set. But I can add it to a CSS class and then conditionally add it to the element.

>Also wouldn't it make sense to use something like this instead?
> ```
> background-position: center;
> background-size: cover;
>```
>
> The cover would be a bit more accurate to what happens on LMS, no?
> Also currently there is a repeat effect is you apply an image that does not match the >expected aspect ratio. It doesn't look elegant.

The issue here is that the preview component doesn't have the same dimensions as the LMS hero section. So there is no way to accurately represent how the image of a given size will look in the LMS. While centering the background image and making it stretch to fit will work better here, it will likely not match how it will look on the instance itself.

Do you have any suggestions for dealing with this issue?

frontend/src/console/components/Hero/Hero.tsx
lgp171188: @Agrendalath, the issue here is not the `this.activeInstance.data.staticContentOverrides.homepageOverlayHtml` being undefined, but `this.activeInstance.data` itself being undefined which causes this condition to return `false`. This appears to be a component lifecycle issue which is also affecting a few other pages, if not all of them.

I added a `componentDidUpdate` method similar to the one in the `InstanceSettings` component and that didn't help. I also tried adding a `getDerivedStateFromProps` method and while that solved this issue, it also made the text input fields read-only for some reason.

I asked @giovannicimolin about this data loading issue and he mentioned that he will create a separate task for investigating and fixing this issue.

frontend/src/console/components/Hero/Hero.tsx
lgp171188: @Agrendalath, can you give more details on what doesn't look correct?

lgp171188:
@Agrendalath,

> However on an old instance the preview doesn't show at all. Should we create a migration that would make existing instances compatible with the required format or just ignore this and deprecate them for now?

I have already mentioned this in the `Author notes and concerns` section of the PR description :slightly_smiling_face: 

>The data migration to migrate the existing instances to pre-populate the default values for the draft_static_content_overrides field which are needed for the changes in this PR hasn't been created in this PR. This is because the changes in this PR also require migrating the instance theme configuration to the v1 version and that hasn't been done yet.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        }
+                        homepageOverlayHtml={
+                          staticContentOverrides!.homepageOverlayHtml
+                        }
+                      />
+                    )}
+                </Col>
+              </Row>
+            )}
+            <CollapseEditArea initialExpanded>
```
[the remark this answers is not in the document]
lgp171188: @Agrendalath, I added it because there are 2 separate sections/components in the page - the preview component and the fields to edit the data and it made sense to add the latter under a `CollapseEditArea`.

registration/api/v2/views.py
```diff
@@ -320,16 +332,22 @@ def static_content_overrides(self, request, pk=None):
...
+            current_value = ''
         else:
-            merged_values = serializer.validated_data
-            if 'version' not in merged_values:
-                merged_values['version'] = 0
+            current_value = merged_values['homepage_overlay_html']
+
+        merged_values['homepage_overlay_html'] = fill_default_hero_title_or_subtitle_text_if_missing(
+            current_value
+        ).format(application.instance_name)
```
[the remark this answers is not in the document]
lgp171188: @Agrendalath, in the original version that I wrote for doing this, the `.format()` method was called only with the default value (known good value in code and not set by user). However, while optimizing it to not repeat code, I ended up adding this issue. Fixed it now.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        Remove
+                      </button>
+                    )}
+                </Col>
+                {themeData &&
+                  themeData.version === 1 && [
+                    <Col md={4} key="heroTitleColor">
+                      <ColorInputField
+                        fieldName="homePageHeroTitleColor"
+                        initialValue={themeData.homePageHeroTitleColor}
```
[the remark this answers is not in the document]
lgp171188: @Agrendalath, the `home-page-hero-title-color` and the `home-page-hero-subtitle-color` variables are part of the default theme and should always be set. So does it still make sense to add the ` || ''` here?

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        onChange={this.onChangeColor}
+                        messages={messages}
+                        loading={instance.loading.includes('draftThemeConfig')}
+                      />
+                    </Col>,
+                    <Col md={4} key="heroSubtitleColor">
+                      <ColorInputField
+                        fieldName="homePageHeroSubtitleColor"
+                        initialValue={themeData.homePageHeroSubtitleColor}
```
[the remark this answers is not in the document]
lgp171188: https://github.com/open-craft/opencraft/pull/559#discussion_r403655648

frontend/src/console/components/Hero/Hero.tsx
lgp171188: @Agrendalath, I have come up with a way to fix this data loading issue - will update the PR shortly. However, that has introduced a bug when the user tries to empty either text box. Since I have spent way more time than estimated on this task already, I think we should try to fix this in a follow-up task. Is that okay @giovannicimolin?

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
lgp171188: @Agrendalath, the `.exec()` method requires a `string` and that is why the check for the existence of `props.homepageOverlayHtml` is required. The other 2 `if` conditions in this component check and set styles, so I don't think those can be removed.

Please let me know if I misunderstood your comment :)

frontend/src/console/components/Hero/Hero.tsx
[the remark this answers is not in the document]
Agrendalath: @lgp171188, ah, my initial check was with the previous version of v1 schema. If we include setting default hero values in the migration, it makes sense to leave it as is.

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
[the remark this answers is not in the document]
Agrendalath: @lgp171188,

> The reason I added it here because the background-size property is needed only when the background-image property is set. But I can add it to a CSS class and then conditionally add it to the element.

As far as I'm aware, moving them to CSS won't have any downsides and we'll have more organized styling, so +1 from me.

> The issue here is that the preview component doesn't have the same dimensions as the LMS hero section. So there is no way to accurately represent how the image of a given size will look in the LMS. While centering the background image and making it stretch to fit will work better here, it will likely not match how it will look on the instance itself.

LMS (and preview component) will also have different dimensions on different screens (e.g. I'm personally using vertical WQHD or horizontal 4k for this), so using fixed background size will not be deterministic, unless we enforce users to upload image with specific proportions or allow them to crop it by themselves. This is how it looks for me - you can see the repeated part on the bottom:
image

I think that the best thing at the moment would be having it with `center` and `cover`, just like in LMS. We could tweak this later, if needed.

frontend/src/console/components/Hero/Hero.tsx
Agrendalath: @lgp171188, ah, I should have been more precise here - the `Hero/styless.scss` is missing `background-color: white;` under `.form-control`.

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
Agrendalath: @lgp171188, I was referring to the `homePageHeroTitleColor`, `homePageHeroSubtitleColor`, but especially to `heroCoverImage`. CSS just falls back to another option if it receives something like `url("")`. Anyway, that's just a nit.

Agrendalath:
@lgp171188,
> I have already mentioned this in the `Author notes and concerns` section of the PR description slightly_smiling_face

Ah, I missed this one. Sorry for this!

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
lgp171188: > CSS just falls back to another option if it receives something like url(""). Anyway, that's just a nit.

@Agrendalath, I wasn't aware of that. But would prefer to keep it explicit :)

frontend/src/console/components/Hero/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+      margin-left: 0;
+      margin-right: 0;
+    }
+  }
+  .info-icon {
+    padding: 5px 10px;
+    color: $primary-1;
+  }
+
+  .image-upload-field {
```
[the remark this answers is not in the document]
lgp171188: @Agrendalath, I have moved this inside the `.hero-customizations` selector.

frontend/src/console/components/Hero/Hero.tsx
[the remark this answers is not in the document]
lgp171188: @Agrendalath, I will leave it as-is unless this is a dealbreaker. We can always change this later if required. CC @giovannicimolin

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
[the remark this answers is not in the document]
lgp171188: @Agrendalath, I have set the background position to `center` and the size to `cover`.

lgp171188:
@Agrendalath, I have updated the PR with all the fixes. Can you let me know if anything is still unresolved or if there are any new issues?

frontend/src/ui/components/HeroPreview/HeroPreview.tsx
Agrendalath: @lgp171188, sure, makes sense!

frontend/src/console/components/Hero/Hero.tsx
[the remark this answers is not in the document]
Agrendalath: @lgp171188, sounds good to me :+1:

Agrendalath:
Please fix [the linting error](https://circleci.com/gh/open-craft/opencraft/14522) before merging.

:+1:
* [x]  I tested this: followed testing instructions, tested empty title and subtitle cases with direct `PATCH` requests to `/v2/instances/openedx_config/{id}/static_content_overrides/`
* [ ]  <s>I checked for accessibility issues</s>
* [ ]  <s>Includes documentation</s>

frontend/packages/api_client/src/runtime.ts
giovannicimolin: No problem, I will update my local image.
As long as it doesn't break the deployments I'm fine with it :smile:

registration/api/v2/views.py
giovannicimolin: @lgp171188 Thanks for implementing this! Looks good.

frontend/src/console/components/Hero/Hero.tsx
giovannicimolin: Sure! We can leave this for a follow up task. Can you add a note to the epic explaining the issue and how to reproduce it?

lgp171188:
@giovannicimolin, @Agrendalath, the integration tests for this PR have failed due to the usual flakiness. I have restarted the failed job but it looks unlikely to complete before the meeting today. So I will wait till they pass before merging this PR.

PR #562, opened 2020-04-13
Set timestamp when DNS records are set/updated.

mtyaka:
This updated adds a new `dns_records_updated` datetime field to `OpenEdXInstance` model. 
The field is initially `None` and gets set to current time every time DNS records in Gandi are set or updated. That typically happens when spawning a new app server.

The patch includes a data migration which iterates through all existing OpenEdXInstances and sets the timestamp of `dns_records_updated` to the creation datetime of the first app server belonging to the instance. If instance has no app servers, the field is left `None`.

The value of the field is added to the instance metadata stored in consul in the form of a unix timestamp.

Note that this doesn't really change the behaviour of Ocim by itself. The extra metadata field will be used by cert-manager only.

This PR is part of [SE-2431](https://tasks.opencraft.com/browse/SE-2431) which also includes:
- https://github.com/open-craft/cert-manager/pull/13
- https://github.com/open-craft/ansible-playbooks/pull/166

**Test instructions**:

1. Check out this branch.
1. Stop Ocim.
1. Run `make migrate`.
1. Run `make manage update_metadata`.
1. Restart Ocim: `make run`.
1. In the Ocim UI, find an instance with at least one appserver ([example](https://stage.console.opencraft.com/instance/4850/)) and check that the value under "DNS records set at" field in the instance Settings tab is populated. Find another instance which does not have any appservers ([example](https://stage.console.opencraft.com/instance/5000/)) and verify that the "DNS records set at" field is blank.


instance/migrations/0128_fill_dns_records_updated_field.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 2.2.9 on 2020-04-15 19:50
```
giovannicimolin, line 1: Can you merge these 2 migrations into one and rebase it on master? 
There are a few conflicting migrations.

mtyaka: @giovannicimolin I already rebased on top of master yesterday and manually resolved migration dependency conflicts. Are you still running into issues with the migrations?
I did not merge the two migrations into a single one though. Would there be any benefit in doing that?

giovannicimolin: >Would there be any benefit in doing that?

Not really, just having less migrations. :stuck_out_tongue_closed_eyes:
This is just a nit so it's fine to keep it as is.

PR #563, opened 2020-04-14
BB-2241: Bug fixes

giovannicimolin:
This PR fixes a few bugs in the new console:

1. Header customization background was pointing to the wrong variable.
2. External domain registration wasn't working, internal subdomain generation moved to python.

**Testing instructions for 2:**
1. Checkout this branch.
2. Create a new instance with an external domain, it should succeed.
3. Log in Ocim adming and check that a subdomain slug was properly generated (in this format `<domain-slug>-<random-hash>`).

**Reviewer:**
- [ ] @lgp171188


registration/api/v2/views.py
```diff
@@ -178,6 +181,27 @@ def validate(self, request):
...
+            # Hash is added to avoid name clashes when generating slug
+            random_hash = ''.join(random.choices(string.ascii_lowercase + string.digits, k=5))
+            subdomain = f"{slugify(external_domain)[:30]}-{random_hash}"
+            # Set request data
+            self.request.data.update({
+                "subdomain": subdomain
+            })
+
+        # Perform create as usual
+        return super(OpenEdXInstanceConfigViewSet, self).create(request, *args, **kwargs)
```
lgp171188, line 203: @giovannicimolin, should we check for conflict with an existing value and try regenerating the slug again if there is a conflict?

giovannicimolin: It's very unlikely that this will cause any clashes, since the domain name is appended first and we do a check for duplicated external domains. 
The internal domain is the slugified external domain truncated to 32 characters + 5 random digits to make it even harder to have a conflict (for cases with huge domain names where 2 instances have the same first 32 digits).  

Can we merge this without the retry mechanism and see if this causes any issues?

PR #565, opened 2020-04-16
[BB-2290] Add password recovery

Agrendalath:
This adds [django-rest-passwordreset](https://github.com/anexia-it/django-rest-passwordreset), which provides API for resetting password and adds the password recovery workflow to the new Ocim frontend.

**Testing instructions:**
1. Checkout this branch.
1. Create a new account in console.
1. Go to the `Login` page and click on `I forgot my password`.
1. Test the following:
   1. non-email value,
   1. non-existing value,
   1. existing value.
1. The last one from the previous point will output an email in the terminal that runs Ocim. Optionally you can copy the whole email and create HTML file from it to check that it looks correctly.
1. Copy the reset link from an email and paste into the browser. Try the following things on the password reset page:
   1. non-matching passwords,
   1. non-correct passwords,
   1. correct passwords.
1. After changing the password, you should be redirected to the login page with an appropriate `success` alert.
1. Use the copied link again and see that it's not valid anymore.

**Reviewers:**
- [ ] @giovannicimolin


requirements.txt
```diff
@@ -53,11 +53,12 @@ django-grappelli==2.12.3  # via -r requirements.in
 django-inlinecss==0.3.0   # via -r requirements.in
 django-libsass==0.7       # via -r requirements.in
 django-redis==4.10.0      # via -r requirements.in
+django-rest-passwordreset==1.1.0  # via -r requirements.in
```
giovannicimolin, line 56: @Agrendalath It looks like this repo is getting abandoned, with no commits for the past 3 months.
Do you think it's possible to replace it with one of the DRF recommended libraries from https://www.django-rest-framework.org/api-guide/authentication/#django-rest-auth?

Agrendalath: @giovannicimolin, the quick grep shows two candidates:
- [Tivix/django-rest-auth](https://github.com/Tivix/django-rest-auth) - this one is abandoned and we had to make [a fork of this](https://github.com/open-craft/django-rest-auth) to support SimpleJWT in Sprints.
-  [sunscrapers/djoser](https://github.com/sunscrapers/djoser), which [doesn't have](https://djoser.readthedocs.io/en/latest/base_endpoints.html#reset-password-confirmation) an endpoint to verify tokens.

`django-rest-passwordreset` actually supports Django==3.0.0 (cf. https://github.com/anexia-it/django-rest-passwordreset/issues/78), but, indeed, it doesn't seem to be maintained. However I haven't found a better option that would support our use-case in the expected timebox.

giovannicimolin: Let's use this one for now then.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <TextInputField
+            fieldName="email"
+            value={this.state.email}
+            onChange={this.onChange}
+            messages={messages}
+            type="email"
+          />
+
+          {this.props.error && (
+            <Alert variant="danger">{this.props.error}</Alert>
```
giovannicimolin, line 87: Nit: Can you check if it's possible to add the styles from the mockups to these alerts for the entire app (with a generic selector)? 

danger
success

api/auth.py
```diff
@@ -31,7 +31,7 @@
 )
 
 
-class TokenSerializer(serializers.Serializer):
+class JwtTokenSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
Agrendalath: This was conflicting in Swagger with [this serializer](https://github.com/anexia-it/django-rest-passwordreset/blob/master/django_rest_passwordreset/serializers.py#L21).

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+  public render() {
+    return (
+      <ContentPage title="Forgot password">
+        <Form
+          className="login" // FIXME: make SCSS more generic to properly reuse this style.
+          onKeyPress={(event: any) => {
+            this.onKeyPress(event);
+          }}
+        >
+          {this.props.succeeded && (
```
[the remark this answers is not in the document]
Agrendalath: Oh, I thought it was intentional. Sure, I'll change this.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+          <TextInputField
+            fieldName="email"
+            value={this.state.email}
+            onChange={this.onChange}
+            messages={messages}
+            type="email"
+          />
+
+          {this.props.error && (
```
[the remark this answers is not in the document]
Agrendalath: Good catch. I see that the `success` message was also affected. I added clearing this to both `mount` and `unmount`.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
Agrendalath: Done.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const { value } = e.target;
+
+    this.setState({
+      [field]: value
+    });
+  };
+
+  public render() {
+    return (
+      <ContentPage title="Forgot password">
```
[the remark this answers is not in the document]
Agrendalath: Done. Please see if it looks better.

api/auth.py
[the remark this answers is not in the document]
giovannicimolin: Got it.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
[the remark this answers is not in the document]
giovannicimolin: This probably went unnoticed while reviewing the mockups.

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
giovannicimolin: @Agrendalath They need some tweaking. 
I thought you had access to the styles from Zeplin, let me know if that's not the case.

Here are the styles:
**Danger**
* Font:
```
  font-family: Chivo;
  font-size: 14px; // You can adjust this (probably 12px is going to look closer to the mockups)
  color: #ffffff;
```
* BG:
```
background-color: #dd0000;
```

**Success**
* Font:
```
  font-family: Chivo;  // $chivo-font if you import the global styles
  font-size: 16px;  // Adjust as you see fit, probably having the same for both is better
  color: #ffffff;
```
* BG:
```
 background-color: #00a556;
```

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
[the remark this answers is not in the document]
giovannicimolin: Works perfectly now!

frontend/src/auth/components/PasswordForgottenPage/PasswordForgottenPage.tsx
Agrendalath: @giovannicimolin, you didn't have enough privileges to give me this access ([cf. this comment](https://tasks.opencraft.com/browse/BB-2219?focusedCommentId=143230&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-143230)). Though agreed, it would be nice to have.

Agrendalath: Fixed. IMO 14px looked well-balanced for both.

PR #567, opened 2020-04-18
[BB-2218] Build the LMS theme preview component

Agrendalath:
This adds the preview component to `Preview & colors` sectionof the new Ocim frontend.

**Notes**:
- the tests are currently failing due to extra dependencies needed by `PreviewComponent` in `ThemePreviewAndColors`; I'll fix them in the meantime,
- there is no "course preview" component, so I used a dummy image fo this and added `TODO` there,
- I used standard instance as a reference for the preview component, because there were some differences between mockups and actual look of the Open edX instance (e.g. navbar was missing the left part, which was already present in the `Navigation` customization, `Hero` was missing padding at the sides and top, footer didn't have the top border, etc.); the styling is not "pixel-perfect", but I tried making it as close to the real one as possible.

**Testing instructions:**
1. Checkout this branch.
1. Create a new account in console or use an exiting one.
1. Go to `Theme -> Preview & colors`.
1. Check that the theme preview is displaying correctly.

**Reviewers:**
- [x] @giovannicimolin


frontend/src/console/components/PreviewComponent/styles.scss
```diff
@@ -0,0 +1,110 @@
+.theme-preview {
+  border: 2px solid black;
```
giovannicimolin, line 2: Can you change this to `1px`? The tick border doesn't look good in HD and mobile screens

giovannicimolin:
@Agrendalath Found a few issues:
* Setting main and link colors has no effect in the preview.
* The course card is not being updated (since there's no component for it, but it is affected by main and link colors).
* The 2px border looks too thick in HD and mobile screens.

frontend/src/console/components/CoursesListingItem/CoursesListingItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+  props: CoursesListingItemProps
+) => {
+  const { themeData } = props;
+  const textStyle = { color: themeData.mainColor };
+  // TODO: Make text customizable.
+
+  return (
+    <Row
+      className="courses-listing-item"
+      style={{ borderBottomColor: themeData.linkColor }}
```
Agrendalath, line 25: @giovannicimolin, should this be altered as well? I see alternative color for the bottom border of the `courses listing item` on the mockups, but I'm not sure if it's related to `linkColor` variable.

frontend/src/console/components/PreviewComponent/styles.scss
Agrendalath: Done!

frontend/src/console/components/CoursesListingItem/CoursesListingItem.tsx
giovannicimolin: @Agrendalath The thing is that we're not 100% sure how the generic variables (`main` and `link` colors) are applied to the instances, so we should schedule a follow up and fix any issues when assigning color.

For this ticket, I'm happy to merge as the preview component was implemented.

PR #568, opened 2020-04-21
BB-2327: Fix color picker and Hero page

giovannicimolin:
This PR fixes a few bugs in the color picker component and hero customization pages.

**Setup instructions:**
1. Checkout this branch, install API client.
2. Run Ocim & frontend.

**Testing instructions:**
42c376bc3e37a88494370be2bfd110fba8124e48, f26960da13d0c7d662fddf90d6662e5c1fabcd3e:
1. Check the pages that have the color picker and make sure they are being correctly displayed on HD, FullHD and 4k. It looks a little broken in mobile, but it's usable.
2. Check that the color preview has the correct color set after picking a color.
3. Try opening multiple color pickers at the same time, try to reproduce the issue from the previous implementation.
4. Check that the correct values are being saved into the backend.

From d8ceebf2eb92c8e92624810c1e03a1468406c05a onwards:
1. Go to the hero customization page.
2. Check that the styling is improved and that the fields are aligned with the container.
3. Check that it's possible to set empty values for the Hero text fields.
4. Check that the reset button works as expected :)

**Reviewer:**
- [x] @Agrendalath


lgp171188:
> Check that it's possible to set empty values for the Hero text fields.

@giovannicimolin, CC @Agrendalath, I don't think it is a good idea to allow the value of the hero text to be set to an empty string in the backend because that will be the same as allowing the user to hide the title/subtitle and I don't think we want to let the user do that.

Using a known magic string [here](https://github.com/open-craft/opencraft/pull/568/files#diff-8d2c787a3561d97515102272dd26a521R146) to let the backend set the default value might be okayish, but the backend shouldn't allow setting `''` for either field. We know that with the current implementation the frontend will automatically send the magic string for the empty string situation, but the backend shouldn't rely on the frontend to send it valid values and instead should validate all the values from all the clients. For example, it is possible to make an API request directly in the context of a user session that sets the value to `''` because the backend allows it due to the changes in this PR.

Agrendalath:
> because that will be the same as allowing the user to hide the title/subtitle and I don't think we want to let the user do that.

@lgp171188, is there a specific reason to disallow hiding the title/subtitle? How does it differ from just setting a whitespace character there?

lgp171188:
> @lgp171188, is there a specific reason to disallow hiding the title/subtitle? How does it differ from just setting a whitespace character there?

@Agrendalath, I think without the changes in this PR, the backend trims the string and then checks for an empty string before setting the default value. Afaik, the scope of the functionality to be exposed to the users was primarily around colors and branding and I'd say hiding the title/subtitle doesn't come within that. While hiding the subtitle might be acceptable as things look okay with just the title. But hiding both causes the LMS home page to look ugly and nothing the like the vanilla LMS theme with just color and branding customizations? So I think the backend should validate and disallow invalid values or reset them to the defaults.

instance/schemas/static_content_overrides.py
```diff
@@ -79,8 +82,17 @@ def fill_default_hero_text(text=None):
         )
         match = homepage_overlay_html_regex.match(text)
         if match:
-            title = match.group('title').strip() or DEFAULT_HERO_TITLE_TEXT
-            subtitle = match.group('subtitle').strip() or DEFAULT_HERO_SUBTITLE_TEXT
+            # Stores whatever value matched and overrides with default
+            # value when the frontend passes `reset_default`.
+            title = match.group('title')
+            if title == 'reset_default':
```
Agrendalath, line 88: That's a bit hacky. Let's hope that no user wouldn't want to have `reset_default` in the title nor subtitle :P
Maybe we could leave it as it was before? Or should we actually allow users to disable the title/subtitle?

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -47,39 +46,59 @@ export class HeroComponent extends React.PureComponent<Props, State> {
...
-  public componentDidMount() {
-    this.checkAndUpdateState();
-  }
+  public componentDidUpdate(prevProps: Props) {
+    const oldState = this.getHeroContents(prevProps);
+    const newState = this.getHeroContents(this.props);
 
-  public componentDidUpdate() {
-    this.checkAndUpdateState();
+    // Only update internal state if something changed after props updated.
```
Agrendalath, line 56: There is an broken edge case here - if you set title/subtitle to `reset_default` twice, it will stay like this.

Agrendalath:
@lgp171188, ah, right, we were doing `strip` here.

instance/schemas/static_content_overrides.py
giovannicimolin: Yup, I'll revert this!

I'll make the reset button send back and empty text, that way the default value is filled.

frontend/src/console/components/Hero/Hero.tsx
giovannicimolin: I'll revert the `reset_default` thing back to how it worked before, and check if there's any other edge case.

Agrendalath: @giovannicimolin, I already tested this and there is a similar one - when you remove everything from this field twice.

frontend/src/ui/components/ColorInputField/ColorInputField.tsx
```diff
@@ -58,28 +55,66 @@ export const ColorInputField: React.SFC<ColorInputFieldProps> = (
...
         <FormControl
           className="input-field-color"
           name={props.fieldName}
-          value={selectedColor}
+          value={fieldValue()}
           disabled={props.loading}
-          onFocus={showColorPicker}
-          style={{ color: selectedColor }}
+          onClick={toggleColorPicker}
+          onKeyPress={toggleColorPicker}
```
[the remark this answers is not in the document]
giovannicimolin: If you navigating using the keyboard (using tabs, for example) the color picker will open up.

frontend/src/console/components/Hero/Hero.tsx
giovannicimolin: Reverted back to @lgp171188 initial implementation. 
I've spend quite some time trying to fix this state management issue already.

frontend/src/ui/components/ColorInputField/ColorInputField.tsx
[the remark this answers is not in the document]
Agrendalath: It does, but you cannot apply the changed style with only the keyboard - when you close the picker with a key, the value is not saved. Also you can open multiple color pickers by tabbing and pressing keys. That's just a nit, but wanted to note it here.

frontend/src/console/components/Hero/Hero.tsx
Agrendalath: I feel that your approach was actually more user-friendly :P
Though it's not in the scope of this ticket.

frontend/src/ui/components/ColorInputField/ColorInputField.tsx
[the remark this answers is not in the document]
giovannicimolin: Ah, I'll remove this line then, since it introduces this buggy behavior.

frontend/src/ui/components/ImageUploadField/ImageUploadField.tsx
```diff
@@ -50,9 +51,27 @@ export const ImageUploadField: React.FC<ImageUploadFieldProps> = (
...
+        </p>
+      )}
+
+      {props.reset !== undefined && (
+        <p>
+          <button
+            className="reset-image"
+            type="button"
+            onClick={() => {
+              if (props.reset !== undefined) {
```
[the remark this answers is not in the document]
giovannicimolin: Typescript was complaining, and I'm not sure why.
I've swapped this for `props.reset!()` + a comment to justify the use of `!`.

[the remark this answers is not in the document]
giovannicimolin: I've done the same for the other occurrence too.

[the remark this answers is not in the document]
Agrendalath: Sounds good, thank you!

PR #570, opened 2020-04-30
[BB-2326] OCIM Bugs

rusrushal13:
Issue: [BB-2326](https://tasks.opencraft.com/browse/BB-2326)

Testing Instructions:
* First Bug:
Try to login with opened Redux/React Dev tools(to have better clarity on states) and check for the console state being refreshed or not

* Second Bug:
Added a Go To Console button in congratulation Page, just create a new instance and click on Go To Console button to check if it's working or not.
Screenshot: go_to_console

* Third Bug:
Added a Django atomic operation for Theme and static updates.

Reviewers:
- [x] @giovannicimolin


frontend/src/registration/components/CongratulationsPage/CongratulationsPage.tsx
```diff
@@ -45,6 +47,11 @@ export class CongratulationsPage extends React.PureComponent<Props, State> {
           <p>
             <WrappedMessage messages={messages} id="congratsMessage2" />
           </p>
+          <NavLink exact to={ROUTES.Console.HOME}>
+            <Button>
```
giovannicimolin, line 51: Can you use the larger button, and center it in the view? 

```suggestion
            <Button size="lg">
```

frontend/src/console/reducers.ts
```diff
@@ -10,6 +10,18 @@ export function consoleReducer(
   let activeInstanceId: number;
 
   switch (action.type) {
+    case Actions.Types.USER_REFRESH_DATA:
+      return update(state, {
```
giovannicimolin, line 14: You should use the clean state defined on `models.ts` instead of manually cleaning up all variables.
See `initialConsoleState` on `console/models.js`:
https://github.com/open-craft/opencraft/blob/b58f24ce71fca95696dd2f9a083077a6ff365afb/frontend/src/console/models.ts#L41

You don't need to worry about merging things here, so just return the entire clean state.
```suggestion
      return initialConsoleState;
```

frontend/src/console/actions.ts
```diff
@@ -11,6 +11,7 @@ export enum Types {
   // Support action to update root state and clean error messages when users change fields
   CLEAR_ERROR_MESSAGE = 'CLEAR_ERROR_MESSAGE',
   // To handle multiple user instances
+  USER_REFRESH_DATA = 'USER_REFRESH_DATA',
```
giovannicimolin, line 14: This does not match what the type is used for.
Please rename it to `CLEAR_CONSOLE_DATA` or something along those lines.

The same goes for the action (`UserRefreshData` -> `ClearConsoleData`).

frontend/src/registration/components/CongratulationsPage/CongratulationsPage.tsx
giovannicimolin: To center the component, you can wrap a div around it with `className="text-center"`.

frontend/src/console/actions.ts
rusrushal13: Thanks for the suggestions, I have addressed it and refactored the code :)

frontend/src/console/reducers.ts
rusrushal13: Thanks for the suggestion, I have refactored the code :)

frontend/src/registration/components/CongratulationsPage/CongratulationsPage.tsx
rusrushal13: Thanks for it, I have added a container which centers the button and used large class for enlarging the button

registration/api/v2/views.py
```diff
@@ -320,17 +323,19 @@ def image(self, request, pk: str):
...
-
-            if 'logo' in request.data:
-                application.logo = request.data['logo']
-                application.save()
-
-            if 'hero_cover_image' in request.data:
-                application.hero_cover_image = request.data['hero_cover_image']
-                application.save()
+            # static updates protected by atomic operations
+            with transaction.atomic():
```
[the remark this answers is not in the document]
rusrushal13: I have fixed the code, and using `transaction.atomic` on the specific methods for theme and static content updates

registration/api/v2/views.py
```diff
@@ -206,20 +207,22 @@ def perform_create(self, serializer):
...
-                static_content_overrides['homepage_overlay_html']
-            ).safe_substitute(
-                instance_name=instance.instance_name
-            )
-            instance.draft_static_content_overrides = static_content_overrides
-        instance.save()
-        # Send verification emails
-        verify_user_emails(instance.user, instance.public_contact_email)
+        # theme update protected by atomic operation
+        with transaction.atomic():
```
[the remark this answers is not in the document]
rusrushal13: I have updated the code. :)

rusrushal13:
> @rusrushal13 I've left a few comments in the PR.
> I can do another pass before the end of the sprint if you address all comments.

@giovannicimolin I have addressed your comments please check once and let me know if there is anything which is missing :smiley:

frontend/src/registration/components/CongratulationsPage/CongratulationsPage.tsx
```diff
@@ -45,6 +47,13 @@ export class CongratulationsPage extends React.PureComponent<Props, State> {
           <p>
             <WrappedMessage messages={messages} id="congratsMessage2" />
           </p>
+          <NavLink exact to={ROUTES.Console.HOME}>
```
giovannicimolin, line 50: The `NavLink` component here is wrapping around the entire div, so clicking anywhere on the div will trigger the link.
This isn't a good accessibility practice, can you move the `div` outside the `Navlink`?

frontend/src/console/actions.ts
```diff
@@ -164,8 +165,14 @@ export interface CancelDeploymentFailure extends Action {
...
 }
 
+export interface UserRefreshData extends Action {
+  readonly type: Types.CLEAR_CONSOLE_DATA;
+  readonly data: Array<InstanceSettingsModel>;
+}
+
 export type ActionTypes =
   | ClearFeedbackMessage
+  | UserRefreshData
```
giovannicimolin, line 175: Here this is still `UserRefreshData` instead of something related to `ClearConsoleData`. 
`UserRefreshData` isn't representative of what this function does.

Make sure this doesn't conflict with the `clearConsoleData` function below.

rusrushal13: Refactored the code and `UserRefreshData` changed to `ClearConsoleData` for better understanding(missed in the last commit, Thanks for this)

frontend/src/registration/components/CongratulationsPage/CongratulationsPage.tsx
rusrushal13: Oh yeah :open_mouth: 
I have missed that completely, thanks a lot for catching this :smiley:

rusrushal13:
> Left a few small comments.

@giovannicimolin  Thanks a lot for catching those issues, I have addressed them, I guess we can merge the PR after the CI checks :)

registration/api/v2/views.py
```diff
@@ -238,6 +240,7 @@ def partial_update(self, request, *args, **kwargs):
         responses={**VALIDATION_RESPONSE, 200: ThemeSchemaSerializer, },
     )
     @action(detail=True, methods=["patch"])
+    @transaction.atomic
```
giovannicimolin, line 243: @rusrushal13 
I've just tested this and this change alone won't do proper row-locking of the object retrieved by `self.get_object`.
To achieve that, you'll need to use the `.select_for_update` queryset method ([django docs](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update)).
```
application = BetaTestApplication.objects.select_for_update().get(id=pk)
with transaction.atomic:
    # Do atomic operation here
```

rusrushal13: Added a fix for this using select_for_update and transaction atomic both

registration/api/v2/views.py
```diff
@@ -252,58 +254,59 @@ def theme_config(self, request, pk=None):
...
-                data={
-                    "non_field_errors": "Schema validation failed."
-                }
-            )
-
-        application.draft_theme_config = safe_merged_theme
-        application.save()
-
-        return Response(status=status.HTTP_200_OK, data=application.draft_theme_config)
+        application = BetaTestApplication.objects.select_for_update().get(id=pk)
```
giovannicimolin, line 257: @rusrushal13 Did you test this change?

I'm getting this error when using this endpoint:
```
TransactionManagementError at /api/v2/instances/openedx_config/2/theme_config/
select_for_update cannot be used outside of a transaction.
```
This doesn't happen in the tests because they use a transaction wrapper around the test.
Can you create a fix for this and test it before sending to review again?

rusrushal13: I tested it, and it's working now :+1:

giovannicimolin:
Overriding merge restrictions since the changes don't affect the integration tests and there's a deadline closing up.

PR #571, opened 2020-05-01
Document how to run a devstack in LXC

arbrandes:
As an alternative to Vagrant, document how a developer can set up a devstack using an LXD container on Ubuntu.


documentation/installation.md
```diff
@@ -79,11 +79,129 @@ the development server at http://localhost:5000/ using your web browser. Log in
...
+
+    SRC=~/src/ocim
+    CONTAINER=ocim
+
+Check the necessary repositories out:
+
+    mkdir -p $SRC
+    cd $SRC
+    git clone https://github.com/open-craft/opencraft ocim
+    git clone https://github.com/open-craft/ansible-playbooks
```
giovannicimolin, line 109: @arbrandes We need to clone the playbooks inside the Ocim folder anyway, so why not do it in this step?
See the vagrant instructions here: https://github.com/open-craft/opencraft/blame/master/documentation/installation.md#L35

arbrandes:
> In any case I think that having instructions on deploying this on Docker is better than having them for LXC containers, mainly because Docker is more widely used.

I'd be fine with having Docker instead of LXC, but there should be at least one officially supported container deployment option.  Can we work on merging Docker support, then?

In any case, even if it's not officially supported, it would be a waste to throw away this documentation.  Where would be a good place for it to live?  Is there a wiki, somewhere?

documentation/installation.md
arbrandes: For what its worth, that's not true for LXC, where you don't run the playbooks from bare metal.  You run them from within the container itself.

bradenmacdonald:
I'd like to see us ditch vagrant and use docker for development. I'm not familiar with LXC/LXD so can't speak to the merits of that, but I do think it makes sense to focus our time on a common deployment method, and my preference would be docker. It's my hope that we can ditch Vagrant everywhere and just use docker for all the types of devstacks we have, keeping our tooling simple.

I'm fine with putting these docs in an "lxc-readme" as long as it has some messaging that it's just documenting an alternative in case it's ever handy but we want people to generally use the official method, whether that's docker or vagrant.

xitij2000:
I think the value of Vagrant is that it's closer to how production is deployed. i.e. on xenial running in a VM. The advantage of Docker is it's faster and simpler in many ways. I'm not sure this is either. 

The Docker stuff @giovannicimolin is mentioning is here: https://github.com/open-craft/opencraft/compare/kshitij/docker_support

 I remember it working long ago. I haven't used it in a while.

arbrandes:
@bradenmacdonald:
> I'd like to see us ditch vagrant and use docker for development.

+1.  I can convincingly make the argument that this will be necessary for when Ocim is migrated over to a container deployment method _for production_.  As a matter of fact, it is probably the best idea for starting in that direction: make a Docker devstack work, then use the same containers in prod. :shrug:

This also addresses @xitij2000's point about Vagrant, which is a good one.  But once a VM is no longer the prod target, we'll need a better devstack.

As for LXC/LXD, the docs suggested here are for it being used as a system container (as opposed to an application container, which is what Docker does), which is basically a Vagrant drop-in replacement without all the unnecessary virtualization overhead.  It is thus less portable than a VM, which is its biggest disadvantage, but if you're already on Linux (and particularly if you're on a CoW filesystem), it's faster.

Which is why I'll then take Braden's suggestion and move this to README.lxc and reword it.   I'll then run it by @giovannicimolin again. :)

arbrandes:
@giovannicimolin, how's this? :)

PR #572, opened 2020-05-03
[BB-2299] Fix hero and instance setting text field component

rusrushal13:
Issue: [BB-2299](https://tasks.opencraft.com/browse/BB-2299)

OCIM UX Fixes
* Instance name and public contact email appearing on Instance settings component
* Hero component empty string text field fixed 

Testing Instructions: 

- For instance, setting component - just visit that page and try to refresh, the instance name and contact email should appear; also try to change the instance name and public contact email and then refresh.
- For hero component fix - just visit the hero component page and empty all title and subtitle to see that it works. 

Reviewers:
- [ ] @lgp171188


frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -30,6 +30,7 @@ interface State {
 
   title: string;
   subtitle: string;
+  emptyFields: string;
```
giovannicimolin, line 33: Use `boolean` instead of `string`.

giovannicimolin: @rusrushal13 ^

rusrushal13: @giovannicimolin thanks, I have updated it :+1:

rusrushal13:
> @rusrushal13, I tried testing the changes in this PR but I noticed the following issue.
> 
> The title/subtitle text box can now be emptied by the user without any issues. But the `onBlur` event on those fields should trigger an API call to update the text on the backend and re-render the returned value. That is not happening. I can see that the `onBlur` handler for an empty string and makes the API call only when it is not empty. We don't want that behaviour.
> 
> I tried removing that check and the API call went through and the updated result was seen in the preview component but the affected text box is still empty. Can you check and fix this?

@lgp171188 Thanks for this, I have checked it and now it's been fixed :+1:

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -66,6 +69,24 @@ export class InstanceSettingsComponent extends React.PureComponent<
...
+      this.setState({
+        instanceName:
+          prevProps.activeInstance.data.instanceName ||
+          this.props.activeInstance!.data!.instanceName,
+        publicContactEmail:
+          prevProps.activeInstance.data.publicContactEmail ||
+          this.props.activeInstance!.data!.publicContactEmail,
+        emptyFields: '0'
+      });
+    }
```
Agrendalath, line 89: Unfortunately it seems to be refreshing the content only after receiving the response from `api/v2/instances/openedx_deployment/1/`. However the required data is already available, as it's returned by `api/v2/instances/openedx_config/`.
Please see the recording below:

settings

The gif below represents what happens when you disable Ocim after the initial page load and before receiving the response to `api/v2/instances/openedx_config/`. The `InstanceSettingsComponent` already has the instance name and contact email in place:
settings2

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -160,6 +181,11 @@ export class HeroComponent extends React.PureComponent<Props, State> {
         homepageOverlayHtml
       );
     }
+    if (this.state.title === '' || this.state.subtitle === '') {
```
Agrendalath, line 184: We should `trim()` these values, because if we input whitespace here it won't be sent to backend and the hero preview will not match this field's value.

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -82,35 +81,51 @@ export class HeroComponent extends React.PureComponent<Props, State> {
...
-  public componentDidUpdate() {
+  public componentDidUpdate(prevProps: Props) {
     this.checkAndUpdateState();
+    if (
+      this.activeInstanceDataExists() &&
+      this.staticContentOverridesExists() &&
+      prevProps.activeInstance.data &&
+      this.props.activeInstance.data &&
+      prevProps.activeInstance.data.draftStaticContentOverrides !==
+        this.props.activeInstance.data.draftStaticContentOverrides
```
Agrendalath, line 101: There are redundant checks here. `this.staticContentOverridesExists()` already calls `activeInstanceDataExists()`, which checks `this.props.activeInstance.data`.
Could we simplify this by doing something like this?
```ts
public componentDidUpdate(prevProps: Props) {
  this.checkAndUpdateState();
  if (
    this.staticContentOverridesExists() &&
    this.staticContentOverridesExists(prevProps) &&
    prevProps.activeInstance.data!.draftStaticContentOverrides !==
      this.props.activeInstance.data!.draftStaticContentOverrides
  ) {
    this.checkNewAndUpdateState();
  }
}

...

private activeInstanceDataExists = (props = this.props) => {
  return props.activeInstance && props.activeInstance.data;
};

private themeConfigExists = (props = this.props) => {
  return (
    this.activeInstanceDataExists(props) &&
    props.activeInstance.data!.draftThemeConfig
  );
};

private staticContentOverridesExists = (props = this.props) => {
  return (
    this.activeInstanceDataExists(props) &&
    props.activeInstance.data!.draftStaticContentOverrides
  );
};
```

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -66,6 +69,24 @@ export class InstanceSettingsComponent extends React.PureComponent<
         publicContactEmail: this.props.activeInstance!.data!.publicContactEmail
       });
     }
+    // if current state is empty and pre-filling
```
Agrendalath, line 72: Maybe we could reword this to something along `Set an initial state or restore empty values`?

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -26,10 +26,9 @@ import messages from './displayMessages';
 import './styles.scss';
 
 interface State {
-  [key: string]: string;
-
   title: string;
   subtitle: string;
+  renderBool: boolean;
```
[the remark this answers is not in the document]
rusrushal13: Thanks, @Agrendalath , I have gone through comments, you can check now :+1:

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -82,35 +81,51 @@ export class HeroComponent extends React.PureComponent<Props, State> {
...
+  };
+
   private checkAndUpdateState = () => {
     if (
       this.homePageOverlayHtmlExists() &&
-      // FIXME: The below condition causes issues when the user tries to empty either text box.
-      //  Investigat and come up with a way to fix this.
-      (this.state.title === '' || this.state.subtitle === '')
+      (this.state.title === '' || this.state.subtitle === '') &&
+      this.state.renderBool === true
```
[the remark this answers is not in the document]
rusrushal13: Thanks, I have updated it :+1:

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -16,6 +16,7 @@ interface State {
   [key: string]: string;
   instanceName: string;
   publicContactEmail: string;
+  emptyFields: string;
```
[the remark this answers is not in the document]
rusrushal13: I have changed this to boolean now, this is same as Hero Component for checking the empty state of instance name and public contact email and re-render accordingly

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
rusrushal13: @Agrendalath Thanks for this thorough explanation, I have updated it now, whenever the component receiving the props I am updating it right away now :+1:

frontend/src/console/components/Hero/Hero.tsx
rusrushal13: Yes, this makes sense, I have updated it, now every method taking props as an argument and performing the instructions based on the props received

Agrendalath: @rusrushal13, it would be more convenient to have the default value for `props` for these checks. That's why I put `(props = this.props)` above. This way we won't need to pass `this.props` all the time, which will make it easier to read and minimize the changes of the unrelated lines in this PR.

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -66,6 +69,24 @@ export class InstanceSettingsComponent extends React.PureComponent<
...
+      this.props.activeInstance.data &&
+      this.state.emptyFields === '1'
+    ) {
+      this.setState({
+        instanceName:
+          prevProps.activeInstance.data.instanceName ||
+          this.props.activeInstance!.data!.instanceName,
+        publicContactEmail:
+          prevProps.activeInstance.data.publicContactEmail ||
+          this.props.activeInstance!.data!.publicContactEmail,
```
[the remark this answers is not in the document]
rusrushal13: I updated it, since I am checking them above, not adding the non-null operator now :+1:

frontend/src/console/components/Hero/Hero.tsx
rusrushal13: Yeah, added the trim method around the strings to remove the whitespaces, will remember this too from next time :+1:

frontend/src/console/components/Hero/Hero.tsx
rusrushal13: Thanks, @Agrendalath, having default arguments surely helps, I have updated it in my latest commit :smiley:

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -72,17 +94,38 @@ export class InstanceSettingsComponent extends React.PureComponent<
...
+    // Only make update request if public contact email changed
+    if (
+      instance.data &&
+      this.state.publicContactEmail !== instance.data.publicContactEmail
+    ) {
+      this.props.updateFieldValue(
+        instance.data.id,
+        'publicContactEmail',
+        value
+      );
```
Agrendalath, line 128: Did you rewrite this because it was saying that the `string` cannot be used for indexing `State`?

If so, this should work:
```ts
if (instance.data && this.state[fieldName as keyof State] !== instance.data[fieldName]) {
  this.props.updateFieldValue(instance.data.id, fieldName, value);
}
```

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -146,9 +162,15 @@ export class HeroComponent extends React.PureComponent<Props, State> {
...
-    });
+    if (field === 'title') {
+      this.setState({
+        title: value
+      });
+    } else if (field === 'subtitle') {
+      this.setState({
+        subtitle: value
+      });
+    }
```
Agrendalath, line 173: We don't need to rewrite it. It would be sufficient to do something like this:
```ts
this.setState({
  [field]: value
} as Pick<State, 'title' | 'subtitle'>);
```

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -82,62 +82,78 @@ export class HeroComponent extends React.PureComponent<Props, State> {
...
   }
 
-  public componentDidUpdate() {
+  public componentDidUpdate(prevProps: Props) {
     this.checkAndUpdateState();
+    if (
+      this.staticContentOverridesExists() &&
+      this.staticContentOverridesExists(prevProps) &&
+      prevProps.activeInstance.data &&
+      this.props.activeInstance.data &&
```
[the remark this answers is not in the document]
rusrushal13: I tried doing it without the checks, and it doesn't compiles, so that's why I left them there :)

frontend/src/console/components/Hero/Hero.tsx
```diff
@@ -244,7 +271,7 @@ export class HeroComponent extends React.PureComponent<Props, State> {
                 <Row>
                   <Col>
                     {this.themeConfigExists() &&
-                      this.staticContentOverridesExists() && (
+                      this.staticContentOverridesExists(this.props) && (
```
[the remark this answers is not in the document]
rusrushal13: Yeah, there is no need, I removed it, thanks for pointing out :smile:

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
```diff
@@ -33,22 +34,43 @@ export class InstanceSettingsComponent extends React.PureComponent<
...
+  private updateInitialState = (props: Props) => {
+    if (
+      (this.state.instanceName.trim() === '' ||
+        this.state.publicContactEmail.trim() === '') &&
+      props.activeInstance.data &&
+      this.state.renderBool
+    ) {
+      this.setState({
+        instanceName: props.activeInstance.data!.instanceName,
+        publicContactEmail: props.activeInstance.data!.publicContactEmail,
```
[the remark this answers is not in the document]
rusrushal13: I missed removing them, thanks :)

frontend/src/console/components/InstanceSettings/InstanceSettings.tsx
rusrushal13: I was confused about how to fix it, and I don't know that I can use it this way, thanks
I changed it :)

frontend/src/console/components/Hero/Hero.tsx
rusrushal13: I learned this too, thanks @Agrendalath :smile:

frontend/src/console/components/Hero/Hero.tsx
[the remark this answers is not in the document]
Agrendalath: That's because TypeScript doesn't resolve this nested check (despite it actually exists) and that's why we should use the `!` here, as indicated in [this comment](https://github.com/open-craft/opencraft/pull/572#discussion_r421738649). Related part of that snippet:
```ts
if (
    this.staticContentOverridesExists() &&
    this.staticContentOverridesExists(prevProps) &&
    prevProps.activeInstance.data!.draftStaticContentOverrides !==
      this.props.activeInstance.data!.draftStaticContentOverrides
  ) {
    this.checkNewAndUpdateState();
  }
```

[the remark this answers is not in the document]
rusrushal13: @Agrendalath  thanks for the comment, and the snippet!
I fixed it :smile: 
Now it works perfect :+1:

PR #573, opened 2020-05-04
[BB-2291] New mechanism for tracking deployments

xitij2000:


registration/api/v2/views.py
```diff
@@ -561,10 +557,15 @@ def destroy(self, request, *args, **kwargs):
                     'details': "Can't cancel deployment while server is being prepared."
                 }
             )
-        # TODO: Prevent appservers provisioned by staff/redeployments to be cancelled
+        deployment = instance.get_latest_deployment()
```
giovannicimolin, line 560: Thanks for fixing this TODO! :+1:

registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -38,14 +38,14 @@ def fix_invalid_domain_names(apps, schema_editor):
...
 
-    BetaTestApplication.objects.update(subdomain=clean_domain_name('subdomain'))
-    OpenEdXInstance.objects.update(
-        internal_lms_domain=clean_domain_name('internal_lms_domain'),
-        internal_lms_preview_domain=clean_domain_name('internal_lms_preview_domain'),
-        internal_studio_domain=clean_domain_name('internal_studio_domain'),
-        internal_discovery_domain=clean_domain_name('internal_discovery_domain'),
-        internal_ecommerce_domain=clean_domain_name('internal_ecommerce_domain'),
-    )
+    # BetaTestApplication.objects.update(subdomain=clean_domain_name('subdomain'))
```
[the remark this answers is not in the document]
xitij2000: This was erroring out, and I just wanted to get past it and see what the issue is. I will undo this. 
That said I think we can clean up out migrations a lot. We are the only users of Ocim so we can gain a lot by squashing migrations and cleaning up old python migrations. It will make some things faster, like tests. We will not need to run this operations again on production or otherwise.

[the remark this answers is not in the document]
giovannicimolin: Ah no problem! I'll consider this change as a WIP.

>That said I think we can clean up out migrations a lot. We are the only users of Ocim so we can gain a lot by squashing migrations and cleaning up old python migrations. It will make some things faster, like tests. We will not need to run this operations again on production or otherwise.

I think we can use the `squashmigrations` feature from Django. 
It'll probably make the tests a little bit faster too. :)

frontend/package.json
```diff
@@ -76,118 +29,49 @@
...
-      "jest-watch-typeahead/testname"
-    ]
-  },
-  "babel": {
-    "presets": [
-      "react-app"
-    ]
+    "typescript": "3.8.3"
   }
 }
```
xitij2000, line 77: I reorganised this a bit since some dependencies were duplicated between dependencies an devDependencies, causing confusion.

instance/models/deployment.py
```diff
@@ -0,0 +1,73 @@
...
+    pr = 'Deployment for GitHub PR'
+    periodic = 'Deployment for periodic build'
+    registration = 'Deployment created during registration'
+    unknown = 'Deployment created unknown or legacy reasons'
+
+
+# Models ######################################################################
+
+
+class Deployment(TimeStampedModel):
```
giovannicimolin, line 55: Can you add this to the Django Admin?

frontend/scripts/update-api-client.sh
```diff
@@ -5,7 +5,7 @@ rm -f packages/api_client/apis/*.ts
 rm -f packages/api_client/models/*.ts
 
 docker run --rm --network="host" -v ${PWD}:/local \
-    openapitools/openapi-generator-cli generate \
+    openapitools/openapi-generator-cli:v4.3.1 generate \
```
xitij2000, line 8: I've added the version tag here so that there won't be unnecessary format changes, and we can investigate new versions if we need to.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -53,7 +59,7 @@ export const RedeploymentToolbar: React.FC<Props> = (props: Props) => {
           variant="primary"
           size="lg"
           onClick={() => {
-            props.performDeployment();
+            performDeployment();
```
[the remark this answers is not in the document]
xitij2000: Both the old and new codes are equivalent. 
A funciton like ``function fun({a,b,c}) {}`` basically translates to ``function fun(params) {let a=params.a, b= params.b, c=params.c}``.
This way you avoid typing `params.` again and again.

frontend/scripts/update-api-client.sh
giovannicimolin: Nice fix!

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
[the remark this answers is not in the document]
giovannicimolin: Nice to know! Thanks for improving this.

frontend/src/auth/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -2,17 +2,15 @@ import * as React from 'react';
...
   const [isAuthenticated, setIsAuthenticated] = React.useState(false);
 
   React.useEffect(() => {
-    checkAuthAndRefreshToken().then((authenticated: boolean) => {
-      setIsAuthenticated(authenticated);
-      setIsLoading(false);
-    });
+    const accessToken = localStorage.getItem('token_access');
+    setIsAuthenticated(!!accessToken);
+    setIsLoading(false);
```
xitij2000, line 20: The private route was refreshing the token too often, each time you'd access a private route. 
With this change we only check if an access token is set, and if it is, we consider the user logged in.

frontend/src/global/api.ts
```diff
@@ -16,7 +17,10 @@ const config = new Configuration({
         if (context.response.status === 403) {
           // Failed requests return 403 and mean that the access token is
           // expired, so we trigger a page refresh.
-          window.location.reload(false);
+          const authenticated = await checkAuthAndRefreshToken();
+          if (!authenticated) {
+            window.location.reload(false);
+          }
```
xitij2000, line 23: With this change if you ever get a 403 error, the app will assume it's because the token has expired, and just use the refresh token to get a new one. It should work seamlessly, without refresh. 
You can test this by logging in, and then editing your access token to an invalid value, with the next request, your login token will be refreshed.

frontend/src/ui/components/Header/Header.tsx
```diff
@@ -1,74 +1,51 @@
...
       <Navbar bg="transparent" expand="md" variant="dark">
         <Navbar.Brand className="logo-container mx-auto">
-          <svg className="site-logo">
-            <use xlinkHref={`${logo}#opencraft_logo`} />
-          </svg>
+          <Nav.Link onClick={navigateTo(ROUTES.Console.HOME)}>
+            <svg className="site-logo">
+              <use xlinkHref={`${logo}#opencraft_logo`}/>
+            </svg>
+          </Nav.Link>
```
xitij2000, line 23: The logo is now linked to the home page.

frontend/src/ui/components/Header/Header.tsx
```diff
@@ -1,74 +1,51 @@
...
+              <Nav.Link>Customize</Nav.Link>
+              <Nav.Link onClick={() => window.open(CONTACT_US_LINK, '_blank')}>
+                Support Request
+              </Nav.Link>
+              <Nav.Link disabled>Status & Notifications</Nav.Link>
+              <Nav.Link onClick={navigateTo(ROUTES.Auth.LOGOUT)}>
+                Log out
+              </Nav.Link>
+            </Route>
+          </Nav> </Navbar.Collapse>
```
xitij2000: Cleaned up this page to remove duplication, and used the router to show things instead of custom code.

frontend/src/auth/components/PrivateRoute/PrivateRoute.tsx
```diff
@@ -19,11 +24,11 @@ export const PrivateRoute: React.FC<RouteProps> = (props: RouteProps) => {
...
     const renderComponent = () => (
-      <Redirect to={{ pathname: ROUTES.Auth.LOGOUT }} />
+      <Redirect to={{ pathname: fallbackRedirect }} />
     );
     return (
       <Route path={props.path} component={renderComponent} render={undefined} />
     );
   }
-  return <Route path={props.path} component={props.component} />;
+  return <Route {...props} />;
```
xitij2000, line 35: With this change you can now provide a fallback path for PrivateRoutes that is different from the regular logout route.

frontend/src/routes/main.tsx
```diff
@@ -6,15 +6,18 @@ import {
...
 import { RegistrationContainer } from 'registration/components';
 import { ConsoleContainer } from 'console/components';
 import { ROUTES } from '../global/constants';
 
 export const MainRoutes = () => (
   <Switch>
-    <Redirect from="/" to={ROUTES.Registration.HOME} exact />
+    <PrivateRoute exact path="/" ifUnauthorizedRedirectTo={ROUTES.Registration.HOME}>
+      <Redirect to={ROUTES.Console.HOME} />
+    </PrivateRoute>
```
xitij2000, line 24: This will take you to the Console home if you are logged in, but take you to the registration home if you are logged out.

frontend/src/global/api.ts
giovannicimolin: So, there's a few failing edge cases here:
* Any requests made with data will lose context
* If you go to the console after the access is expired, the interface just keeps on loading forever.

I've done a few tests locally, and manage to add a bit of code that **retries** the request if the token is successfully updated. Here is the entire snippet:
```
const getApiKey = () => {
  const authToken = localStorage.getItem('token_access');
  // The current version of the API client requires us to specify the word "Bearer"
  // The previous version did not.
  return authToken ? `Bearer ${authToken}` : '';
}

const config = new Configuration({
  basePath: `${OCIM_API_BASE}/api`,
  apiKey: (name: string, scopes?: string[]) => {
    return getApiKey();
  },
  middleware: [
    {
      post: async context => {
        if (context.response.status === 403) {
          // Failed requests return 403 and mean that the access token is
          // expired, so we trigger a page refresh.
          const authenticated = await checkAuthAndRefreshToken();
          if (!authenticated) {
            window.location.reload(false);
          }
          else {
            // Retry request after updating API key
            const requestContext = context.init;
            requestContext.headers = {
              Authorization: getApiKey()
            };
            let retry = context.fetch(context.url, requestContext);
            return retry;
          }
        } else if (context.response.status === 401) {
          // If the refresh endpoint fails, it means that the refresh
          // is expired, and the user is effectively logged out.
          performLogout();
        }
        return context.response;
      }
    }
  ]
});
``` 

Can you include this in your diff?

frontend/src/auth/components/PrivateRoute/PrivateRoute.tsx
giovannicimolin: Nice! That's perfect, but see my other comment regarding access token refresh issues.
I think I've found a good solution.

frontend/src/routes/main.tsx
giovannicimolin: @xitij2000 Can you change the console home to the `Theme preview and colors` page on https://github.com/open-craft/opencraft/blob/6df9c2e14cf0b859159fc847dcf0d820a8576fc0/frontend/src/routes/console.tsx#L22?

frontend/src/ui/components/Header/Header.tsx
```diff
@@ -1,72 +1,48 @@
...
+            <Route path={ROUTES.Registration.HOME}>
+              <Nav.Link onClick={navigateTo(ROUTES.Auth.LOGIN)}>Login</Nav.Link>
+            </Route>
+            <Route path={ROUTES.Auth.LOGIN}>
+              <Nav.Link onClick={navigateTo(ROUTES.Registration.HOME)}>
                 Create your account
               </Nav.Link>
-            )}
+            </Route>
+            <Route path={ROUTES.Console.HOME}>
```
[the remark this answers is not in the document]
xitij2000: Yup, in any page you can have a route component that renders different things based on the current page. This can also allow (for instance) for us to have a Help component that is page specific since it has a different help message on each page.

registration/migrations/0007_fix_invalid_domain_names.py
```diff
@@ -59,5 +59,4 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(fix_invalid_domain_names, migrate_noop),
```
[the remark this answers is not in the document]
xitij2000: Yup. I tried squashing migration but ran into too many issues with interdependencies. It's no small task. 

Instead I quickly set up pytest and used that to run the tests without migrations.

frontend/src/global/api.ts
xitij2000: I tried the exact same thing but it wasn't working. Now I realise the obvious stupid mistake I made, which is that obviously the old request will have the old token so it needs to be updated!

frontend/src/ui/components/Header/Header.tsx
xitij2000: This was the main reason for doing that! Otherwise, redirecting would either go to the login page when logged out, or the registration page when logged in.

instance/models/deployment.py
```diff
@@ -0,0 +1,73 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# Enums #######################################################################
+
+class DeploymentType(DjangoChoiceEnum):
+    """Enumeration of all types of deployments"""
+    user = 'Deployment initiated by user'
+    batch = 'Deployment created by batch redeplpoyment script'
```
giovannicimolin, line 44: Typo
```suggestion
    batch = 'Deployment created by batch redeployment script'
```

PR #574, opened 2020-05-12
BB-2328: Add new frontend redirects

giovannicimolin:
This PR adds a simple redirection mechanism to direct users to the new frontend until we do the domain migration.

**Testing instructions:**
1. Checkout this branch.
2. Run Ocim.
3. Go to http://localhost:5000/registration and check that the form is displaying as usual.
4. Add these to your `.env` file:
```
USER_CONSOLE_FRONTEND_URL="https://app.console.opencraft.com"
NEW_USER_CONSOLE_REGISTRATION_ENABLED=True
```
5. Open your browser inspetion tool.
6. Restart Ocim, check that http://localhost:5000/registration redirects to the new frontend.
7. Check that the redirect was a `302 Found` and not a permanent redirect. 

**Reviewer:**
- [x] @Agrendalath 

**Notes:**
This is just a temporary change, until Ocim gets migrated to the new domain (manage.opencraft.com). It will be removed soon, so that's why I didn't add any tests.

### DIFF
--- a/registration/views.py
+++ b/registration/views.py
@@ -22,8 +22,10 @@ Registration views
 
 # Imports #####################################################################
 
+from django.conf import settings
 from django.contrib.auth import authenticate, login
 from django.db import transaction
+from django.shortcuts import redirect
 from django.urls import reverse_lazy
 from django.utils.decorators import method_decorator
 from django.views.decorators.debug import sensitive_post_parameters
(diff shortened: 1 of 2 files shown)


registration/views.py
```diff
@@ -81,6 +83,22 @@ def form_valid(self, form):
         verify_user_emails(user, user.email, self.object.public_contact_email)
         return response
 
+    # pylint: disable=arguments-differ
+    def get(self, *args, **kwargs):
+        """
+        Get registration form.
+
+        Retirects to new frontend if flag is set, otherwise
```
Agrendalath, line 91: Nit: typo:
```diff
- Retirects
+ Redirects
```

Agrendalath: Also we could add the name of the variable from settings (currently `REDIRECT_TO_NEW_CONSOLE`) to the docstring to be aware of this condition without reading the code.

giovannicimolin:
Skipping integration tests since the change doesn't touch any code run by that and there's a pressing deadline coming up.

PR #578, opened 2020-05-21
[BB-2444] Remove public contact email from registration

rusrushal13:
**Issue**: [BB-2444](https://tasks.opencraft.com/browse/BB-2444)

**Screenshots**(Updated):

image

image

Testing instructions:
* Go through the registration process by creating a new instance, check it works perfectly or not
* At the congratulation page, the message should be right.
* Go through by signing in and checking that the first time the public contact email should be similar to the user email you set when creating the instance.

**Reviewers**:
- [ ] @giovannicimolin

### DIFF
--- a/frontend/packages/api_client/src/models/OpenEdXInstanceConfig.ts
+++ b/frontend/packages/api_client/src/models/OpenEdXInstanceConfig.ts
@@ -71,9 +71,7 @@ export interface OpenEdXInstanceConfig {
      */
     instanceName: string;
     /**
-     * The email your instance of Open edX will be using to send emails, and where your users should send their support requests.
      * 
-     * This needs to be a valid email.
      * @type {string}
      * @memberof OpenEdXInstanceConfig
      */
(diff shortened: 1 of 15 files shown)


frontend/packages/api_client/src/models/OpenEdXInstanceConfig.ts
```diff
@@ -70,14 +70,6 @@ export interface OpenEdXInstanceConfig {
...
      */
     instanceName: string;
-    /**
-     * The email your instance of Open edX will be using to send emails, and where your users should send their support requests.
-     * 
-     * This needs to be a valid email.
-     * @type {string}
-     * @memberof OpenEdXInstanceConfig
-     */
-    publicContactEmail: string;
```
giovannicimolin, line 80: @rusrushal13 You should not remove `publicContactEmail` from the API endpoint, just make it an optional parameter and auto-fill it with the user's email if it's not filled in.

giovannicimolin: Users should be able to edit this field in the console, we just want to remove it on the registration form.

giovannicimolin:
@rusrushal13 Btw, from your screenshot, it looks like the styling of that page is broken. Can you make sure it's fixed before the final version?

See the console in production:
image

frontend/packages/api_client/src/models/OpenEdXInstanceConfig.ts
rusrushal13: Thanks @giovannicimolin , I have updated the PR :)

PR #579, opened 2020-05-22
BB-2426: Add email verification page

giovannicimolin:
This PR replaces the Ocim email confirmation page by a page in the new frontend.

**Testing instructions:**
1. Checkout this branch.
2. Add these variables to your `.env`:
```
USER_CONSOLE_FRONTEND_URL="http://localhost:3000/"
NEW_USER_CONSOLE_REGISTRATION_ENABLED=True
```
3. Make sure you have email set up on your devstack:
```
EMAIL_BACKEND='django.core.mail.backends.smtp.EmailBackend'
# SMTP configuration
EMAIL_HOST=smtp.gmail.com
EMAIL_USE_TLS=True
EMAIL_PORT=587
EMAIL_HOST_USER=YOUR-ACCOUNT@opencraft.com
EMAIL_HOST_PASSWORD=<PASSWORD>
``` 
4. Register a new instance.
5. Click the link in your email, check that you were taken to the new frontend, and wait until the confirmation page appears. 
6. Check that the instance starts provisioning.
7. Try a random link: http://localhost:3000/verify-email/1234/ and check that the email verification fails.

**Reviewer:**
- [ ] @lgp171188


email_verification/api/v2/views.py
```diff
@@ -0,0 +1,79 @@
...
+    public_actions=["retrieve"],
+)
+class VerifyEmailViewset(RetrieveModelMixin, GenericViewSet):
+    """
+    Public view used by the frontend to activate email addresses.
+    """
+    authentication_classes = []
+    permission_classes = [AllowAny]
+    # We don't need a queryset or serializer
+    queryset = ''
```
lgp171188, line 52: @giovannicimolin, should this be set to `None` instead of `''`?

email_verification/api/v2/views.py
```diff
@@ -0,0 +1,79 @@
...
+        responses={**VALIDATION_AND_AUTH_RESPONSES},
+        security=[],
+    )
+    # pylint: disable=arguments-differ
+    def retrieve(self, request, pk=None):
+        """
+        Return a list of all users.
+        """
+        try:
+            EmailAddress.objects.confirm(pk).email
```
lgp171188, line 66: @giovannicimolin, the link works without any errors even when I open the confirmation link multiple times. Shouldn't the link expire after a single use or the expiry time whichever happens first?

email_verification/tests/test_api.py
```diff
@@ -0,0 +1,78 @@
...
+from django.contrib.auth.models import User
+from django.test import TestCase, override_settings
+from django.urls import reverse
+
+from simple_email_confirmation.models import EmailAddress
+
+
+# Tests #######################################################################
+
+class VerifyEmailAPITestCase(TestCase):
```
lgp171188, line 36: @giovannicimolin, depending on the result of the discussion above about having single-use links, you may have to add tests for that.

registration/tests/test_views.py
```diff
@@ -100,9 +100,26 @@ def _assert_registration_succeeds(self, form_data):
         with patch('registration.provision._provision_instance', autospec=True) as mock_handler:
             expected_call_count = 0
             for verification_email in mail.outbox:
-                verify_url = re.search(r'https?://[^\s]+',
-                                       verification_email.body).group(0)
-                self.client.get(verify_url)
+                # Extract validation code from email
+                verify_code = re.search(
```
lgp171188, line 104: @giovannicimolin, this is ugly. [Obligatory reference](http://regex.info/blog/2006-09-15/247).
> Some people, when confronted with a problem, think “I know, I'll use regular expressions.”  Now they have two problems.

:stuck_out_tongue: 

But I don't think we can do anything else that is better here :man_shrugging: 
.

frontend/src/auth/components/EmailVerificationPage/styles.scss
```diff
@@ -0,0 +1,19 @@
...
+    padding-left: 15px;
+}
+
+.email-verification-console-button {
+    margin-bottom: 20px;
+
+    i {
+        padding-left: 10px;
+    }
+}
```
lgp171188, line 19: @giovannicimolin, missing newline at the end of the file.

frontend/src/auth/components/EmailVerificationPage/displayMessages.ts
```diff
@@ -0,0 +1,32 @@
...
+  followUp: {
+    defaultMessage:
+      'Your instance should be ready within the next 2 hours. ' +
+      'In the meantime, you can start customizing your instance.',
+    description: ''
+  },
+  contact: {
+    defaultMessage:
+      "At any point, don't hesitate to write to us at " +
+      'contact@opencraft.com if you have any questions or issues.',
```
lgp171188, line 19: @giovannicimolin, can we not hard-code `contact@opencraft.com`?

email_verification/api/v2/views.py
giovannicimolin: It's not possible because the Viewset expects a queryset or a valid queryset return (and `None` isn't one of them).
Running with `None` makes Swagger fail to genenerate documentation.

This was the fastest way to properly add this view to the DRF Routers. Do you have any suggestions?

email_verification/api/v2/views.py
giovannicimolin: I'm just reusing the code from the email verification library. And it has the same behavior of the previous one. Changing the email verification library is not in the scope of this task.

email_verification/tests/test_api.py
giovannicimolin: See comment above. I'm not modifying the email verification library on this task.

registration/tests/test_views.py
giovannicimolin: I don't like this approach too (seems hacky to fetch the link from the email and run it through the browser).
Hopefully we get rid of this when doing e2e test of the new MFE.

giovannicimolin: Do you have any suggestions on how to improve this?

frontend/src/auth/components/EmailVerificationPage/styles.scss
giovannicimolin: Done!

frontend/src/auth/components/EmailVerificationPage/displayMessages.ts
giovannicimolin: Done!

email_verification/api/v2/views.py
lgp171188: I'm not sure why this behaviour is there. But okay to leave it as-is for the time being as I will need time to figure out if there is a better way.

email_verification/api/v2/views.py
lgp171188: @giovannicimolin, can you add a comment mentioning this behaviour so that this is documented? Okay to just do it here as the old registration form code is going to be removed.

registration/tests/test_views.py
lgp171188: You can get the confirmation key from the `EmailAddress` model, generate the URL using the key and assert that the email contains the URL and also verify that the verification works using that URL. What do you say?

giovannicimolin: Yup, definitively a better approach!

email_verification/api/v2/views.py
giovannicimolin: Done!

giovannicimolin:
Merging as per https://chat.opencraft.com/opencraft/pl/7rune54ik7futy7bys9hc1ifnr.

@lgp171188 Please leave your approval here later.

PR #580, opened 2020-05-25
Small Improvements in OCIM devstack setup docs

ahmedbilal:
* Add a note to use Python 3.5 or 3.6 for ansible virtual environment
* Add instructions to run `migrate` before attempting to create User.

### DIFF
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -40,6 +40,11 @@ pointing to it instead:
 
     ln -s <cloned_repo_path> deploy
 
+
+**Note:** Make sure to use Python 3.5 or 3.6 as the ansible-playbooks are not yet compatible 
+with the newer Python versions. You can use pyvenv to setup Python 3.5 or 3.6. For more information
+see https://hackernoon.com/reaching-python-development-nirvana-bb5692adf30c
+
 Create a new virtualenv to install the dependencies of the `ansible-playbooks`
 repository – most notably Ansible:
 
@@ -61,8 +66,12 @@ command:
     vagrant ssh
 
 Inside the virtual machine, set HUEY_ALWAYS_EAGER to false in
-/home/vagrant/opencraft/.env. Also, create a superuser account which will
-be used to log in to Ocim:
+/home/vagrant/opencraft/.env. Run all the pending migrations:
+
+    make manage migrate
+
+
+Also, create a superuser account which will be used to log in to Ocim:
 
     make manage createsuperuser
 
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -68,7 +68,7 @@ command:
 Inside the virtual machine, set HUEY_ALWAYS_EAGER to false in
 /home/vagrant/opencraft/.env. Run all the pending migrations:
 
-    make manage migrate
+    make migrate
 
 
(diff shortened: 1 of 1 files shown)


documentation/installation.md
```diff
@@ -40,6 +40,11 @@ pointing to it instead:
 
     ln -s <cloned_repo_path> deploy
 
+
+**Note:** Make sure to use Python 3.5 or 3.6 as the ansible-playbooks are not yet compatible
```
lgp171188, line 44: @ahmedbilal, in case you are wondering why we are still using Python 3.5 or 3.6 when 3.9 will be out soon - it is because Ubuntu 16.04 is the base OS supported by the Open edX platform (ships with Python 3.5.x) and we use the same base OS throughout our infrastructure to minimize compatibility issues. However, we had to upgrade Ocim to run on Python 3.6 (set up in production using `pyenv`, `pyenv-virtualenv`) as one of the dependencies we use required Python 3.6

documentation/installation.md
```diff
@@ -61,8 +66,12 @@ command:
     vagrant ssh
 
 Inside the virtual machine, set HUEY_ALWAYS_EAGER to false in
-/home/vagrant/opencraft/.env. Also, create a superuser account which will
-be used to log in to Ocim:
+/home/vagrant/opencraft/.env. Run all the pending migrations:
+
+    make manage migrate
```
lgp171188, line 71: @ahmedbilal, there is a `make migrate` shortcut to run the migrations. Not to be confused with `make migrations` which translates to the `makemigrations` Django management command.

```suggestion
    make migrate
```

documentation/installation.md
ahmedbilal: @lgp171188 Aah. Make sense now. Thanks :+1:

lgp171188:
@ahmedbilal, thanks for these improvements! :+1: 

It is probably a good idea to make a note of all the issues that you ran into and the details of how you fixed them and mention them on your onboarding task so that we can collect similar feedback from other newcomers and address those issues in batches. It is *not* a good idea to use your onboarding budget to fix issues like this. Please ask your mentor more about this if you have any questions :)

ahmedbilal:
@lgp171188 Thanks for review. Should I squash commits?

documentation/installation.md
```diff
@@ -40,6 +40,11 @@ pointing to it instead:
 
     ln -s <cloned_repo_path> deploy
 
+
+**Note:** Make sure to use Python 3.5 or 3.6 as the ansible-playbooks are not yet compatible 
+with the newer Python versions. You can use pyvenv to setup Python 3.5 or 3.6. For more information
+see https://hackernoon.com/reaching-python-development-nirvana-bb5692adf30c
```
lgp171188, line 46: @ahmedbilal, nit: wouldn't it be better to link to the [official README](https://github.com/pyenv/pyenv) instead of some random link on the internet? :) It's okay if you want to retain this and also add the link to the official README file :)

ahmedbilal: @lgp171188 Yep, that logical. Actually, I substituted `virtualenv` in my mind and forgot to mention it.

PR #581, opened 2020-05-26
[BB-2446] Added a message on console if the emails not been verified

rusrushal13:
**Issue**: [BB-2446](https://tasks.opencraft.com/browse/BB-2446)

**Screenshots**(Updated):
image

**Testing Instructions**:
- Create a new user, in starting it won't have the verified, the alert should be showing on the console.
- Next, verify the mail and the info should not show on the console.

**Reviewers**:
- [ ] @xitij2000 
- [ ] @giovannicimolin


frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -25,6 +25,10 @@ export const RedeploymentToolbar: React.FC<Props> = ({
 }: Props) => {
   const [show, setShow] = React.useState(false);
 
+  const [alertShow, setAlertShow] = React.useState(true);
+
+  const handleCloseAlert = () => setAlertShow(false);
```
xitij2000, line 30: I think this functionality is meaningless since it's not persisted on page refresh. Let it be a persistent notification till the email is verified, since we want to gate access on this basis.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -33,11 +37,13 @@ export const RedeploymentToolbar: React.FC<Props> = ({
   let undeployedChanges: number = 0;
   let deploymentStatus: DeploymentStatus | null = null;
   let deploymentType: DeploymentType | null = null;
+  let isEmailVerified: boolean = false;
```
xitij2000, line 40: This should default to true, otherwise all users will see this message flash while the details are being loaded from the API. That is not a good experience and can be confusing.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -70,73 +76,90 @@ export const RedeploymentToolbar: React.FC<Props> = ({
...
+          <Alert
+            className="alert-info"
+            onClose={() => handleCloseAlert()}
+            variant="info"
+            dismissible
+          >
+            <Alert.Heading>Please verify all your emails!</Alert.Heading>
+          </Alert>
+        </div>
+      ) : null}
```
xitij2000, line 162: I don't think the redeployment toolbar is the right place for this. I think it's better to move this to its own component and have it directly on the ConsolePage.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -70,73 +76,90 @@ export const RedeploymentToolbar: React.FC<Props> = ({
...
+      </div>
+      {!isEmailVerified && alertShow ? (
+        <div className="redeployment-toolbar">
+          <Alert
+            className="alert-info"
+            onClose={() => handleCloseAlert()}
+            variant="info"
+            dismissible
+          >
+            <Alert.Heading>Please verify all your emails!</Alert.Heading>
```
xitij2000, line 159: Please move this message to a WrappedMessage so it can be translated in the future.

xitij2000: We don't need an alert heading tag here since this is all the content. A heading would be useful if there was more here.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -70,73 +76,90 @@ export const RedeploymentToolbar: React.FC<Props> = ({
...
+                messages={messages}
+              />
+            </Button>
+          </Modal.Footer>
+        </Modal>
+      </div>
+      {!isEmailVerified && alertShow ? (
+        <div className="redeployment-toolbar">
+          <Alert
+            className="alert-info"
```
xitij2000, line 154: Make this an alert-warning, since it prevents users from accessing certain functionality.

frontend/src/console/components/RedeploymentToolbar/styles.scss
```diff
@@ -57,3 +56,15 @@
...
+  width: 90%;
+  font-family: "Roboto", sans-serif;
+  background-color: #6d9cae;
+  align-items: center;
+  display: flex;
+  margin: 0rem;
+  justify-content: center;
+  border-radius: 15px;
+  padding: 0.5rem;
+}
```
xitij2000, line 70: Let's keep custom styling to a minimum, keep whatever we get from bootstrap and add only the bare minimum needed. 

Also, don't directly use font names or colour values. Instead, use the defined variables for our theme colours. Here just use the default colours for alert. The theme should set them, you shouldn't override them.

registration/api/v2/serializers.py
```diff
@@ -367,6 +367,7 @@ class OpenEdXInstanceDeploymentStatusSerializer(serializers.Serializer):
     undeployed_changes = serializers.JSONField()
     deployed_changes = serializers.JSONField()
     deployment_type = serializers.ChoiceField(choices=DeploymentType.choices())
+    is_email_verified = serializers.BooleanField()
```
xitij2000, line 370: The email verification status has nothing to do with a particular deployment. It has to do with a user profile or user's openedx instance. Please move this to "AccountSerializer" and have the API include information about which particular email addresses are unverified.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
giovannicimolin: @xitij2000 We need to hide the deployment toolbar before the user activates the email, otherwise it'll wrongfully show that there are pending changes (since there's no changes).

So I though of turning the deployment toolbar into a yellow container with the message stating that the user needs to active the emails before the instance is set up.

xitij2000: OK. In that case this can just replace the deployment toolbar?

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
rusrushal13: Thanks, I have updated it :)

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
rusrushal13: Thanks, yeah it makes sense. I have changed it :+1:

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
rusrushal13: I have made the same :+1:

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
rusrushal13: I have changed it to a component

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
rusrushal13: Thanks, I have changed it :+1:

frontend/src/console/components/RedeploymentToolbar/styles.scss
rusrushal13: I have changed it to a component :+1:

frontend/src/console/components/AlertMessage/AlertMessage.tsx
```diff
@@ -0,0 +1,14 @@
...
+import messages from './displayMessages';
+
+export const AlertMessage: React.FC = () => (
+  <Alert
+    className="d-flex justify-content-center align-middle alert-warning"
+    variant="warning"
+  >
+    <WrappedMessage id="verifyEmail" messages={messages} />
+  </Alert>
+);
```
xitij2000, line 14: Since this component has a generic name like `AlertMessage` it shouldn't be fixed to just showing an alert message about email. Either the component should be called `EmailVerificationAlert` or it should be generic and allow any alert message. 

Can you move this `WrappedMessage` to a parameter to this component?
i.e. 

```suggestion

interface Props {
  children: React.ReactNode;
}

export const AlertMessage: React.FC = (props: Props) => (
  <Alert
    className="d-flex justify-content-center align-middle alert-warning"
    variant="warning"
  >
    {props.children}
  </Alert>
);
```

frontend/src/console/components/AlertMessage/styles.scss
```diff
@@ -0,0 +1,8 @@
+@import '~styles/theme';
+
+.alert-warning {
+  background-color: #6d9cae;
+  border-radius: 15px;
+  border: 0px;
+  color: white;
+}
```
xitij2000, line 8: Don't hard-code colour or font values.  If you want to use this colour, introduce a new variable for it. I think it's better to use the default bootstrap theme colour. If that needs to be changed, change it at styles/_theme.scss.

frontend/src/console/components/ConsolePage/styles.scss
```diff
@@ -71,8 +71,6 @@
   }
 }
 
-
-
 .header-link {
   color: white;
```
xitij2000, line 76: Please remove unnecessary changes from the diff.

frontend/src/console/components/RedeploymentToolbar/RedeploymentToolbar.tsx
```diff
@@ -70,73 +70,78 @@ export const RedeploymentToolbar: React.FC<Props> = ({
...
+                id="confirmCancelDeployment"
+                messages={messages}
+              />
+            </Button>
+          </Modal.Footer>
+        </Modal>
+      </div>
     </div>
   );
 };
```
xitij2000, line 147: If there are no longer any changes in this file, please remove it from the commit.

frontend/src/console/components/RedeploymentToolbar/styles.scss
```diff
@@ -40,7 +39,7 @@
     font-family: $playfair-font;
     font-size: 24px;
     color: #0f1f24;
-    margin-bottom: -15px
+    margin-bottom: -15px;
   }
 
   .modal-body {
```
xitij2000, line 45: Remove these changes.

frontend/src/console/components/ConsolePage/__snapshots__/ConsolePage.spec.tsx.snap
```diff
@@ -307,41 +307,10 @@ exports[`Console Page Correctly renders page with data 1`] = `
...
-      <button
-        className="float-right loading btn btn-primary btn-lg"
-        disabled={true}
-        onClick={[Function]}
-        type="button"
-      >
-        Deploy (0 updates)
-      </button>
-    </div>
+    Some of your email addresses are pending confirmation. Please confirm them to begin managing your Open edX instance.
```
lgp171188, line 313: @xitij2000, @rusrushal13, there is just one email address for the user to confirm here, right? If so, can this be updated to `Please confirm your email address to begin managing your Open edX instance.`?

CC @giovannicimolin

frontend/src/console/components/ConsolePage/displayMessages.ts
```diff
@@ -2,6 +2,11 @@ const messages = {
   editCourses: {
     defaultMessage: 'Edit courses (Studio)',
     description: 'Label to Studio link.'
+  },
+  verifyEmail: {
+    defaultMessage:
+      'Some of your email addresses are pending confirmation. Please confirm them to begin managing your Open edX instance.',
```
lgp171188, line 8: @xitij2000, @rusrushal13 (CC @giovannicimolin), [this comment](https://github.com/open-craft/opencraft/pull/581#discussion_r431072778) is relevant here. Posted there by mistake.

frontend/src/ui/components/EmailActivationAlertMessage/styles.scss
```diff
@@ -0,0 +1,26 @@
...
+
+
+.email-activation-alert {
+  background-color: #fff3cd;
+  color: #856404;
+  padding: 15px;
+  margin: 0 -15px;
+
+  display: flex;
+  align-items: center;
```
rusrushal13, line 11: @giovannicimolin a small nit pick, we don't need to explicitly mention `display: flex` and `align-items: center` as we are using bootstrap classes in our `div` container.
Also, as @xitij2000 pointed out we can put the hex in the `themes.scss` :)

rusrushal13:
@giovannicimolin LGTM
:+1:

- [x] I tested this: (describe what you tested)
- [x] I read through the code
- [ ] ~I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~

PR #585, opened 2020-06-01
Disable users associated with archived instance

pkulkark:
This PR fixes the issue [SE-2486](https://tasks.opencraft.com/browse/SE-2486)

Testing Instructions:

1. Checkout this branch on your ocim devstack.
2. Register an user and verify the login works.
3. Create an instance and ensure the registered user is added to the lms_users list.
4. Archive the instance.
5. Verify that the login for the user associated with the instance doesn't work.


instance/models/openedx_instance.py
```diff
@@ -464,9 +464,20 @@ def archive(self, **kwargs):
...
+        self.disable_users()
         super().archive()
         self.logger.info('Archiving instance finished.')
 
+    def disable_users(self):
+        """
+        Deactivate users associated with this instance.
+        """
+        lms_users = self.lms_users.all()
+        for user in lms_users:
```
s0b0lev, line 476: Hi, what do you think if instead for/loop use orm, like:
```
self.lms_users.filter(is_superuser=False).update(is_active=False)
```

PR #589, opened 2020-06-09
[SE-2666] Fix unexpected prefix purge

pkulkark:
This PR fixes the issue of consul purging all keys containing a prefix thus leading to unexpected deletions. 

**Issue:** [SE-2666](https://tasks.opencraft.com/browse/SE-2666)

**Testing Instructions:**
1. Checkout this branch and create an ocim instance such that it's id can be a prefix to already existing instance (use a non-critical instance)
2. Archive the instance and verify that only the consul data of the archived instance is removed and not the prefixed instance.

**Reviewers**

- [x] @pcockwell 
- [ ] @lgp171188

### DIFF
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -586,13 +586,9 @@ class ConsulAgent:
     def purge(self):
         """
         Will remove the key/prefixed-key from Key-Value store.
-        If no key/prefixed-key is specified, all keys/prefixed-keys will be removed.
         :return: True if the operation succeeded, False otherwise.
         """
-        if not self.prefix:
-            return self._client.kv.delete(self.prefix, recurse=True)
-        else:
-            return self._client.kv.delete(self.prefix)
+        return self._client.kv.delete(self.prefix)
 
     @staticmethod
     def _cast_value(value):
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -572,12 +572,6 @@ class ConsulAgent:
         stored['version'] = stored['version'] + 1
         return self._client.kv.put(self.prefix, json.dumps(stored).encode('utf-8'))
 
-    def remove_dict(self):
-        """
-        Remove a config dict using this instance's prefix.
-        """
-        self._client.kv.delete(self.prefix)
-
     def delete(self, key, **kwargs):
         """
         Will delete the given key/prefixed-key value from Consul's Key-Value store.
@@ -598,7 +592,7 @@ class ConsulAgent:
         if not self.prefix:
             return self._client.kv.delete(self.prefix, recurse=True)
         else:
-            return self._client.kv.delete(self.prefix, recurse=False)
+            return self._client.kv.delete(self.prefix)
 
     @staticmethod
     def _cast_value(value):
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -591,10 +591,14 @@ class ConsulAgent:
 
     def purge(self):
         """
-        Will simply removes all keys/prefixed-keys from Key-Value store.
+        Will remove the key/prefixed-key from Key-Value store.
+        If no key/prefixed-key is specified, all keys/prefixed-keys will be removed.
         :return: True if the operation succeeded, False otherwise.
         """
-        return self._client.kv.delete(self.prefix, recurse=True)
+        if not self.prefix:
+            return self._client.kv.delete(self.prefix, recurse=True)
+        else:
+            return self._client.kv.delete(self.prefix, recurse=False)
 
     @staticmethod
     def _cast_value(value):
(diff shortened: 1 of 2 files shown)


pcockwell:
@pkulkark @clemente After speaking with @giovannicimolin about testing this PR, he gave me access to the OCIM Stage UI and ssh access to `stage.console.opencraft.com` and informed me of how to test. 

Here are the steps I took:
- Login to OCIM Stage UI and create KV pairs under ocim/instances. The keys were `9` and `99`,  both with a value of `test`
- SSH to into `stage.console.opencraft.com`, and join the screen
- Checked out this branch, reset the Django app and python console
- Ran this code:
```
from instance.models.utils import ConsulAgent
agent = ConsulAgent(prefix='9')
agent.purge()
```
Upon doing so, I received an `ACLPermissionDenied` error (see below)
consul_test_acl_permission_denied

As of right now, until this is resolved, I am unable to properly test this.

clemente:
```
agent = ConsulAgent(prefix='9')
```
@pcockwell Here's the issue: Ocim's new ACL grants write access only to `/ocim/instances` (and maybe to a few others like `/ocim/certs/`, `/lock`, not-relevant here). You created your test key in `/ocim/instances/9`. However, with the line above you're asking to delete `/9`, and the ACL doesn't allow it. Try `ocim/instances/9`.

(By the way don't trust my syntax regarding the beginning slash, I write it because it makes the path clearer, but I think you don't need to write it)

pcockwell:
After following @clemente's instructions, I have managed to test this and verify it does indeed work. 

LGTM :+1:

instance/models/utils.py
```diff
@@ -591,10 +591,14 @@ def delete(self, key, **kwargs):
 
     def purge(self):
         """
-        Will simply removes all keys/prefixed-keys from Key-Value store.
+        Will remove the key/prefixed-key from Key-Value store.
```
lgp171188, line 594: @pkulkark, this method is only called from one place in the code and that is specific to an instance. Since the configuration data for an instance is stored in a single k-v pair, I don't see a reason why we need the recursive deletion at all.

Before, we used a prefix like `ocim/instances/<id>/` and stored multiple k-v pairs under that. The recursive deletion would have been needed only in that case. What do you say?

There is an unused `remove_dict` method which already does this. So we can move its logic to the `purge()` method, remove the `remove_dict()` method and the associated unit test.

CC @clemente, @pcockwell

instance/tests/models/test_utils.py
```diff
@@ -851,21 +851,22 @@ def test_purge_no_prefix(self):
 
     def test_purge_with_prefix(self):
```
lgp171188, line 852: Based on [this comment](https://github.com/open-craft/opencraft/pull/589/files#r437483538), this test can be removed since the `purge()` method will just call the `<consul client>.kv.delete()` method and we will be testing the `consul` client library.

instance/models/utils.py
clemente: I don't follow very well how prefixes are used; there seem to be many equivalent ways to operate with keys. I think our current case doesn't need recursive, but imagine that in the future we switch
- from `ocim/instances/123`
- to `ocim/instances/123/encrypted_data`, `ocim/instances/123/plaintext_data`

Then by asking „please delete 123 recursively“, the difference between recursive and non-recursive is relevant (I guess the 2nd case would delete more).
If we just store 1 key called 123, we don't need to care in practice (though the code may be untested in the cases we don't use).
Anyway, I'm not familiar with how we use prefixes and I don't care about the implementation. Just make sure that the tests cover the dangerous parts…

lgp171188: @clemente, afaik, we use prefixes in two contexts - the first to restrict all operations to the `ocim/instances` prefix. The second is to delete all the keys under an instance's prefix - `ocim/instances/1234/*`. We migrated away from the second a while ago.

We only store one key per instance now and so there is no need to recursively delete anything at all from the only code that calls `.purge()`, which is in the context of an instance.

pkulkark: @lgp171188 From the unit tests it looked like the expected behaviour is to delete all key-value pairs if no prefix is sent to the purge method. Should that be kept as is?

pkulkark: I've removed the `remove_dict()` method and the corresponding test case.

lgp171188: > @lgp171188 From the unit tests it looked like the expected behaviour is to delete all key-value pairs if no prefix is sent to the purge method. Should that be kept as is?

@pkulkark, I think this is not needed any more because:
* We don't store the instance configuration in multiple keys under a prefix.
* Other than the previous scenario, the only other scenario where deleting all key-pairs will be needed is if we have to delete all the instance data from consul. I'm sure we don't want that happening either.

So I think it should be okay to not expect or test the behaviour in question.

pkulkark: Thanks @lgp171188. I've made the required updates.

PR #591, opened 2020-06-15
[BB-2450] Notice board

0x29a:
This PR adds new page with name `Notice Board` that contains notifications about deployments and implements list endpoint for Open edX deployments.

Mockup itself: https://app.zeplin.io/project/5de4bac9fca2e1782d7245af/screen/5de4c06d0a6f456dd207ed62

**Issue:** [BB-2450](https://tasks.opencraft.com/browse/BB-2450)

**Screenshots**:

_Notification expand and collapse:_

Peek 2020-07-01 11-19

_Different types of notifications:_

Screenshot_2020-07-01_11-32-54

**Testing Instructions:**
1. Checkout this branch.
1. Enter new frontend subdirectory: `cd frontend`
1. Build Ocim API client: `./scripts/build-api-client.sh`
1. Reinstall dependencies: `npm install`
1. Run frontend, `npm start`.
1. Go to `http://localhost:3000/`
1. Log in as user that have instance and at least one deployment.
1. You should see, that `Status & Notifications` navigation link is enabled now:
    Screenshot_2020-06-15_12-51-46
    Click on it, you should see at least one deployment notification.
1. **[optional]**
To test different types of notifications, consider this change:
```diff
diff --git a/registration/api/v2/views.py b/registration/api/v2/views.py
index 2e580b40..527831d9 100644
--- a/registration/api/v2/views.py
+++ b/registration/api/v2/views.py
@@ -211,7 +211,30 @@ class NotificationsViewSet(GenericViewSet):
 
         undeployed_changes = build_instance_config_diff(application)
 
-        notifications = []
+        diff = [
+            ("change", "some_var1", ("test", "test1")),
+            ("change", "theme_config.btn-secondary-bg", ("#000000", "#FFFFFF")),
+            ("add", "theme_config", [("added-bg", "#FFFFFF")]),
+            ("remove", "theme_config", [("removed-bg", "#FFFFFF")]),
+            (
+                "add",
+                "static_content_overrides",
+                [
+                    ("version", 0),
+                    (
+                        "homepage_overlay_html",
+                        "<h1>Welcome to Wonderland</h1><p>It works! Powered by Open edX®</p>",
+                    ),
+                ],
+            ),
+            ("change", "some_dict.version", (0, 1)),
+            ("add", "", [("instance_name", "Wonderland")]),
+            ("remove", "", [("some_var2", "test")]),
+        ]
+        notifications = [
+            {"deployed_changes": diff, "date": application.instance.created, "status": status}
+            for status in DeploymentState
+        ]
         for deployment in deployments:
             deployment_status = deployment.status()
             notifications.append({
```
[Gist with code of endpoint with this change](https://gist.github.com/0x29a/4295e9864110723a59aa830ad612eb03).

**Reviewers**

- [x] @xitij2000  
- [x] @Agrendalath


frontend/src/console/components/NoticeBoard/NoticeBoard.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const [detailsCollapsed, setDetailsCollapsed] = React.useState(true);
+
+  const deploymentStatus = props.alert.status;
+
+  const caretClass = detailsCollapsed ? 'fa-caret-down' : 'fa-caret-up';
+  const collapsedClass = detailsCollapsed ? 'collapsed' : '';
+  const iconClass = iconClassMap[deploymentStatus];
+  const statusClass = statusClassMap[deploymentStatus];
+
+  const month = props.alert.date.toLocaleString('default', { month: 'long' });
```
Agrendalath, line 57: I believe that this would produce exactly what we want?
```js
const date = props.alert.date.toLocaleDateString('default', {month: 'long', year: 'numeric', day: 'numeric'})
```

frontend/src/console/components/NoticeBoard/NoticeBoard.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <p className="m-0 deployment-alert-date">
+            <small>{`${month} ${day}, ${year}`}</small>
+          </p>
+        </div>
+        <div // eslint-disable-line jsx-a11y/interactive-supports-focus
+          role="button"
+          className="d-flex expand-caret-button ml-auto"
+          onClick={() => setDetailsCollapsed(!detailsCollapsed)}
+          onKeyDown={() => setDetailsCollapsed(!detailsCollapsed)}
+        >
```
Agrendalath, line 86: Could we reuse [this](https://github.com/open-craft/opencraft/blob/1ea068526157dfcce6960115f42c6de9ff2f7db5/frontend/src/ui/components/CollapseEditArea/CollapseEditArea.tsx#L20-L27) or make something similar (using proper links instead of `div` button)?

frontend/src/console/components/NoticeBoard/NoticeBoard.tsx
```diff
@@ -0,0 +1,127 @@
...
+    </div>
+  );
+};
+
+const NoticeBoardComponent: React.FC<Props> = (props: Props) => {
+  React.useEffect(
+    () => props.getAlerts(),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
```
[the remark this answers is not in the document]
xitij2000: I think this is OK. I think it can be avoided by doing:
```jsx
React.useEffect(
    () => {props.getAlerts()},
    // eslint-disable-next-line react-hooks/exhaustive-deps
    []
  );
```

frontend/src/console/components/NoticeBoard/displayMessages.ts
```diff
@@ -0,0 +1,39 @@
...
+  [DeploymentStatus.Preparing]: {
+    defaultMessage: 'Deployment in-progress: preparing.',
+    description: 'Redeployment alert message.'
+  },
+  [DeploymentStatus.Offline]: {
+    defaultMessage: 'Deployment offline.',
+    description: 'Redeployment alert message.'
+  },
+  NoticeBoard: {
+    defaultMessage: 'Notice Board.',
```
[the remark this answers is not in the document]
xitij2000: Additionally, I should say we should maintain the standard convention of keeping the first letter small in cases like these.

frontend/src/console/components/NoticeBoard/styles.scss
```diff
@@ -0,0 +1,110 @@
+@import '~styles/theme';
+
+.deployment-alert {
+  .healthy,
+  .changes_pending {
+    background-color: $light-blue;
```
xitij2000, line 6: Instead of overriding colours like this, use the classes in jsx. i.e.
bg-primary, text-primary etc.

registration/api/v2/views.py
```diff
@@ -433,17 +433,22 @@ def get_queryset(self):
...
+
         if not user.is_authenticated:
-            return BetaTestApplication.objects.none()
+            return queryset.none()
         elif user.is_staff:
-            return BetaTestApplication.objects.all()
+            return queryset.all()
         else:
-            return BetaTestApplication.objects.filter(user=self.request.user)
+            return queryset.filter(user=self.request.user)
```
xitij2000, line 443: :+1: Good cleanup!

registration/api/v2/views.py
```diff
@@ -433,17 +433,22 @@ def get_queryset(self):
...
 
     def get_serializer_class(self):
         if self.action == "create":
             return OpenEdXInstanceDeploymentCreateSerializer
 
+        if self.action == "list":
+            return OpenEdXInstanceDeploymentAlertSerializer
+
```
xitij2000, line 451: Rather than this being the list action for deployments, I think we should create a completely new View. The reasoning here is:
- Eventually we might need a real deployment list, and this will be the natural place for it. 
- Eventually we will have more kinds of alerts, notifications etc than just those about instances/deployments.

0x29a:
> Could you please also add the instructions how to prepare different types of alerts on the page? There are some steps needed between point 7. and 8. from your instructions. For now I just manually the state [here](https://github.com/open-craft/opencraft/blob/baac851b31e7bf9239eea94426ddf2786e35682e/registration/api/v2/views.py#L554) to test this, as I don't have full context on this.

@Agrendalath 
I tried to prepare json redux state dump, but it seems that deserialization of string dates doesn't work for redux devtools plugin.
So, I included in PR description diff of view change, that allows to see different types of alerts.

frontend/src/console/components/NoticeBoard/styles.scss
0x29a: @xitij2000 
I can't find class alternatives for some colors, like this one (`$light-blue`), because they defined in our code (`frontend/src/styles/_theme.scss`).

So I'm not sure how to fix this.

0x29a:
And here is some sort of change log since last review:

1. Added `alertLoading` field to console state, use it to show loading overlay now.
2. Added test for `NoticeBoard.tsx`.
3. Fixed routing issue.
4. Moved new backend code to separate endpoint.
5. Fixed almost all mentioned issues.

frontend/src/console/components/NoticeBoard/styles.scss
0x29a: @xitij2000 Yay! I managed to remove almost all color overriding. Thanks for that suggestion.

0x29a:
Changelog:
1. Added the last portion of requested changes.
2. Renamed `alert` to `notification` everywhere.
3. Removed deprecated files from `api-client`.

registration/api/v2/views.py
```diff
@@ -133,6 +133,98 @@ def get_queryset(self):
...
+                "deployed_changes": [],
+                "status": DeploymentState.preparing.name,
+                "date": application.instance.created,
+            })
+
+        last_notification = notifications[-1]
+        if last_notification['status'] == DeploymentState.healthy:
+            undeployed_changes = build_instance_config_diff(application)
+            if undeployed_changes:
+                last_notification['status'] = DeploymentState.changes_pending
```
Agrendalath, line 218: Could you please add a comment explaining why we're doing this?

registration/api/v2/views.py
```diff
@@ -133,6 +133,98 @@ def get_queryset(self):
...
+        """
+        Get latest `BetaTestApplication` instance owned by current user.
+        """
+
+        user = self.request.user
+
+        applications = BetaTestApplication.objects.select_related('instance')
+
+        if user.is_staff:
+            applications = applications.all()
```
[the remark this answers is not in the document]
giovannicimolin: @Agrendalath 
>should we assume that the new console is used only by non-staff users?

Currently, many things work on this assumption, so it's ok to consider it's only used by non-staff (as long as nothing breaks when we log in it's fine :stuck_out_tongue:).
Both suggestions you've made are valid though, can you add this into the new follow up epic?

registration/api/v2/views.py
```diff
@@ -133,6 +133,98 @@ def get_queryset(self):
             return UserProfile.objects.filter(user=self.request.user)
 
 
+class NotificaitonsViewSet(GenericViewSet):
```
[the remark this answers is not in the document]
0x29a: Ouch!

0x29a:
@Agrendalath thanks!

And here is changelog:
- Fixed typos
- Added fetching of limit constant from environment
- Added handling of malformed limit
- Added comment about last deployment status change

frontend/src/console/components/NoticeBoard/NoticeBoard.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const collapsedClass = detailsCollapsed ? 'collapsed' : '';
+  const iconClass = iconClassMap[deploymentStatus];
+  const statusClass = statusClassMap[deploymentStatus];
+
+  const month = props.alert.date.toLocaleString('default', { month: 'long' });
+  const day = props.alert.date.getDate();
+  const year = props.alert.date.getFullYear();
+
+  const details = props.alert.deployedChanges ? (
+    JSON.stringify(props.alert.deployedChanges)
```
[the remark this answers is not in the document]
0x29a: Hey @giovannicimolin, do you have ideas about this?

PR is near to acceptable state, so if you have any suggestions, I can implement this.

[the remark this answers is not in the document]
giovannicimolin: @Agrendalath @0x29a Sorry for missing this ping. 
My idea was to have a parsed text here with the values changed (nothing fancy, just not a RAW json file).
It can be a simple list with items like this:
- Changed `variable_name` to `#ffaa00`
- Changed `variable_name` to `#00ee33`

Do you know have an estimate of how hard it is to implement that? Do you have enough time allocated for this sprint for this? 
If not, then expand the budget a bit and move it to blocker, and finish it early next sprint.

_Note:_ ignore the color preview from github.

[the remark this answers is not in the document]
0x29a: @giovannicimolin 
Thanks!

Well, it appears much easier than I thought :+1:

I think I can finish this before end of day today.

0x29a:
@giovannicimolin 
Here is how changes look now:

_No changes:_
Screenshot_2020-06-30_19-14-20

_Plenty of changes:_
Screenshot_2020-06-30_19-06-19

Is that what we want?

xitij2000:
> Is that what we want?
Perhaps in the future we can have a mapping that maps these to a friendly name?
So `btn-register-color` would show up as "Register button Colour". etc. 
Anything not in the mapping just wouldn't show up even if changed. This can be used for internal fields like version. 
CC: @giovannicimolin

giovannicimolin:
@0x29a Looks great to me.

@xitij2000 
>Perhaps in the future we can have a mapping that maps these to a friendly name?
So btn-register-color would show up as "Register button Colour". etc.
Anything not in the mapping just wouldn't show up even if changed. This can be used for internal fields like version.

That's a good idea, and will improve the UI even more here. 
@0x29a Can you add this improvement as a requirement in the BB-2625 epic?

0x29a:
@giovannicimolin 
>Can you add this improvement as a requirement in the BB-2625 epic?

Done.

>You can pass parameters to a WrappedMessage like this (code, message) instead of composing using small messages.

Updated. Thanks!

PR #592, opened 2020-06-18
[SE-2729] OCIM add instance description field

fferrin:
This PR adds a `notes` field to the instances to record the comments related to them.

**JIRA ticket**: [SE-2729](https://tasks.opencraft.com/browse/SE-2729)

**Screenshots**: 
- Notes view
image

- Edit notes view
image

1. Migrations must be done previously
2. UX design is OK?

**Reviewers**
- [x] @swalladge


instance/api/instance.py
```diff
@@ -117,3 +118,21 @@ def app_servers(self, request, pk):
         Get this Instance's entire list of AppServers
         """
         return Response(InstanceAppServerSerializer(self.get_object(), context={'request': request}).data)
+
+    @detail_route(methods=['post'])
+    def update_attrs(self, request, pk):
```
swalladge, line 123: I'm not sure we want the ability to update any attribute of an instance reference here. Perhaps a more focused `set_notes` handler?

instance/static/html/instance/details.html
```diff
@@ -28,6 +28,22 @@ <h3 ng-if="instance">
...
+<div class="panel" ng-controller="Details">
+  <div class="row">
+    <div class="large-10 columns">
+      <h3>Notes <a ng-click="is_editing_notes=true"><i class="fa fa-edit"></i></a></h3>
+
+      <div ng-if="!is_editing_notes">{{ instance.notes }}</div>
+
+      <form name="editNotesForm" ng-if="is_editing_notes">
+        <textarea rows="4" cols="50" ng-model="instance.notes">{{ instance.notes }}</textarea>
+        <a ng-click="is_editing_notes=false">Cancel</a>
```
swalladge, line 40: I also don't know enough angular to know why setting is_editing_notes here doesn't work, but does work in the other place above. Maybe add a `cancel_edit_notes` function to call?

instance/static/js/src/instance.js
```diff
@@ -175,6 +177,19 @@ app.controller("Details", ['$scope', '$state', '$stateParams', 'OpenCraftAPI',
             });
         };
 
+        $scope.edit_notes = function() {
+            let data = {"notes": $scope.instance.notes};
+
+            OpenCraftAPI.one("instance", $scope.instance.id).customPOST(data, "update_attrs").then(function () {
+                $scope.notify('Instance notes updated.');
```
swalladge, line 184: Nice use of notifications :+1:

instance/static/html/instance/details.html
```diff
@@ -28,6 +28,22 @@ <h3 ng-if="instance">
   </div>
 </div>
 
+<div class="panel" ng-controller="Details">
+  <div class="row">
+    <div class="large-10 columns">
+      <h3>Notes <a ng-click="is_editing_notes=true"><i class="fa fa-edit"></i></a></h3>
+
+      <div ng-if="!is_editing_notes">{{ instance.notes }}</div>
```
swalladge, line 36: Can we support markdown here? Maybe via sending it through [marked.js](https://marked.js.org) here in the frontend? We should also consider sanitizing, otherwise it leaves it open to XSS.

instance/api/instance.py
fferrin: I was thinking about using this endpoint for future instance updates, but I can limit the scope for now.

instance/static/html/instance/details.html
fferrin: That works, but I didn't want to create unnecessary functions if the same could be accomplished using the variable. I'm going to try a little more and if it doesn't work I add that function

instance/static/html/instance/details.html
fferrin: Definetly! I'll take a look at it

instance/static/js/src/instance.js
```diff
@@ -216,6 +248,8 @@ app.controller("Details", ['$scope', '$state', '$stateParams', 'OpenCraftAPI',
                     $scope.is_spawning_appserver = false;
                 }
                 $scope.old_appserver_count = instance.appserver_count;
+                $scope.originalNotes = $scope.instance.notes;
+                $scope.renderMarkDownElements();
```
swalladge, line 252: @fferrin, can we take a more functional approach here - something like 

```
<div ng-if="!isEditingNotes" class="markdown">{{ render(instance.notes) }}</div>
```

```
render = (text) => dompurify.sanitize(marked.parse(text))
```

instance/static/js/src/instance.js
```diff
@@ -175,6 +178,35 @@ app.controller("Details", ['$scope', '$state', '$stateParams', 'OpenCraftAPI',
             });
         };
 
+        $scope.editNotes = function() {
+            let data = {"notes": DOMPurify.sanitize($scope.instance.notes)};
```
swalladge, line 182: We want to sanitize the output, not the input, since we're doing it on the frontend - this sanitizing should be part of the function that renders the markdown text to html.

instance/static/js/src/instance.js
fferrin: @swalladge That looks better by the way, but I think we should either set the string to the `innerHTML` attribute or create a new child, right?

instance/static/js/src/instance.js
fferrin: Shouldn't we also sanitize the content for the backend to avoid SQL injections? Or is it not necessary?

swalladge: If we want to sanitize input, then we should do that on the backend - frontend sanitation is easily bypassed.

instance/static/js/src/instance.js
swalladge: Yep, I'm not that familiar with how angular does it, but I'm sure it has a method to output the raw html instead of escaping it. We want to find a function that lets angular handle it; we don't want to maintain other functions that manually set the innerhtml.

fferrin: I'll look for it then

fferrin:
@swalladge Could you give me some hints of why JS testings are failing? I just added `ngSanitize` in `instance/static/js/src/instance.js` and two modules in `Makefile`

instance/static/js/src/instance.js
```diff
@@ -23,7 +24,8 @@ var app = angular.module('InstanceApp', [
     'ngRoute',
     'ui.router',
     'restangular',
-    'mm.foundation'
+    'mm.foundation',
+    'ngSanitize'
```
swalladge, line 28: @fferrin why is ngSanitize added? I don't see it used anywhere.

swalladge: Oh I see, perhaps that was an alternative to dompurify? Looks like we can strip out ngSanitize if we're going with dompurify?

swalladge:
I can't see what's causing the js tests to fail yet... :/

static/external/js/Makefile
```diff
@@ -24,8 +24,8 @@ CDNJS = https://cdnjs.cloudflare.com/ajax/libs
 # Files #######################################################################
 
 all: angular.min.js angular.min.js.map angular-mocks.js angular-route.min.js angular-route.min.js.map \
-	 angular-ui-router.min.js django-angular.min.js icheck.min.js jquery.min.js jshint.js \
```
swalladge, line 27: @fferrin looks like jshint was accidentally removed? This may be causing issues.

instance/static/js/src/instance.js
fferrin: You were right. DOMPurify is an alternative to ngSanitize. I'll remove it

swalladge:
@fferrin :+1: great work! Please squash and merge. :D

- [x] I tested this: followed test instructions; checked for xss issues (ng-bind-html uses ngSanitize, so we're good here)
- [x] I read through the code
- [x] Includes documentation

PR #594, opened 2020-06-21
SE-2293 Switch to S3 as the default instance storage type for unit tests

nizarmah:
Switches default storage instance type to S3 and adds Swift storage type override wherever necessary.

**JIRA tickets**: SE-2293

**Dependencies**: None

**Merge deadline**: None

**Testing instructions**:

1. Use `make test` in the OCIM virtual machine.

> CircleCI build status should be enough.

**Reviewers**
- [ ] @mtyaka


nizarmah:
To be honest, I have no idea why it failed again. That's the error I am dealing with right now.
```
ERROR: test_spawn_appserver (instance.tests.integration.integration_instance.InstanceIntegrationTestCase)
  File "/home/********/project/venv/lib/python3.6/site-packages/django/test/utils.py", line 373, in inner
    return func(*args, **kwargs)
  File "/home/********/project/instance/tests/integration/integration_instance.py", line 389, in test_spawn_appserver
  File "/home/********/project/instance/tests/integration/integration_instance.py", line 71, in f_retry
    return f(*args, **kwargs)
  File "/home/********/project/instance/tests/integration/integration_instance.py", line 112, in assert_instance_up
    check_url_accessible(url)
  File "/home/********/project/instance/tests/integration/utils.py", line 63, in check_url_accessible
    get_url_contents(url, auth, attempts, delay)
  File "/home/********/project/instance/tests/integration/utils.py", line 48, in get_url_contents
    res.raise_for_status()
  File "/home/********/project/venv/lib/python3.6/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://0700c4.integration.**********/error
```

The unfortunate thing is that the integration test was reverted back to master, and only the unit tests were changed (removed s3 overrides), as requested in the JIRA task mentioned in this pull request. 

Anyway, I guess I'll open it for review. But please let me know if you have any idea on what is happening exactly.

mtyaka:
Thanks for this @nizarmah! Removing the setting from `.env.test` looks like a good first step, but I was hoping we could go further for this ticket.

To explain a bit of background on this ticket:

`swift` used to be the default instance storage in Ocim in the past, but because OVH Swift service that we've been using has been very unstable, we decided to switch to S3 (which is also what edX use). The only reason we still support `swift` is because of a few old instances that we haven't migrated to S3 yet (and I'm not even sure that's still true). 

Swift is no longer supported in recent versions of edx-platform and we'll be completely removing support for it from Ocim in the future as well. For this reason we want *all* tests to be using `s3` instead of `swift`. The only exceptions would theoretically be tests that explicitly test if the deprecated `swift` integration still works correctly - but I'm not sure if we have any such tests at all.

So what I'd like you to do is to try to remove all instances of `@override_settings(INSTANCE_STORAGE_TYPE='swift')` except perhaps in a few cases where using the deprecated `swift` storage might make sense.
If tests start failing when you do that, we should look into why they are failing. We might have to mock out some S3 APIs for unit tests.

mtyaka:
> Anyway, I guess I'll open it for review. But please let me know if you have any idea on what is happening exactly.

It looks like this could have just been a temporary issue with the S3 API. I triggered a test re-run to see if it happens again. Unfortunately integration tests are often flaky because of the amount of external services involved.

nizarmah:
@mtyaka Thanks a lot for the great explanation. The acceptance criteria on the Jira issue was a little bit misleading. Now that I know exactly what you want, I converted everything back to S3 and removed the Swift overrides. I will only add Swift overrides whenever needed, as you mentioned.

The CircleCI build should fail, but I just wanted to "refresh" the PR with the proper commits. I'll be working on fixing the errors that happen right now. :slightly_smiling_face: I'll keep you updated.

mtyaka:
Thanks @nizarmah, sorry for the confusion!

nizarmah:
@mtyaka The changes I just did should finally pass all the unit tests; however, [there seems to be the following lines in the `test_openedx_instance.py`.](https://github.com/open-craft/opencraft/blob/71b2b463480c1786d7c4f8b3ad9d17ad703d00af/instance/tests/models/test_openedx_instance.py#L247)

```
storage_settings = {
    'EDXAPP_DEFAULT_FILE_STORAGE: swift.storage.SwiftStorage',
    'EDXAPP_GRADE_STORAGE_CLASS: swift.storage.SwiftStorage',
    'VHOST_NAME: openstack',
    'XQUEUE_SETTINGS: openstack_settings'
}
```

Should that also be changed to `S3` for the class `OpenEdXInstanceTestCase`?

registration/tests/test_views.py
```diff
@@ -472,8 +470,7 @@ def test_invalid_privacy_url(self, privacy_url):
         })
 
     @override_settings(
-        DEFAULT_PRIVACY_POLICY_URL='http://example.com:5000/test-no-privacy',
-        INSTANCE_STORAGE_TYPE='s3'
+        DEFAULT_PRIVACY_POLICY_URL='http://example.com:5000/test-no-privacy'
```
mtyaka, line 473: Nit: this could fit into a single line now (`@override_settings(...)`)

instance/tests/integration/integration_instance.py
```diff
@@ -345,6 +343,9 @@ def test_spawn_appserver(self):
                 'static_template_about_content': 'Hello world!',
                 'homepage_overlay_html': '<h1>Welcome to the LMS!</h1>',
             },
+            s3_access_key='test-s3-access-key',
+            s3_secret_access_key='test-s3-secret-access-key',
+            s3_bucket_name='test-s3-bucket-name',
```
mtyaka, line 348: This is an integration test that should be using real AWS credentials (which are already set as environment variables in CircleCI), so I don't think we should be setting these s3 fields here.
I think that's why the integration test is currently failing.

instance/tests/models/test_openedx_instance.py
```diff
@@ -213,6 +213,9 @@ def test_spawn_appserver(self, mocks, privacy_policy_url, mock_consul, mock_prov
         instance = OpenEdXInstanceFactory(
             sub_domain='test.spawn',
             privacy_policy_url=privacy_policy_url,
+            s3_access_key='test',
+            s3_secret_access_key='test',
+            s3_bucket_name='test-s3-bucket-name',
```
mtyaka, line 218: I would like to avoid having to pass the s3_* fields in explicitly like this. Do tests fail if we remove them? If they're required for tests to work, maybe it would be better to set them in the mocked `provision_s3` method (which is [what would happen](https://github.com/open-craft/opencraft/blob/4b050bfba879dc9c26b2d1490079590374ad4169/instance/models/mixins/storage.py#L297-L298) if the method wasn't mocked out).

instance/tests/integration/integration_instance.py
nizarmah: Alright, thanks for telling me. However, I am having the `503` issue.

instance/tests/models/test_openedx_instance.py
nizarmah: They are needed, otherwise the `get_storage_settings()` doesn't return the proper output. Which in result causes some errors.

nizarmah: I removed them, I'll see what error I get and fix it accordingly, without adding those explicit `s3_*` fields.

mtyaka:
Nice work @nizarmah :+1: This is good to merge once the tests pass.

The changes looks good to me. I am not sure why the test keeps failing. I've seen that test sometimes fail on different PRs, so it seems like a flaky test, but the error is also generic enough that it could mask a real underlying problem.

- [x] I tested this: I observed the tests on Circle CI
- [x] I read through the code
- [ ] Includes documentation

PR #596, opened 2020-06-22
[SE-2599] Fix flaky integration tests

fferrin:
This PR increases the retries for spawning new OpenedX appservers by 2, to avoid some integration test failures in CircleCI.

**JIRA tickets**: [SE-2599](https://tasks.opencraft.com/browse/SE-2599)

1. This is the first approach to solve this issue given by @lgp171188.
2. Not sure how to test it given the flaky nature of the failures. Maybe leave it open for a month or something like that to see if it happens again?

**Reviewers**
- [x] @mtyaka

### DIFF
--- a/instance/tests/integration/integration_instance.py
+++ b/instance/tests/integration/integration_instance.py
@@ -77,6 +77,31 @@ def retry(f, exception=AssertionError, tries=5, delay=10):
     return f_retry  # true decorator
 
 
+def spawn_openstack_server(name_prefix, retries=1):
+    """
+    Spawn an OpenStack server.
+
+    :param name_prefix: Name prefix for OpenStack server
+    :param retries: Number of times it retry spawn a server, defaults to 1
+    :return: OpenStack server instance
+    :raises: TimeoutError
+    """
+    while retries >= 0:
+        server = OpenStackServer(name_prefix=name_prefix)
+        server.save()
+        server.start()
+
+        try:
+            server.sleep_until(lambda: server.status.accepts_ssh_commands, timeout=120)
+        except TimeoutError as e:
+            server.os_server.delete()
+            retries -= 1
+            if retries < 0:
+                raise e
+        else:
+            return server
+
+
 class InstanceIntegrationTestCase(IntegrationTestCase):
     """
     Integration test cases for instance high-level tasks
@@ -489,10 +514,7 @@ class InstanceIntegrationTestCase(IntegrationTestCase):
         """
         Test that OpenStackServer detects if the VM was terminated externally.
         """
-        server = OpenStackServer(name_prefix="integration_test")
-        server.save()
-        server.start()
-        server.sleep_until(lambda: server.status.accepts_ssh_commands, timeout=120)
+        server = spawn_openstack_server("integration_test")
         server.os_server.delete()
         self.assert_server_terminated(server)
 
--- a/instance/models/utils.py
+++ b/instance/models/utils.py
@@ -572,12 +572,6 @@ class ConsulAgent:
         stored['version'] = stored['version'] + 1
         return self._client.kv.put(self.prefix, json.dumps(stored).encode('utf-8'))
 
-    def remove_dict(self):
-        """
-        Remove a config dict using this instance's prefix.
-        """
-        self._client.kv.delete(self.prefix)
-
     def delete(self, key, **kwargs):
         """
         Will delete the given key/prefixed-key value from Consul's Key-Value store.
@@ -591,10 +585,10 @@ class ConsulAgent:
 
     def purge(self):
         """
-        Will simply removes all keys/prefixed-keys from Key-Value store.
+        Will remove the key/prefixed-key from Key-Value store.
         :return: True if the operation succeeded, False otherwise.
         """
-        return self._client.kv.delete(self.prefix, recurse=True)
+        return self._client.kv.delete(self.prefix)
 
     @staticmethod
     def _cast_value(value):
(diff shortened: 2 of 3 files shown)


lgp171188:
@fferrin, the `periodic_builds_retries` field that you have changed in this PR has nothing to do with retries in the integration tests. We cannot and should not change the default value of `periodic_builds_retries`. CC @mtyaka as the task reviewer.

fferrin:
@lgp171188 I will put my comment here to discuss it with @mtyaka. 

I consulted Guruprasat because I don't have a deep understanding of how OCIM works and sometimes I confuse some terms. Since this is not the parameter to edit to increase retries, could it be that we need to change the default value of [this task](https://github.com/open-craft/opencraft/blob/31fcd99f4fefeca7ced89ebd2bb735ea0e4b8bf4/instance/tasks.py#L59)?

mtyaka:
@fferrin The `num_attempts` parameter of the `create_new_deployment` function does control the number of retry attempts, but the test that was failing was testing a lower level function (spawning a single appserver rather than a full-blown deployment).

The problem with [the failing test](https://app.circleci.com/pipelines/github/open-craft/opencraft/910/workflows/6f7ca8c1-ded0-4572-9deb-7236a8d0b24e/jobs/10831/steps) was that after a new VM was requested on OVH, it remained unreachable over ssh for more than 2 minutes. However that failure is from January and I checked recent appservers and integration tests and haven't seen this particular error. Instead it would be better to try and fix [this failure](https://app.circleci.com/pipelines/github/open-craft/opencraft/1710/workflows/89666d71-d1b5-4c05-8649-2aa77e358fa6/jobs/17089), which has been occurring quite often lately.

I'll update [SE-2599](https://tasks.opencraft.com/browse/SE-2599) with some more details.

fferrin:
@mtyaka I added retries to the test that is currently failing. I understand that is what you meant by your comment, right? If so, do you think setting `retries = 1` is ok or should this value be increased? Would the docstring also need to be updated to explain the reason for the retry referring to the ticket?

I understand that this solves the problem of the test that fails but does not guarantee that in normal execution it will not happen, right? Should we also provide a solution for this case or would it be solved manually (spawning anothar appserver maybe)?

fferrin:
@mtyaka I'm having test failures in another PR also. For example, `test_ansible_settings_use_replica_set_1___open_release_ginkgo____opencraft_release_ginkgo__` test is failing but I change nothing in that file.

It could be that I need to rebase my branches with `master`?

instance/tests/integration/integration_instance.py
```diff
@@ -23,6 +23,7 @@
 
 import os
 import re
+import subprocess
```
mtyaka, line 26: I don't think we need this.

instance/tests/integration/integration_instance.py
```diff
@@ -489,12 +490,22 @@ def test_openstack_server_terminated(self):
...
         Test that OpenStackServer detects if the VM was terminated externally.
         """
-        server = OpenStackServer(name_prefix="integration_test")
-        server.save()
-        server.start()
-        server.sleep_until(lambda: server.status.accepts_ssh_commands, timeout=120)
-        server.os_server.delete()
-        self.assert_server_terminated(server)
+        retries = 1
+        while retries >= 0:
```
mtyaka, line 494: Can we move this code into a helper function? Call it `spawn_openstack_server` and make it take an optional `retries` parameter, which defaults to 1?

This test will then look like this:
```python
server = spawn_openstack_server()
server.os_server.delete()
self.assert_server_terminated(server)
```

instance/tests/integration/integration_instance.py
```diff
@@ -489,12 +490,22 @@ def test_openstack_server_terminated(self):
...
+            server = OpenStackServer(name_prefix="integration_test")
+            server.save()
+            server.start()
+            try:
+                server.sleep_until(lambda: server.status.accepts_ssh_commands, timeout=120)
+            except TimeoutError:
+                server.os_server.delete()
+                retries -= 1
+            else:
```
mtyaka, line 503: I'd like to take the `else` part out of the try/except block to make it clear that verifying that the server had been terminated is the main part of the test.

instance/tests/integration/integration_instance.py
```diff
@@ -489,12 +490,22 @@ def test_openstack_server_terminated(self):
...
+        retries = 1
+        while retries >= 0:
+            server = OpenStackServer(name_prefix="integration_test")
+            server.save()
+            server.start()
+            try:
+                server.sleep_until(lambda: server.status.accepts_ssh_commands, timeout=120)
+            except TimeoutError:
+                server.os_server.delete()
+                retries -= 1
```
mtyaka, line 502: If at this point all retries have already been used up, we should just re-raise the `TimeoutError`.

instance/tests/integration/integration_instance.py
```diff
@@ -489,12 +490,22 @@ def test_openstack_server_terminated(self):
...
+            else:
+                server.os_server.delete()
+                self.assert_server_terminated(server)
+                break
+
+        self.assertTrue(retries >= 0)
```
mtyaka, line 508: We don't need to assert on the number of retries since that's just a test implementation detail.

instance/tests/integration/integration_instance.py
```diff
@@ -77,6 +77,31 @@ def f_retry(*args, **kwargs):
     return f_retry  # true decorator
 
 
+def spawn_openstack_server(name_prefix, retries=1):
+    """
+    Spawn an OpenStack server.
+
+    :param name_prefix: Name prefix for OpenStack server
+    :param retries: Number of times it retry spawn a server, defaults tp 1
```
mtyaka, line 85: typo: "defaults *to* 1"

fferrin: The font of my IDE is not the most appropriate...

PR #599, opened 2020-06-29
SE-2870 - Changed Heartbeat App Addition

nizarmah:
The heartbeat app addition was done using a condition that checked if the `openedx_release` is *not* `master`. However, releases after `ironwood`, such as `juniper` will have the heartbeat app added by default. 

Therefore, the condition was changed to check for old releases instead of checking if *not* `master`. However, not all old releases are included, only the following:
* `ironwood`
* `hawthorn`
* `ginkgo`

**JIRA tickets**: SE-2870

**Merge deadline**: None

**Testing instructions**:

1. Clone `open-craft/opencraft` and checkout the branch `nizar/change_heartbeat_app_addition`
2. Following the [OCIM docs](https://ocim.opencraft.com/en/latest/installation/) to install OCIM on your machine.
3. Run OCIM and create a new instance running `open-release/juniper.1`
4. Launch a new app server and make sure it configures successfully.

There is a problem with the `integration_tests_group_1` as mentioned in SE-2872. So it will probably fail. However, we can place this into `External / Blocker` after review until SE-2872 is resolved.

**Reviewers**
- [ ] @pkulkark


instance/models/mixins/openedx_config.py
```diff
@@ -425,10 +425,14 @@ def _get_configuration_variables(self):
             forum_hb_path = "lms.lib.comment_client.utils.check_forum_heartbeat"
         template["EDXAPP_LMS_ENV_EXTRA"]["HEARTBEAT_EXTENDED_CHECKS"].append(forum_hb_path)
 
-        # master has djangoapps.heartbeat installed by default now.
+        # master and newer releases djangoapps.heartbeat installed by default now.
```
pkulkark, line 428: Nit: might be better to say `master and juniper release onwards has djangoapps.heartbeat installed by default`

pkulkark:
@nizarmah LGTM :+1: . I was able to successfully spawn a juniper instance with this change.

- [x]  I tested this: created a juniper instance and spawned appserver successfully.
- [x]  I read through the code
- [ ]  ~~I checked for accessibility issues~~ N/A

PR #600, opened 2020-06-29
[SE-2892] Fix issues with instance reference admin form and creator assignment

fferrin:
This PR:
- changes how the inlines instance references of the OpenedX instance are displayed on the admin page. Now you should see them expanded.
- allows you to save an instance without providing and `owner` or `creator`. By default, it sets these fields to the currently logged user.

**JIRA ticket**: [SE-2892](https://tasks.opencraft.com/browse/SE-2892)

**Testing**: 
- Go to an OpenedX instance on the admin page, edit it and verify that the instance reference at the bottom of the form is expanded
- Save a new Instance without providing an owner or created. You should be able to do it.

**Reviewers**
- [ ] @mtyaka


registration/models.py
```diff
@@ -371,6 +371,7 @@ def commit_changes_to_instance(self, deploy_on_commit=False, retry_attempts=2, c
         instance.name = self.instance_name
         instance.privacy_policy_url = self.privacy_policy_url
         instance.email = self.public_contact_email
+        instance.creator_id = creator
```
mtyaka, line 374: This is not the best place to do this since it will overwrite the instance creator every time a new deployment is triggered from the new UI. While I think right now only the instance creator can trigger deployments from the new UI, that might not be the case in the future.

The appropriate place to set the creator would be when the instance is first created, which happens inside [this method](https://github.com/open-craft/opencraft/blob/ebd8c3666388173174f3fdf7c289a39104f61f90/registration/provision.py#L55).

instance/models/appserver.py
```diff
@@ -173,7 +173,8 @@ class AppServer(ValidateModelMixin, TimeStampedModel):
     name = models.CharField(max_length=250, blank=False)
     server = models.OneToOneField(OpenStackServer, on_delete=models.CASCADE, related_name='+')
     # The Instance that owns this. InstanceReference has related_name accessors like 'openedxappserver_set'
-    owner = models.ForeignKey(InstanceReference, on_delete=models.CASCADE, related_name='%(class)s_set')
+    owner = models.ForeignKey(InstanceReference, null=True, blank=True, on_delete=models.CASCADE,
```
mtyaka, line 176: This change should be removed. The appserver's `owner` field is completely different from instance `owner` and should never be blank.

instance/migrations/0130_add_blank_fields_to_appserver_and_deployment.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.AlterField(
+            model_name='deployment',
+            name='creator',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='userprofile.UserProfile'),
+        ),
+        migrations.AlterField(
+            model_name='openedxappserver',
+            name='owner',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='openedxappserver_set', to='instance.InstanceReference'),
+        ),
```
mtyaka, line 23: It looks like you need to re-generate migrations ; these seem to be outdated as they do not correspond to the changes. I believe that's also why the `quality_migrations_tests` test is currently failing.

instance/admin.py
```diff
@@ -67,19 +68,9 @@ class OpenEdXInstanceAdmin(admin.ModelAdmin): # pylint: disable=missing-docstrin
...
+    inlines = (
         InlineInstanceReferenceAdmin,
-    ]
-
-    def get_inline_instances(self, request, obj=None):
-        """
-        Hides inlines while creating ``OpenEdXInstance``.
-        """
-        # Doesn't show the inline instance for new objects since we have custom
-        # logic for creating InstanceReference objects
```
mtyaka, line 79: Is this check no longer needed? It might not be required anymore, but we need to double check that removing this doesn't break anything.

registration/models.py
mtyaka: Also when we move this code to the more appropriate place, can we add a unit test to verify it works?

fferrin: Done! I didn't know the flow of execution but you are, it doesn't make sense to add it on every new deployment. I also add the test you asked for

instance/admin.py
fferrin: I think you were right about this. I misinterpreted "hide" with "collapse".

registration/provision.py
```diff
@@ -77,6 +78,9 @@ def _provision_instance(sender, **kwargs):
             privacy_policy_url=application.privacy_policy_url,
         )
         application.instance.lms_users.add(user)
+        creator_profile = UserProfile.objects.get(user=user)
```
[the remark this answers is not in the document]
fferrin: Changed!

mtyaka:
Tested, it works well! Thanks @fferrin :+1:

Please rebase on top of master & squash the commits. You can then proceed and deploy to stage.

- [x] I tested this: Verified that the instance reference form now works consistently. Verified that the creator is set automatically for beta instances.
- [x] I read through the code
- [ ] Includes documentation

PR #602, opened 2020-07-07
SE-2865 Disable the new OCIM console for staff users

mavidser:
This PR disables the new OCIM console for staff users

**JIRA tickets**: [SE-2865](https://tasks.opencraft.com/browse/SE-2865)

**Screenshots**:
Screen Shot 2020-07-07 at 12 33 39

**Testing instructions**:

- Checkout this branch.
- Enter new frontend subdirectory: `cd frontend`
- Build Ocim API client: `./scripts/build-api-client.sh`
- Reinstall dependencies: `npm install`
- Run tests - `npm test`
- Run frontend - `npm start`
- Go to `http://localhost:3000/` and login as a staff user
- You should see the above error page.
- Logout and login as a normal user - you should be able to see the OCIM console as usual.

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/frontend/packages/api_client/src/apis/V2Api.ts
+++ b/frontend/packages/api_client/src/apis/V2Api.ts
@@ -580,7 +580,7 @@ export class V2Api extends runtime.BaseAPI {
     }
 
     /**
-     * Open edX Instance Configuration API.  This API can be used to manage the configuration for Open edX instances owned by clients.
+     * Endpoint for getting the list of instances owned by the current user.  If the user is staff, return a 400 error. This is used by the frontend to prevent staff users from accidentally accessing and updating other users\' instances (see SE-2865).
      * Get all instances owned by user
      */
     async instancesOpenedxConfigListRaw(): Promise<runtime.ApiResponse<Array<OpenEdXInstanceConfig>>> {
@@ -606,7 +606,7 @@ export class V2Api extends runtime.BaseAPI {
     }
 
     /**
-     * Open edX Instance Configuration API.  This API can be used to manage the configuration for Open edX instances owned by clients.
+     * Endpoint for getting the list of instances owned by the current user.  If the user is staff, return a 400 error. This is used by the frontend to prevent staff users from accidentally accessing and updating other users\' instances (see SE-2865).
      * Get all instances owned by user
      */
     async instancesOpenedxConfigList(): Promise<Array<OpenEdXInstanceConfig>> {
--- a/frontend/src/console/components/ConsolePage/ConsolePage.tsx
+++ b/frontend/src/console/components/ConsolePage/ConsolePage.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { CustomizationSideMenu, RedeploymentToolbar } from 'console/components';
-import { EmailActivationAlertMessage } from 'ui/components';
+import { EmailActivationAlertMessage, ErrorPage } from 'ui/components';
+import { OCIM_API_BASE } from 'global/constants';
 import { Row, Col, Container } from 'react-bootstrap';
 import { WrappedMessage } from 'utils/intl';
 import { InstancesModel } from 'console/models';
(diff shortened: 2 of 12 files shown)


frontend/packages/api_client/src/apis/V2Api.ts
```diff
@@ -584,7 +584,9 @@ export class V2Api extends runtime.BaseAPI {
      * Get all instances owned by user
      */
     async instancesOpenedxConfigListRaw(): Promise<runtime.ApiResponse<Array<OpenEdXInstanceConfig>>> {
-        const queryParameters: runtime.HTTPQuery = {};
+        const queryParameters: runtime.HTTPQuery = {
+            fail_if_staff: true,
```
mtyaka, line 588: You shouldn't change files in packages/api_client directly, they are automatically build from python sources with the `scripts/update-api-client.sh` script. We should remove this from this file. Since the v2 api is currently only used from the frontend, I think we can just remove the `fail_if_staff` parameter and always fail when accessed by a staff user.

By the way while examining how the client is built, I noticed that [these two lines in the build script](https://github.com/open-craft/opencraft/blob/3707c93680532a19068fbba4014d149df8159b98/frontend/scripts/update-api-client.sh#L4-L5) are trying to remove wrong files. It should be `packages/api_client/src/apis/*.ts` and `packages/api_client/src/models/*.ts`. Could you add that change to this PR?

registration/api/v2/views.py
```diff
@@ -401,6 +401,22 @@ def static_content_overrides(self, request, pk=None):
...
             return Response(status=status.HTTP_200_OK, data=application.draft_static_content_overrides)
 
+    def list(self, request, *args, **kwargs):
+        """
+        Endpoint for getting the list of instances owned by the current user
+
+        If `fail_if_staff` is set for a staff user, the request returns a 400 error. This is used by the frontend to
+        prevent staff users from accidentally accessing and updating other users' instances.
+        """
+        if request.query_params.get("fail_if_staff", False) and request.user.is_staff:
```
mtyaka, line 411: As mention in a comment above, we can probably remove the query parameter altogether and always fail with the error if user is staff.

registration/api/v2/views.py
```diff
@@ -401,6 +401,22 @@ def static_content_overrides(self, request, pk=None):
...
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data={
+                    'details': "Not allowed for staff users"
```
mtyaka, line 415: Can we also add a code; something like `code`: 'NOT_ACCESSIBLE_TO_STAFF'` so that on the frontend we can match on the code instead of the error message string? It's not that important, but it would feel a bit cleaner to not have to rely on the exact error message.

frontend/src/console/components/ConsolePage/ConsolePage.tsx
```diff
@@ -115,6 +116,18 @@ export class ConsolePageComponent extends React.PureComponent<Props> {
       return this.props.children;
     };
 
+    if (this.props.error === 'Not allowed for staff users') {
```
mtyaka, line 119: As mentioned elsewhere, I would prefer if we could match on an error code instead of a full error message.

frontend/src/ui/components/ErrorPage/ErrorPage.tsx
```diff
@@ -4,15 +4,38 @@ import { WrappedMessage } from 'utils/intl';
...
+    <div className="error-page">
+      <div className="error-container">
+        <Row className="error-page-content">
+          <p>
+            <WrappedMessage id="somethingWrong" messages={messages} />
+          </p>
+          <WrappedMessage id="errorMessage" messages={messages} />
+        </Row>
+      </div>
     </div>
```
mtyaka, line 39: Nit: since most of the returned HTML structure is the same, can we pull the inner bits that are different into variables so that we can do something like:

```javascript
let body;
if (props.messsages && props.messageId) {
  body = <WrappedMessage id={...} ... />;
} else {
  body = <>
    <p>
      ...
    </p>
    <WrappedMessage ... />
  </>;
}

return (
  <div className="error-page">
    <div className="error-container">
      <Row className="error-page-content">
        {body}
      </Row>
    </div>
  </div>
);
```

frontend/packages/api_client/src/apis/V2Api.ts
mavidser: Ah, didn't notice the warning at the top - have removed the query param, and updated the buildscript.

mavidser:
@mtyaka - Thanks for the review - have addressed the feedback now.

mtyaka:
Nice work @mavidser! :+1:

- [x] I tested this: I verified that if I am logged in as staff and I tried to access the new frontend, I get an error with a link to the old interface. I verified that when logged in as a regular user I can access my account and configure my instance using the new frontend.
- [x] I read through the code
- [ ] Includes documentation

mavidser:
Sure @mtyaka. Can you also mark it approved in GH's UI (I prefer merging it here for better logging)?

PR #604, opened 2020-07-09
Fix Ocim integration test failures

lgp171188:
This PR
* Adds changes to only start/stop the celery workers on appserver activation/deactivation. This prevents the unit tests related to Consul from being skipped. This has to be tested and verified on stage.
* Fixes the consul configuration which was breaking unit and integration tests - enables consul encryption for the integration tests and the cleanup job and runs consul in dev mode for everything else.
* Increases the retries and the delay between retries when checking whether the integration test instance is up or not. Also adds logging to show the retries.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -169,11 +169,16 @@ jobs:
       - run:
           name: Run Consul
           command: |
-            echo Running consul agent to connect to the integration consul server
             sudo wget -P /tmp https://releases.hashicorp.com/consul/1.2.1/consul_1.2.1_linux_amd64.zip
             sudo unzip /tmp/consul_1.2.1_linux_amd64.zip -d /usr/local/bin
             mkdir -p /tmp/consul-data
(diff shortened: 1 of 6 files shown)


nizarmah:
@lgp171188 I was testing the changes on Stage. Sorry it took so long, I was launching a new app server in order to activate and deactivate the app servers to test your change.

It seems like _deactivating_ an app server isn't working.

lgp171188:
> It seems like deactivating an app server isn't working.

@nizarmah, can you provide more details on what you mean by "not working"? The deactivation process involves running an ansible playbook on the deactivated appserver and it is expected to take a few minutes (3-5 minutes maximum) to complete.

I tested this on a stage instance and could see that the `edxapp_workers:*` services got stopped one by one without any issues.

nizarmah:
> @nizarmah, can you provide more details on what you mean by "not working"? 

I was about to when it ended up getting deactivated. Everything seems to be good. I'm just rechecking it again. Sorry about that. 
I left it for a few minutes, but maybe I should've given it more time.

PR #605, opened 2020-07-10
SE-2868 Prevent users from spawning many appservers in parallel from the new UI

mavidser:
This PR disables the new OCIM console for staff users

**JIRA tickets**: [SE-2868](https://tasks.opencraft.com/browse/SE-2868)

**Testing instructions**:

- Checkout this branch.
- Add `DEPLOYMENT_START_DELAY=600` to your .env (~10 mins is also my recommendation for prod)
- Start the webserver and huey workers `make run.dev`
- Start the ocim console - `cd frontend && ./scripts/build-api-client.sh && npm install && npm start`
- On the ocim console, create a new user, or login with an existing one.
- Do a couple of deployments in 10 mins by doing changes on the instances and clicking 'Deploy' a few times.
- Observe huey worker logs - only the last deployment should have started, and the rest should have no-oped by logging 'cancelled'.
- Start a deployment after 10 mins (when an appserver has started provisioning). The provisioning should get cancelled, and a new deployment should be scheduled. 
- Verify instances on [OVH](https://horizon.cloud.ovh.net/project/instances/) too to see that cancelled appservers get deleted when new ones are provisioned.
- Run all tests - `make test`

**Reviewers**
- [ ] @mtyaka


instance/models/openedx_appserver.py
```diff
@@ -576,6 +578,15 @@ def accepts_ssh_commands():
             # Provisioning (ansible)
             self.logger.info('Provisioning server...')
             self._status_to_configuring_server()
+
+            # Check if deployment was cancelled before starting the playbooks.
+            # Terminating here in in case any previous terminations were no-op due to VM being in a pre-ready state.
+            if self.deployment:
+                deployment = OpenEdXDeployment.objects.get(pk=self.deployment_id)
```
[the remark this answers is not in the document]
mavidser: I'm considering making it `if self.deployment_id` in the next pass. Not using `self.deployment` as I noticed that in some calls, the variable was an instance of 'Deployment', while being 'OpenEdxDeployment' in others. I decided not to investigate this discrepancy, and just import the intended object. Can do a type check instead if that'd be the preferred approach.

instance/utils.py
```diff
@@ -218,3 +219,44 @@ def choices(cls):
...
+        if deployment.status() in (DeploymentState.changes_pending, DeploymentState.provisioning):
+            deployment.cancel_deployment()
+    deployment = OpenEdXDeployment.objects.create(
+        instance_id=instance.ref.id,
+        creator=creator_profile,
+        type=deployment_type,
+        changes=changes,
+    )
+    # We schedule a deployment instead of starting right now, which gives user a few mins to update the instance
+    # without us having to setup and teardown the appserver.
```
mtyaka, line 257: Can we add some logging here so that the log is visible under the instance "Log" tab in the "old" manage interface?

instance/utils.py
```diff
@@ -218,3 +219,44 @@ def choices(cls):
...
+    start_deployment.schedule(
+        args=(instance.ref.id, deployment.id),
+        kwargs=kwargs,
+        delay=settings.DEPLOYMENT_START_DELAY
```
mtyaka, line 261: Would it make sense to only use the delay when deploying from the new end-user interface?

I think it could be confusing for admins if they click the "Launch new appserver" button and nothing seems to happen when using the manage.opencraft.com interface (we'll have to update that interface to show deployments rather than appservers at some point).

For deployments triggered by PR watchers and periodic builds the delay probably also isn't neccessary.

mavidser: Good catch - fixed. Adding delay only for console-triggered deployments

instance/utils.py
```diff
@@ -218,3 +219,44 @@ def choices(cls):
...
+    from instance.models.openedx_deployment import DeploymentState, OpenEdXDeployment
+    from instance.tasks import start_deployment
+    from userprofile.models import UserProfile
+
+    changes = None
+    if instance.betatestapplication_set.exists():
+        changes = build_instance_config_diff(instance.betatestapplication_set.get())
+
+    deployment_type = deployment_type or DeploymentType.unknown
+    creator_profile = creator and UserProfile.objects.get(user_id=creator)
```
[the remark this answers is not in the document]
mavidser: True, updated.

PR #606, opened 2020-07-10
[SE-2865] Update URL for old OCIM

mavidser:
See #602 for complete details

**JIRA tickets**: [SE-2865](https://tasks.opencraft.com/browse/SE-2865)

This is a follow-up to that PR, where the URL has been updated from `/` to `/instance`, as the former one redirects back to the new console on stage

**Reviewers**: @mtyaka


mtyaka:
Thanks @mavidser :+1:

- [x] I tested this: I verified that while stage.manage.opencraft.com redirects to the new UI, stage.manage.opencraft.com/instance doesn't. I didn't actually test the change in Ocim locally because I'm currently using my spare laptop where I don't have the Ocim devstack set up, and the change is trivial.
- [x] I read through the code
- [ ] Includes documentation

PR #608, opened 2020-07-11
[SE-2869] Forbid registering Ocim custom domains that conflict with current or past clients' domains

fferrin:
This PR prevents users from using invalid or conflicting domains

**JIRA tickets**: [SE-2869](https://tasks.opencraft.com/browse/SE-2869)

**Testing instructions**:

This PR already provides tests, but for manual testing you can try:
- Using ``opencraft.hosting`` as a domain
- Using as domain a subdomain from an instance. For example, using ``test.domain.com`` when ``domain.com`` already exists
- Using a domain such that, when creating ``studio``, ``ecommerce`` or ``studio`` subdomains, it overrides an existing domain. For example, try using ``domain.com`` when ``studio.domain.com`` was used as a domain.

**Reviewers**
- [ ]  @swalladge

### DIFF
--- a/registration/api/v2/serializers.py
+++ b/registration/api/v2/serializers.py
@@ -25,6 +25,7 @@ from typing import Dict
 from django.contrib.auth.models import User
 from django.contrib.auth.password_validation import validate_password
 from django.core.validators import RegexValidator
+from django.db.models import Q
 from django.utils import timezone
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


registration/api/v2/serializers.py
```diff
@@ -315,6 +316,37 @@ def validate_user(self, value):
...
+    #         - Domain in use or such that it is a subdomain of an existing domain
+    #         - Domain such that, when studio, ecommerce and preview subdomains are created, they
+    #         override a domain in use
+    #     """
+    #     is_opencraft_domain = BetaTestApplication.BASE_DOMAIN == value
+    #     if is_opencraft_domain:
+    #         raise ValidationError("Can not use OpenCraft domain.")
+    #
+    #     _, _, domain = value.partition('.')
+    #     is_domain_in_use = BetaTestApplication.objects.filter(external_domain=domain).exists()
```
swalladge, line 334: We need to check all OpenEdxInstances, not just the betatest applications - ie. all these fields:

 https://github.com/open-craft/opencraft/blob/cd0595d19effdb1b14cd5aaec6c3f22b6c11a50a/instance/models/mixins/domain_names.py#L83-L88

PR #612, opened 2020-07-15
[SE-2948] Remove redirect to preliminary page

mavidser:
When no appservers are active, instance LMS and Studio URLs should not redirect the user, but just display contents of default.opencraft.com. Have applied the patch manually to the below instance on stage too, for testing purposes.

**JIRA tickets**: [SE-2948](https://tasks.opencraft.com/browse/SE-2948)

**Sandbox URL**: https://sid-test-2.stage.opencraft.hosting (The DNS has 3 IP addresses, so you might see the old deployment on IPs other than 54.36.101.120)

**Testing instructions**:

1. Run `curl --insecure -v --header "Host:se-2870-test.opencraft.hosting" https://54.36.101.120` and verify 301 Moved Permanently response.
2. Run `curl --insecure -v --header "Host:sid-test-2.stage.opencraft.hosting" https://54.36.101.120` and verify 200 OK is returned, with content of default.opencraft.com
3. Verify that `backend be-edxins-sid-test-2sta-567` section of `/etc/haproxy/haproxy.cfg` on stage haproxy has the updated code.

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/instance/models/mixins/load_balanced.py
+++ b/instance/models/mixins/load_balanced.py
@@ -108,7 +108,11 @@ class LoadBalancedInstance(models.Model):
             return [], []
         self.logger.info("Configuring load balancer to point to the preliminary page.")
         backend_name = "be-preliminary-page-{}".format(primary_key)
-        config = "    server preliminary-page {}:80".format(settings.PRELIMINARY_PAGE_SERVER_IP)
+        ca_file_path = "/etc/ssl/certs/ca-certificates.crt"
+        config = "    server preliminary-page {}:443 ssl verify required ca-file {}".format(
+            settings.PRELIMINARY_PAGE_SERVER_IP,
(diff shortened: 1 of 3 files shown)


mtyaka:
Thanks for this @mavidser! You'll have to update the failing test, then we can merge this. :+1:

- [x] I tested this: I didn't manage to actually test this because the integration load balancer that I'm using locally doesn't have wasn't set up with the old load-balancer-ansible role (it's using the new load-balancer-v2 role from ansible-playbooks instead). The change is trivial enough and covered by a unit test that I don't have to test this manually. I created [SE-3012|https://tasks.opencraft.com/browse/SE-3012] to look into how we could move the devstack and integration to the new v2 role.
- [x] I read through the code
- [ ] Includes documentation

PR #618, opened 2020-07-28
SE-3057: Address several problems found following Vagrant install instructions.

Kelketek:
[SE-3057](https://tasks.opencraft.com/browse/SE-3057)

* Clarified Python version requirements.
* Removed redundant instruction for HUEY_ALWAYS_EAGER environment variable, now handled by Ansible.
* Added missing instruction to perform migrations as part of initial setup.
* Added override_settings to tests which assume that AWS keys are set, which they are not by default.

@symbolist I did end up creating the above ticket since while getting my devstack up and running I found enough problems to feel like an issue should be created, but as time went on I was able to fix all of them and put them into a PR. I'm not sure if this means I should just attach it to one of my existing assigments instead, or if I should pull in the above ticket into the sprint, or, perhaps, it should be put into a future sprint with the pre-assigned status 'In Review', and assigned a reviewer later.

Please let me know your thoughts, and if there's anyone else we should have review this ticket.


nizarmah (approved):
:+1:

- [x] I tested this:
  - Set up Ocim locally
  - Attempt to Launch new App Server with a User without UserProfile
  - Checked CircleCI builds to validate tests are passing
  - Launched Docs and made sure changes don't break anything and have proper format
- [x] I read through the code
- [x] Includes documentation

PR #623, opened 2020-08-04
Use NRQL conditions instead of Synthetics.

mtyaka:
NRQL makes alerting much more flexible.

The main change for now is that alerts are triggered only after any of the Synthetic monitors fails three times in a row to avoid alerts due to temporary networks blips.

**Test instructions**

1. If you don't already have `ADMINS` setting set in your local `.env` file, set it to your email (ie. `ADMINS='[["Matjaz Admin", "matjaz@opencraft.com"]]'`.
1. Set `NEWRELIC_ADMIN_USER_API_KEY` and `NEWRELIC_LICENSE_KEY` in your local `.env` file. You can find them on the production Ocim VM's `.env` file.
1. Start the django shell (`make shell`).
1. Get a reference to an `OpenEdXInstace` (or create a new one if you don't have an existing instance in your devstack).
1. Run `instance.enable_monitoring()`. Make sure that there are no errors.
1. Log into NewRelic and go to 'Alerts & AI -> Policies'. Search by the URL of your test instance and verify that the policy alerts look correct.
1. Run `instance.disable_monitoring()`. 
1. Verify that the policy for your instance has been removed from NewRelic.
1. Don't forget to remove `NEWRELIC_ADMIN_USER_API_KEY` and `NEWRELIC_LICENSE_KEY` from your local `.env` file!

### DIFF
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -768,6 +768,9 @@ NEWRELIC_ADMIN_USER_API_KEY = env('NEWRELIC_ADMIN_USER_API_KEY', default=None)
 # The basic auth password needed to access the node exporter.
 NODE_EXPORTER_PASSWORD = env('NODE_EXPORTER_PASSWORD', default=None)
 
+# Thresholds for NRQL alert conditions
+NEWRELIC_NRQL_ALERT_CONDITION_DURATION = env('NEWRELIC_NRQL_ALERT_CONDITION_DURATION', default='16')
+
 # Load balancing ##############################################################
 
 # The load-balancing server given in the form ssh_username@server.domain will be created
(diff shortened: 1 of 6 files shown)


opencraft/settings.py
```diff
@@ -764,6 +764,9 @@
 # The basic auth password needed to access the node exporter.
 NODE_EXPORTER_PASSWORD = env('NODE_EXPORTER_PASSWORD', default=None)
 
+# Thresholds for NRQL alert conditions
+NEWRELIC_NRQL_ALERT_CONDITION_DURATION = env('NEWRELIC_NRQL_ALERT_CONDITION_DURATION', default='16')
```
giovannicimolin, line 768: Nice idea of setting the threshold using a setting, we can easily adjust it after some testing.

mtyaka:
Thanks @giovannicimolin!

PR #626, opened 2020-08-07
[BB-2558] Temporarily set default `FORUM_MONGO_AUTH_MECH` to `MONGODB-CR`

Agrendalath:
This is a temporary change that sets the default authentication mechanism to MONGODB-CR. It will be reverted once we upgrade to [SCRAM](https://docs.mongodb.com/manual/core/security-scram/).

This option is ignored in Ironwood, so it affects only Juniper instances, which have it set to SCRAM by default.

This does not include testing the fix itself, because we're already using it, cf. [relevant discussion](https://discuss.openedx.org/t/discussion-forum-errors-in-latest-master-with-mongo-3-2/1843/7) and [example instance](https://manage.opencraft.com/instance/19980/edx-appserver/13102/).

## Testing instructions
1. Check that [the test AppServer](https://stage.manage.opencraft.com/instance/5450/edx-appserver/2077/) contains `FORUM_MONGO_AUTH_MECH: :mongodb_cr` in its `Configuration`.

## Manual testing instructions
1. Check the branch out on stage, [create an instance](https://ocim.opencraft.com/en/latest/provisioning-sandboxes/#manual-provisioning) and spawn new AppServer.
1. In Ocim go to `AppServer -> Configuration` and confirm that `FORUM_MONGO_AUTH_MECH: :mongodb_cr` is present there.

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -131,6 +131,8 @@ def production_instance_factory(**kwargs):
         "demo_test_users": [],
         # Disable certificates process to reduce load on RabbitMQ and MySQL
         "SANDBOX_ENABLE_CERTIFICATES": False,
+        # HACK: This is a temporary change for BB-2558. It will be removed after upgrading MongoDB to SCRAM.
+        "FORUM_MONGO_AUTH_MECH": ":mongodb_cr",
     }
     configuration_extra_settings = kwargs.pop("configuration_extra_settings", "")
     if configuration_extra_settings:
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -87,8 +87,26 @@ def instance_factory(**kwargs):
     # Ensure caller provided required arguments
     assert "sub_domain" in kwargs
 
+    # HACK: This is a temporary change for BB-2558. It will be removed after upgrading MongoDB to SCRAM.
+    sandbox_settings = {
+        "FORUM_MONGO_AUTH_MECH": ":mongodb_cr",
+    }
+    configuration_extra_settings = kwargs.pop("configuration_extra_settings", "")
+    if configuration_extra_settings:
+        configuration_extra_settings = yaml.load(configuration_extra_settings, Loader=yaml.SafeLoader)
+    else:
+        configuration_extra_settings = {}
+    extra_settings = yaml.dump(
+        ansible.dict_merge(sandbox_settings, configuration_extra_settings),
+        default_flow_style=False
+    )
+    instance_kwargs = dict(
+        configuration_extra_settings=extra_settings,
+    )
+    instance_kwargs.update(kwargs)
+
     # Create instance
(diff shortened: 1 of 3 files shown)


instance/factories.py
```diff
@@ -87,8 +87,26 @@ def instance_factory(**kwargs):
...
+        default_flow_style=False
+    )
+    instance_kwargs = dict(
+        configuration_extra_settings=extra_settings,
+    )
+    instance_kwargs.update(kwargs)
+
     # Create instance
-    instance = OpenEdXInstance.objects.create(**kwargs)
+    instance = OpenEdXInstance.objects.create(**instance_kwargs)
```
Agrendalath, line 109: This is just a copy-paste from `production_instance_factory` below.

instance/factories.py
```diff
@@ -87,8 +87,26 @@ def instance_factory(**kwargs):
...
+    else:
+        configuration_extra_settings = {}
+    extra_settings = yaml.dump(
+        ansible.dict_merge(sandbox_settings, configuration_extra_settings),
+        default_flow_style=False
+    )
+    instance_kwargs = dict(
+        configuration_extra_settings=extra_settings,
+    )
+    instance_kwargs.update(kwargs)
```
xitij2000, line 106: Wouldn't it make more sense to add this to instance.models.mixins.openedx_database:OpenEdXDatabaseMixin._get_mongo_settings ? Otherwise, this will add a permanent configuraiton param to the extra settings that will need to be removed once we upgrade.

instance/factories.py
```diff
@@ -131,6 +149,8 @@ def production_instance_factory(**kwargs):
         "demo_test_users": [],
         # Disable certificates process to reduce load on RabbitMQ and MySQL
         "SANDBOX_ENABLE_CERTIFICATES": False,
+        # HACK: This is a temporary change for BB-2558. It will be removed after upgrading MongoDB to SCRAM.
+        "FORUM_MONGO_AUTH_MECH": ":mongodb_cr",
```
lgp171188: @Agrendalath, @xitij2000, this particular change has to be merged before we start doing the Ocim Juniper upgrades in the next sprint (CC @pomegranited ) as the forum errors out without this being set on the Juniper instances.

instance/factories.py
Agrendalath: Right, that's a good point! I changed this.

instance/factories.py
Agrendalath: @lgp171188, @pomegranited described this really nicely [here](https://tasks.opencraft.com/browse/BB-2558?focusedCommentId=160445&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-160445).
That's why I asked for FF review here.

xitij2000:
@Agrendalath  Sorry for the delay! I thought I had already approved it, but it seems I never submitted it.

PR #628, opened 2020-08-10
SE-3098: Use multiple threads per worker, remove deployment monitoring task.

Kelketek:
This pull request enables multiple threads per worker. This should solve the immediate problem of the workers being blocked without increasing memory usage or CPU in any significant sense, as all threads can share current process memory and our workloads are almost entirely IO-based. This takes us from an effective worker count of 3 to a count of 18.

To test this PR: Start up your local dev stack and try to provision a server. If the server provisions, it works. For contrast, if you were to do this on master, it would hang indefinitely as the single process would only have one thread, and a minimum of two workers is needed for provisioning.

A second test would be to verify that a new multi-appserver deployment runs smoothly and deactivates the old servers.

This PR also removes the `reconfigure_load_balancers` task, which was requested to be removed by @lgp171188 due to obsolescence.

This is a part 1 fix that will solve the immediate problem of not having enough workers to process the provisioning of servers. A set of follow-up PRs are coming which will handle the implementation of Huey priorities to further clear the technical debt surrounding this problem.


Procfile
```diff
@@ -1,5 +1,5 @@
 web: gunicorn opencraft.wsgi --timeout 60 --workers 4 --log-file -
 websocket: daphne -p 2001 opencraft.asgi:application
-worker: python3 manage.py run_huey --no-periodic
+worker: python3 manage.py run_huey --no-periodic --workers=6
```
lgp171188, line 3: @Kelketek, instead of hardcoding this, can we make this configurable?

Procfile.dev
```diff
@@ -1,3 +1,3 @@
 web: python3 manage.py runserver 0.0.0.0:5000
-worker: python3 manage.py run_huey --no-periodic
+worker: python3 manage.py run_huey --no-periodic --workers=6
```
lgp171188, line 2: @Kelketek, same as [this comment](https://github.com/open-craft/opencraft/pull/628/files#r469852862). In fact, this may not be needed in a dev environment at all.

instance/management/commands/migrate_ephemeral.py
```diff
@@ -59,11 +59,14 @@ def handle(self, *args, **options):
         for ref in refs:
             # Disable ephemeral support and spawn a new AppServer
             LOG.info('Migrating %s to use persistent databases', ref.instance.name)
+            old_server_ids = list(ref.instance.appserver_set.filter(_is_active=True).values_list('id', flat=True))
```
lgp171188, line 62: @Kelketek, how are the changes in this file relevant to this PR? Also @viadanna (who I remember working on this a long time ago), can we remove this management command and all the related code if we don't need it any more?

instance/models/openedx_instance.py
```diff
@@ -374,7 +374,6 @@ def spawn_appserver(self,
             app_server.make_active()
 
         appserver_spawned.send(sender=self.__class__, instance=self, appserver=app_server)
-
```
lgp171188, line 377: @Kelketek, nit: unnecessary change. If there isn't a reason for changing this, can you revert it?

instance/tasks.py
```diff
@@ -145,37 +169,34 @@ def spawn_appserver(
...
     """
-    appserver = OpenEdXAppServer.objects.get(pk=appserver_id)
+    appserver = OpenEdXAppServer.objects.get(id=appserver_id)
     if not appserver.server.status.is_healthy_state:
         appserver.server.update_status()
         if not appserver.server.status.is_healthy_state:
             logger.info('Not %s %s: ID=%s since the server is not Ready.', "Activating" if active else "Deactivating",
                         appserver, appserver_id)
             return False
-
```
lgp171188, line 170: @fox, nit: I'd recommend leaving this newline intact as it improves readability.

instance/tasks.py
```diff
@@ -78,45 +76,71 @@ def start_deployment(
...
+        old_server_ids: List[int] = None,
+        target_count: int = None,
+):
+    """
+    Called in pipeline with make_appserver_active when the current deployment is meant to deactivate old servers
+    upon success. Since any particular server could be the last one, this checks to make sure that the correct number
+    of servers which are new are now available, and that the current deployment hasn't been cancelled.
 
-    # For redundancy - if app server deployment was successful even after cancelling, we prevent the change in
-    # active appserver here
```
lgp171188, line 96: @fox, should this comment be moved to above the

```python
if deployment.cancelled:
``` 
check?

instance/tasks.py
```diff
@@ -78,45 +76,71 @@ def start_deployment(
...
+    If this is the case, it deactivates all the old app servers.
+    """
+    if not result:
+        # This function is chained after mark_appserver_active. If that function failed, we already know things aren't
+        # OK and bail out.
+        return
```
lgp171188, line 115: @fox, though this scenario will be obvious because the previous task failed, it will still be useful to log a message that states what is happening.

instance/tasks.py
```diff
@@ -78,45 +76,71 @@ def start_deployment(
...
-        return False
+        return
 
+    instance = OpenEdXInstance.objects.get(ref_set__pk=instance_ref_id)
     # If this deployment is to be marked active on success, others should be deactivated automatically
-    if mark_active_on_success:
-        other_appservers = instance.appserver_set.filter(_is_active=True).exclude(pk__in=appserver_ids)
+    current_count = instance.appserver_set.exclude(id__in=old_server_ids).filter(_is_active=True).count()
+    all_ready = current_count == target_count
+    if all_ready:
```
lgp171188, line 126: @fox, it will be useful to log a message when this case isn't true, for example when 1 of the 2 expected appservers have been spawned,

instance/tasks.py
```diff
@@ -145,37 +169,34 @@ def spawn_appserver(
     )
 
     if appserver and mark_active_on_success:
-        make_appserver_active(appserver, active=True, deactivate_others=deactivate_old_appservers)
-
-    # Huey doesn't seem to resolve properly if the result of a task is so return False
```
lgp171188, line 150: @fox, is this still an issue? If yes, we should return `False`.

lgp171188:
> This pull request enables multiple threads per worker.

@fox, can we update the log message format to include the thread id if it is available? With that it is easier to identify and filter log messages by thread.

opencraft/settings.py
```diff
@@ -296,7 +296,7 @@
     'immediate': env.bool('HUEY_ALWAYS_EAGER', default=False),
 
     # Options to pass into the consumer when running ``manage.py run_huey``
-    'consumer': {'workers': 1, 'loglevel': logging.INFO},
+    'consumer': {'workers': 1, 'loglevel': logging.INFO, 'worker_type': 'thread'},
```
lgp171188, line 299: @fox, can we also make this configurable?

instance/management/commands/migrate_ephemeral.py
Kelketek: I'm not familiar with why this management command exists, but I do know how my changes affected what the function calls actually do, so when I found this file, which referenced the functions I changed, I knew this code would crash if not updated, and I updated it.

Happy to remove the command altogether while I'm here if it's no longer needed, though.

instance/tasks.py
Kelketek: We're no longer running this task in a map() function and doing a result check, so it shouldn't be a problem.

Procfile.dev
Kelketek: > > This pull request enables multiple threads per worker.
> 
> @fox, can we update the log message format to include the thread id if it is available? With that it is easier to identify and filter log messages by thread.

```
13:25:04 worker.1              | [2020-08-13 13:25:04,569] INFO:instance.models.load_balancer:Worker-5:load_balancer=1 (haproxy-dev.ope) | PLAY [all] *****************************************************************************************************************************
```

While Honcho considers this worker 1, Huey considers it as Worker-5 (as this task was executed in the fifth thread), and marks that in the logs.

> @Kelketek, same as this comment. In fact, this may not be needed in a dev environment at all.

Perhaps not, after this change, but it'd probably be good to have more than one, if someone wants to test multi-appserver deployment locally, otherwise they'll be stuck with serial provisioning. While local deployment may not be working yet, that is a target of my other task.

> @Kelketek, instead of hardcoding this, can we make this configurable?

I should be able to find a way to do this. Will update once I've got it. :)

opencraft/settings.py
Kelketek: This is configurable-- you can always override when calling the management command by changing the args (we do this for the worker count in the procfile, though THAT should be made configurable as you've noted).

In fact this code change has no impact, it's just there to make it more explicit which worker type is being chosen, and to make sure if upstream changes the default, our default is not changed.

Kelketek: Looking closer at the way we're invoking Huey in the procfile, and the way we do not access environment variables there, perhaps we SHOULD actually make it configurable here. I'll add that now.

Kelketek:
@lgp171188 @mavidser Updated to address notes.

lgp171188:
@Kelketek, I noticed this error in the output of `integration_tests_group_1` (which runs the only test in `integration/tests/integration_instance.py` that actually spawns an appserver) and this appears to be from the code you have modified in this PR. Can you check and fix this issue?

```
13/Aug/2020 19:53:14 |     INFO | process=2268  | huey                      | Executing instance.tasks.check_deactivation: cfe88c52-da74-41b9-8103-fdc29ab2a9dc
13/Aug/2020 19:53:14 |    ERROR | process=2268  | huey                      | Unhandled exception in task cfe88c52-da74-41b9-8103-fdc29ab2a9dc.
  File "/home/********/project/venv/lib/python3.6/site-packages/huey/api.py", line 342, in _execute
    task_value = task.execute()
  File "/home/********/project/venv/lib/python3.6/site-packages/huey/api.py", line 690, in execute
    return func(*args, **kwargs)
  File "/home/********/project/venv/lib/python3.6/site-packages/huey/contrib/djhuey/__init__.py", line 128, in inner
    return fn(*args, **kwargs)
  File "/home/********/project/instance/tasks.py", line 123, in check_deactivation
    deployment = OpenEdXDeployment.objects.get(pk=deployment_id)
  File "/home/********/project/venv/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/********/project/venv/lib/python3.6/site-packages/django/db/models/query.py", line 408, in get
    self.model._meta.object_name
instance.models.openedx_deployment.OpenEdXDeployment.DoesNotExist: OpenEdXDeployment matching query does not exist.
```

Kelketek:
@lgp171188 Sure! I'm pretty sure I know what the cause is-- some tests did not previously require real database models to be used because they all stayed within one mocked task. I should be able to do something similar to this test that I did to the others.

Kelketek:
@lgp171188 This should now be resolved. The branch has also been updated so that it can be merged in as soon as you have verified all is well and give it the green light.

instance/tasks.py
```diff
@@ -76,47 +74,81 @@ def start_deployment(
...
 
+    # For redundancy - if app server deployment was successful even after cancelling, we prevent the change in
+    # active appserver here
     if deployment.cancelled:
         logger.info('Deployment %s was cancelled, returning.', deployment.id)
-        return False
+        return
 
     logger.info('Spawning servers for deployment %s [%s]', deployment, deployment.id)
+    old_server_ids = list(instance.appserver_set.filter(_is_active=True).values_list('id', flat=True))
```
lgp171188, line 84: @Kelketek, since `old_server_ids` is only used by the `check_deactivation()` method and it already has access to the `instance_ref_id`, can't we avoid getting it here and passing it two levels down?

instance/tasks.py
```diff
@@ -76,47 +74,81 @@ def start_deployment(
...
+        # NOTE: Though this does provision multiple servers in parallel, issues have been known to occur
+        # for deploying multiple instances at once in the case of outstanding migrations.
+        spawn_appserver(
+            instance_ref_id=instance_ref_id,
+            mark_active_on_success=mark_active_on_success,
+            num_attempts=num_attempts,
+            success_tag=success_tag,
+            failure_tag=failure_tag,
+            deployment_id=deployment.id,
+            old_server_ids=old_server_ids,
```
lgp171188, line 96: See [this comment](https://github.com/open-craft/opencraft/pull/628/files#r470525805)

instance/tasks.py
```diff
@@ -76,47 +74,81 @@ def start_deployment(
...
+        )
+
 
-    appserver_ids = appserver_spawn_tasks.get(blocking=True)
+@db_task()
+def check_deactivation(
+        result: bool,
+        instance_ref_id: int = None,
+        deployment_id: int = None,
+        old_server_ids: List[int] = None,
```
lgp171188, line 106: See [this comment](https://github.com/open-craft/opencraft/pull/628/files#r470525805)

instance/tasks.py
```diff
@@ -76,47 +74,81 @@ def start_deployment(
     logger.info('Retrieving deployment: ID=%s', instance_ref_id)
     deployment = OpenEdXDeployment.objects.get(pk=deployment_id)
 
+    # For redundancy - if app server deployment was successful even after cancelling, we prevent the change in
+    # active appserver here
```
lgp171188, line 78: @Kelketek, in [this comment](https://github.com/open-craft/opencraft/pull/628/files#r469862596) I was referring to moving this comment to the condition in the `check_deactivation()` function where this is relevant.

Here there is no change to the active appserver because the deployment has not yet been started.

Kelketek: The reason for grabbing this data during deployment was to avoid any chance of race conditions changing what's there. While it's unlikely to happen, it is possible for someone to change how many app servers are intended to be in an instance before one of the servers has been created. It's also possible for one server to not have started while another one has activated. This could cause data inconsistencies, so I just made the checks all at the beginning.

In unit tests, this problem will almost certainly happen because Huey is always eager, so one server will certainly activate before the rest and the old IDs will  be out of date.

lgp171188: @Kelketek, that makes sense. I will revert my changes to the PR and we can go ahead with the merge and deployment after the checks pass.

PR #629, opened 2020-08-13
SE-3056 Update marketing links

s0b0lev:
This PR is about updating marketing links for edx instances.

### 1. Blog link removed
It is enough to update ENV variable `MKTG_URL_OVERRIDES` to have empty 'BLOG' key.

### 2. Contact page url changed
It require to add such value `"CONTACT_US_CUSTOM_LINK":"/contact"` to the `SiteConfiguration` ([related edx-platform code here](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/branding/api.py#L285)) 

#### Issue
When OpenEdXInstance create full configuration object, it use ansible.dit_merge function to merge all configs. 

[This function ignore list and just replace it](https://github.com/open-craft/opencraft/blob/master/instance/ansible.py#L78) without going through list items and copying one by one.

So if instance.configuration_extra_settings doesn't have  `EDXAPP_SITE_CONFIGURATION` [hardcoded value](https://github.com/open-craft/opencraft/pull/629/files#diff-acb4f1b5a942f182074cf3b6e59a66e3R41) will works as expected.

But if instance.configuration_extra_settings have `EDXAPP_SITE_CONFIGURATION` it will completely replace list, and we need manually set `CONTACT_US_CUSTOM_LINK` to point to the proper page.

**Example configuration**
Default SiteConfiguration values.
Also here we don't have site_id, wasn't able to test if it will work, need to run it on staging (or at least on configured OCIM).
```
'EDXAPP_SITE_CONFIGURATION': [
    {
        'values': {
            'CONTACT_US_CUSTOM_LINK': '/contact'
        }
    }
]
```

`Instance.configuration_extra_settings` have `EDXAPP_SITE_CONFIGURATION`

```
'EDXAPP_SITE_CONFIGURATION': [
     {
        'site_id': 1,
        'values': {
            'ACTIVATION_EMAIL_SUPPORT_LINK': '',
            'CONTACT_US_ENABLE': False
        }
    }
]
```

## Testing instruction
- Deploy new appServer
- Check that footer links were changed as expected

## Author notes and concerns
1. Contact page will be changed for all instances, which doesn't have EDXAPP_SITE_CONFIGURATION in Instance.configuration_extra_settings
2. If Instance.configuration_extra_settings have EDXAPP_SITE_CONFIGURATION we need to add `"CONTACT_US_CUSTOM_LINK":"/contact"` manually
3. I was thinking to update [ansible.dict_merge()](https://github.com/open-craft/opencraft/blob/master/instance/ansible.py#L78) function to properly merge list items, but it is dangerous to do, because we don't know which list_1 item should be merged with list_2 item.
4. Not sure if we need site_id for default object (need to test it somewhere):
```
'EDXAPP_SITE_CONFIGURATION': [
    {
        # 'site_id': 1, by default? or it will be populated?
        'values': {
            'CONTACT_US_CUSTOM_LINK': '/contact'
        }
    }
]
```

## Reviewers:
- [ ] @mavidser

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -45,7 +45,6 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                 "8.8.8.8",
                 "8.8.4.4"
             ],
-
             # HTTP authentication
             "COMMON_ENABLE_BASIC_AUTH": True,
             "COMMON_HTPASSWD_USER": self.instance.http_auth_user,
@@ -77,8 +76,8 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                     # because we also add other extended checks
                     "openedx.core.djangoapps.heartbeat.default_checks.check_celery",
                 ],
+                # rewrite marketing links:
                 "MKTG_URL_OVERRIDES": {
-                    # rewrite marketing links:
                     # remove blog link - not supported yet
                     "BLOG": "",
                     # use different contact page
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -436,11 +436,18 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                 ],
             })
 
+        if "MKTG_URL_OVERRIDES" not in template["EDXAPP_LMS_ENV_EXTRA"]:
+            template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"] = {}
+
+        # rewrite marketing links:
+        # remove blog link - not supported yet
+        template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"]["BLOG"] = ""
+        # use different contact page
+        template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"]["CONTACT"] = "/contact"
+
         if self.privacy_policy_url:
             # Custom Privacy Policy
-            template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"] = {
-                "PRIVACY": self.privacy_policy_url,
-            }
+            template["EDXAPP_LMS_ENV_EXTRA"]["MKTG_URL_OVERRIDES"]["PRIVACY"] = self.privacy_policy_url
 
         # The dotted import path to the forum heartbeat function has changed in Juniper.
         # So use the old path only for the older releases.
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -77,6 +77,13 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                     # because we also add other extended checks
                     "openedx.core.djangoapps.heartbeat.default_checks.check_celery",
                 ],
+                "MKTG_URL_OVERRIDES": {
+                    # rewrite marketing links:
+                    # remove blog link - not supported yet
+                    "BLOG": "",
+                    # use different contact page
+                    "CONTACT": "/contact",
+                }
             },
 
             "EDXAPP_LMS_NGINX_PORT": 80,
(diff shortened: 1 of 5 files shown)


mavidser:
@s0b0lev 

> So if instance.configuration_extra_settings doesn't have EDXAPP_SITE_CONFIGURATION hardcoded value will works as expected.

This would pose a problem as instances often override EDXAPP_SITE_CONFIGURATION ([example](https://stage.manage.opencraft.com/instance/5290/edx-appserver/2083/)).

I think a better way would be to not use EDXAPP_SITE_CONFIGURATION and set up something like EDXAPP_CONTACT_US_CUSTOM_LINK and map that in the [opencraft/configuration](http://github.com/open-craft/configuration) repo.

cc @mtyaka any thoughts on this? Does this introduce additional challenges when applying changes to existing instances?

s0b0lev:
@mavidser sounds like a good plan :+1:

mtyaka:
> Not sure if we need site_id for default object (need to test it somewhere)

It should be safe to leave it empty. When no site_id is present, the default site_id is automatically [pulled out from django settings](https://github.com/edx/configuration/blob/5a1d2454a5426e8b2a75d7fe1ee536e337f58d48/playbooks/roles/edxapp/tasks/site_configuration.yml#L24-L27) by the playbook.

> I think a better way would be to not use EDXAPP_SITE_CONFIGURATION and set up something like EDXAPP_CONTACT_US_CUSTOM_LINK and map that in the opencraft/configuration repo.

Yes, it would be good if we had a way to provide defaults for SiteConfiguration in Ocim while providing ability to override `EDXAPP_SITE_CONFIGURATION` without having to repeat all the defaults.

I would prefer if it was generic to allow us to set up defaults for any SiteConfiguration setting, not just the marketing link.

One way to do that would be to implement `EDXAPP_SITE_CONFIGURATION_EXTRA` (similar to `EDXAPP_ENV_EXTRA` which would get merged with existing `EDXAPP_SITE_CONFIGURATION` by matching up `site_id`s). 

> any thoughts on this? Does this introduce additional challenges when applying changes to existing instances?

I think it is fine to leave it as it is for this ticket since it won't break existing instances that use `EDXAPP_SITE_CONFIGURATION` (but it won't fix the links either).

If you could create a ticket to implement a mechanism to be able to provide defaults for site configuration either in edx/configuration or in Ocim, that would be great.

instance/models/mixins/openedx_config.py
```diff
@@ -77,6 +76,13 @@ def _get_configuration_variables(self):
                     # because we also add other extended checks
                     "openedx.core.djangoapps.heartbeat.default_checks.check_celery",
                 ],
+                # rewrite marketing links:
+                "MKTG_URL_OVERRIDES": {
+                    # remove blog link - not supported yet
+                    "BLOG": "",
+                    # use different contact page
+                    "CONTACT": "/contact",
```
mavidser, line 84: Is this variable required now, given that CONTACT_US_CUSTOM_LINK is the one being picked for the url? If not, we can remove this (if required - what happens if both the variables have conflicting values?)

s0b0lev: That's right, so:
1. `CONTACT_US_CUSTOM_LINK` is used for rendering footer link
2. `MKTG_URL_OVERRIDES['CONTACT']` used for email here: [djangoapps/student/views/management.py#L749](https://github.com/edx/edx-platform/blob/98af9ce4185bde0a38030923a88403a101e011ab/common/djangoapps/student/views/management.py#L749 ) and for home page footer (as I understand for custom theme: [red-theme/lms/templates/footer.html#L47](https://github.com/edx/edx-platform/blob/752b3072c4854752a0e6f4827c959917b555992f/themes/red-theme/lms/templates/footer.html#L47))

So we have to set `MKTG_URL_OVERRIDES['CONTACT']` this key, edx may also use it somewhere else as well.

mavidser (approved):
:+1:

- [x] I tested this: Ran tests, verified the sandbox config and the deployed appserver.
- [x] I read through the code
- [NA] I checked for accessibility issues
- [x] Includes documentation
- [NA] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #630, opened 2020-08-19
[SE-3040] Enables Auto Commit for MYSQLdb When Retrieving Cursor for DB

nizarmah:
Previously, any preupgrade sql commands ran would not be committed to the MySQL database. So, this PR enables auto commit in order to commit any changes during `instance_redeploy` to the MySQL database. 

**JIRA tickets**: SE-3040

**Testing instructions**:

TBD

**Reviewers**
- [ ] TBD

### DIFF
--- a/instance/models/mixins/openedx_database.py
+++ b/instance/models/mixins/openedx_database.py
@@ -198,6 +198,7 @@ class OpenEdXDatabaseMixin(MySQLInstanceMixin, MongoDBInstanceMixin, RabbitMQIns
             port=self.mysql_server.port,
             database=db_name,
         )
+        conn.autocommit(True)
         return conn.cursor()
 
     def _get_mysql_user_name(self, suffix):


instance/models/mixins/openedx_database.py
```diff
@@ -198,6 +198,7 @@ def get_mysql_cursor_for_db(self, db_suffix):
             port=self.mysql_server.port,
             database=db_name,
         )
+        conn.autocommit(True)
```
pomegranited, line 201: I wouldn't put this command here..  it's a bit extreme, since this cursor is used everywhere, and we only do the `--preupgrade-sql-commands` rarely.

Could you move it to the `instance_redeploy` [`_do_mysql_commands`](https://github.com/open-craft/opencraft/blob/nizar/instance_redeploy_mysql_autocommit/instance/management/commands/instance_redeploy.py#L283) method instead?

nizarmah: @pomegranited I had to include the changes in the `get_mysql_cursor_for_db` because committing changes is done through the connection.

If I were to move it to `instance_redeploy`, we wouldn't be able to setup the cursor using `get_mysql_cursor_for_db`. It would require that we setup a new connection and grab the cursor from that connection, since committing would be done using `conn.commit()` upon `cursor.execute()`. And auto committing is limited under a certain connection, so that would also no be possible.

Instead, I went for an argument `autocommit=False` which has a default `False` value.

pomegranited: Ok.. I think there was a way to do it  in `_do_mysql_commands` itself.. before closing `cursor.close()`, you could add this to effectively do the same thing:

```python
cursor.execute("COMMIT")
```

But if you want to keep it here, then it needs unit tests added, to ensure coverage of the change.

(I merged latest master to trigger the test re-runs, since the integration test issue has been repaired.)

PR #631, opened 2020-08-19
SE-3154 + SE-3175 Updates for Juniper auth and forum issues

pomegranited:
Configuration changes to address issues noted during Juniper upgrades.

**Use secure cookies with default SameSite=None policy**

7401806 Updates the default instance configuration to set:

```yaml
EDXAPP_CSRF_COOKIE_SECURE: true
EDXAPP_SESSION_COOKIE_SECURE: true
```

**Forum dependencies build error**

Loss of a pinned dependency in cs_comments_service requires using our custom branch until the issue is merged upstream.

679dd78 Applies this for juniper and ironwood deployments.

**Sandbox**

* LMS: https://se3154.sandbox.stage.opencraft.hosting/
* Studio: https://studio.se3154.sandbox.stage.opencraft.hosting/

**Testing instructions**

Deployed to Ocim stage, see sandbox links above.  Since we don't have LetsEncrypt certificates set up for stage instances, you'll have to accept the self-signed cert to do this testing.

Using Chrome, Firefox, and Safari:

1. Register a new user account.
1. Activate the account (see email).
1. Logout
1. Login again
1. Enrol in the DemoX course, and navigate to the discussion forum
1. Ensure you can post a message there.

**Author Notes & Concerns**

1. Initially added `DCS_SESSION_COOKIE_SAMESITE: "Lax"`, but reverted due to [unknown unintended side effects on auth to 3rd party apps like LTI](https://github.com/open-craft/opencraft/pull/631#discussion_r473127288).
1. Have hard-coded a `FORUM_VERSION` on open-craft's own fork, but this is intended to be temporary until upstream changes merge.

**Reviewers**

- [ ] @nizarmah 

CC @clemente

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -96,6 +96,10 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             "EDXAPP_LOGIN_REDIRECT_WHITELIST": [self.instance.studio_domain, ],
             "EDXAPP_SESSION_COOKIE_DOMAIN": '.{}'.format(self.instance.domain),
 
+            # Use secure SameSite cookies
+            "EDXAPP_CSRF_COOKIE_SECURE": True,
+            "EDXAPP_SESSION_COOKIE_SECURE": True,
+
             # Run a command to delete expired sessions once a day. The time is random and different in each server
             # to avoid the case when all servers connect to the database at exactly the same time.
             # It's fine to use random numbers here because this function is called just once per appserver and the
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -29,6 +29,7 @@ class OpenEdXConfigMixin(ConfigMixinBase):
         """
         return any(release_name in self.openedx_release for release_name in releases)
 
+    # pylint: disable=too-many-branches, useless-suppression
     def _get_configuration_variables(self):
         """
         Creates and returns a dictionary of ansible configuration variables for the instance
@@ -456,6 +457,19 @@ class OpenEdXConfigMixin(ConfigMixinBase):
             hb_app = "openedx.core.djangoapps.heartbeat"
             template["EDXAPP_CMS_ENV_EXTRA"]["ADDL_INSTALLED_APPS"].append(hb_app)
             template["EDXAPP_LMS_ENV_EXTRA"]["ADDL_INSTALLED_APPS"].append(hb_app)
+
+        # See https://github.com/edx/cs_comments_service/pull/323
+        if self._is_openedx_release_in(['ironwood']):
+            # See https://github.com/open-craft/cs_comments_service/pull/5
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["forum_version"] = "opencraft-release/ironwood.2"
+        if self._is_openedx_release_in(['juniper']):
+            # See https://github.com/open-craft/cs_comments_service/pull/6
+            # TODO: Once https://github.com/edx/cs_comments_service/pull/323 merges,
+            # update to use edx for forum_source_repo, and "open-release/juniper.master" for forum_version(s).
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["forum_version"] = "opencraft-release/juniper.2"
+            template["FORUM_VERSION"] = "opencraft-release/juniper.2"
 
         return template
 


instance/models/mixins/openedx_config.py
```diff
@@ -453,6 +461,14 @@ def _get_configuration_variables(self):
...
             template["EDXAPP_LMS_ENV_EXTRA"]["ADDL_INSTALLED_APPS"].append(hb_app)
 
+        # See https://github.com/edx/cs_comments_service/pull/323
+        if self._is_openedx_release_in(['ironwood']):
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["forum_version"] = "opencraft-release/ironwood.2"
+        if self._is_openedx_release_in(['juniper']):
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["FORUM_VERSION"] = "opencraft-release/juniper.2"
+
```
pomegranited, line 471: @swalladge quality failure 😦 

> pylint: too-many-branches / Too many branches

swalladge: :joy:

clemente: This shouldn't be hard-coded to OpenCraft branches, if we still expect that other people use Ocim to manage their instances. But I don't know how to override this branch. Couldn't we modify our playbooks instead? (Not sure).

Right now the `openedx_release` field [is used as default for the forum, notifier, etc.](https://github.com/open-craft/opencraft/blob/734bac60816da3796c1e99e504ffd767279b9f5a/instance/models/openedx_appserver.py#L84).
An opposite way would be to override `forum_version` through `extra_configuration_settings` in every instance we create. Manually.
I don't know whether we have any option to automatically apply some temporary variables specific to our instances. Something like a default `configuration_extra_settings`, that makes it clear that it is overriding the defaults, while we keep simple/sane defaults. This isn't trivial. 

So, apart from modifying the playbooks instead, or doing the `forum_version` override manually in each instance, or finding another way to override defaults… Another way to do this would be to do this hard-coded override *only* for instances built off OpenCraft branches (adding a condition to the `if` here). But we'd still be hard-coding OpenCraft and OpenCraft's forks in the code.

pomegranited:
Sandbox is running now -- ready for review. Do you have time @pkulkark or @clemente ?

instance/models/mixins/openedx_config.py
```diff
@@ -66,6 +67,8 @@ def _get_configuration_variables(self):
                 "BROKER_HEARTBEAT": 0,
                 "ADDL_INSTALLED_APPS": [
                 ],
+                # Required for Safari auth with secure SameSite cookies
+                "DCS_SESSION_COOKIE_SAMESITE": "Lax",
```
clemente, line 71: For reference. I also left a comment at SE-3154 about `Lax` vs. `None`; we can discuss it there.

nizarmah (approved):
:+1:

- [x] I tested this as suggested in the _Testing Instructions_
- [x] I read through the code
- [x] Includes documentation

# 

This approval is a bit "pre-mature". It is just a verification that the changes have been tested and work properly. Please don't merge this pull request before addressing @clemente's comments :slightly_smiling_face: 

But once the comments have been resolved, and there are no major code changes that need to be tested in a sandbox, it can be merged :+1: 

If there's any major code changes and a refreshed / new sandbox instance, please re-request my review :slightly_smiling_face:

instance/models/mixins/openedx_config.py
swalladge: > Something like a default configuration_extra_settings,

that is what this openedx_config is supposed to be - a safe, tested set of defaults for instances. I think this is fine, because if anyone else uses ocim, they can add their own extra settings overrides.

instance/models/mixins/openedx_config.py
```diff
@@ -453,6 +462,15 @@ def _get_configuration_variables(self):
...
             template["EDXAPP_LMS_ENV_EXTRA"]["ADDL_INSTALLED_APPS"].append(hb_app)
 
+        # See https://github.com/edx/cs_comments_service/pull/323
+        if self._is_openedx_release_in(['ironwood']):
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["forum_version"] = "opencraft-release/ironwood.2"
+        if self._is_openedx_release_in(['juniper']):
+            template["forum_source_repo"] = "https://github.com/open-craft/cs_comments_service.git"
+            template["forum_version"] = "opencraft-release/juniper.2"
+            template["FORUM_VERSION"] = "opencraft-release/juniper.2"
```
pomegranited, line 472: FYI @swalladge We have to specify both `forum_version` and `FORUM_VERSION` here, because [we set `"forum_version": self.openedx_release` in Ocim above](https://github.com/open-craft/opencraft/blob/7ff66c37fcd4dd52486f54772081a807281a9fa0/instance/models/mixins/openedx_config.py#L186-L187), and [edx:configuration juniper.2 uses `forum_version`](https://github.com/edx/configuration/blob/open-release/juniper.2/playbooks/roles/forum/tasks/deploy.yml#L45) still. [edx:configuration master drops `forum_version` and uses `FORUM_VERSION`](https://github.com/edx/configuration/blob/642db0e17837d758748982e7b95bcd399bb39bae/playbooks/roles/forum/tasks/deploy.yml#L45), so post-juniper, we can drop these lowercase variables.

To demonstrate, [this sandbox failed to provision](https://stage.manage.opencraft.com/instance/5530/edx-appserver/2116/), because it ended up with: 
```yaml
FORUM_VERSION: opencraft-release/juniper.2
forum_source_repo: https://github.com/open-craft/cs_comments_service.git
forum_version: open-release/juniper.2  # does not exist on this fork ^
```



the [configuration code for juniper is dumb](https://github.com/edx/configuration/blob/open-release/juniper.2/playbooks/roles/forum/defaults/main.yml#L92-L93) (it's fixed in [master]()):

instance/models/mixins/openedx_config.py
clemente: I don't object to merging it; I just think it's bad practice to hardcode OpenCraft-specific logic or branches in the code. It creates more possible bugs (e.g. if edX releases juniper.3 and we deploy it, this temporary snippet would force juniper.2 for the forum instead of auto-using juniper.3), it creates more maintenance work, …
There should be at least some comment asking to remove this temporary code.

Other ways, like creating a new setting to specify FORUM_VERSION, are more complex.

instance/models/mixins/openedx_config.py
pomegranited: @clemente I've read your notes, and since I don't know what the full effects of this change might be, I'm going to remove this line, and we can look into it more fully if clients report issues with Safari.

[`SameSite=Lax` is the default for the devstack](https://github.com/edx/edx-platform/blob/e7099518d92dddbcc671b014045cde800d23f2d1/lms/envs/devstack.py#L400), which indicated to me that this was a relatively safe option, but I haven't tested every case, so can't be sure.

I found that it fixed SSO logins on Safari for my client, so have put this into their config, but maybe it's something particular to their setup?

instance/models/mixins/openedx_config.py
pomegranited: @clemente I agree it's bad practice, so have added a TODO to remind us to fix this when/if we can.

But since this fix is breaking juniper.2 redeployments and new self-hosted services, and this code fix isn't available in any upstream branches yet, I think it's all we can do right now.

Plus, even once https://github.com/edx/cs_comments_service/pull/323 is ported to `juniper`, it'll only be on `juniper.master`, so we can't just use our usual [`openedx_release`](https://github.com/open-craft/opencraft/blob/7ff66c37fcd4dd52486f54772081a807281a9fa0/instance/models/mixins/openedx_config.py#L186-L187) for forums either while we're on `juniper.2`.

A better long-term approach might be to add additional `FORUM_STABLE_REF` and `FORUM_STABLE_REPO` environment variables, and configure it for our deployment, but as noted, anyone trying to use Ocim with `juniper.2` or `ironwood.2` right now will be unable to provision.

pomegranited (commented):
I've addressed the comments as best I can, so am going to merge this, and will discuss follow-up tickets on SE-3175.

Apologies for combining these two issues into one; I thought it would speed the release process, not hinder it, but I was mistaken!

PR #634, opened 2020-08-24
[BB-2801] Fix integration tests for Juniper

0x29a:
This PR fixes four issues that I've encountered during running CI using Juniper instances.

**JIRA tickets**:
- [BB-2801](https://tasks.opencraft.com/browse/BB-2801)

**Dependencies**:
https://github.com/open-craft/configuration/tree/integration-juniper

**Testing instructions**:

Passing of all CI checks is enough.

**Reviewers**
- [ ] @lgp171188

### DIFF
--- a/instance/tests/integration/integration_instance.py
+++ b/instance/tests/integration/integration_instance.py
@@ -354,7 +354,7 @@ class InstanceIntegrationTestCase(IntegrationTestCase):
         homepage_html = get_url_contents(instance.url)
         self.assertIn(instance.static_content_overrides['homepage_overlay_html'], homepage_html)
 
-        page_url = '{}/{}'.format(instance.url, page)
+        page_url = instance.url + page
         server_html = get_url_contents(page_url)
         self.assertIn(instance.static_content_overrides['static_template_{}_content'.format(page)], server_html)
 
--- a/playbooks/collect_activity/collect_activity.yml
+++ b/playbooks/collect_activity/collect_activity.yml
@@ -22,11 +22,12 @@
       become_user: www-data
 
     - name: Run collect_activity script
-      command: >
+      shell: >
+        . /edx/app/edxapp/edxapp_env &&
         /edx/bin/python.edxapp /tmp/collect_activity.py
-          --config-section {{ config_section }}
-          --out {{ remote_output_filename }}
-          {{ extra_script_arguments }}
+        --config-section {{ config_section }}
+        --out {{ remote_output_filename }}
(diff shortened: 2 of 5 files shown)


instance/tests/integration/integration_instance.py
```diff
@@ -354,7 +354,7 @@ def assert_static_content_overrides_work(self, instance, appserver, page):
         homepage_html = get_url_contents(instance.url)
         self.assertIn(instance.static_content_overrides['homepage_overlay_html'], homepage_html)
 
-        page_url = '{}/{}'.format(instance.url, page)
+        page_url = instance.url + page
```
pcockwell, line 357: Is there a particular reason for this change? Will the previous syntax not work with the Python version upgrades?

playbooks/collect_activity/collect_activity.yml
```diff
@@ -22,11 +22,12 @@
       become_user: www-data
 
     - name: Run collect_activity script
-      command: >
+      shell: >
+        . /edx/app/edxapp/edxapp_env &&
```
pcockwell, line 26: I think `source /edx/app/edxapp_env` would be more explicit here

pcockwell (changes requested):

instance/tests/integration/integration_instance.py
0x29a: Yes, @pcockwell, by some reason `instance.url` contains trailing slash in Juniper unlike previous release, so this was breaking test.

playbooks/collect_activity/collect_activity.yml
0x29a: @pcockwell `shell` module uses `/bin/sh` under the hood, which has no `source` command.

pcockwell (approved):
Good to go assuming tests complete successfully :+1:

- [ ] I read through the code

instance/tests/integration/integration_instance.py
lgp171188: @0x29a, `instance.url` had a trailing slash even on Ironwood. So fixing this is good. 👍🏼

lgp171188 (approved):
:+1: @0x29a, this is good to merge. While it will be ideal to merge after addressing https://github.com/open-craft/configuration/commit/72144b3e0615c5f154191757f64354753f66221b#r41717195, we can address that in a separate PR as there isn't a lot of time before the sprint ends.
- [x] I tested this works by checking the integration test run.
- [x] I read through the code

PR #637, opened 2020-08-31
BB-2883 Allow instantce_statistics_csv to receive multiple domains; Create trial_instances_report on cronjob

pcockwell:
Modifies the `instance_statistics_csv` command to allow for multiple domains. Also uses this new version in a new task to send a trial instances report to emails set in `settings.TRIAL_INSTANCES_REPORT_RECIPIENTS`

**JIRA tickets**:
[BB-2883](https://tasks.opencraft.com/browse/BB-2883)

**Screenshots**:

trial_instance_report_email
trial_instance_report_csv

**Testing instructions**:

Test on functionality on staging:

1. Generate the public SSH key for `www-data@stage.manage.opencraft.com`  using `ssh-keygen -y -f ~/.ssh/id_rsa`
2. Add public key to `authorized_keys` file on `logs.opencraft.com`
3. Set `INSTANCE_LOGS_SERVER_HOST` to `logs.opencraft.com` in `.env` file on `stage.manage.opencraft.com`
4. Run
```
make manage -- instance_statistics_csv --domains daniel.stage.opencraft.hosting,piotr.stage.opencraft.hosting,gabrieltesting.stage.opencraft.hosting
```
5. Wait for resulting output

Test task on staging:

1. Generate the public SSH key for `www-data@stage.manage.opencraft.com`  using `ssh-keygen -y -f ~/.ssh/id_rsa`
2. Add public key to `authorized_keys` file on `logs.opencraft.com`
3. Set `INSTANCE_LOGS_SERVER_HOST` to `logs.opencraft.com` in `.env` file on `stage.manage.opencraft.com`
4. Set `TRIAL_INSTANCE_REPORT_RECIPIENTS` to `'["YOUR_EMAIL@opencraft.com"]'` in `.env` file on `stage.manage.opencraft.com`
5. Edit `reports/tasks.py` and change this line
```
@db_periodic_task(crontab(day='1', hour='2', minute='0'))
```
to this
```
@db_periodic_task(crontab(minute='*/5'))
```
6. Start the web process with `make run` (in the web screen tab) and wait 5-6 minutes
7. Check your email for a resulting CSV

**Reviewers**
- [ ] @lgp171188

### DIFF
--- a/instance/management/commands/instance_statistics_csv.py
+++ b/instance/management/commands/instance_statistics_csv.py
@@ -69,10 +69,10 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            '--domain',
+            '--domains',
             '-d',
             default=None,
-            help='The fully qualified domain of the instance to get the data for',
+            help='Comma-separated list of fully qualified domains of the instances to get the data for',
             required=True
         )
         parser.add_argument(
@@ -115,6 +115,8 @@ class Command(BaseCommand):
                 ))
                 sys.exit(1)
 
+        domains = options['domains'].split(',')
+
         start_date = datetime.strptime(options['start_date'], "%Y-%m-%d").date()
         end_date = datetime.strptime(options['end_date'], "%Y-%m-%d").date()
 
@@ -129,36 +131,58 @@ class Command(BaseCommand):
             ))
             sys.exit(1)
 
-        self.collect_instance_statistics(out, options['domain'], start_date, end_date)
+        self.collect_instance_statistics(out, domains, start_date, end_date)
 
-    def get_instance_from_domain_name(self, domain_name):
+    def get_instances_from_domain_names(self, domain_names):
         """ Get an instance object for a given domain name """
-        try:
-            instance = OpenEdXInstance.objects.get(
-                Q(external_lms_domain=domain_name) | Q(internal_lms_domain=domain_name)
+        instances = OpenEdXInstance.objects \
+            .filter(ref_set__openedxappserver_set___is_active=True) \
+            .filter(successfully_provisioned=True) \
+            .filter(
+                Q(external_lms_domain__in=domain_names) | Q(internal_lms_domain__in=domain_names)
             )
-        except OpenEdXInstance.DoesNotExist:
+
+        if not instances:
             self.stderr.write(self.style.ERROR(
-                'No OpenEdXInstance exists with an external or internal '
-                'domain of {domain_name}'.format(
-                    domain_name=domain_name
+                'No OpenEdXInstances exist with an external or internal '
+                'domain of {domain_names}'.format(
+                    domain_names=', '.join(domain_names)
                 )
             ))
             sys.exit(1)
 
-        # If there are no active appservers for the instance, we should error out
-        if not instance.successfully_provisioned or not instance.get_active_appservers():
-            self.stderr.write(self.style.ERROR(
-                'No active OpenEdXAppServers exist for the instance with '
-                'external or internal domain of {domain_name}'.format(
-                    domain_name=domain_name
-                )
-            ))
-            sys.exit(1)
+        return instances
 
-        return instance
+    def get_instance_metadata(self, instances):
+        """ Get a mapping of name_prefix to public_ip for active appservers """
+        instance_metadata = {}
+        for instance in instances:
+            appserver = instance.get_active_appservers().first()
+            name_prefix = appserver.server_name_prefix
+            domain = instance.domain
+            ref_name = instance.ref.name
 
-    def get_elasticsearch_hits_data_summary(self, playbook_output_dir, name_prefix, start_date, end_date):
+            try:
+                betatestapplication = instance.betatestapplication_set.get()
+                email = betatestapplication.user.email
+                status = betatestapplication.status
+            except BetaTestApplication.DoesNotExist:
+                email = 'N/A'
+                status = 'N/A'
+
+            instance_metadata[name_prefix] = {
+                'domain': domain,
+                'ref_name': ref_name,
+                'name_prefix': name_prefix,
+                'public_ip': appserver.server.public_ip,
+                'instance_age': datetime.now(instance.created.tzinfo) - instance.created,
+                'email': email,
+                'status': status
+            }
+
+        return instance_metadata
+
+    def get_elasticsearch_hits_data_summary(self, playbook_output_dir, name_prefixes, start_date, end_date):
         """ Execute the collect_elasticsearch_data playbook to gather statistics """
         if not settings.INSTANCE_LOGS_SERVER_HOST:
             self.stderr.write(self.style.WARNING(
@@ -179,6 +203,8 @@ class Command(BaseCommand):
         log_line.info = log_line
         log_line.error = log_line
 
+        self.stderr.write(self.style.SUCCESS(','.join(name_prefixes)))
+
         # Launch the collect_elasticsearch_data playbook, which places a file into the `playbook_output_dir`
         # on this host.
         ansible.capture_playbook_output(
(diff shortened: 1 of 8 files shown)


reports/tasks.py
```diff
@@ -0,0 +1,226 @@
...
+
+    for instance in instances:
+        betatest_application = instance.betatestapplication_set.get()
+        appserver = instance.get_active_appservers().first()
+        name_prefix = appserver.server_name_prefix
+        public_ip = appserver.server.public_ip
+
+        with ansible.create_temp_dir() as playbook_output_dir:
+            get_elasticsearch_hits_data_summary(playbook_output_dir, name_prefix, beginning_of_last_month, end_of_last_month)
+            get_instance_usage_data(playbook_output_dir, name_prefix, public_ip)
```
pcockwell, line 178: Currently, this is looping through relevant instances and calling both playbooks synchronously. This should probably be streamlined/optimized to run as much in parallel as possible. The only way I can think of doing so thus far would be to gather the public IP addresses of each instance, and then run `get_instance_usage_data` with the entire list of public IPs. Even then, Elasticsearch queries would still likely need to be run sequentially.

Do you have any thoughts/suggestions?

reports/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+        ),
+        playbook_path=playbook_path,
+        username=settings.INSTANCE_LOGS_SERVER_SSH_USERNAME,
+        logger_=log_line,
+    )
+
+
+# Run on the 1st of every month
+@db_periodic_task(crontab(day='1', hour='2', minute='0'))
+def send_trial_instance_report(recipients=["patrick@opencraft.com"]):
```
viadanna, line 136: It would be nice to have the recipients added to the settings/env.

pcockwell: Added this. Thanks for the suggestion

reports/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    Generate and send a trial instance data report
+
+    This task runs on the first of every month at 2AM
+    """
+
+    # Get instances with an active betatestapplication and
+    # whose InstanceReference has an active OpenEdXAppServer
+    instances = OpenEdXInstance.objects \
+        .exclude(betatestapplication__isnull=True) \
+        .filter(ref_set__openedxappserver_set___is_active=True)
```
lgp171188, line 45: @pcockwell, you can break the line after a `{` and that looks cleaner than using trailing `\`.

reports/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+
+    csv_filename = '/tmp/trial_instances_report.csv'
+
+    call_command(
+        'instance_statistics_csv',
+        out=csv_filename,
+        domains=','.join(domains),
+        start_date=beginning_of_last_month.strftime("%Y-%m-%d"),
+        end_date=end_of_last_month.strftime("%Y-%m-%d")
+    )
```
lgp171188, line 66: @pcockwell, we should handle any exceptions coming out of this call and send an email notifying the `TRIAL_INSTANCE_REPORT_SENDER` about it.

opencraft/settings.py
```diff
@@ -874,6 +874,10 @@
 INSTANCE_LOGS_SERVER_HOST = env('INSTANCE_LOGS_SERVER_HOST', default=logs_server_host_default)
 INSTANCE_LOGS_SERVER_SSH_USERNAME = env('INSTANCE_LOGS_SERVER_SSH_USERNAME', default='ubuntu')
 
+# Trial Instances Report ######################################################
+
+TRIAL_INSTANCE_REPORT_RECIPIENTS = env.json('TRIAL_INSTANCE_REPORT_RECIPIENTS', default=['billing@opencraft.com'])
```
lgp171188, line 879:
```suggestion
TRIAL_INSTANCES_REPORT_RECIPIENTS = env.json('TRIAL_INSTANCES_REPORT_RECIPIENTS', default=['billing@opencraft.com'])
```

reports/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+# Logging #####################################################################
+
+logger = logging.getLogger(__name__)
+
+
+# Constants ###################################################################
+
+TRIAL_INSTANCE_REPORT_SENDER = getattr(settings, 'TRIAL_INSTANCE_REPORT_SENDER',
```
lgp171188, line 26: @pcockwell, this is not defined in `settings.py` at all and hence will never get overridden. I'd recommend dropping this and just using `settings.DEFAULT_FROM_EMAIL`.

reports/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+
+    domains = [instance.external_lms_domain or instance.internal_lms_domain for instance in instances]
+
+    # Start with current date
+    today = datetime.utcnow().date()
+
+    # Get last month by subtracting the day of the month from
+    # the actual date to get the last day of last month
+    # Then replace the day with 1, to get the first day of last month
+    end_of_last_month = (today - timedelta(days=today.day))
```
lgp171188, line 51:
```suggestion
    end_of_last_month = (today - timedelta(days=1))
```
is better and explicit, no?

reports/tasks.py
```diff
@@ -0,0 +1,85 @@
...
+    instances = OpenEdXInstance.objects.exclude(
+        betatestapplication__isnull=True
+    ).filter(
+        ref_set__openedxappserver_set___is_active=True
+    )
+
+    domains = [instance.external_lms_domain or instance.internal_lms_domain for instance in instances]
+
+    # Start with current date
+    today = datetime.utcnow().date()
```
lgp171188, line 46: @pcockwell, it will be better to not rely on `today` being the first of a month and instead explicitly get the first day of the current month.

reports/tasks.py
```diff
@@ -0,0 +1,85 @@
...
+
+# Logging #####################################################################
+
+logger = logging.getLogger(__name__)
+
+
+# Tasks #######################################################################
+
+# Run on the 1st of every month
+@db_periodic_task(crontab(day='1', hour='2', minute='0'))
```
lgp171188, line 27: @pcockwell, it might be a good idea to make this configurable from a settings string and keep this time as a default.

reports/tasks.py
pcockwell: I would disagree. Doing `- timedelta(days=today.day)` will give you last month's data regardless of what day it is run, allowing the schedule for this task to change independently of the code. Happy to add a comment for this though

playbooks/collect_instance_statistics/query_elasticsearch.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # pylint: skip-file
 
+from __future__ import print_function
```
lgp171188, line 4: @pcockwell, why is `print_function` needed when the script runs on Python 3?

pcockwell:
@lgp171188 

Do we want to have a different default for `TRIAL_INSTANCES_REPORT_RECIPIENTS` or `TRIAL_INSTANCES_REPORT_SCHEDULE` on staging than on production?

instance/management/commands/instance_statistics_csv.py
```diff
@@ -244,27 +279,35 @@ def log_line(line):
...
+            )
+            self.get_instance_usage_data(
+                playbook_output_dir,
+                instance_metadata
+            )
 
             csv_writer = csv.writer(out, quoting=csv.QUOTE_NONNUMERIC)
             csv_writer.writerow([
                 'Fully Qualified Domain Name',
                 'Server Name Prefix',
```
lgp171188, line 307: @pcockwell, I think this is useful only inside this task and not useful in any report.

lgp171188:
@pcockwell, I tested this on stage and it works okay. Approving this PR now.

lgp171188 (approved):
:+1: 
- [x] I tested this and verified the management command updates and the periodic task work okay.
- [x] I read through the code

PR #638, opened 2020-09-02
Add a health check endpoint to ocim

mavidser:
Adds a health check endpoint to OCIM which verifies the postgres, redis and consul connections.

**JIRA tickets**: [SE-3021](https://tasks.opencraft.com/browse/SE-3021)

**Testing instructions**:

- Run `make test.unit`
- Make sure postgres, redis and consul are running, then visit `/health_check` URL on ocim. Now take down services one by one and verify the non-200 response from the health check page.

**Author notes**:
- If postgres is unavailable then the view isn't even called as ocim fails at trying to log the request in the database. I've still added the verification code to the view in case the logging behaviour ever changes in the future.

**Reviewers**
- [ ] @mtyaka 
- [ ] @lgp171188

### DIFF
--- a/opencraft/views.py
+++ b/opencraft/views.py
@@ -22,8 +22,16 @@ OpenCraft views
 
 # Imports #####################################################################
 
+from consul import Consul
+from django.db import connection
+from django.http import HttpResponse
 from django.views.generic.base import RedirectView
+from django.views.generic import View
 from django.urls import reverse
+from huey.contrib.djhuey import HUEY
+from psycopg2 import OperationalError
+from redis.exceptions import ConnectionError as RedisConnectionError
+from requests.exceptions import ConnectionError as RequestsConnectionError
 
 from instance.models.instance import InstanceReference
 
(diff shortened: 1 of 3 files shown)


opencraft/views.py
```diff
@@ -45,3 +53,46 @@ def get_redirect_url(self, *args, **kwargs):
...
             return reverse('registration:register')
+
+class HealthCheckView(View):
+    """
+    Health Check view
+    """
+    def get(self, request):
+        try:
+            # Verify that the postgres database backend is reachable
+            self._check_postgres_connection()
```
lgp171188, line 64: @mavidser, will the app even be able to serve requests when the database is down? Just thinking out loud. CC @mtyaka

opencraft/views.py
```diff
@@ -45,3 +53,46 @@ def get_redirect_url(self, *args, **kwargs):
...
+        except OperationalError:
+            return HttpResponse('Postgres unreachable', status=500)
```
mtyaka, line 66: Nit: edx-platform uses 503 status for their [heartbeat endpoint](https://github.com/edx/edx-platform/blob/c87df89be5f1d0bda8135253b21590a3b92a738a/openedx/core/djangoapps/heartbeat/views.py#L21), which I think is somewhat more appropriate than 500.

opencraft/views.py
mtyaka: This is acknowledged under "Author notes", but I think it would be good to also add a comment explaining that to the code.

mtyaka (approved):
Nice work @mavidser! :+1: You'll have to fix some lint/quality issues and might want to address the nits, but otherwise this looks good to merge :)

- [x] I tested this: I verified that the health_check endpoint works correctly on my devstack.
- [x] I read through the code
- [ ] Includes documentation

PR #640, opened 2020-09-10
[SE-3136] Add client instance name to New Relic monitoring condition name

gabor-boros:
This PR is adding client instance name to New Relic monitoring condition name to be able to find root cause of incidents faster and improve the investigation experience. In case the name is too long, it will be truncated due to limitations.

**Dependencies**: None

**Screenshots**:

Screenshot 2020-09-11 at 9 53 11

**Merge deadline**: None

**Testing instructions**:

1. If you don't already have `ADMINS` setting set in your local `.env` file, set it to your email (ie. `ADMINS='[["Gabor Admin", "gabor@opencraft.com"]]'`.
2. Set `NEWRELIC_ADMIN_USER_API_KEY` and `NEWRELIC_LICENSE_KEY` in your local `.env` file. You can find them on the production Ocim VM's `.env` file.
3. Start the django shell (`make shell`).
4. Get a reference to an `OpenEdXInstace` (or create a new one if you don't have an existing instance in your devstack).
5. Run `instance.enable_monitoring()`. Make sure that there are no errors.
6. Log into NewRelic and go to 'Alerts & AI -> Policies'. Search by the URL of your test instance and verify that the policy alerts look correct.
7. Run `instance.disable_monitoring()`. 
8. Verify that the policy for your instance has been removed from NewRelic.
9. Don't forget to remove `NEWRELIC_ADMIN_USER_API_KEY` and `NEWRELIC_LICENSE_KEY` from your local `.env` file!

1. The format of **the condition name was not finalized beforehand**, so I'm opened to discuss other ideas too.
2. In case the condition name reaches the maximum (64) characters and the instance name is one long word, the instance name will be replaced by `...`. Example faulty condition name: `Preview of ...` [1] [2]

[1] In my opinion, this shouldn't be an issue since it would mean that the instance name should be at least 42 characters long without any whitespace. Number 42 is built up from: length of the longest "prefix" with formatting extracted from 64, the current maximum length. So: `64 - len("Extended heartbeat of ")`
[2] The 64 length is not confirmed by NewRelic developers yet. Currently, I'm investigating the max length with the help of New Relic, but **I think, this shouldn't be a blocker**, since an institution name which is equal to or longer than 42 chars is really rare.

### DIFF
--- a/instance/models/mixins/openedx_monitoring.py
+++ b/instance/models/mixins/openedx_monitoring.py
@@ -22,9 +22,12 @@ Open edX instance monitoring mixin
 
 # Imports #####################################################################
 
+import textwrap
+
 from django.conf import settings
 from django.db import models
 from django.db.models import ProtectedError
+from django.utils.html import escape
 
 from instance import newrelic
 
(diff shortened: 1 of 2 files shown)


gabor-boros:
@toxinu Added an extra test to ensure that the condition name is not exceeding the (assumed) maximum length of the field.

toxinu (approved):
Great work! You can merge it 👍🏻 

- [x] I tested this: test locally with testing instructions
- [x] I read through the code
- [ ] I checked for accessibility issues N/A
- [ ] Includes documentation N/A

PR #642, opened 2020-09-11
[SE-2497]  Avoid sending deployment failure emails if deployment triggered by OpenCraft members

gabor-boros:
Avoid sending deployment failure emails if deployment triggered by OpenCraft members.
Apply changes made in #633 to implement conditional email sending.

**Dependencies**: None

**Merge deadline**: None

**Testing instructions**:

*Assuming you already installed and setup OCIM locally and you registered to your local OCIM installation as a regular user would do.*

1. Create a new deployment from the new interface, make it fail and check email is sent
2. Redeploy instances using instance_redeploy command and check that email is not sent
3. Create a new deployment from the manage site and check that no email is sent if it fails

* Added logging for skipped emails to ensure no email sending skipped without any traces.
* Unit tests are **not really unit tests** for this scenario, since the signal itself is not tested and during the instance tests, the signal is not mocked, but I can live with this. The possible solution: refactor the tests for the `OpenEdxInstance` model but **that would blow up the scope** of this task. This is valid for the management command test as well.

Closes #633

### DIFF
--- a/instance/models/mixins/utilities.py
+++ b/instance/models/mixins/utilities.py
@@ -166,6 +166,10 @@ def send_urgent_alert_on_permanent_deployment_failure(sender, **kwargs) -> None:
     if deployment_id is not None:
         deployment = Deployment.objects.get(pk=deployment_id)
         if deployment.type in ignorable_deployment_types:
+            logger.warning(
+                "Skip sending urgent alert e-mail after instance %s didn't provision initiated by OpenCraft member",
+                instance,
+            )
             return
 
     if appserver is None and instance.provisioning_failure_notification_emails:
--- a/instance/models/mixins/utilities.py
+++ b/instance/models/mixins/utilities.py
@@ -167,7 +167,8 @@ def send_urgent_alert_on_permanent_deployment_failure(sender, **kwargs) -> None:
         deployment = Deployment.objects.get(pk=deployment_id)
         if deployment.type in ignorable_deployment_types:
             logger.warning(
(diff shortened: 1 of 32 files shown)


instance/models/mixins/utilities.py
```diff
@@ -147,11 +148,30 @@ def send_urgent_alert_on_permanent_deployment_failure(sender, **kwargs):
...
     if not instance.betatestapplication_set.exists():
         return
 
+    # In case a deployment initiated by an OpenCraft member
+    # and failed, skip sending urgent email
+    if deployment_id is not None:
+        deployment = Deployment.objects.get(pk=deployment_id)
+        if deployment.type in ignorable_deployment_types:
+            logger.warning(
+                "Skip sending urgent alert e-mail after instance %s didn't provision initiated by OpenCraft member",
```
pkulkark, line 170: nit: Maybe change the msg text to "Skip sending urgent alert e-mail after instance %s provisioning failed since it was initiated by OpenCraft member".

pkulkark (approved):
@gabor-boros Just the small nit about the message text but otherwise LGTM :+1: 

- [x] I tested this: I verified that the email notification is sent only if deployment is initiated from the new UI and not if initiated via the admin or shell as described in the testing instructions.
- [x] I read through the code
- [ ] ~~I checked for accessibility~~ N/A

gabor-boros:
Although the commits after the approval looks strange, the only change was the suggestion made during approval.

instance/models/mixins/utilities.py
gabor-boros: Resolved by 76541d7

PR #643, opened 2020-09-16
[SE-2869] Forbid registering domains/subdomains which would trick our DNS settings

gabor-boros:
This PR implements a domain/subdomain validation for custom and managed past/archived/active instances. The implementation uses three approaches to validate the domain:

1. If the domain starts with a reserved word (preview/studio/ecommerence/discovery), the validation will fail
2. If the subdomain starts with a reserved word (preview/studio/ecommerence/discovery), the validation will fail
3. If the domain is a subdomain of an existing instance's domain, the registration will fail
4. If the subdomain is already in our DNS settings (like haproxy.net.opencraft.com or ldp.opencraft.hosting), the validation will fail

**Dependencies**: None

**Screenshots**:

Requiring ack from the user before submission
Screenshot 2020-09-18 at 13 35 12

(1) Domain starts with a reserved word
Screenshot 2020-09-16 at 22 16 32

(2) Subdomain starts with a reserved word
Screenshot 2020-09-16 at 22 14 34
Screenshot 2020-09-16 at 22 14 44
Screenshot 2020-09-16 at 22 14 52

(3) Subdomain of an existing instance's domain
Screenshot 2020-09-16 at 22 11 03
Screenshot 2020-09-16 at 22 11 40

(4) Subdomain is already in our DNS settings
Screenshot 2020-09-17 at 8 22 47
Screenshot 2020-09-17 at 8 22 55

**Merge deadline**: None

**Testing instructions**:

1. Start the backend service
2. Start the frontend by executing `REACT_APP_INTERNAL_DOMAIN_NAME=".<YOUR_DOMAIN>" npm start` where `YOUR_DOMAIN` is equal to the `DEFAULT_INSTANCE_BASE_DOMAIN` and be aware of the `.` at the beginning.  
3. Navigate to `http://localhost:3000/registration/domain`
4. Try any of the scenarios shown on the screenshots above (please note that the `DEFAULT_INSTANCE_BASE_DOMAIN` defaults to `plebia.net` where only some DNS records are set)

* **BetaTestApplication's `BASE_DOMAIN` is set to the `DEFAULT_INSTANCE_BASE_DOMAIN`** as it seems that's the correct approach, since setting `opencraft.hosting` domain for every beta test app does not make sense if someone sets `DEFAULT_INSTANCE_BASE_DOMAIN` to something else (like `plebia.net`) and has no control over `opencraft.hosting` domain's DNS records.
* The scenario discussed in `test_external_domain_forbidden_domain_not_causing_issue` does not exists for subdomains
* In case we cannot access the `DEFAULT_INSTANCE_BASE_DOMAIN` DNS records, **we won't allow** registering users with the given subdomain **to make sure we are not overriding possibly existing DNS records** because of a network glitch or similar.
* Watch out for betatestapplication 0018 migration! (that's only a help text and validator error code change though)
* The domain accept text is **required for subdomains as well** on the registration page when not registering with custom domain (this could make sense in some cases)
* The js/ts client is updated (`npm run update-api-client`), hence some changes which may not be related completely to this change
* There is a really-really small chance that we will raise an error for the user if an OC member manually registers a subdomain on Gandi but that's not in the cache (since we have no callback from Gandi to OCIM)


registration/api/v2/serializers.py
```diff
@@ -22,16 +22,21 @@
 import logging
 from typing import Dict
 
+import tldextract
```
toxinu, line 25: Please add `tldextract` into `base.in`([file](https://github.com/open-craft/opencraft/blob/master/requirements/base.in)) since we are using it directly it is better to directly pin it.

gabor-boros: Oh, that's a fair point! Thanks for noticing 👌

frontend/packages/api_client/src/apis/V1Api.ts
```diff
@@ -227,14 +232,58 @@ export class V1Api extends runtime.BaseAPI {
...
+     * Uses InstanceReference to iterate all types of instances, and serializes them.  The fields that are returned for each instance depend on its instance_type and whether you are listing all instances (returns fewer fields) or just one instance (returns all fields).  The only fields that are available for all instances, regardless of type, are the fields defined on the InstanceReference class, namely:  * `id` * `name` * `notes` * `created` * `modified` * `instance_type` * `is_archived`  Note that IDs used for instances are always the ID of the InstanceReference object, which may not be the same as the ID of the specific Instance subclass (e.g. the OpenEdXInstance object has its own ID which should never be used - just use its InstanceReference ID). This detail is managed by the API so users of the API should not generally need to be aware of it.
      * API to list and manipulate instances.
      */
     async instanceRead(requestParameters: InstanceReadRequest): Promise<InstanceReferenceDetailed> {
         const response = await this.instanceReadRaw(requestParameters);
         return await response.value();
     }
 
+    /**
```
clemente, line 243: @gabor-boros The _instance notes_  parts of the code (changing notes through a new API) don't seem directly related to this PR. Could you mention in the description why and how this is being changed? (and how to test it)? (or maybe remove this part of the code)

gabor-boros: @clemente The js/ts client is updated (`npm run update-api-client`), hence some changes which may not be related completely to this change. (Also added this to the description).

.env.test
```diff
@@ -9,7 +9,14 @@ OPENSTACK_PASSWORD='pass'
...
 OPENSTACK_AUTH_URL='http://example.com/auth/url'
 OPENSTACK_REGION='test-region'
-DEFAULT_INSTANCE_BASE_DOMAIN='example.com'
+
+# Since the DEFAULT_INSTANCE_BASE_DOMAIN is used for defining the help
+# text of the BetaTestApplication as well, we need to set this value to
+# our production hosting url. Although this sounds risky, since the API
+# key of Gandi is fake, we cannot modify anything on the production records
+# using unit tests.
+DEFAULT_INSTANCE_BASE_DOMAIN='opencraft.hosting'
```
swalladge, line 18: @gabor-boros 

> Since the DEFAULT_INSTANCE_BASE_DOMAIN is used for defining the help text of the BetaTestApplication as well,

Why is this a problem? What's wrong with it defaulting to example.com?

frontend/packages/api_client/src/apis/V1Api.ts
swalladge: Yep looks like this regen was missed when we added the instance notes api endpoint. :+1:

registration/api/v2/serializers.py
```diff
@@ -315,6 +344,88 @@ def validate_user(self, value):
...
+
+        for domain in domains:
+            try:
+                managed_domains = gandi_api.filter_dns_records(domain)
+                records = {tldextract.extract(record["content"]) for record in managed_domains}
+            except Exception as exc:
+                logger.warning("Unable to retrieve the domains for %s: %s.", domain, str(exc))
+                raise ValidationError("The domain cannot be validated.")
+
+            # Because manually registered CNAMEs may have a dot (.) in their subdomain
```
swalladge, line 419: note that subdomain nesting is infinite... nothing is stopping someone registering `a.b.c.d.e.f.g.h.i.example.com`.

registration/api/v2/serializers.py
```diff
@@ -315,6 +344,88 @@ def validate_user(self, value):
...
+            raise ValidationError("This domain is already taken.")
+
+        domains = set([
+            settings.DEFAULT_INSTANCE_BASE_DOMAIN,
+            settings.GANDI_DEFAULT_BASE_DOMAIN,
+        ])
+
+        for domain in domains:
+            try:
+                managed_domains = gandi_api.filter_dns_records(domain)
```
swalladge, line 413: This is going to return a _lot_ of records in production. Not sure we can do this. It may be easier and just as affective to have a blacklist of subdomains we definitely don't want used (eg. `*.net.opencraft.hosting` covers most cases) - expanding on the `_validate_subdomain_not_contains_reserved_word` method.

swalladge (changes requested):
Looking pretty good in general, however I have some reservations about the validation algorithm in general... (see comments here and on jira)

instance/gandi.py
```diff
@@ -136,6 +136,36 @@ def add_dns_record(self, record):
...
         return result
 
+    def list_dns_records(self, record):
+        """
+        List all records of a domain name for a given type.
+        """
+        lexicon_config = self._get_base_config()
+        lexicon_config['domain'] = record["domain"]
+        lexicon_config['action'] = 'list'
+        lexicon_config['type'] = record["type"]
```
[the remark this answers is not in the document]
gabor-boros: Fair point, I used to write double quotes so I'll need to keep myself to single quotes. Nice finding!

registration/api/v2/serializers.py
```diff
@@ -288,6 +306,17 @@ class OpenEdXInstanceConfigSerializer(serializers.ModelSerializer):
     public_contact_email = serializers.EmailField(required=False)
     is_email_verified = serializers.BooleanField(source='email_addresses_verified', read_only=True)
 
+    def _validate_subdomain_not_contains_reserved_word(self, subdomain):
```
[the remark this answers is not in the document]
gabor-boros: I'll adjust it soon.

registration/api/v2/serializers.py
gabor-boros: Hm, the wording here maybe not the best, though there is an existing test case (as far as I remember) and an attached sample for that ((3) Subdomain of an existing instance's domain <- in the description) so that's checked as well and working. I'll rephrase the comment

registration/api/v2/serializers.py
gabor-boros: I left a comment regarding this on the Jira task, but in general:

I don't feel that filtering over those records should be a big issue in the production environment considering the number of records, response time of the API and the ability to use a cache (if we want).

The reason I'd check the existing records anyway is that the effect on our DNS settings is much bigger if someone uses a not blacklisted record (which should be) than for the users to wait ~300 ms on the API request. I'm thinking about a scenario when a user is capable to overwrite for example a k8s cluster what we are building (in theory), because we forgot to list that in the blacklist.

I think it would be safer on the long run to check the records than facing with issues on the production env caused by a user.

registration/api/v2/serializers.py
```diff
@@ -315,6 +344,88 @@ def validate_user(self, value):
...
+            # No need to check for subdomain, since it will match anyway
+            | Q(extra_custom_domains__contains=domain)
+        ).exists()
+
+        if is_taken:
+            raise ValidationError('This domain is already taken.')
+
+        return value
+
+    def validate_subdomain(self, value) -> str:
```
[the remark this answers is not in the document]
gabor-boros: @swalladge I thought about this as well. The endpoint will do the `validate_*` checks first, then (when saving the db instance) runs the model validators too.

Right now OpenCraft members would be able to create subdomains which may overwrite existing domain records, though we don't create instances from shell for customers (or testing purposes) in production environment (note that stage has a different base domain so, that wouldn't affect the DNS records).

What is your opinion @swalladge, should I merge the available subdomain validator to the existing validators? (I'm up for it since that would result in a more secure solution)

registration/api/v2/serializers.py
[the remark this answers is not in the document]
gabor-boros: I decided to use staticmethod decorator here, since we have no other places where we would like to validate subdomains. (except if we decide to use that on model validator level as well)

.env.test
gabor-boros: Fair point, I wrote that comment before I hardcoded the help text of the model field. I'll adjust the code accordingly.

gabor-boros: Resolving this conversation since https://github.com/open-craft/opencraft/pull/643/commits/7444bddb2d8a9d66fac7880e4d2ba9c2ae382917 is committed

registration/api/v2/serializers.py
gabor-boros: This is what you thought about, right?

instance/gandi.py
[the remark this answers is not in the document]
gabor-boros: Resolving conversation since I changed the double quotes to single ones in the file (other files left alone, to not mix up quotation styles)

registration/api/v2/serializers.py
swalladge: Yep, resolved on Jira. This is fine. :+1:

registration/api/v2/serializers.py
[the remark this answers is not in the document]
swalladge: Yep probably better to merge so all the validators are in the same place. 

> Right now OpenCraft members would be able to create subdomains which may overwrite existing domain records

I think, based on our workflows, we should restrict everyone here. This is only for the beta test application model, so if on opencraft member needs to do set a domain that wouldn't validate for any reason, they can do that through the openedx instance model.

registration/api/v2/serializers.py
swalladge: @gabor-boros oh something I just thought of - is this serializer also used for updates? If so, is this then going to fail validation if the domain name wasn't changed, because that domain name will now be in the dns records?

gabor-boros: @swalladge Nice thought, I'll double check it 👍

registration/api/v2/serializers.py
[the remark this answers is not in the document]
gabor-boros: Consider my thoughts about validation call order invalid. I mixed up. The model validator will run first, then the serializer one.

registration/models.py
```diff
@@ -305,7 +303,7 @@ def domain(self):
         """
         The full domain requested for this application.
         """
-        return '{0}.{1}'.format(self.subdomain, self.BASE_DOMAIN)
+        return '{0}.{1}'.format(self.subdomain, settings.DEFAULT_INSTANCE_BASE_DOMAIN)
```
clemente, line 306: I don't know why `self.BASE_DOMAIN` was here in addition to being in settings (`DEFAULT_INSTANCE_BASE_DOMAIN`), but I guess it's because of Ocim stage: Ocim stage has the .env setting `DEFAULT_INSTANCE_BASE_DOMAIN=stage.opencraft.hosting`, but the code was hardcoding `opencraft.hosting` (in `self.BASE_DOMAIN`) maybe as a way to make Ocim Stage show the same behaviour as Ocim production, which can be desirable.
With the new code, Ocim Stage users would get `mysubdomain.stage.opencraft.hosting` by default whereas they were getting `mysubdomain.opencraft.hosting` before. I think this difference  doesn't matter because we don't have users registering in Stage, much less with custom domains. But it will make a difference when we test this PR in Stage.

gabor-boros: Yes, I agree. We will need to make sure testing on stage properly, to not break even our own users experience on stage. 😇

registration/api/v2/serializers.py
gabor-boros: Rephrased

registration/api/v2/serializers.py
gabor-boros: The final solution regarding serializers: The serializer will validate the field if and only if the field is changed compared to its previous state.

registration/api/v2/serializers.py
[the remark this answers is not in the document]
gabor-boros: This is fixed in 56747bf

registration/api/v2/serializers.py
gabor-boros: Also fixed in 56747bf

instance/gandi.py
```diff
@@ -136,6 +136,36 @@ def add_dns_record(self, record):
         result = client.execute()
         return result
 
+    def list_dns_records(self, record):
```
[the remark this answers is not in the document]
gabor-boros: I'm thinking about this change @swalladge, I can change this, but then it would break the convention we use at setting and removing DNS records (`set_dns_record` and `remove_dns_record`). What are your thoughts about it? Should we still change the `list_dns_records`' signature?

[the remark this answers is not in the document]
swalladge: Nah that's fine - better to be consistent here.  If we refactor things in the future it would be nice to define the types more clearly.

registration/models.py
gabor-boros: By the way, I just checked. The domain names we registered for OCIM stage environments are ending with `stage.opencraft.hosting`. Soooo, this may not be an issue

Screenshot 2020-09-28 at 7 57 45

registration/api/v2/serializers.py
```diff
@@ -338,9 +339,10 @@ def validate_external_domain(self, value):
         """
         Prevent users from registering with an external domain which was or currently in use.
         """
-        is_changed = self.instance and self.instance.external_domain == self.initial_data.get("external_domain")
+        is_new_instance = self.instance is None
+        is_changed = not is_new_instance and self.instance.external_domain == self.initial_data.get("external_domain")
```
swalladge, line 343: Is checking equality correct here? `self.instance.external_domain == self.initial_data.get("external_domain")` Shouldn't it be checking if it's not equal (`is_changed`)?

swalladge:
@gabor-boros :+1: awesome work! Please squash your commits to one or more logical changesets, and we can merge. :)

- [x] I tested this: followed test instructions, tested on stage
- [x] I read through the code
- [x] Includes documentation
- [ ] NA I made sure any change in configuration variables is reflected in the corresponding

gabor-boros:
@swalladge Thank you for the review and the approve!

Since the PR cannot be merged until it is not approved, could you please do that too? I'm planning to "squash and merge" if that's good for you too ☺️

swalladge (approved):

swalladge:
@gabor-boros done, sorry I missed that.

registration/api/v2/serializers.py
gabor-boros: Well, yes. That should be `!=` instead of `==`. Fixing, adding test and the I'll do the merge 😊

PR #644, opened 2020-09-18
Add settings flag to enable/disable trial instances report

pcockwell:
Followup to #637 that adds a settings variable to enable/disable the new task


lgp171188:
@pcockwell, as discussed on the task, this is not needed. Closing this now.

PR #647, opened 2020-09-22
[BB-3014] Disable the custom external domain feature in the Ocim frontend

shimulch:
This PR removes the ability to add external domain from OCIM console UI. As that feature requires more discovery and technical design to be done, we are temporarily disabling it.

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-3014

~~**Discussions**:~~ 

**Dependencies**: None

**Screenshots**: 
localhost_3000_registration_custom-domain (1)

~~**Sandbox URL**:~~ 

**Merge deadline**: "None" 

**Testing instructions**:

1. Run OCIM
2. Run console UI from ``frontend`` directory
3. Clicking on ``I want to use my own domain`` moves to the new contact page
4. ``/registration/custom-domain`` page doesn't exist

N/A

**Reviewers**
- [ ] @kaizoku 

~~**Settings**~~


kaizoku:
@shimulch, I still see the link for using a custom domain: 
Screenshot_2020-09-23 OpenCraft Instance Manager Console

The handler is gone, so clicking the link does nothing, but it would be confusing for users who see a link which doesn't work.

It looks like the frontend tests are still failing. We should fix those as well.


Also, I think we're better off deleting code rather than commenting. If we want to reinstate this feature later we can reference or revert these commits that removed it.

shimulch:
@kaizoku Thanks for your review.
> The handler is gone, so clicking the link does nothing, but it would be confusing for users who see a link which doesn't work.

I kept that to put contact link there. Suggest me if I should change the text to something else.

> It looks like the frontend tests are still failing. We should fix those as well.

Actually it was not failing but 2 obsolete tests were there. Those were related to custom domain and I previously commented those. Removed them. Now js tests seems fine. But the current failure seems to be something unrelated and due to network timeout from subdomain form submission.

> Also, I think we're better off deleting code rather than commenting. If we want to reinstate this feature later we can reference or  revert these commits that removed it.

Removed all commented code. But files related to external domain feature are still there, just not being called.

kaizoku:
Thanks @shimulch, this is looking pretty good.
The styling or spacing looks a bit different to me than in your screenshot though:
Screenshot_2020-10-01 OpenCraft Instance Manager Console

Could we change this instead to "To use custom domain, please contact us at contact@opencraft.com"? We won't have to worry about the spacing around the hyphen in that case either.

kaizoku:
Thanks for updating the text @shimulch. Looks good to me :+1:

- [✓] I tested this: (deployed in devstack, checked custom domain link and contact page)
- [✓] I read through the code
- [✓] I checked for accessibility issues

kaizoku (approved):

PR #648, opened 2020-09-24
[SE-3340] Introduce separate stage and prod deploys

gabor-boros:
This PR changes the CI/CD logic to be able to deploy to stage and production environments separately in addition that the production deploys will require manual approval as well.

**Dependencies**:

- [x] Create stage branch called `stage`
- [x] Create production branch called `production`
- [ ] After approval but before merge, notice everyone about the upcoming change

**Merge deadline**: None

**Testing instructions**:

N/A

* Production branch must have the same branch protection settings that master has
* I have **no idea how to test this change before merging it**, but if you have one, please let me know and I'll adjust the Testing section.
* I have **no rights yet** to create the branches mentioned above, so I'll need a core member for that


circle.yml
```diff
@@ -332,7 +332,7 @@ workflows:
           context: ocim-frontend-stage
           requires:
             - frontend-test
-      # Frontend deployment (only on master branch)
+      # Frontend deployment (only on master and stage branches)
       - frontend-deploy:
           name: frontend-deploy-stage
```
giovannicimolin, line 337: Deploy stage should only happen in the `stage` branch, not on `master` too. Otherwise merges to `master` will overwrite changes in `stage`

circle.yml
```diff
@@ -369,5 +375,6 @@ workflows:
             branches:
               only:
                 - master
+                - stage
```
giovannicimolin, line 378: Cleanup workflow should only run on `master`, this is related to integration tests and automatically clears all dangling VMs and resources.

gabor-boros: @giovannicimolin Fare point. Reverting this part.

circle.yml
gabor-boros: @giovannicimolin For the first time I had the same opinion, though in case we deploy backend from master to staging automatically and deploy frontend by merging to staging branch, that would make really hard to follow what is happening in the staging environment. 

In case we deploy completely separately (backend automatically by merging to master, frontend manually by merging to stage) we could encounter with the same issue what we have right now, but for the other direction, or am I missed something?

gabor-boros: I overlooked this, since backend is not deployed automatically to stage.. Updating the PR in minutes to cover the suggestion.

gabor-boros: Changes are covered by 1899b54

mtyaka (approved):
Changes look good to me, but please remember to update the [frontend/README](https://github.com/open-craft/opencraft/blob/0e0ff261c9cabbae3bfa3e93bd85aab0a47c56ec/frontend/README.md#deployment-process) file before merging :+1:

- [x] I tested this: I created a `stage` branch and pushed the changes form this branch to it. The [frontend-build-test-and-deploy](https://app.circleci.com/pipelines/github/open-craft/opencraft/2286/workflows/13259a11-c288-4ddd-bbe9-fa0b705c8773) job was triggered, as expected.
- [x] I read through the code
- [ ] Includes documentation - **pending**

circle.yml
```diff
@@ -341,7 +341,7 @@ workflows:
           filters:
             branches:
               only:
-                - master
+                - stage
       # Deploy production
       - frontend-build:
           name: frontend-build-production
```
mtyaka, line 347: @gabor-boros I just noticed that the `frontend-build-production` step still requires `frontend-deploy-stage`, which I don't think is appropriate now that stage and production branches are separate. I think it should require `frontend-test` instead.

gabor-boros: My initial idea was to have at least one deploy of the change on stage to be able to deploy on production, but that would cause more confusion, so I just changed this in 1f74190 for the behaviour you just wrote.

mavidser:
Just noting that this is still in my queue, and will get to reviewing this tomorrow :)

mavidser:
@gabor-boros I tested this by pushing this branch (rebased on master) on stage, and the workflow deployed as expected.
However creating [`release-se-3340_test1`](https://github.com/open-craft/opencraft/releases/tag/release-se-3340_test1) tag just created a pipeline with no workflow [(#2494)](https://app.circleci.com/pipelines/github/open-craft/opencraft) - I'm not completely sure why, but this would need some looking into.

gabor-boros:
So, the tag creation triggering is fixed now. It turned out that **ALL** dependencies **must** have the tags filtering as well. Otherwise those won't be started. Without those dependencies the jobs won't start, so nothing will start at all. The documentation on this behaviour wasn't the clearest. Sorry about the mess.

gabor-boros:
Screenshot 2020-11-10 at 13 56 57

FTR: I stopped that job to not deploy to prod

mavidser (approved):
Great work @gabor-boros, this is good :+1:

- [x] I tested this: Deployed to stage and prod using the new pipelines
- [x] I read through the code

PR #650, opened 2020-09-28
[SE-3362] Send periodic build failure notifications

gabor-boros:
Make sure to send periodic build failure notifications if periodic builds are enabled on the instance.

**JIRA tickets**: https://openedx.atlassian.net/browse/BTR-8

**Screenshots**: 
N/A

**Sandbox URL**:
N/A

**Testing instructions**:

1. Create an instance
2. Enable periodic builds on the instance
3. Add your email address to the Periodic build notification emails text field
4. Set `00:05:00` as build interval (5 minutes)
5. Misconfigure it to no be able to provision (for example change the configuration repo branch to `notexistsatall`)
6. Wait for the email (check your spam folder as well) - It should arrive in 5-10 minutes based on the delay in the queue

1. The testing of this functionality is time consuming. In case you would like to check logs instead of emails, you should look out for the "`Sending notification e-mail to %s after instance %s didn't provision`" log line.

**Reviewers**
- [ ] @toxinu


instance/models/mixins/utilities.py
```diff
@@ -137,6 +137,46 @@ def _mail_admins_with_attachment(
 
 # Functions #####################################################################
 
+def send_urgent_deployment_failure_email(recipients: List[str], instance_name: str) -> None:
+    '''
```
toxinu, line 141: Can you use triple double quotes please (for the docstring)?

instance/models/mixins/utilities.py
```diff
@@ -137,6 +137,46 @@ def _mail_admins_with_attachment(
...
+    Send notification email about failed periodic deployments.
+    '''
+    logger.warning(
+        "Sending notification e-mail to %s after instance %s didn't provision",
+        recipients,
+        instance_name,
+    )
+
+    send_mail(
```
toxinu, line 170: Since the email will be sent to a public mailing list, most of user won't have access to the web interface. Do you think we can append the actual error without worrying about leaking internal stuff?

toxinu (approved):

instance/models/mixins/utilities.py
gabor-boros: Oh, sure. I overlooked this.

instance/models/mixins/utilities.py
gabor-boros: @toxinu This is what I'm thinking about as well. Since the error messages are in logs I would feel uncomfortable what kind of logs we share automatically. It would be bad to leak some secrets or internal stuff to public mailing lists, though that would be extremely helpful for them to see at least one error message.

I'll check how could we filter those log lines and share some with them in a secure way (if we can).

toxinu: Thanks @gabor-boros. Yes please go ahead and spend some time on it otherwise this feature will be very limited for the community.

gabor-boros: @toxinu I checked it.

* As far as I can see, we are not masking the log output, so in case a sensitive data is there, we would forward it.
* Although I could filter those logs which are starting with "fatal:" or "failed:" the issue is the same (sensitive data filtering)
* If there is no sensitive data in fatal/failed, that maybe more misleading to sand that instead of sending nothing, because those logs are rarely meaningful without context (other info logs as well)

We would be contacted anyway to get more info, since to figure out the root cause usually requires the context. As an example - this is a fatal log line:

```
fatal: [149.202.185.232]: FAILED! => {"before": "****", "changed": false, "msg": "Local modifications exist in repository (force=no)."}
```

Without context this is not really helpful, I think. (the **** masking is made by me)

gabor-boros: I could dig deeper in this topic, but from what I've seen, without being able to mask potential sensitive data (1), it would be risky to share the raw logs.

Notes:
1. It is not confirmed that the log contains any sensitive data, this is an assumption

gabor-boros: @toxinu I made some research. It seems that usernames will be in the output for sure (like mongodb username) but as far as I can see, this is not true for passwords (because the 3rd party tools are not printing them to stdout/stderr). If we can trust 3rd parties (like mongodb, ansible, etc) that they wont print secrets we are good to send the lines to public mailing lists. Else, I can write a simple masking based on a collection of regular expressions, but that will be far from perfect or secure.

What do you think Geoffrey?

toxinu: Thanks for your investigation @gabor-boros. I think we prefer to send less logs (or not at all) than taking the risk to leak secrets. At least OpenCraft employees and community will know when a periodic build is failing and then an OpenCraft employee will be able to share additional logs and context to the mailing list.
I agree that this is not perfect but it sounds more reasonable to me, what are your thoughts @gabor-boros?

gabor-boros: @toxinu I agree with you! In that case when I'm getting back to this later today, I'll merge and deploy this change. Also, I'll take care of the community Jira task as well

PR #651, opened 2020-10-02
SE-3310 Ensure prod instances email urgent on monitoring fail and alerts auto close once resolved

samuelallan72:
Otherwise we only get paged on a provision failure; ping monitoring
previously would only email the settings.ADMINS emails.

**Test instructions**:

- register a new account, choose instance settings, and confirm your email
- view the settings of the new instance
- verify that additional_monitoring_emails has the PROD_APPSERVER_FAIL_EMAILS added
- verify that the corresponding newrelic alert policy for the new instance has ADMINS _and_ PROD_APPSERVER_FAIL_EMAILS set in the notification channels
- do something to take the instance offline (change the dns records, stop the lms service, etc.)
- verify that newrelic alerts open
- put the instance back online
- verify the newrelic alerts/incidents auto close after a few minutes.

**Reviewers**:

- [x] @giovannicimolin 
- [ ] @lgp171188

### DIFF
--- a/registration/provision.py
+++ b/registration/provision.py
@@ -93,6 +93,7 @@ def _provision_instance(sender, **kwargs):
 
         if settings.PROD_APPSERVER_FAIL_EMAILS:
             application.instance.provisioning_failure_notification_emails = settings.PROD_APPSERVER_FAIL_EMAILS
+            application.instance.additional_monitoring_emails = settings.PROD_APPSERVER_FAIL_EMAILS
             application.instance.save()
         application.save()
         # At this point we know the user has confirmed their email and set up an instance.
--- a/instance/newrelic.py
+++ b/instance/newrelic.py
@@ -184,7 +184,7 @@ def add_alert_nrql_condition(policy_id, monitor_url, name):
                 'type': 'static',
                 'name': name,
                 'enabled': True,
-                'value_function': 'single_value',
+                'value_function': 'sum',
                 'terms': [{
                     'duration': settings.NEWRELIC_NRQL_ALERT_CONDITION_DURATION,
                     'threshold': '1',
--- a/instance/tests/test_newrelic.py
+++ b/instance/tests/test_newrelic.py
@@ -254,7 +254,7 @@ class NewRelicTestCase(TestCase):
                 'type': 'static',
                 'name': condition_name,
                 'enabled': True,
-                'value_function': 'single_value',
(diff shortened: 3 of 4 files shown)


registration/provision.py
```diff
@@ -93,6 +93,7 @@ def _provision_instance(sender, **kwargs):
 
         if settings.PROD_APPSERVER_FAIL_EMAILS:
             application.instance.provisioning_failure_notification_emails = settings.PROD_APPSERVER_FAIL_EMAILS
+            application.instance.additional_monitoring_emails = settings.PROD_APPSERVER_FAIL_EMAILS
```
lgp171188, line 96: @swalladge, afaik, the monitoring SLA is applicable only for paying customers - P&T and institutional clients and that is why we add `urgent@` manually to those instances.

giovannicimolin: @lgp171188 This is the `production_instance_factory`, so it's fine. If a developer ends up spawning an appserver that fails, it's his responsibility of cleaning it up/addressing the page. 
I'd prefer keep this explicit mechanism instead of having to manually set the monitoring.

giovannicimolin: Btw @swalladge, take a look at this: https://github.com/open-craft/opencraft/pull/489 (your previous PR here).
Do we really need this change here? I assumed we needed to set the emails manually, but for P&T, they are already added, so no problem.

swalladge: @giovannicimolin according to the original requirements, yes we did need this change.

> Doing this because we only want to auto set PROD_APPSERVER_FAIL_EMAILS for end user instances.

You can check SE-1832, but I believe the reason to move things to here, is that now they will be autoset up for new instances that are set up through the betatest registration process (our standard documented process for setting up client and maintained instances).

swalladge: Also, provisioning_failure_notification_emails != additional_monitoring_emails, so I'm not sure what you're asking. :thinking:

giovannicimolin: Yup, I think I've mixed up those two.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Tested the change on a live monitor in Ocim production.
- [x] I read through the code: simple diff.

@swalladge Good to merge & test on production monitors.

swalladge:
Merging because the integration test failures are unrelated.

PR #653, opened 2020-10-02
[BB-2164] Erase Ocim trial user data after X months

shimulch:
This PR introduces a new periodic task, that deletes old beta test users.

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-2164

~~**Discussions**:~~

**Dependencies**: None

~~**Screenshots**:~~

~~**Sandbox URL**:~~

**Merge deadline**: "None"

**Testing instructions**:

1. Create couple of beta test instances
2. Set ``INACTIVE_OLD_BETATEST_USER_DAYS = 1``
3. Set ``DELETE_OLD_BETATEST_USER_DAYS = 1`` 
4. Manually change ``BetaTestApplication.created`` to an old date
5. Manually change corresponding user's ``last_login`` to an old date
6. Delete deployed instances, keep ``BetaTestApplication`` model instance
7. Run Huey 
8. Users should be inactive now.
9. Change ``UserProfile.modified`` to an old date
10. Run Huey
11. Users will be deleted, also all data with ``on_delete=CASCADE`` will be deleted.

1. Not sure if this is what we should do

**Reviewers**
- [ ] @0x29a 

~~**Settings**~~

### DIFF
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -305,8 +305,7 @@ if settings.CLEANUP_OLD_BETATEST_USERS:
             days=settings.DELETE_OLD_BETATEST_USER_DAYS
         )
 
-        User = get_user_model()
-        queryset = User.objects.filter(
+        queryset = get_user_model().objects.filter(
             # has a beta test application
             betatestapplication__isnull=False,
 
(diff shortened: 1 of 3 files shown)


instance/tasks.py
```diff
@@ -268,3 +270,73 @@ def delete_old_logs():
...
+
+            # are not staff or superuser
+            is_staff=False,
+            is_superuser=False,
+
+            # don't belong to any organization (ex. opencraft)
+            profile__organization=None,
+
+            # not paid user
+            profile__accept_paid_support=False,
```
0x29a, line 320: Looks like this field doesn't mean paid user, and contains `True` for all users: https://github.com/open-craft/opencraft/blob/19465bda29ee9c4f26f2267dca3c65bc54f6bac5/registration/api/v2/serializers.py#L192-L198

instance/tasks.py
```diff
@@ -268,3 +270,73 @@ def delete_old_logs():
...
+
+            # instances doesn't exists or already deleted
+            profile__instancereference__isnull=True,
+            openedxinstance__isnull=True,
+            openedxappserver__isnull=True,
+            betatestapplication__instance__isnull=True,
+
+            # at least specified days old
+            betatestapplication__created__lte=inactive_cutoff,
+            last_login__lte=inactive_cutoff
```
0x29a, line 330: I found user in my local OCIM, that meets all other criterias, but has `last_login` == `None`. Probably, we should cover this case also?

0x29a (changes requested):
Looks good. Though, I left a couple of minor notes. Please address them, write test for `cleanup_old_betatest_users` and this is good to go.

instance/tasks.py
shimulch: Ok. I didn't notice that. Removed that condition.

instance/tasks.py
shimulch: Added in the condition

shimulch:
@0x29a, thanks for the confirmation and review. Added tests and addressed your reviews.

0x29a (approved):
:+1:

- [x] I tested that task works properly and deletes users, that meet conditions described in task docstring.
- [x] I read through the code
- [x] Includes documentation

PR #658, opened 2020-10-05
Revert "[BB-3014] Disable the custom external domain feature in the Ocim frontend"

gabor-boros:
Reverting open-craft/opencraft#647 since it removes the external domain registration instead of hiding that.


lgp171188 (approved):
:+1: 
- [x] I checked that this is a plain revert of the commit which broke the production registration form.

lgp171188:
@gabor-boros, CC @shimulch, @giovannicimolin force merging this before the CI checks pass (frontend tests passed) to fix production. FYI.

PR #662, opened 2020-10-13
Unify section/header formatting

deepbrook:
Clean up the use of header formats ('#' vs '-').
This should make it easier to differentiate between the
local-setup specific commands, and common operations
on vagrant boxes.

Please let me know if the sections I separated are in fact
part of the local setup after all - I'll adjust the headers
accordingly.

### DIFF
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -130,8 +130,11 @@ everything is working correctly:
 
     make test.unit
 
-Migrations
-----------
+## Common Operations
+The following commands are meant to be run from within your `vagrant` box,
+after logging into them via `vagrant ssh`.
+
+### Migrations
 
 To run database migrations:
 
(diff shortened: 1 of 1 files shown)


giovannicimolin:
@deepbrook Which ticket is this for? The Ocim UX one?

deepbrook:
@giovannicimolin ah no! This was just something I stumbled upon while reading the docs, which had me confused at first. It's not related to any ticket per se (perhaps my onboarding ticket, since that's the one I assigned my time to for reading this).

gabor-boros:
@giovannicimolin What do you think if we merge it?

giovannicimolin:
@gabor-boros I think this will break our documentation, no? Not all places use this formatting.
If it doesn't break ocim.opencraft.com as is we can merge it, otherwise let's just close it.

PR #663, opened 2020-10-14
[SE-3487] enable LOS for NRQL alert conditions

gabor-boros:
Based on NewRelic's [announcement](https://docs.newrelic.com/docs/release-notes/platform-release-notes/alerts-release-notes/09-28-2020-action-required-new-relic-one-streaming-alerts-nrql-conditions), we must change our NRQL alert condition to enable loss of signal detection. Also, based on my understanding, in our use case it worth to fill the gap with static "0" values to indicate we did not receive data during that time period. This way we would remain "backward compatible" in terms of the shape of the data.

This change will require us to "re-install" **all** conditions we have, which means that first we will delete them, then add it again. An alternate solution could be to update all conditions with the "Migrator" application NewRelic provided to help the migration. Both directions can work, but I'd go with re-installing the conditions OCIM knows.

For those conditions which are added manually (like which checks OCIM itself) we need to use the Migrator application.

**Screenshots**: 

Screenshot 2020-10-14 at 13 50 18

**Sandbox URL**: N/A

**Testing instructions**:

1. Go to [NewRelic](https://one.newrelic.com/launcher/nrai.launcher?pane=eyJuZXJkbGV0SWQiOiJhbGVydGluZy11aS1jbGFzc2ljLnBvbGljaWVzIiwibmF2IjoiUG9saWNpZXMiLCJzZWxlY3RlZEZpZWxkIjoidGhyZXNob2xkcyIsInBvbGljeUlkIjoiMTA4NTk4MSIsImNvbmRpdGlvbklkIjoiMTYyMzQyMzYifQ==&sidebars[0]=eyJuZXJkbGV0SWQiOiJucmFpLm5hdmlnYXRpb24tYmFyIiwibmF2IjoiUG9saWNpZXMifQ==&platform[accountId]=1346955) and check the policy created from stage env and check the "Thresholds" section. It must contain the new "Loss of Signal" related changes (Signal lost after settings)
2. Go to OCIM production and get the NewRelic related settings
3. Go to OCIM stage, checkout this branch and add the copied settings to .env
4. Restart the shell on stage
5. Get a random (successfully provisioned) instance's ID
6. Get the instance object from OpenEdXInstance model
7. Call the `enable_monitoring()` function on the instance
8. Check that NewRelic created the monitoring policy and set the desired parameters
9. Revert the changes on stage to use master branch - including deleting the copied NewRelic settings

1. The REST documentation is not updated yet, but the API already knows the new options. For the APIs request scheme, please check [it's playground](https://rpm.newrelic.com/api/explore/alerts_nrql_conditions/create). For the documentation of NerdGraph - which already contains the documentation of the new functionality - , you can check [this documentation](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/alerts-nerdgraph/nerdgraph-api-loss-signal-gap-filling#h3-create-a-new-condition-with-loss-of-signal-settings).
2. Based on the UI (I did not find documentation about that though) the `expiration_duration` must be at least `60` seconds.
3. I set `expiration.closeViolationsOnExpiration` to `False` to be more explicit, though its default value is `False`

**Reviewers**
- [ ] @Agrendalath 
- [ ] @mtyaka


Agrendalath (approved):
:+1:
* [x]  I tested this: followed the testing instructions
* [x]  I checked for accessibility issues: n/a
* [x]  Includes documentation: n/a

PR #665, opened 2020-10-16
[SE-3470] Add failure context to periodic build notifications

gabor-boros:
This PR is adding more context for to the periodic provisioning failure notifications. Also, it adds an extra functionality to send "ack" email when a provisioning succeeds after a failure. The following extra information will be provided:

- [x] branch
- [x] last commit date
- [x] configuration branch
- [x] extra config parameters (filtered/masked)
- [x] last successful deployment (to see if it is continuously broken or a one-time issue -> during a weekend, the periodic build is ran 4 times at least)
- [x] last relevant ansible failure log (filtered/masked)

**Screenshots**: N/A

**Sandbox URL**: N/A

**Testing instructions**:

1. Create an instance
2. Enable periodic builds on the instance
3. Add your email address to the Periodic build notification emails text field
4. Set `00:05:00` as build interval (5 minutes)
5. Misconfigure it to no be able to provision (for example change the configuration repo branch to `notexistsatall`)
6. Wait for the email (check your spam folder as well) - It should arrive in 5-10 minutes based on the delay in the queue
7. Check the email has no sensitive data and has the desired (listed above) content.
8. Fix the configuration
9. Wait for the ack email - It should arrive in 5-10 minutes based on the delay in the queue

1. The sensitive filtering is based on best effort, hence it won't replace the matching text parts but the whole text. This was safer then experimenting with something else.

**Reviewers**
- [x] @Kelketek


instance/tests/models/test_mixin_utilities.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        Test nothing is sensitive in the given data.
+        """
+        with SensitiveDataFilter(data) as filtered_data:
+            self.assertEqual(data, filtered_data)
+
+    @ddt.data(
+        ("username:password", SensitiveDataFilter.FILTERED_TEXT),
+        ("api-abc", SensitiveDataFilter.FILTERED_TEXT),
+        ("api_abc", SensitiveDataFilter.FILTERED_TEXT),
```
Kelketek, line 48: Can we add tests to make sure this is case insensitive?

instance/tests/models/test_mixin_utilities.py
```diff
@@ -0,0 +1,296 @@
...
+
+        task_name, log_entry = get_ansible_failure_log_entry(self.appserver.log_entries_queryset)
+
+        self.assertEqual(task_name, expected_task_name)
+        self.assertDictEqual(log_entry, expected_log_entry)
+
+    @override_settings(LOG_LIMIT=2)
+    def test_entry_and_task_name_not_found_within_the_log_limit(self):
+        """
+        Test even the log entries contains the information needed, the log limit would be violated
```
Kelketek, line 284: The wording here's a bit confusing. Would the following work and be accurate?

"Test that the log entries don't contain information past the cutoff point even if those entries might have matched otherwise, and it would result in the log being empty."

Kelketek (changes requested):
Looks great, @gabor-boros ! Just a couple of notes.

gabor-boros:
Thank you @Kelketek for the awesome comments! I'll adjust the code in an hour

Kelketek:
:+1:

- [ ] ~~I tested this: (describe what you tested)~~
- [X] I read through the code

toxinu:
Merging it since @Kelketek approved it.

PR #666, opened 2020-10-27
[SE-3470] Adjust periodic build failure email content

gabor-boros:
Fix the most relevant log collection and attach all the Ansible produced logs to the email. Also, make sure that the configuration is sent as an attachment instead of a plain text within the email.

**Structure of the email**

```

Subject: Deployment failed at instance: {instance_name}

Body:
The periodic deployment of {edx_platform_release} failed. Please see the details below.
 
Ansible task name: {ansible_task_name}
Relevant log lines:
{relevant_log_entry}
 
AppServer ID:                     {appserver_id}
Latest successful provision date: {latest_successful_provision_date}
Configuration source repo:        {configuration_source_repo_url}
Configuration version:            {configuration_version}
OpenEdX platform source repo:     {edx_platform_repository_url}
OpenEdX platform release:         {edx_platform_release}
OpenEdX platform commit:          {edx_platform_commit}
 
Attachments:
- build_log.txt -> This contains all the Ansible build logs
- configuration.json -> This contains the configuration we created the appserver with
```

**Screenshots**: Always include screenshots if there is any change to the UI. Can be useful for people that are not reviewer but want to know what's going on.

**Sandbox URL**: N/A

**Testing instructions**:

1. Checkout this code on stage environment
2. Create an instance
3. Enable periodic builds on the instance
4. Add your email address to the Periodic build notification emails text field
5. Set 00:05:00 as build interval (5 minutes)
6. Misconfigure it to no be able to provision (for example change a database password or similar to have build logs and not fail immediately)
7. Wait for the email (check your spam folder as well) - It should arrive in 5-10 minutes based on the delay in the queue
8. Check the email has no sensitive data and has the desired (listed above) content.
9. Check the email has a `build_log.txt` attachment and it has content (with no secrets)
10. Check the email has a `configuration.json` attachment and it has content (with no secrets)
11. Fix the configuration
12. Wait for the ack email - It should arrive in 5-10 minutes based on the delay in the queue

1. Testing the flow could take from 5 minutes to 1 hour, depending on how fast the periodic task picked up by the workers.
2. Checking the whole build log is near impossible since it may have several thousand log records

**Reviewers**
- [ ] @Kelketek / @toxinu

### DIFF
--- a/instance/models/mixins/utilities.py
+++ b/instance/models/mixins/utilities.py
@@ -22,6 +22,7 @@ Instance app model mixins - Utilities
 
 # Imports #####################################################################
 
+import ast
 import re
 import logging
 import sys
@@ -72,7 +73,7 @@ class SensitiveDataFilter:
 
     SENSITIVE_VALUE_PATTERNS: list = COMMON_PATTERNS + [
         # This will match is extremely broad, but the safest
-        re.compile(r".*\:.*"),
+        re.compile(r"^(?!/).*\w+\:[\w\#\=\_\-\*\!\+\$\@\&\%\^]+"),
     ]
 
     def __init__(self, data: DataType):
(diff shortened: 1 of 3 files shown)


Kelketek:
@gabor-boros The code looks solid to me, but I'm unable to test this at the moment due to the ongoing problems with OVH. I'm putting this in the blocker status for now. If you find out before I do that it's clear to review, let me know.

FYI, I'm expecting to be out tomorrow, so if this is pressed for time, you may need to ping a firefighter. Otherwise, if deployments aren't working in time for me to review today, I expect I'll be able to hit it Friday.

nedbat:
The email seems great: move the config and build log into attachments so the are available but not distracting.

BTW: the PR description says:
```
- build_log.txt -> This contains the
```
I assume it meant to say, "This contains the complete build log"

gabor-boros:
@nedbat , Thank you for reviewing the new structure of the email!

Yes, I meant that "This contains all the Ansible build logs". Fixed it in the description.

Kelketek (approved):
:+1: Pending successful build.

- [X] I tested this: I was able to get a proper failure email with more useful context and logs.
- [X] I read through the code

gabor-boros:
@nedbat The additional build logs will contain the following ansible messages parsed parameters:

* `cmd` (if there is any for the log)
* `msg` (if there is any for the log)
* `stdout` (if there is any for the log)
* and `stderr` (if there is any for the log)

In my opinion, the most relevant log should be enough though since it is the last error message which broke the build. The reason I not included every log line and tried to filter those which has the more chance having relevant information is that ~20.000 lines of log generated during a build, out of this ~2400 ansible related regardless of the existence of stdout/err or cmd/msg fields. Based on this sending every log line could be a noise rather a help. Depending on where the build fails, the additional logs should be around few hundred lines (so ansible logs which produced the 4 kind of fields above).

In case the provided logs are not enough for a possible debugging or to reproduce a failure, I'm more than happy to include more logs, feel free to ping me anytime!

PR #670, opened 2020-11-12
Increase servername prefix length to 30 chars

mavidser:
Some instances with very similar names are generating the same servernames (Ironwood and Juniper periodic instances). This aims to be a fix for those. Keeping a reasonable limit to prevent excessively large hostnames, but a 10 character increase should do fine.

**JIRA tickets**: [SE-3484](https://tasks.opencraft.com/browse/SE-3484)

**Testing instructions**:

1. Make sure the unit tests pass (in the CI, or via `make test`)
2. Verify that stage works as expected. [This](https://stage.manage.opencraft.com/instance/5290/edx-appserver/2614/) is a sample instance generated using this branch.

**Reviewers**
- [ ] @Kelketek


Kelketek:
Can we have some update to the documentation or a comment somewhere in the code where someone might find it that could explain what's likely to be the cause if this happens in the future?

mavidser:
@Kelketek Sure, I missed updating that. Does this look okay now?

Kelketek:
@mavidser I'm not sure what our policy is for linking to internal tickets for open source documentation. If it's not a concern I'm good with that. Otherwise I'd remove that bit. I'll leave it to you to determine and fix as necessary.

:+1:

- [X] Verified the longer name length
- [X] I read through the code
- [X] Includes documentation

mavidser:
@Kelketek Thanks. The repo does refer to multiple SE-XXXX tickets as reference points in comments, so I guess it should be fine to link to those in the documentation too. I'll merge these once the CI completes the check :+1:

PR #673, opened 2020-11-24
[SE-3598] Set NRQL LOS to 10 mins from 1 minute

gabor-boros:
Based on the consultation with NewRelic, it seems that the loss of signal alerting is too sensitive. As they suggested, hereby I'm setting its value to 10 minutes. This would mean that if a network blip happens or the something happens on their side, that won't result in almost immediate notifications on our side.

**Screenshots**: N/A

**Sandbox URL**: N/A

**Testing instructions**:

N/A, though it could be tested by doing the following:

1. Enable monitoring for an instance on stage
2. Shut down the app server
3. In 10-12 minutes the notification should arrive

1. We will need to disable and enable notifications for all instances which already has monitoring to adjust the threshold.
2. Related conversation is linked to the ticket

**Reviewers**
- [ ] @mtyaka


mtyaka (approved):
Looking good @gabor-boros :+1:

- [ ] I tested this: I didn't test this, but based on the info provided by NewRelic staff, this should do the trick.
- [x] I read through the code
- [x] Includes documentation

PR #674, opened 2020-11-25
Prevent new deployments until first P&T instance is provisioned

mavidser:
This fixes a bug where deployments where the first deployments allowed to be cancelled. This PR also adds a is_successfully_deployed flag which is used to prevent user deployments if the first deployment never succeeded.

**JIRA tickets**: [SE-3517](https://tasks.opencraft.com/browse/SE-3517)

**Testing instructions**:

1. Register as a new P&T user on OCIM
2. Verify the email to start provisioning
3. Verify that you are not allowed to make new changes while the instance is being provisioned (neither before the appserver was created, nor after the appserver was created as there's a few mins of gap between the two depending on the queue size).
4. Terminate the first appserver being created - and make sure that the UI still doesn't allow the user to deploy changes (we wat a staff member to fix the first instance first).

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/instance/models/openedx_deployment.py
+++ b/instance/models/openedx_deployment.py
@@ -67,6 +67,7 @@ class OpenEdXDeployment(Deployment):
     # Field which denotes if the deployment was cancelled by the user
     cancelled = models.BooleanField(default=False)
 
+     # pylint: disable=too-many-return-statements
     def status(self):
         """
         Current state of deployment.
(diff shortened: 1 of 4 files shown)


mtyaka:
Changes look good @mavidser. Can you fix the tests and deploy to stage, so I can test there?

mtyaka:
Looks good @mavidser :+1: I wasn't able to test the UI part because it hasn't been deployed to stage, but I did confirm that the backend change prevents user from canceling an existing deployment or starting a new one even if the frontend appears to allow it.

- [x] I tested this: Verified that the changes prevent regular users from running redeployments before the first successful deployment.
- [x] I read through the code
- [ ] Includes documentation

PR #676, opened 2020-12-03
Update Celery configuration to keep using RabbitMQ as the backend

mtyaka:
This updates instance settings so that instances continue to work correctly using RabbitMQ as the Celery backend. This change is required since https://github.com/edx/configuration/pull/6147 switched to local Redis for native install.

We will switch to Redis in Ocim when we implement support for a Redis backend with ACLs, but for now this makes it possible to keep using RabbitMQ.

**Test instructions**:

I spawned two servers on Ocim stage:
- Juniper: https://stage.manage.opencraft.com/instance/6540/
- master: https://stage.manage.opencraft.com/instance/6530/

Check the instances to make sure they work correctly and the configuration looks good. You can use the default `staff@example.com` account.

Note: Celery tasks don't seem to be working on the master sandbox, but I think that's related to the problems on edx-platform master where tasks don't get registered properly (see https://github.com/edx/edx-platform/pull/25479#issuecomment-736073800).

**Reviewers**:

- [x] @0x29a

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -317,6 +317,8 @@ class OpenEdXConfigMixin(ConfigMixinBase):
 
             # RabbitMQ disabled locally
             "SANDBOX_ENABLE_RABBITMQ": False,
+            # Redis disabled locally
+            "SANDBOX_ENABLE_REDIS": False,
 
             # Ecommerce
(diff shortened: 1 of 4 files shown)


0x29a (approved):
:+1:

- [x] I tested this:
    1. Checked configuration of both instances.
    2. Signed up, enrolled in demo courses, checked that all tabs are working on both instances.
    3. Checked that Redis is not installed.
- [x] I read through the code

PR #679, opened 2020-12-19
[BB-3331] Add ability to configure appserver.yml version

0x29a:
## What this PR is about?
It adds the ability to configure default playbook options per instance.

Consider this snippet:
https://github.com/open-craft/opencraft/blob/2adf770f010f7a0c3a7e89a6d2da73052a3d28cb/opencraft/settings.py#L589-L599

Currently, to use different version of `appserver.yml` playbook, you have to change some of these settings, then restart Ocim. Then, after deployment, you have to change settings back because they're used by every new instance.

To be able to set this per instance, I added four new fields to Open edX instance model:
https://github.com/open-craft/opencraft/blob/b24a63891bbae1304ad851210a73bc045cae4c4a/instance/models/openedx_appserver.py#L127-L152

So, now we can set them from admin panel:
Screenshot from 2020-12-20 19-22-19

"Settings" tab of instance now has new settings listed:

image

## Why do we need to set different versions of playbooks for instances?

Because Ansible version is [hardcoded](https://github.com/open-craft/ansible-playbooks/blob/43b81c5d3e7e3c2bcd686d7380506004c6d66020/requirements.txt#L1) in `ansible-playbooks` repo. And unfortunately, version `2.5.1` doesn't work with Ubuntu 20.04, which Koa depends on. So, we have no ability to deploy Koa instances without reloading Ocim now.

With this change, we'll be able to use a different version of `ansible-playbooks` repo for Koa instances.

Another benefit is that we don't need to test all our playbooks to upgrade clients to Koa.

## Does this affect existing instances somehow?

No, they'll be using the same `appserver.yml` playbook as before. 

## Other information

**JIRA tickets**: 
- [BB-3331](https://tasks.opencraft.com/browse/BB-3331).
- It will unblock [BB-3266](https://tasks.opencraft.com/browse/BB-3266) (periodic builds of `open-release/koa.master`).
- It can make [BB-3309](https://tasks.opencraft.com/browse/BB-3309) less critical.

**Sandbox URL**
With this fix, I successfully deployed and activated a new app server for https://juniper-koa.stage.opencraft.hosting/ instance.

Each commit in this PR has detailed description.

**Testing instructions**:
1. Open https://stage.manage.opencraft.com/admin/instance/openedxinstance/675/change/
2. Change `Ansible appserver version` to `master`.
3. Try to create a new app server for this instance. Deployment should fail, since `master` version uses Ansible 2.5.1, which isn't compatible with Ubuntu 20.04.


gabor-boros:
👍  🎉 

One minor nit: It would be useful if the ansible version could be visible on the frontend, under the "Configuration" dropdown of the instance. If that's too much effort, feel free to ignore this.

- [x] I read through the code
- NA I checked for accessibility issues
- NA Includes documentation
- NA I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

gabor-boros (approved):

gabor-boros (approved):
Thank you for adding the extra changes!

0x29a:
Thanks for reviewing this, @gabor-boros! :slightly_smiling_face:

0x29a:
@gabor-boros
https://stage.manage.opencraft.com/instance/6380/edx-appserver/2968/ failed as expected with error as described in https://github.com/open-craft/opencraft/pull/679/commits/5b2624f6a95495aae2b1400668aee41c891cbb0d commit message.

gabor-boros:
@0x29a I did revert the change in the config (master back to the special branch) and saved the config.

Feel free to merge and go ahead 😊🤞

0x29a:
@gabor-boros, I did additional tests of new change and found a bug on `Settings` tab:

Screenshot from 2020-12-21 20-06-59

It's because new fields were not listed in serializer fields. I included fix in last commit:

image

PR #683, opened 2021-01-04
[SE-3825] Make ansible appserver values configurable

gabor-boros:
This PR resolves the issues identified during the investigation of why Watched Forks related PRs are not working properly for master builds by adding ansible configuration values to Watched Forks too.

**Dependencies**: None

**Screenshots**: 

Screenshot_2021-01-04 Add watched fork OpenCraft

**Merge deadline**: None, but sooner is better

**Testing instructions**:

1. Checkout changes locally (or on stage)
2. Run migrations
3. Start/Restart OCIM
4. Go to watched forks admin (`/admin/pr_watch/watchedfork/add/`)
5. Create a new Watched Fork to see the model can be saved
6. Create a new Watched Fork entry based on what we have in production for `edx-platform` but for `open-craft/edx-platform`
7. Open a PR for `open-craft/edx-platform`
8. Validate that the instance is created with the expected ansible settings you set on the Watched Fork
9. Revert the changes on Stage (if tested on stage)
10. Unapply migrations (if tested on stage)

N/A

**Reviewers**
- [x] @eLRuLL


pr_watch/migrations/0012_auto_20210104_1044.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 2.2.12 on 2021-01-04 10:44
```
eLRuLL, line 1: @gabor-boros I think there is already a 0012 migration, please rebase on top of that one. Also could you please set a name to this migration as well? So it can be more understandable (`--name`)

pr_watch/models.py
```diff
@@ -72,6 +73,37 @@ class WatchedFork(models.Model):
         db_index=True,
         help_text='Github fork name that will be watched for PRs. E.g.: open-craft/edx-platform'
     )
+    ansible_appserver_repo_url = models.URLField(
+        max_length=256,
+        blank=True,
+        null=True,
+        help_text=('The repository to pull the default Ansible playbook from.')
```
eLRuLL, line 80: @gabor-boros not entirely sure about this, but it looks pretty fragile to allow this to be blank and null, I guess this is done so it doesn't affect previous instances.
I think we can create instances from django admin, this means that these new fields are not necessary when doing that? And if that's the case, why were those fields added on the tests

eLRuLL (changes requested):

pr_watch/models.py
gabor-boros: @eLRuLL I'm not entirely sure I got your question right, but these fields are for "overriding" the default configuration of instances. Hence it is required to allow blank and null values (otherwise we couldn't use default values at all

pr_watch/migrations/0012_auto_20210104_1044.py
gabor-boros: 👌 will rename tomorrow morning

pr_watch/models.py
eLRuLL: yeah no problem, I think I was getting confused with instance configuration, but this is just for watching PRs

pr_watch/migrations/0012_auto_20210104_1044.py
gabor-boros: Renamed 😊

eLRuLL:
@gabor-boros 👍
- [x] I read through the code
- NA I checked for accessibility issues
- NA Includes documentation
- NA I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

eLRuLL (approved):

PR #691, opened 2021-01-14
Revert "[FAL-23] Enable ORGANIZATIONS_APP feature flag by default."

mtyaka:
Enabling the organizations app means that users have to go to django admin and create organizations before they can start creating new courses in the Studio. We don't want users to have to fiddle with django admin before being able to use the Studio.

This reverts commit 9bd5fe47177d87af54889e0b3b63e52df585a8fe.


nizarmah (approved):
@mtyaka :+1:

- [x] I tested this by launching a new instance and verifying that the `ORGANIZATIONS_APP` doesn't get set, which means that the instance will read it's default value from the `commons.py` environment file, which is set to `False`.
- [x] I read through the code
- [ ] ~I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~

mtyaka:
Thanks @nizarmah!

PR #693, opened 2021-01-14
BB-3517 Change build success/fail email subject

arjunsinghy96:
Changes the build success/fail email subject as requested by Ned [here](https://discuss.openedx.org/t/how-do-we-build-continuous-integration-for-the-supported-installation/3065/30?u=0x29a)

**JIRA Tickets**: BB-3517

**Discussions**: https://discuss.openedx.org/t/how-do-we-build-continuous-integration-for-the-supported-installation/3065/30?u=0x29a

**Merge deadline**: None

**Testing Instructions**:
1. Checkout to this PR
2. ssh to the vagrant box
3. run `make test.one instance.tests.models.test_openedx_instance.OpenEdXInstanceTestCase` and check all the tests pass
4. In file instance/models/mixins/utilities.py check that the email subjects are according to [this comment](https://discuss.openedx.org/t/how-do-we-build-continuous-integration-for-the-supported-installation/3065/30?u=0x29a)

**Reviewers**:
- [ ] @farhaanbukhsh 
- [ ] @0x29a

### DIFF
--- a/instance/models/mixins/utilities.py
+++ b/instance/models/mixins/utilities.py
@@ -311,13 +311,14 @@ def get_ansible_failure_log_entry(entries) -> Tuple[str, Dict[str, Any], List[st
     return task_name, log_entry, other_ansible_logs
 
 
-def send_periodic_deployment_success_email(recipients: List[str], instance_name: str) -> None:
+def send_periodic_deployment_success_email(recipients: List[str], instance) -> None:
     """
     Send notification email about successful periodic deployments.
 
     This email sending should be called, when the previous deployment failed, but the
     latest passed.
     """
+    instance_name = str(instance)
     logger.warning(
         "Sending acknowledgement e-mail to %s after instance %s didn provision",
         recipients,
@@ -325,7 +326,7 @@ def send_periodic_deployment_success_email(recipients: List[str], instance_name:
     )
 
     send_mail(
-        'Deployment back to normal at instance: {}'.format(instance_name),
+        '{} CI: Passed'.format(instance.name),
         'The deployment of a new appserver was successful. You can consider any failure notification '
         'related to {} as resolved. No further action needed.'.format(instance_name),
         settings.DEFAULT_FROM_EMAIL,
@@ -346,8 +347,8 @@ def send_urgent_deployment_failure_email(recipients: List[str], instance_name: s
 
     send_mail(
         'Deployment failed at instance: {}'.format(instance_name),
-        'The deployment of a new appserver failed and needs manual intervention. '
-        'You can find the logs in the web interface.',
+        'The deployment of a new appserver "{}" failed and needs manual intervention. '
+        'You can find the logs in the web interface.'.format(instance_name),
         settings.DEFAULT_FROM_EMAIL,
         recipients,
         fail_silently=False,
@@ -399,7 +400,7 @@ def send_periodic_deployment_failure_email(recipients: List[str], instance) -> N
     )
 
     email = EmailMultiAlternatives(
-        'Deployment failed at instance: {}'.format(instance_name),
+        '{} CI: Failed'.format(instance.name),
         'The periodic deployment of {edx_platform_release} failed. Please see the details below.\n\n'
         'Ansible task name:\t{ansible_task_name}\n'
         'Relevant log lines:\n{relevant_log_entry}\n\n'
@@ -524,5 +525,5 @@ def send_acknowledgement_email_on_deployment_success(sender, **kwargs) -> None:
 
     send_periodic_deployment_success_email(
         periodic_build_failure_emails,
-        str(instance)
+        instance
     )
--- a/instance/tests/models/test_openedx_instance.py
+++ b/instance/tests/models/test_openedx_instance.py
@@ -517,8 +517,8 @@ class OpenEdXInstanceTestCase(TestCase):
         self.assertEqual(outbox.to, failure_emails)
         self.assertEqual(
             outbox.body,
-            'The deployment of a new appserver failed and needs manual intervention. '
-            'You can find the logs in the web interface.',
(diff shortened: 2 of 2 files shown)


0x29a:
@arjunsinghy96, could you please sign up in Circle CI with your GitHub account and trigger pipeline for this branch?

0x29a (changes requested):
@arjunsinghy96, @farhaanbukhsh, I've reviewed changes and have a couple of suggestions.

First, I don't think we have to change `send_urgent_deployment_failure_email`, since we use this function only for registered user instances. They are not CI. I accidentally pointed to this function in the ticket description, sorry.

Second, with `'{} CI: Failed'.format(instance.name)` subject we'll have to change periodic build instances names to `Koa`, `Juniper`, `Ironwood` and `Master` to get desired email subject. IMO, instance with just `Master` name is a bit confusing.

Could we add two new fields to the `OpenEdXPeriodicBuildsMixin` instead? Then, we could do something like:
```python
if instance.periodic_build_success_email_subject:
    email_subject = instrance.periodic_build_success_email_subject
else:
    email_subject = 'Deployment back to normal at instance: {}'.format(instance)'
```

What do you guys think?

farhaanbukhsh:
> @arjunsinghy96, @farhaanbukhsh, I've reviewed changes and have a couple of suggestions.
> 
> First, I don't think we have to change `send_urgent_deployment_failure_email`, since we use this function only for registered user instances. They are not CI. I accidentally pointed to this function in the ticket description, sorry.

I think the function that you should be touching are https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/utilities.py#L314 and https://github.com/open-craft/opencraft/blob/master/instance/models/mixins/utilities.py#L357

> 
> Second, with `'{} CI: Failed'.format(instance.name)` subject we'll have to change periodic build instances names to `Koa`, `Juniper`, `Ironwood` and `Master` to get desired email subject. IMO, instance with just `Master` name is a bit confusing.
> 
> Could we add two new fields to the `OpenEdXPeriodicBuildsMixin` instead? Then, we could do something like:
> 

What are the two fields that you are referring to?

1. periodic_build_success_email_subject
2. Name of the release? How do we avoid master is not very clear to me.

> ```python
> if instance.periodic_build_success_email_subject:
>     email_subject = instrance.periodic_build_success_email_subject
> else:
>     email_subject = 'Deployment back to normal at instance: {}'.format(instance)'
> ```
> 
> What do you guys think?

The Idea seems legit to me :+1:

arjunsinghy96:
> First, I don't think we have to change send_urgent_deployment_failure_email, since we use this function only for registered user instances. They are not CI. I accidentally pointed to this function in the ticket description, sorry.

Are only the periodic builds considered CI? If yes, then this does make sense . I will revert changes in `send_urgent_deployment_failure_email`.

> What are the two fields that you are referring to?
 >   1. periodic_build_success_email_subject
 >   2. Name of the release? How do we avoid master is not very clear to me.

@farhaanbukhsh I think the two fields `Demid` is refering to are:
 1. periodic_build_success_email_subject
 2. periodic_build_failure_email_subject

I think instead of adding them as fields we can add them as `property` which would not require migrations. Also, adding them to `OpenEdXInstance` feels more appropriate to me as accessing `self.name` (for instance name) in `OpenEdXPeriodicBuildsMixin` might create confusion in future (accessing unrelated property. Correct me if we can do this in mixins.)

I am thinking something similar to following in `OpenEdXInstance`:

```python
@cached_property
def success_email_subject(self):
    if self.periodic_builds_enabled:
        return "{} CI: Passed".format(self.name)
   return "Deployment back to normal at instance: {}'.format(self)"

@cached_property
def failure_email_subject(self):
    if self.periodic_builds_enabled:
        return "{} CI: Failed".format(self.name)
    return "Deployment failed at instance: {}".format(self)
```
@0x29a Do we have to make the subject exactly as in [this comment](https://discuss.openedx.org/t/how-do-we-build-continuous-integration-for-the-supported-installation/3065/30)?

The subjects with above approach will be:
1. Periodic build Koa CI: Passed (or Failed)
2. Periodic build Juniper CI: Passed (or Failed)
3. Periodic build Ironwood CI: Passed (or Failed)
4. Periodic build Master CI: Passed (or Failed)

I think these satisfy the `shorten the subject and move the important information closer to the front` criteria mentioned by `Ned`

@farhaanbukhsh @0x29a What do you guys think? Does this make sense?

0x29a:
@arjunsinghy96
>Are only the periodic builds considered CI?

Yes, correct.

>Does this make sense?

Yep, property approach looks good to me.

0x29a (approved):
>I think these satisfy the shorten the subject and move the important information closer to the front criteria mentioned by Ned

Well, yes, you right. I just thought we could make subject to be like `Koa CI: Passed (or Failed)`, but it's not worth introducing new fields. PR looks good to me as is.

Thanks!

farhaanbukhsh:
@arjunsinghy96  can you update your PR :) thanks

0x29a:
@nedbat, email subjects will look like: `Periodic build Koa CI: Passed (or Failed)`. Is it fine?

farhaanbukhsh (approved):
👍 

- [x] I tested this: It's a subject line change I tried comparing it with the requirement
- [x] I read through the code
~~- [ ] I checked for accessibility issues~~
~~- [ ] Includes documentation~~
~~- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

nedbat:
Looking at the code, it seems like the subject line will be "Koa CI: Passed" or "Koa CI: Failed", without the "Periodic Build" prefix. I like the way the code is: short and sweet, with the important words prominent.

nedbat (approved):

arjunsinghy96:
_Rebased branch to master to include latest changes._

PR #696, opened 2021-01-21
Clear s3_bucket_name when bucket doesn't exist.

mtyaka:
When deprovisioning S3 and the instance's S3 bucket does not exist, we should treat the 404 error the same way as a successful delete and clear the bucket name.

This also fixes a bug where if deleting the bucket fails for whatever reason, Ocim no longer goes on to delete the IAM user, policy, and access key, because that makes it impossible to automatically delete the bucket on a retry.

**Test instructions**:

1. Create some test instances on Ocim stage. Archived them. You can use the management command: `make manage -- archive_instances --domains <mysubdomain>.stage.opencraft.hosting`. Note: Feel free to use "Delete Test 1" through "Delete Test 4" that I already created in Ocim stage for testing purposes.
2. Log into the Ocim Stage S3 AWS account (find info on how to log in in vault under "OpenCraft Tools : Resources : AWS : AWS S3 Stage").
3. Go to the S3 dashboard. Verify that you can see the S3 buckets belonging to your archived instances.
4. Go to the IAM dashboard. Verify that you can see the users belonging to your archived instances.
5. Start a django shell on Ocim stage, get a reference to your first archived instance: `instance = OpenEdXInstance.objects.get(internal_lms_domain='<myinstance>.stage.opencraft.hosting')`.
6. Delete the instance with `instance.delete()`. Verify that the instance was successfully deleted and that the S3 bucket and the IAM user no longer exist. _Note: these instances were not all created cleanly, so you may run into unrelated issues when trying to delete the instance. Please ignore errors not related to S3/IAM users._
7. Get a reference to your second archived instance in the django shell.
8. In the S3 dashboard, manually delete the S3 bucket belonging to your instance.
9. Delete the instance. Verify that it succeeds and that the IAM user gets removed.
10. Repeat the same process with another instance, this time delete **both** the S3 bucket and the IAM user. **In the django shell, blank out the instance's `s3_bucket_name` field with `instance.s3_bucket_name = ''; instance.save()`, otherwise Ocim will still try to delete the bucket and fail because we manually removed the user.** Verify deletion succeeds and S3 bucket gets removed.

### DIFF
--- a/instance/models/mixins/storage.py
+++ b/instance/models/mixins/storage.py
@@ -459,13 +459,11 @@ class S3BucketInstanceMixin(models.Model):
         response = self.s3.list_object_versions(Bucket=self.s3_bucket_name)
         return response.get('Versions', []) + response.get('DeleteMarkers', [])
 
-    def deprovision_s3(self):
+    def _delete_s3_bucket(self):
         """
-        Deprovision S3 by deleting S3 bucket and IAM user
+        Deletes the S3 bucket.
         """
-        self.logger.info('Deprovisioning S3 started.')
-        if (not self.storage_type == self.S3_STORAGE or
-                not (self.s3_access_key or self.s3_secret_access_key or self.s3_bucket_name)):
+        if not self.s3_bucket_name:
             return
 
         try:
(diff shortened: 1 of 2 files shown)


instance/models/mixins/storage.py
```diff
@@ -481,26 +479,86 @@ def deprovision_s3(self):
...
+            self.s3_bucket_name = ""
             self.save()
-        except ClientError:
+
+        try:
+            # Access keys and policies need to be deleted before removing the user
+            self._delete_s3_access_key()
+            self._delete_s3_user_policy()
+            self._delete_s3_user()
+        except Exception:
```
toxinu, line 553: @mtyaka We may need some debug here like which step failed and the error message, right?

toxinu (approved):
:+1:

I just left one minor comment. Feel free to ignore it if you think this is not relevant. :)

- [x] I tested this: testing instructions on instances 2 to 4
- [ x I read through the code
- [ ] I tested that the UI can be used with a keyboard only (tab order, keyboard controls).
- [ ] Includes documentation

instance/models/mixins/storage.py
mtyaka: It should be possible to see which step failed from the stacktrace (we re-raise the exception), and since errors like this would rarely occur, I don't think we need to do anything special.

PR #698, opened 2021-01-25
FAL-78 Fix the MKTG_URL_LINK_MAP override 

s0b0lev:
This PR is about resolving conflicts for the **MKTG_URL_LINK_MAP** configuration.

### Description

The issue with the previous implementation happens for juniper.3 release. 
- https://github.com/open-craft/edx-platform/blob/opencraft-release/juniper.3/common/djangoapps/edxmako/shortcuts.py#L93

For testing master branch was used. 
- https://github.com/edx/edx-platform/blob/master/common/djangoapps/edxmako/shortcuts.py#L96

This PR was tested on both environments and work fine for master and juniper.3 .
 

**Testing instruction:**
- Follow testing instruction in the ticket details 

**Authors notes:**
- Included API client updates 
- Instead of injecting the full list into MKTG_URL_LINK_MAP - this code will add only disabled pages

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -83,19 +83,7 @@ class OpenEdXConfigMixin(ConfigMixinBase):
                     # use different contact page
                     "CONTACT": "/contact",
                 },
-                # default values
-                "MKTG_URL_LINK_MAP": {
-                    "FAQ": "help",
-                    "PRESS": "press",
-                    "SITEMAP.XML": "sitemap_xml",
-                    "COURSES": "courses",
-                    "ROOT": "root",
-                    "WHAT_IS_VERIFIED_CERT": "verified-certificate",
-                    # "BLOG": "blog",  not supported yet
-                    **self.instance.get_mktg_url_link(),
-                },
             },
-
             "EDXAPP_LMS_NGINX_PORT": 80,
             "EDXAPP_LMS_SSL_NGINX_PORT": 443,
             "EDXAPP_LMS_BASE_SCHEME": 'https',
--- a/instance/models/openedx_appserver.py
+++ b/instance/models/openedx_appserver.py
@@ -279,12 +279,20 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
         'configuration_storage_settings',
         'configuration_theme_settings',
         'configuration_secret_keys',
+        'configuration_marketing_links',
         # The extra settings should stay at the end of this list to allow manual overrides of all settings.
         'configuration_extra_settings',
     ]
 
     class Meta(AppServer.Meta):
         verbose_name = 'Open edX App Server'
+
+    @property
+    def configuration_marketing_links(self):
+        """
+        Get `EDXAPP_MKTG_URL_LINK_MAP` values.
+        """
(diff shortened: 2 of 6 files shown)


instance/models/openedx_appserver.py
```diff
@@ -279,13 +279,21 @@ class OpenEdXAppServer(AppServer, OpenEdXAppConfiguration, AnsibleAppServerMixin
         'configuration_storage_settings',
         'configuration_theme_settings',
         'configuration_secret_keys',
+        'configuration_marketing_links',
```
s0b0lev, line 282: Added before `configuration_extra_settings`. To support override.

instance/models/mixins/openedx_config.py
```diff
@@ -83,19 +83,7 @@ def _get_configuration_variables(self):
                     # use different contact page
                     "CONTACT": "/contact",
                 },
-                # default values
-                "MKTG_URL_LINK_MAP": {
```
s0b0lev, line 87: Completly removed from `EDXAPP_LMS_ENV_EXTRA`. 

MKTG_URL_LINK_MAP - will contain a list of only disabled pages.

nizarmah:
@s0b0lev 

> Instead of injecting the full list into MKTG_URL_LINK_MAP - this code will add only disabled pages

https://github.com/edx/edx-platform/blob/70cc6d860814a8447b95d223066d4ada0b1c70bc/common/djangoapps/edxmako/shortcuts.py#L48-L59

The variables mentioned above indicate that it should be allowed to not just disable certain marketting links, but also override them.

Accordingly, this seems to be eliminating functionality. 

Would you mind, please, changing the approach you took in order to resolve the issue? 

cc @lgp171188 I'd like to hear your thoughts on this please

nizarmah (changes requested):
@s0b0lev 

> Instead of injecting the full list into MKTG_URL_LINK_MAP - this code will add only disabled pages

https://github.com/edx/edx-platform/blob/70cc6d860814a8447b95d223066d4ada0b1c70bc/common/djangoapps/edxmako/shortcuts.py#L48-L59

The variables mentioned above indicate that it should be allowed to not just disable certain marketting links, but also override them.

Accordingly, based on your description this seems to be eliminating functionality. 

Would you mind, please, changing the approach you took in order to resolve the issue? 

cc @lgp171188 I'd like to hear your thoughts on this please

nizarmah (commented):
:+1:

- [ ] I tested this:
  - [x] I provisioned a new instance and verified that customization using the extra configuration were not affected.
  - [x] I verified that Studio does not crash in newly provisioned instances
  - [ ] I disabled the pages from the frontend app and made sure the changes got reflected
- [x] I read through the code

---

@s0b0lev I had trouble testing the third point, I mentioned earlier, on my machine and on Ocim stage. Can you please find someone who would be able to review that last bullet point?

nizarmah (approved):
@s0b0lev thanks for creating the sandbox instance for me to test it! I really appreciate it! 

👍🏼 

- [x] I tested this:
  - [x] I provisioned a new instance and verified that customization using the extra configuration were not affected.
  - [x] I verified that Studio does not crash in newly provisioned instances
  - [x] I disabled the pages from the frontend app and made sure the changes got reflected

PR #702, opened 2021-02-02
BB-3592 Send email on successful instance update

arjunsinghy96:
Sends email to the BetaTestApplication user on successful instance redeployment of changes made by user.

**JIRA Tickets** : [BB-3592](https://tasks.opencraft.com/browse/BB-3592)

**Merge deadline**: None
**Testing instructions**:

1. Checkout to this PR
2. ssh to the vagrant box
3. run `make test.one 'registration.tests.test_api.InstanceDeploymentAPITestCase.test_send_email_on_application_redeployment'`and check all the tests pass

**Reviewers**
- [ ] @farhaanbukhsh

### DIFF
--- a/registration/signals.py
+++ b/registration/signals.py
@@ -25,7 +25,11 @@ from django_rest_passwordreset.models import ResetPasswordToken
 from django_rest_passwordreset.signals import reset_password_token_created
 from django_rest_passwordreset.views import ResetPasswordRequestToken
 
+from instance.models.deployment import Deployment, DeploymentType
+from instance.signals import appserver_spawned
 from opencraft.utils import html_email_helper
+from registration.models import BetaTestApplication
+from registration.utils import send_changes_deployed_success_email
 
 
 # noinspection PyUnusedLocal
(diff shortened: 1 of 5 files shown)


registration/signals.py
```diff
@@ -54,3 +58,38 @@ def password_reset_token_created(
...
+    appserver = kwargs['appserver']
+    deployment_id = kwargs['deployment_id']
+
+    if deployment_id is None or appserver is None:
+        return
+
+    deployment = Deployment.objects.get(pk=deployment_id)
+
+    # Send emails for only user triggered emails
+    if deployment.type != DeploymentType.user.name:
```
farhaanbukhsh, line 79: What about deployment created while doing registration?  i.e  DeploymentType.registration.name

registration/signals.py
```diff
@@ -54,3 +58,38 @@ def password_reset_token_created(
...
+        return
+
+    application = instance.betatestapplication_set.first()  # There should only be one
+    is_appserver_healthy = appserver.status.is_healthy_state
+
+    # Send email only for registered client application where application status is
+    # accepted. Applications with pending status will go through the approval signal
+    # and send welcome email.
+    if not application or application.status != BetaTestApplication.ACCEPTED:
```
farhaanbukhsh, line 88: Can an app server spawn if the BetaTestApplication status is pending?

registration/signals.py
```diff
@@ -54,3 +58,38 @@ def password_reset_token_created(
...
+    if deployment_id is None or appserver is None:
+        return
+
+    deployment = Deployment.objects.get(pk=deployment_id)
+
+    # Send emails for only user triggered emails
+    if deployment.type != DeploymentType.user.name:
+        return
+
+    application = instance.betatestapplication_set.first()  # There should only be one
```
farhaanbukhsh, line 82: can we add a more meaningful comment?

farhaanbukhsh (changes requested):
Overall it looks good, I have left few questions. :) Can we deploy this on staging and check?

registration/signals.py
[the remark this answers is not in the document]
arjunsinghy96: The `DeploymentType.registration` is handled already. We send a welcome email to the user after the deployment and mark the `BetaTestApplication` status as `ACCEPTED`.
In that case, an extra instance update email won't make sense.

registration/signals.py
arjunsinghy96: The first deployment after the registration happens with `BetaTestApplication.PENDING` status (Welcome email sent on success).
This check serves as a check to determine that current deployment is a config update deployment.
In theory, appserver spawns are independent of `BetaTestApplication` status.

registration/signals.py
arjunsinghy96: Sure. I would update it with
```
# The instance should be associated with only one BetaTestApplication by design, thus we can take the first record
```

farhaanbukhsh (approved):
👍 

- [x] I tested this: We have tested this PR on staging. On re-provisioning the instance an email is sent to the user.
- [x] I read through the code

PR #704, opened 2021-02-03
[BB-3616]: Add check for DNS configuration

arjunsinghy96:
If external domain is added to the instance, we need to check that proper CNAME records are created with the DNS by use.

**Related Tickets**: [BB-3616](https://tasks.opencraft.com/browse/BB-3616)

**Testing Instructions**:
1. Checkout to this branch
2. Run migrations using `make migrate`
2. Run dev server using `make run.dev`
3. Open admin panel
4. Create a test `User` with a `BetaTestApplication`
5. Make the following curl request (replace variable with proper values)

```
curl -X PATCH 'http://localhost:5000/api/v2/instances/openedx_config/:instance_id/' \
  -H 'Authorization: Bearer <jwt-token>' \
  -H 'Content-Type: application/json' \
  -H 'Accept: */*' \
  --data-raw '{"external_domain":"a-new-domain.com"}' \
  --compressed
```
6. Verify that a task is scheduled (worker logs)
7. Verify that the task is started after 1 minute (worker logs)
8. Verfiy that the DNS configuration check fails and is logged (worker logs)
9. Verfiy that the Email log is printed (worker logs)

2. Check that `templates/emails/dns_not_configured.txt` and `templates/emails/dns_not_configured.html` have appropriate data corresponding to [this Zeplin design](https://app.zeplin.io/project/5ea850e04326be273e4bcf5a/screen/5fdb0719a491febb0d11d5a2)
3. Run `make test.one registration.tests` and verfiy that all unittests pass.


registration/utils.py
```diff
@@ -80,3 +84,42 @@ def send_account_info_email(application: BetaTestApplication) -> None:
...
+    try:
+        answers = dns.resolver.resolve(domain, 'CNAME')
+        cname_records = [str(rdata) for rdata in answers]  # There should only be one
+        is_configured_properly = [
+            record == 'haproxy.opencraft.hosting'
+            for record in cname_records
+        ]
+        return any(is_configured_properly)
+    except (dns.resolver.NXDOMAIN, dns.resolver.NoAnswer, dns.resolver.NoNameservers):
+        return False
```
farhaanbukhsh, line 125: we should not let the error pass silently, we should log this. Else we will never know if the records were not configured properly or any other exception occured.

registration/utils.py
```diff
@@ -19,6 +19,10 @@
 """
 Utility functions related to registration.
 """
+import binascii
+import os
+
+import dns.resolver
```
farhaanbukhsh, line 25: can we make this as 
```suggestion
import dns.resolver as dns_resolver
```

farhaanbukhsh (changes requested):
I have few suggestions but overall the approach LGTM :+1:

registration/utils.py
```diff
@@ -80,3 +84,42 @@ def send_account_info_email(application: BetaTestApplication) -> None:
...
+def is_dns_configured(domain: str) -> bool:
+    """
+    Checks that the provided domain has a proper DNS configuration.
+    There should be a CNAME record present with value haproxy.opencraft.hosting
+    """
+    try:
+        answers = dns.resolver.resolve(domain, 'CNAME')
+        cname_records = [str(rdata) for rdata in answers]  # There should only be one
+        is_configured_properly = [
+            record == 'haproxy.opencraft.hosting'
```
farhaanbukhsh, line 120: we should get  `haproxy.opencraft.hosting` from a constant since this can change, we might as well use another load ballencer.

kaizoku (approved):
@arjunsinghy96 :+1:

Thanks for the updated test instructions.
I have two very small nits but it's otherwise working for me.

farhaanbukhsh (approved):

PR #707, opened 2021-02-07
[BB-3585] Add initial setup for e2e tests

arjunsinghy96:
This adds the base for e2e test in OCIM with react frontend

**JIRA Tickets**: [BB-3585](https://tasks.opencraft.com/browse/BB-3585)

**Testing Instruction**
1. Checkout to this branch
2. Run `vagrant up` and `vagrant ssh` to checkin into the devstack.
3. Run `make requirements`
4. Run `./bin/install-supported-firefox`
4. Run `make test.e2e`
5. Test that the frontend is build and e2e tests are run and passing.
6. Check that the Circle CI build `e2e-tests` is passing.

**Author Notes**

1. selenium and firefox need to be updated to `3.141.0`(added in requirements) and `85+` (updated in `./bin/install-supported-firefox`) respectively
2. We have `registration.tests.utils.BrowserTestMixin` using `selenium` but I could not find its use anywhere.


kaizoku (changes requested):
@arjunsinghy96, the overall structure of the tests is looking alright, though they're quite flaky on circleci and I'm also having trouble running them locally at the moment.

Also these new tests aren't configured to run on CircleCI, or in the `make test` target. Should we be adding them there as well?
I can help configure `circle.yml` to run the e2e tests if it helps. Our `circle.yml` file for this project has a lot going on at this point.
We should just be able to add another workflow for the e2e tests.

--

I left a few requests for changes, specifically around the `sleep` calls in the tests.

One of the commit messages is incredibly long and duplicates of the same message as well. Since we'll squash commits before we merge, this isn't a huge deal as long as we change the message, but I just thought I should note it since it will clutter the `git log` view.

e2e_tests/tests/utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def setUp(self):
+        """
+        Setup browser for the e2e tests.
+        """
+        super().setUp()
+        options = Options()
+        options.headless = True
+        cap = DesiredCapabilities().FIREFOX
+        firefox_binary = FirefoxBinary("/usr/bin/firefox.ubuntu")
```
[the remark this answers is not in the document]
arjunsinghy96: I ran the tests inside vagrant. The firefox version is `47.0.1` in vagrant box. We need to install the latest version of `firefox` (`85+`) and `selenium` (`3.141.0`) in vagrant to run the test.

For `firefox`, I did it using `sudo apt install firefox`. This installed the binary was available at `/usr/bin/firefox.ubuntu`

I think we can include firefox in `debian_packages.lst` to make sure the appropriate version is installed.

We can also add an environment variable `FIREFOX_BINARY_PATH` to make it environment agnostic.

opencraft/settings.py
```diff
@@ -114,7 +114,8 @@
     'registration',
     'reports',
     'backup_swift',
-    'periodic_builds'
+    'periodic_builds',
+    'e2e_tests',
```
kaizoku, line 118: Do we need to have this configured for all runtimes including production?
It seems like something we wouldn't want to add to the installed apps on runtime systems like production and stage, just while running tests.

Maybe we can split the test settings out into a separate settings file where this is added?

kaizoku:
Thanks for the updates @arjunsinghy96. I have one other comment and there's a merge conflict with dependencies from something else that must have merged in the meantime.

Apart from the remaining comment this looks good.

opencraft/settings.py
arjunsinghy96: This makes sense @kaizoku  and I think splitting the settings file might go outside the scope of this PR.

I think the following approach works with minimal changes.

```python
# This env var is already added for e2e tests, we can utilise this
E2E_TESTS = env.bool('E2E_TESTS', default=True)
...
# If we are running e2e test add e2e_tests application in LOCAL_APPS
if E2E_TESTS:
    LOCAL_APPS += (
        'e2e_tests',
    )
```

kaizoku: That works as well @arjunsinghy96. Thanks for the update to that.

kaizoku (approved):

kaizoku:
Thanks for the revisions @arjunsinghy96! This is looking good.  :+1:

- [✓] I tested this: (examined CircleCI tests run results, ran tests in devstack)
- [✓] I read through the code
- [n/a] ~I checked for accessibility issues~

PR #708, opened 2021-02-08
Revert "Fix Koa integration tests (#682)"

shimulch:
This reverts commit https://github.com/open-craft/opencraft/commit/853f9953ca2abe8c7be3726663f26b19ad6377ca. This is causing integration tests to fail as the default release of OCIM is still Juniper.


0x29a (approved):
:+1:

- [x] I checked that CI checks are green.

PR #711, opened 2021-02-15
[BB-3670] Include only new instances in send_trial_instances_report task

spokerman12:
**Description**

We're fulfilling the following user story:

> "As the person converting Ocim trial users, I need an autogenerated monthly report with a list of trial users for each month. "

**Supporting information**

- [Jira ticket](https://tasks.opencraft.com/browse/BB-3670)

**Testing instructions**

1. Set up a local OCIM environment using this branch
2. Create instances and appservers
3. Edit their created dates with any method available
4. Run `send_trial_instances_report` on `reports/tasks.py` or
5. Run `./manage.py test reports.tests.ReportTaskTestCase` inside vagrant VM

**Snapshots:**

None

**Further considerations:**

- Initially we enabled a flag for the command, but we updated the feature to reflect current requirements as discussed in the review comments.

Reviewer:
[ ] @lgp171188


instance/management/commands/instance_statistics_csv.py
```diff
@@ -98,6 +98,13 @@ def add_arguments(self, parser):
             default=None,
             help='Path to the output file of the new CSV. Leave blank to use stdout.'
         )
+        parser.add_argument(
```
lgp171188, line 101: @spokerman12, the management command should be left as-is and the code [here](https://github.com/open-craft/opencraft/blob/master/reports/tasks.py) should be updated to do the necessary filtering.

spokerman12: I don't understand, on this [comment](https://tasks.opencraft.com/browse/BB-3670?focusedCommentId=188674&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-188674) you said 

> [implement] a flag to the existing command that takes in a date range and filters the instances down to that

- Are domains always required, then?
- Should I remove the --created flag?
- Should I create another db_periodic_task or edit the current send_trial_instances_report? The name is confusing as it suggests I should edit it.

lgp171188: @spokerman12, yes, I said what you have quoted above. But I feel that it is better to leave the command as-is and generic enough and update the periodic task instead to filter and pass only the instances created in the most recent completed month. Does this make sense or do you have any questions?

spokerman12: Makes sense! I implemented the changes suggested plus a test.

reports/tasks.py
```diff
@@ -46,27 +46,19 @@
         day_of_week=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY_OF_WEEK
     )
 )
-def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
+def send_trial_instances_report(
+        recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
     """
-    Generate and send a trial instance data report
+    Generate and send a new instance data report for the past month
```
lgp171188, line 52: @spokerman12, nit: the new instances are still trial instances :)

reports/tasks.py
```diff
@@ -46,27 +46,19 @@
         day_of_week=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY_OF_WEEK
     )
 )
-def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
+def send_trial_instances_report(
+        recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
```
lgp171188, line 50: @spokerman12, is there a reason you changed the formatting of this line? The old one met the code quality checks and didn't cause any issues before.

And if there is some reason to change it, you should format it according to our coding standards, like what you had to do for the django-password-validators PR.

reports/tasks.py
```diff
@@ -46,27 +46,19 @@
...
-    Generate and send a trial instance data report
+    Generate and send a new instance data report for the past month
 
     This task runs on the first of every month at 2AM
     """
 
     if not recipients:
-        logger.warning('No recipients listed for Trial Instances Report. It will not be generated.')
+        logger.warning(
+            'No recipients listed for Trial Instances Report. It will not be generated.')
```
lgp171188, line 59: @spokerman12, similar concerns as https://github.com/open-craft/opencraft/pull/711/files#r579396180

reports/tasks.py
```diff
@@ -76,6 +68,21 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+    # Get instances with an active betatestapplication,
+    # whose InstanceReference has an active OpenEdXAppServer
+    # and was created last month
+    instances = OpenEdXInstance.objects.exclude(
+        betatestapplication__isnull=True
+    ).filter(
+        ref_set__openedxappserver_set___is_active=True
+    ).filter(
+        ref_set__created__range=[
+            beginning_of_last_month.strftime("%Y-%m-%d"),
```
lgp171188, line 80: @spokerman12, since this is just a time range, how much effort would it require to make this configurable instead of hard-coding it to the last completed month?

reports/tasks.py
```diff
@@ -76,6 +68,21 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+            end_of_last_month.strftime("%Y-%m-%d")
+        ])
```
lgp171188, line 82: @spokerman12, nit: closing parenthesis formatting.

reports/tasks.py
```diff
@@ -92,7 +99,8 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
         )
     except SystemExit:
         # The command exited prematurely. Send failure email to recipients
-        logger.error('`instance_statistics_csv` command failed. Sending notification email')
+        logger.error(
+            '`instance_statistics_csv` command failed. Sending notification email')
```
lgp171188, line 103: @spokerman12, similar concerns as https://github.com/open-craft/opencraft/pull/711/files#r579396180

reports/tasks.py
```diff
@@ -76,6 +68,21 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+        betatestapplication__isnull=True
+    ).filter(
+        ref_set__openedxappserver_set___is_active=True
+    ).filter(
+        ref_set__created__range=[
+            beginning_of_last_month.strftime("%Y-%m-%d"),
+            end_of_last_month.strftime("%Y-%m-%d")
+        ])
+    domains = [
+        instance.external_lms_domain or instance.internal_lms_domain for instance in instances
```
lgp171188, line 84: @spokerman12, I think `instance.domain` will always contain the right value and so this `or` condition may not be needed.

reports/tests.py
```diff
@@ -528,3 +533,110 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+
+        user1, _ = get_user_model().objects.get_or_create(
+            username='test1', email='test1@example.com')
+        user2, _ = get_user_model().objects.get_or_create(
+            username='test2', email='test2@example.com')
+        user3, _ = get_user_model().objects.get_or_create(
+            username='test3', email='test3@example.com')
+
+        with freeze_time(date1):
+            instance1 = OpenEdXInstanceFactory(
```
lgp171188, line 581: @spokerman12, the repeated code below for creating multiple users can be refactored into a separate function or a factory function, no?

reports/tests.py
```diff
@@ -528,3 +533,110 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+                successfully_provisioned=True, betatestapplication=BetaTestApplication())
+            make_test_appserver(instance=instance1, is_active=True)
+            BetaTestApplication.objects.create(
+                user=user1,
+                public_contact_email='publicemail1@example.com',
+                subdomain='betatestdomain1',
+                instance_name=instance1.name,
+                status=BetaTestApplication.PENDING,
```
lgp171188, line 589: @spokerman12, I don't think `PENDING` is the correct state. Once an appserver is created, I think the status changes to "approved".

reports/tasks.py
spokerman12: Understood. Changed like so.

reports/tasks.py
spokerman12: I ran this through autopep8, but it's true that while we should adhere to PEP8 per the handbook, the 79 character per line limit is changed to 120.

reports/tasks.py
spokerman12: I ran this through autopep8, but it's true that while we should adhere to PEP8 per the handbook, the 79 character per line limit is changed to 120.

reports/tasks.py
spokerman12: It'd be easy, just adding more arguments to send_trial_instances_report.
But right now I don't think it's convenient to increase the scope as we still have tasks left; I'd be happy to do it on the next sprint, though.

reports/tasks.py
spokerman12: Changed according to previous suggestions.

reports/tasks.py
spokerman12: This part of the logic was already implemented, so I wouldn't want to rework it unnecessarily; plus, I don't think the command is tested thoroughly. As suggested before, I'd be more than happy to work on this on a different task.

reports/tasks.py
spokerman12: Corrected as suggested.

reports/tests.py
spokerman12: Because it's implied that there's an ordering  user1 < user2 < user3, instance1 < instance2 < instance3 ... etc, because we're adding a singular feature I prefer the test to be explicit (or "don't make me think," as suggested in the handbook).

I'd add the factory function if we were adding more tests that could make use of it, but in this case I don't think it's necessary.

reports/tests.py
spokerman12: The command only wants that the referred BetaTestApplication is not null, but I'll change it to Approved to reflect real scenarios.

reports/tasks.py
```diff
@@ -76,6 +66,22 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+    # whose InstanceReference has an active OpenEdXAppServer
+    # and was created last month
+    instances = OpenEdXInstance.objects.exclude(
+        betatestapplication__isnull=True
+    ).filter(
+        ref_set__openedxappserver_set___is_active=True
+    ).filter(
+        ref_set__created__range=[
+            beginning_of_last_month.strftime("%Y-%m-%d"),
+            end_of_last_month.strftime("%Y-%m-%d")
```
lgp171188, line 79: @spokerman12, this has the off-by-one error as mentioned [here](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#range). Can you fix it and also write a test case for this scenario?

reports/tests.py
```diff
@@ -528,3 +533,96 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+            [],
+            sent_mail.attachments
+        )
+        self.assertEqual(
+            ['to@example.com'],
+            sent_mail.to
+        )
+        self.assertEqual(len(mail.outbox), 1)
+
+    def test_send_trial_instances_report(self):
```
lgp171188, line 587: @spokerman12, this test doesn't test the corner cases like the one mentioned [here](https://github.com/open-craft/opencraft/pull/711/files#r580161566).

reports/tasks.py
spokerman12: Thank you for picking up this case. I've modified the filter and added the unit test

reports/tests.py
spokerman12: I added the test case as stated in previous comment

reports/tests.py
```diff
@@ -528,3 +533,160 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+            send_trial_instances_report(recipients=['to@example.com'])
+            sent_mail = mail.outbox[0]
+
+            self.assertEqual(len(mail.outbox), 1)
+
+            self.assertIn(
+                'Please find attached a CSV with the statistics for February 2020',
+                sent_mail.body
+            )
+            print(sent_mail.attachments[0][1])
```
lgp171188, line 664: @spokerman12, nit: this should be removed.

spokerman12: My bad!

reports/tests.py
```diff
@@ -528,3 +533,159 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+        """
+        Returns a trial instance for a user
+
+        :param key: String to append to properties.
+        :return: A trial instance
+        """
+        user, _ = get_user_model().objects.get_or_create(
+            username=key + 'test', email=key + 'test@example.com')
+        instance = OpenEdXInstanceFactory(
+            successfully_provisioned=True, betatestapplication=BetaTestApplication())
```
lgp171188, line 553: @spokerman12, nit: formatting of closing parenthesis doesn't match the coding standards that we follow.

spokerman12: Fixed plus another occurrence.

reports/tests.py
```diff
@@ -528,3 +533,161 @@ def _generate_invoice_date(year=datetime.now().year, month=1, this_month=False):
...
+        """
+        Tests that if send_trial_instances_report succeeds, the recipient receives
+        a correct report with only instances created last month.
+        """
+        date1 = "2020-01-01"
+        date2 = "2020-02-01"
+        date3 = "2020-03-01"
+
+        with freeze_time(date1):
+            instance1 = self.create_user_with_trial_instance(key='1')
```
lgp171188, line 599: @spokerman12, the key doesn't matter as it is not used for any of the checks below. So can't the `create_user_with_trial_instance()` method be a factory method that accepts a count and returns the `count` number of instances, which can be assigned like

```python
instance1, instance2, instance3 = self.user_trial_instance_factory(count=3)
```
?

spokerman12: True. Modified as requested.

lgp171188 (approved):
@spokerman12, :+1: Please wait till the unrelated CI failures are fixed before merging this PR.
- [x] I tested and verified that the periodic job now includes only the instances created in the last completed month.
- [x] I read through the code

lgp171188:
@spokerman12, can you fix the failing quality check before moving the task to blocker?

spokerman12:
@lgp171188 sure. I'd be in favor of adding a lint-fix for python; I don't believe it's implemented.

PR #721, opened 2021-02-26
[BB-3775] Setup periodic task to kill zombie app servers

spokerman12:
We want to set up a periodic task for the `kill_zombies` command.

Requirements summarized after discussion:

- Frequency configurable via ~environment variables~ django settings (opencraft/settings.py)
- Defaults to once a day
- Call the kill_zombies management command using call_command
- Have a reasonable threshold (that is configurable) and if the number of VMs to be deleted exceeds that we should alert by sending an email to the ops@opencraft.com
- Have a settings toggle (configurable via environment variables) to disable the job

`kill_zombies` has no tests, so we convened on testing the following for the periodic task:

- Picks up and applies all the proposed ~env variables~ settings
- Sends email if we're over the threshold
- Sends email on failure
- Invokes the management command to do the actual cleanup

**JIRA tickets**:  [BB-3775](https://tasks.opencraft.com/browse/BB-3775)

**Dependencies**: None

**Screenshots**: See Gitlab issue

**Merge deadline**: "None"

**Testing instructions**:

1. Pull this branch to your local OCIM environment
2. Inside the vagrant VM, run unit tests with
```
./manage.py test instance.tests.management.test_kill_zombies.KillZombiesPeriodicallyTestCase
```
3. On normal execution, the task will overwrite the default arguments with whatever env variables are available. Set these on `opencraft/settings.py` to test manually. These are the defaults:
```
KILL_ZOMBIES_SCHEDULE="0 0 */1 * *"       # Daily at 12AM
KILL_ZOMBIES_ENABLED=False         
OPENSTACK_REGION="gra"                          # This is already set in the file
ADMINS = env.json('ADMINS', default=set())   # Already set, has the form  of (name, email) tuples)
KILL_ZOMBIES_WARNING_THRESHOLD=10
```
These defaults can be changed per the reviewer's request.

4. Edit Procfile.dev to
```
worker_low_priority: HUEY_QUEUE_NAME=opencraft_low_priority python3 manage.py run_huey

```
5. Because `kill_zombies` is not tested, we have to override its behavior and treat is as a black box. To do so, hack this in `instance/management/commands/kill_zombies.Command.handle`:
```
...
# Set options
self.region = options.get("region")
self.dry_run = options.get("dry_run", False)
if self.dry_run:    # This block of code is lower in the handle method
    result = "Would have terminated {} zombies if this weren't a dry run.".format(death_count)
else:
    result = "Terminated {} zombies.".format(death_count)
self.log(result)
return result
...
```
Then manually change `death_count`so you can test.
6. Run `make run.dev` inside the vagrant VM. You should see entries similar to:
```
21:31:01 worker_low_priority.1 | [2021-02-26 21:31:01,880] INFO:root:Worker-1:Executing periodic task `kill_zombies_periodically`
```

1. `kill_zombies` needs some tests.

**Reviewers**
- [ ] @lgp171188

### DIFF
--- a/instance/management/commands/kill_zombies.py
+++ b/instance/management/commands/kill_zombies.py
@@ -71,16 +71,18 @@ class Command(BaseCommand):
 
         if not nova_servers:
             self.log('No servers found in region {}.'.format(self.region))
-            return
+            return None
         self.log('Found {} unterminated servers in region {}.'.format(len(nova_servers), self.region))
 
         # Scan each server for the zombieness.
         death_count = sum(1 for srv in nova_servers if self.not_zombie_or_die(srv))
 
         if self.dry_run:
-            self.log("Would have terminated {} zombies if this weren't a dry run.".format(death_count))
+            result = "Would have terminated {} zombies if this weren't a dry run.".format(death_count)
         else:
-            self.log("Terminated {} zombies.".format(death_count))
+            result = "Terminated {} zombies.".format(death_count)
+        self.log(result)
+        return result
 
     def not_zombie_or_die(self, nova_server):
         """
--- a/instance/tasks.py
+++ b/instance/tasks.py
@@ -23,11 +23,13 @@ Worker tasks for instance hosting & management
 # Imports #####################################################################
 
 from datetime import datetime
-import logging
 from typing import Optional, List
+import logging
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.management import call_command, CommandError
+from django.core.mail import EmailMessage
 from django.db import connection
 from django.db.models import Q
 from django.utils import timezone
@@ -47,6 +49,9 @@ from pr_watch import github
 
 logger = logging.getLogger(__name__)
 
+# Constants  ##################################################################
+
+KILL_ZOMBIES_CRON_SCHEDULE = settings.KILL_ZOMBIES_SCHEDULE.split()
 
 # Tasks #######################################################################
 
@@ -339,3 +344,115 @@ if settings.CLEANUP_OLD_BETATEST_USERS:
         users_to_delete = queryset.filter(is_active=False, profile__modified__lte=delete_cutoff)
         logger.info('Deleting %s users.', users_to_delete.count())
         users_to_delete.delete()
+
+
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.region: str = settings.OPENSTACK_REGION
+        self.recipient: list = [email for _, email in settings.ADMINS]
+        self.threshold: int = settings.KILL_ZOMBIES_WARNING_THRESHOLD
+
+    def send_email(self, subject: str, body: str) -> int:
+        """
+        Utility method for sending emails
+        """
+        if self.recipient:
+            email = EmailMessage(
+                subject, body, settings.DEFAULT_FROM_EMAIL, self.recipient
+            )
+            email.send()
+        else:
+            logging.info("Email recipient is undefined. Skipping email.")
+
+    def trigger_warning(self, num_zombies):
+        """
+        Warns when more than `threshold` VMs will be terminated
+        """
+        subject = "Zombie instances are over the current threshold ({})".format(
+            self.threshold
+        )
+        body = (
+            "The number of zombie OpenStack VMs ({}) in the {} region "
+            "is over the KILL_ZOMBIES_WARNING_THRESHOLD ({}).\n"
+            "These instances will be terminated using the `kill_zombies` command."
+        ).format(num_zombies, self.region, self.threshold)
+        logging.info(
+            "%s\nSending an email notification to: %s",
+            body, self.recipient
+        )
+        self.send_email(subject, body)
+
+    def get_zombie_servers_count(self, stdout: str) -> int:
+        """
+        If there are servers to terminate, a `kill_zombies` dry run will warn:
+        "Would have terminated {} zombies" to stdout.
+        """
+        num_instances = 0
+        if "No servers found in region" in stdout:
+            return num_instances
+        try:
+            num_instances = int(stdout.split("Would have terminated")[1].split()[0])
+        except (IndexError, ValueError):
+            logger.info("Received unexpected input from dry run. Defaulting to zero")
+        return num_instances
+
+    def run(self):
+        """
+        Main method for running the task
+        """
+        try:
+            dry_run_output, output = "", ""
+            dry_run_output = call_command(
+                "kill_zombies",
+                region=self.region,
+                dry_run=True,
+            )
+            num_of_zombies = self.get_zombie_servers_count(dry_run_output)
+            if num_of_zombies > self.threshold:
+                self.trigger_warning(num_of_zombies)
+            if num_of_zombies != 0:
+                output = call_command("kill_zombies", region=self.region)
+                logging.info(
+                    "Task `kill_zombies_periodically` ran successfully "
+                    "and terminated %s zombies", num_of_zombies
+                )
+            else:
+                logging.info(
+                    "Found zero zombies to terminate. "
+                    "Task `kill_zombies_periodically` ran successfully."
+                )
+        except CommandError:
+            logger.error(
+                "Task `kill_zombies` command failed. Sending notification email"
+            )
+            subject = "Terminate zombie instances command failed"
+            body = (
+                "Scheduled execution of `kill_zombies` command failed. "
+                "Log entries are displayed below:"
+                "%s\n%s"
+            ) % (dry_run_output, output)
+            self.send_email(subject, body)
+            return False
+        return True
+
+
+if settings.KILL_ZOMBIES_ENABLED:
+    @db_periodic_task(crontab(
+        minute=KILL_ZOMBIES_CRON_SCHEDULE[0],
+        hour=KILL_ZOMBIES_CRON_SCHEDULE[1],
+        day=KILL_ZOMBIES_CRON_SCHEDULE[2],
+        month=KILL_ZOMBIES_CRON_SCHEDULE[3],
+        day_of_week=KILL_ZOMBIES_CRON_SCHEDULE[4],
+    ))
+    def kill_zombies_task():
+        """
+        A wrapper for `kill_zombies_periodically` so it only
+        exists when KILL_ZOMBIES_ENABLED is true
+        """
+        task_runner = KillZombiesRunner()
+        logging.info("Executing periodic task `kill_zombies_periodically`")
+        task_runner.run()
--- a/instance/tests/management/test_kill_zombies.py
+++ b/instance/tests/management/test_kill_zombies.py
@@ -1,0 +1,176 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2019 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Tests for the 'kill_zombies' management command.
+"""
+# Imports #####################################################################
+
+from unittest.mock import patch, call
+from django.core import mail
+from django.core.management import CommandError
+from django.test import TestCase, override_settings
+from instance.tasks import KillZombiesRunner
+
+# Tests #######################################################################
+
+# pylint: disable=bad-continuation
+
+
+@override_settings(
+    KILL_ZOMBIES_ENABLED=True,
+    ADMINS=(
+        ("admin1", "admin1@localhost"),
+        ("admin2", "admin2@localhost"),
+    ),
+    OPENSTACK_REGION="some-region"
+)
+class KillZombiesPeriodicallyTestCase(TestCase):
+    """
+    Test cases for the `kill_zombies_task` periodic task.
+    """
+    @override_settings(KILL_ZOMBIES_ENABLED=False)
+    @patch("instance.tasks.call_command")
+    def test_toggle_task_false(self, mock_command):
+        """
+        Tests that the task can't be run if disabled
+        through env variables
+        """
+        with self.assertRaises(ImportError):
+            from instance.tasks import kill_zombies_task
+            kill_zombies_task()
+        self.assertFalse(mock_command.called)
+
+    @patch("instance.tasks.call_command")
+    def test_set_env_variables(self, mock_command):
+        """
+        Tests that the task passes env variables to
+        the kill_zombies command call
+        """
+        test_runner = KillZombiesRunner()
+        test_runner.run()
+        first_call = call.mock_command("kill_zombies", region="some-region", dry_run=True)
+        self.assertEqual(mock_command.mock_calls[0], first_call)
+        mock_command.assert_called_with("kill_zombies", region="some-region")
+
+    @patch.object(KillZombiesRunner, "trigger_warning")
+    def test_get_zombie_servers_count(self, mock_trigger):
+        """
+        Tests that KillZombiesRunner.get_zombie_servers_count
+        works as expected
+        """
+        test_runner = KillZombiesRunner()
+        no_instances = """
+        2021-02-26 14:16:12 | Starting kill_zombies
+        2021-02-26 14:16:12 | No servers found in region some-region.
+        """
+        some_instances = """
+        2021-02-26 14:16:12 | Starting kill_zombies
+        2021-02-26 14:16:12 | Found 32 unterminated servers in region some-region.
+        2021-02-26 14:16:12 | Would have terminated 32 zombies if this weren't a dry run.
+        """
+        bad_output = "qwerty"
+        self.assertEqual(test_runner.get_zombie_servers_count(no_instances), 0)
+        self.assertEqual(test_runner.get_zombie_servers_count(some_instances), 32)
+        self.assertEqual(test_runner.get_zombie_servers_count(bad_output), 0)
+        self.assertEqual(test_runner.get_zombie_servers_count(""), 0)
+
+    @override_settings(KILL_ZOMBIES_WARNING_THRESHOLD=3)
+    @patch.object(KillZombiesRunner, "get_zombie_servers_count", return_value=4)
+    @patch.object(KillZombiesRunner, "trigger_warning")
+    @patch("instance.tasks.call_command")
+    def test_trigger_warning_if_over_threshold(
+        self,
+        mock_get_zombie_servers_count,
+        mock_trigger_warning,
+        mock_command
+    ):
+        """
+        Tests that trigger_warning is called when the number
+        of zombies to delete is over the threshold
+        """
+        test_runner = KillZombiesRunner()
+        test_runner.run()
+        self.assertTrue(mock_get_zombie_servers_count.called)
+        self.assertTrue(mock_trigger_warning.called)
+
+    @patch.object(KillZombiesRunner, "get_zombie_servers_count", return_value=0)
+    @patch("instance.tasks.call_command")
+    def test_do_nothing_if_zero(
+        self,
+        mock_get_zombie_servers_count,
+        mock_command
+    ):
+        """
+        Tests that only a dry_run is executed when
+        the number of zombies to terminate is zero.
+        """
+        test_runner = KillZombiesRunner()
+        test_runner.run()
+        mock_get_zombie_servers_count.assert_called_with(
+            "kill_zombies",
+            region="some-region",
+            dry_run=True
+        )
+
+    @override_settings(
+        DEFAULT_FROM_EMAIL="from@site.com",
+        ADMINS=(("admin3", "admin3@localhost"), ("admin4", "admin4@localhost"))
+    )
+    @patch("instance.tasks.call_command")
+    def test_send_warning_email(self, mock_command):
+        """
+        Tests that trigger_warning sends an email when called
+        """
+        test_runner = KillZombiesRunner()
+        test_runner.trigger_warning(5)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn(
+            "is over the KILL_ZOMBIES_WARNING_THRESHOLD",
+            mail.outbox[0].body
+        )
+        self.assertEqual(mail.outbox[0].from_email, "from@site.com")
+        self.assertEqual(mail.outbox[0].to, ["admin3@localhost", "admin4@localhost"])
+
+    @override_settings(
+        DEFAULT_FROM_EMAIL="from@site.com",
+        ADMINS=(("admin3", "admin3@localhost"),)
+    )
+    @patch.object(KillZombiesRunner, "get_zombie_servers_count", return_value=4)
+    @patch.object(KillZombiesRunner, "trigger_warning")
+    @patch("instance.tasks.call_command")
+    def test_send_email_on_failure(
+        self,
+        mock_get_zombie_servers_count,
+        mock_trigger_warning,
+        mock_command
+    ):
+        """
+        When call_command fails, it raises a CommandError.
+        Tests that an email is sent when CommandError is raised
+        """
+        mock_command.side_effect = CommandError()
+        test_runner = KillZombiesRunner()
+        test_runner.run()
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn(
+            "Scheduled execution of `kill_zombies` command failed",
+            mail.outbox[0].body
+        )
+        self.assertEqual(mail.outbox[0].from_email, "from@site.com")
+        self.assertEqual(mail.outbox[0].to, ["admin3@localhost"])
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -910,6 +910,12 @@ TRIAL_INSTANCES_REPORT_RECIPIENTS = env.json('TRIAL_INSTANCES_REPORT_RECIPIENTS'
 # Format is '<minute> <hour> <day> <month> <day_of_week>' like normal crontabs
 TRIAL_INSTANCES_REPORT_SCHEDULE = env('TRIAL_INSTANCES_REPORT_SCHEDULE', default='0 2 1 * *')
 
+# Kill Zombies Periodic task #################################################
+
+KILL_ZOMBIES_ENABLED = env("KILL_ZOMBIES_ENABLED", default=False)
+KILL_ZOMBIES_SCHEDULE = env("KILL_ZOMBIES_SCHEDULE", default="0 0 * * *")
+KILL_ZOMBIES_WARNING_THRESHOLD = env("KILL_ZOMBIES_WARNING_THRESHOLD", default=10)
+
 # User inactive and delete ####################################################
 
 CLEANUP_OLD_BETATEST_USERS = env('CLEANUP_OLD_BETATEST_USERS', default=True)


instance/tasks.py
```diff
@@ -47,6 +50,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Constants  ##################################################################
+
+KILL_ZOMBIES_CRON_SCHEDULE = os.environ.get("KILL_ZOMBIES_SCHEDULE", "0 0 */1 * *").split()
```
lgp171188, line 55: @spokerman12, `0 0 */1 * *` is the same as the simpler `0 0 * * *`, no? Afaik, the `*/1` just means run on every day of the month which is what `*` also means.

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.enabled: bool = bool(os.environ.get("KILL_ZOMBIES_ENABLED", False))
+        self.region: str = os.environ.get("KILL_ZOMBIES_REGION", "gra")
+        self.recipient: str = os.environ.get(
+            "KILL_ZOMBIES_WARNING_RECIPIENT", "ops@opencraft.com"
```
lgp171188, line 359: @spokerman12, I would recommend against using an actual email address specific to OpenCraft in a public repository like this.

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+        )
+        self.threshold: int = int(os.environ.get("KILL_ZOMBIES_THRESHOLD", 10))
```
lgp171188, line 361:
```suggestion
        self.warning_threshold: int = int(os.environ.get("KILL_ZOMBIES_WARNING_THRESHOLD", 10))
```

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+
+
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.enabled: bool = bool(os.environ.get("KILL_ZOMBIES_ENABLED", False))
+        self.region: str = os.environ.get("KILL_ZOMBIES_REGION", "gra")
```
lgp171188, line 357: @spokerman12, this is already available via `settings.OPENSTACK_REGION`.

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+
+    def handle_dry_run_output(self, stdout: str) -> int:
+        """
+        If there are servers to terminate, a `kill_zombies` dry run will warn:
+        "Would have terminated {} zombies" to stdout.
+        """
+        if "No servers found in region" in stdout:
+            return 0
+        try:
+            return int(stdout.split("Would have terminated")[1].split()[0])
```
lgp171188, line 397: @spokerman12, is there a better/cleaner way to get the number from the output? I can think of using a regular expression, but you know what they say
> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

😛 
In this case, it looks like using a regular expression might make this cleaner. The saying warns against overusing/using regular expressions everywhere without assessing if there are better solutions :)

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+    )
+)
+def kill_zombies_periodically():
+    """
+    Runs the 'kill_zombies' command to terminate servers
+    whose manual termination didn't succeed.
+    Defaults schedule is daily at 12AM
+    """
+    task_runner = KillZombiesRunner()
+    if task_runner.enabled:
```
lgp171188, line 418: @spokerman12, this would still a cause a DB task to be created, queued and processed by a worker though the task itself might do nothing. So do you think it will be cleaner to do something like what is done [here](https://github.com/open-craft/opencraft/blob/master/backup_swift/tasks.py#L105-L116)?

instance/tasks.py
```diff
@@ -339,3 +345,103 @@ def cleanup_old_betatest_users():
...
+                dry_run=True,
+            )
+            num_of_zombies = task_runner.handle_dry_run_output(dry_run_output)
+            if num_of_zombies > task_runner.threshold:
+                task_runner.trigger_warning()
+            output = call_command("kill_zombies", region=task_runner.region)
+            logging.info(
+                "Task `kill_zombies_periodically`ran succesfully "
+                "and received the following result: %s",
+                output
```
lgp171188, line 434: @spokerman12, I don't think it is a good idea to clutter the logs with the output of the command. If you still want to log something, it might make sense to just log the number of zombies that were cleaned up.

instance/tests/management/test_kill_zombies.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Tests that the task passes env variables to
+        the kill_zombies command call
+        """
+        kill_zombies_periodically()
+        mock_command.assert_called_with(
+            "kill_zombies", region="some-region"
+        )
+
+    @patch("instance.management.commands.kill_zombies.openstack_utils.get_nova_client")
```
lgp171188, line 75: @spokerman12, as discussed already, the `kill_zombies` command should be treated as a blackbox that works and hence there is no need to test it or anything that it invokes.

instance/tests/management/test_kill_zombies.py
```diff
@@ -0,0 +1,157 @@
...
+        of zombies to delete is over the threshold
+        """
+        kill_zombies_periodically()
+        self.assertTrue(mock_handle_dry_run_output.called)
+        self.assertTrue(mock_trigger_warning.called)
+
+    @override_settings(DEFAULT_FROM_EMAIL="from@site.com")
+    @patch.dict("os.environ", {"KILL_ZOMBIES_RECIPIENT": "to@site.com"})
+    @patch("instance.tasks.call_command")
+    def test_send_warning_email(self, mock_command):
```
lgp171188, line 123: @spokerman12, this test or another one should test that the email is sent to the expected recipient(s).

instance/tasks.py
spokerman12: You are right regarding the schedule.
On the Jira task and all discussions it was always stated "environment variables" rather than Django settings.

instance/tasks.py
spokerman12: Thank you for suggesting this. Added

instance/tasks.py
spokerman12: Modified as suggested.

instance/tasks.py
spokerman12: Thank you for mentioning this. Implemented.

instance/tasks.py
lgp171188: @spokerman12, you must've misunderstood what I meant probably because you still don't have a lot of context about Ocim. Ocim's Django settings are indeed populated via environment variables and that is what I meant. See [here](https://github.com/open-craft/opencraft/blob/master/opencraft/settings.py#L44) for example. Afaik, there is (almost) nothing that reads the environment variables outside of the Django settings.

instance/tasks.py
spokerman12: Oh of course, I love regex and I considered it but because we need to find `Would have terminated` and then the number, most reasonable implementations of this line need at least two calls, that is without using a more complex regex.
I tried once again and it seems a bit unnecessary for such a predictable log output.
Thanks for the suggestion, though!

instance/tasks.py
spokerman12: Aha, thanks for pointing this out.

instance/tasks.py
spokerman12: Thank you for suggesting this. Modified as suggested.

instance/tests/management/test_kill_zombies.py
spokerman12: Alright. Thanks for clarifying this. Modified to fit requirements.

instance/tests/management/test_kill_zombies.py
spokerman12: True. I added the assertion on both email testcases

instance/tasks.py
spokerman12: After reading a bit, the suggestion seems like the wrong approach as it should be Huey's responsibility to handle the tasks using [revoke](https://huey.readthedocs.io/en/latest/api.html#TaskWrapper.revoke) and pause.
I'll implement the proposal but without a test for toggle=true test because that would be testing Huey.

spokerman12: I tested it manually:

```
(opencraft) vagrant@ubuntu-xenial:~/opencraft$ echo $KILL_ZOMBIES_ENABLED
false
...
16:14:02 worker.1              | [2021-03-01 16:14:02,844] INFO:huey.consumer:MainThread:The following commands are available:
16:14:02 worker.1              | + instance.tasks.start_deployment
16:14:02 worker.1              | + instance.tasks.check_deactivation
16:14:02 worker.1              | + instance.tasks.spawn_appserver
16:14:02 worker.1              | + instance.tasks.make_appserver_active
16:14:02 worker.1              | + instance.tasks.shut_down_obsolete_pr_sandboxes
16:14:02 worker.1              | + instance.tasks.terminate_obsolete_appservers_all_instances
16:14:02 worker.1              | + instance.tasks.clean_up
16:14:02 worker.1              | + instance.tasks.delete_old_logs
16:14:02 worker.1              | + instance.tasks.cleanup_old_betatest_users
16:14:02 worker.1              | + registration.tasks.verify_external_domain_configuration

(opencraft) vagrant@ubuntu-xenial:~/opencraft$ echo $KILL_ZOMBIES_ENABLED
true
16:15:50 worker.1              | [2021-03-01 16:15:50,640] INFO:huey.consumer:MainThread:The following commands are available:
16:15:50 worker.1              | + instance.tasks.start_deployment
16:15:50 worker.1              | + instance.tasks.check_deactivation
16:15:50 worker.1              | + instance.tasks.spawn_appserver
16:15:50 worker.1              | + instance.tasks.make_appserver_active
16:15:50 worker.1              | + instance.tasks.shut_down_obsolete_pr_sandboxes
16:15:50 worker.1              | + instance.tasks.terminate_obsolete_appservers_all_instances
16:15:50 worker.1              | + instance.tasks.clean_up
16:15:50 worker.1              | + instance.tasks.delete_old_logs
16:15:50 worker.1              | + instance.tasks.cleanup_old_betatest_users
16:15:50 worker.1              | + instance.tasks.kill_zombies_periodically_task
16:15:50 worker.1              | + registration.tasks.verify_external_domain_configuration

```

instance/tasks.py
spokerman12: All discussions in the task referred to "environment variables".
I'll apply the requirement change early tomorrow.

spokerman12: Applied

instance/management/commands/kill_zombies.py
```diff
@@ -71,16 +71,18 @@ def handle(self, *args, **options):
...
         death_count = sum(1 for srv in nova_servers if self.not_zombie_or_die(srv))
 
         if self.dry_run:
-            self.log("Would have terminated {} zombies if this weren't a dry run.".format(death_count))
+            result = "Would have terminated {} zombies if this weren't a dry run.".format(death_count)
         else:
-            self.log("Terminated {} zombies.".format(death_count))
+            result = "Terminated {} zombies.".format(death_count)
+        self.log(result)
+        return result
```
lgp171188, line 85: @spokerman12, these changes appear to just be a cleanup of some of the existing code. Are they needed for any of the other changes made in this PR? If not, I would recommend dropping these from the PR.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+
+
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.region: str = getattr(settings, "OPENSTACK_REGION", "gra")
+        self.recipient: list = [email for _, email in getattr(settings, "ADMINS", [])]
```
lgp171188, line 356: @spokerman12, `settings.ADMINS` defaults to `[]` ([documentation](https://docs.djangoproject.com/en/2.2/ref/settings/#admins)). So I'd recommend using `settings.ADMINS` directly.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
         users_to_delete.delete()
+
+
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.region: str = getattr(settings, "OPENSTACK_REGION", "gra")
```
lgp171188, line 355: @spokerman12, `OPENSTACK_REGION` **has** to be set for Ocim to work. So `settings.OPENSTACK_REGION` can be used directly.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        self.recipient: list = [email for _, email in getattr(settings, "ADMINS", [])]
+        self.threshold: int = getattr(settings, "KILL_ZOMBIES_WARNING_THRESHOLD")
```
lgp171188, line 357: @spokerman12, using `getattr()` doesn't make much sense here unless you wanted to specify a default value and missed adding it. `settings.KILL_ZOMBIES_WARNING_THRESHOLD` should be okay otherwise.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        Utility method for sending emails
+        """
+        if self.recipient:
+            email = EmailMessage(
+                subject, body, settings.DEFAULT_FROM_EMAIL, self.recipient
+            )
+            return email.send()
+        else:
+            logging.info("Email recipient is undefined. Skipping email.")
+            return None
```
lgp171188, line 370: @spokerman12, multiple questions about this function.
* It looks like the return value of this function is not used anywhere at all. So why return anything? 
* If `return email.send()` is needed for some reason, the `else` clause is superficial and not needed.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+
+    def trigger_warning(self):
+        """
+        Warns when more than `threshold` VMs will be terminated
+        """
+        subject = "{} Zombie instances over current threshold ({})".format(
+            datetime.now().strftime("%Y-%m-d"), self.threshold
```
lgp171188, line 377: @spokerman12, I don't think it is necessary to include the timestamp in the email subject because it can be found by the time the email was sent/received. If you still want to use it, `timezone.now()` is better as it is timezone-aware.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        )
+        body = (
+            "The number of instances in region {} to be terminated"
```
lgp171188, line 380:
```suggestion
            "The number of zombie OpenStack VMs in the {} to be terminated "
```
Note the trailing space which is needed during string concatenation to separate the words.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+            "is over KILL_ZOMBIES_WARNING_THRESHOLD ({}).\n"
```
lgp171188, line 381:
```suggestion
            "The number of zombie OpenStack VMs to be terminated in the {} region "
            "is over the KILL_ZOMBIES_WARNING_THRESHOLD of {}.\n"
```

```suggestion
            "The number of zombie OpenStack VMs to be terminated in the {} region  "
```
Note the trailing space which is needed during string concatenation to separate the words.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        )
+        self.send_email(subject, body)
+
+    def handle_dry_run_output(self, stdout: str) -> int:
+        """
+        If there are servers to terminate, a `kill_zombies` dry run will warn:
+        "Would have terminated {} zombies" to stdout.
+        """
+        if "No servers found in region" in stdout:
+            return 0
```
lgp171188, line 395: @spokerman12, it will be good to log a warning message in this case.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+    logging.info("Executing periodic task `kill_zombies_periodically`")
+    try:
+        dry_run_output, output = "", ""
+        dry_run_output = call_command(
+            "kill_zombies",
+            region=task_runner.region,
+            dry_run=True,
+        )
+        num_of_zombies = task_runner.handle_dry_run_output(dry_run_output)
+        if num_of_zombies > task_runner.threshold:
```
lgp171188, line 435: @spokerman12, this will choke when the `handle_dry_run_output()` function call returns `None`.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        If there are servers to terminate, a `kill_zombies` dry run will warn:
+        "Would have terminated {} zombies" to stdout.
+        """
+        if "No servers found in region" in stdout:
+            return 0
+        try:
+            return int(stdout.split("Would have terminated")[1].split()[0])
+        except IndexError:
+            logger.info("Received unexpected input from dry run")
+            return None
```
lgp171188, line 400: The code that invokes this function only treats it as an integer. So this will cause trouble there. I'd recommend using a variable `number_of_zombies_to_terminate` with `0` as its default value and returning that.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+            "The number of instances in region {} to be terminated"
+            "is over KILL_ZOMBIES_WARNING_THRESHOLD ({}).\n"
+        ).format(self.region, self.threshold)
+        logging.info(
+            "%s\nSending an email notification to: %s",
+            body, self.recipient
+        )
+        self.send_email(subject, body)
+
+    def handle_dry_run_output(self, stdout: str) -> int:
```
lgp171188, line 389: @spokerman12, wouldn't something like `get_number_of_zombie_servers_to_terminate()` be a better name for this function as it clearly explains what this function does?

instance/tasks.py
lgp171188, line 395: As mentioned [here](https://github.com/open-craft/opencraft/pull/721/files#r587551180), it is a good idea to avoid using `return` statements in multiple places.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        """
+        If there are servers to terminate, a `kill_zombies` dry run will warn:
+        "Would have terminated {} zombies" to stdout.
+        """
+        if "No servers found in region" in stdout:
+            return 0
+        try:
+            return int(stdout.split("Would have terminated")[1].split()[0])
+        except IndexError:
```
lgp171188, line 398: @spokerman12, what if the `int()` call returned a `ValueError`?

instance/tasks.py
spokerman12: Pylint complains otherwise

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+            logger.info("Received unexpected input from dry run")
+            return None
+
+
+if getattr(settings, "KILL_ZOMBIES_ENABLED"):
```
lgp171188, line 403: @spokerman12, what if `KILL_ZOMBIES_ENABLED` is defined and is set to `False`?

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+    @db_periodic_task(crontab(
+        minute=KILL_ZOMBIES_CRON_SCHEDULE[0],
+        hour=KILL_ZOMBIES_CRON_SCHEDULE[1],
+        day=KILL_ZOMBIES_CRON_SCHEDULE[2],
+        month=KILL_ZOMBIES_CRON_SCHEDULE[3],
+        day_of_week=KILL_ZOMBIES_CRON_SCHEDULE[4],
+    ))
+    def kill_zombies_periodically_task():
```
lgp171188, line 411: @spokerman12, the `periodically` part is already implied by the usage of the `db_periodic_task` decorator above and imho, it shouldn't be added in the name.

Also, why is this wrapper needed in the first place instead of adding the implementation directly within this function?

instance/tasks.py
lgp171188: @spokerman12, what do you mean by that? Pylint **shouldn't** complain if you define a variable with a default value and return it **once** at the end.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+    task_runner = KillZombiesRunner()
+    logging.info("Executing periodic task `kill_zombies_periodically`")
+    try:
+        dry_run_output, output = "", ""
+        dry_run_output = call_command(
+            "kill_zombies",
+            region=task_runner.region,
+            dry_run=True,
+        )
+        num_of_zombies = task_runner.handle_dry_run_output(dry_run_output)
```
lgp171188, line 434: Since you have abstracted away all the logic in the `KillZombiesRunner` class, this function shouldn't have to bother about the "how" part and should just ask `task_runner.get_zombie_servers_count()` to give it the number? After all, it is the "runner" class according to its name.

instance/tasks.py
```diff
@@ -339,3 +344,111 @@ def cleanup_old_betatest_users():
...
+        if num_of_zombies > task_runner.threshold:
+            task_runner.trigger_warning()
+        output = call_command("kill_zombies", region=task_runner.region)
```
lgp171188, line 437: @spokerman12, similar concerns as [here](https://github.com/open-craft/opencraft/pull/721/files#r587564362).

instance/tests/management/test_kill_zombies.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertFalse(mock_command.called)
+
+    @patch("instance.tasks.call_command")
+    def test_set_env_variables(self, mock_command):
+        """
+        Tests that the task passes env variables to
+        the kill_zombies command call
+        """
+        kill_zombies_periodically()
+        mock_command.assert_called_with(
```
lgp171188, line 66: @spokerman12, shouldn't you also test if the same command was called twice, once with the flag for the dry run?

instance/management/commands/kill_zombies.py
spokerman12: These are needed as `result` is `call_command`'s return value.

instance/tasks.py
spokerman12: Applied

instance/tasks.py
spokerman12: Applied

instance/tasks.py
spokerman12: Set `10` as default.

instance/tasks.py
spokerman12: Pylint complains if we don't set `return None`.
This was old design. I'll remove both statements.

instance/tasks.py
spokerman12: Removed

instance/tasks.py
spokerman12: Applied. Thanks for the whitespace addendum.

instance/tasks.py
spokerman12: `call_command` for `kill_zombies` already logs this here https://github.com/open-craft/opencraft/blob/f043e87a2fd3ac26cf3a7528d411d0ebda3b693b/instance/management/commands/kill_zombies.py#L75

instance/tasks.py
spokerman12: Addressed as suggested in next comment.

instance/tasks.py
spokerman12: I made it a`return None` because it would fail only if `kill_zombies` should fail. Passing a `0` instead of raising an error will allow `kill_zombies_periodically`'s `call_command` to `kill_zombies` to fail silently.
I applied the suggestion nonetheless.

instance/tasks.py
spokerman12: Applied as suggested

instance/tasks.py
spokerman12: This was a composite issue. I've applied the previous suggestion.

instance/tasks.py
spokerman12: This was modified following previous suggestions.

instance/tasks.py
spokerman12: This is tested here https://github.com/open-craft/opencraft/blob/d4e6dc4d10da72becd3c93d18c928040d84794e2/instance/tests/management/test_kill_zombies.py#L47

instance/tasks.py
spokerman12: This is a design oversight caused by side effects of addressing the first review. I'll implement the request.

spokerman12: The wrapper is necessary because if a periodic task is defined, Huey will run it unless we explicitly tell it not to.

instance/tasks.py
spokerman12: I'll apply this.

instance/tasks.py
spokerman12: Addressed in previous comment.

instance/tests/management/test_kill_zombies.py
spokerman12: Added as suggested

spokerman12: Added as suggested

instance/tasks.py
lgp171188: @spokerman12, I don't see any code changes handling the `ValueError` that might be thrown by the `int()` call now?

instance/tasks.py
lgp171188: @spokerman12, why this `getattr()` call instead of just using `settings.KILL_ZOMBIES_ENABLED`?

instance/tasks.py
lgp171188: @spokerman12, I don't understand your answer. Why can't the body of the `kill_zombies_periodically()` function be directly within this `kill_zombies_periodically_task()` function? That is the wrapping that I asked about in my previous comment.

instance/tasks.py
```diff
@@ -339,3 +344,117 @@ def cleanup_old_betatest_users():
...
+
+class KillZombiesRunner:
+    """
+    Helper class to run `kill_zombies_periodically`
+    """
+
+    def __init__(self):
+        self.region: str = settings.OPENSTACK_REGION
+        self.recipient: list = [email for _, email in settings.ADMINS]
+        self.threshold: int = getattr(settings, "KILL_ZOMBIES_WARNING_THRESHOLD", 10)
```
lgp171188, line 357: @spokerman12, is there a reason for using `getattr()` calls everywhere instead of just using `settings.SETTING_NAME`? In this case, the default is already defined in `settings.KILL_ZOMBIE_WARNING_THRESHOLD`.

instance/tasks.py
```diff
@@ -47,6 +49,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Constants  ##################################################################
+
+KILL_ZOMBIES_CRON_SCHEDULE = getattr(settings, "KILL_ZOMBIES_SCHEDULE").split()
```
lgp171188, line 54: @spokerman12, is there a reason for using `getattr()` instead of just `settings.KILL_ZOMBIES_SCHEDULE`, which is defined in the Django settings and has a default value defined.

instance/tests/management/test_kill_zombies.py
lgp171188: @spokerman12, this is now a duplicate of the assertion in the last line of this test method, no?

instance/tests/management/test_kill_zombies.py
```diff
@@ -0,0 +1,156 @@
...
+        no_instances = """
+        2021-02-26 14:16:12 | Starting kill_zombies
+        2021-02-26 14:16:12 | No servers found in region some-region.
+        """
+        some_instances = """
+        2021-02-26 14:16:12 | Starting kill_zombies
+        2021-02-26 14:16:12 | Found 32 unterminated servers in region some-region.
+        2021-02-26 14:16:12 | Would have terminated 32 zombies if this weren't a dry run.
+        """
+        self.assertEqual(test_runner.get_zombie_servers_count(no_instances), 0)
```
lgp171188, line 89: @spokerman12, can you add an assertion to verify that in case the output doesn't match the expected pattern, the returned count is `0`?

instance/tasks.py
```diff
@@ -339,3 +344,117 @@ def cleanup_old_betatest_users():
...
+            dry_run_output, output = "", ""
+            dry_run_output = call_command(
+                "kill_zombies",
+                region=self.region,
+                dry_run=True,
+            )
+            num_of_zombies = self.get_zombie_servers_count(dry_run_output)
+            if num_of_zombies > self.threshold:
+                self.trigger_warning()
+            output = call_command("kill_zombies", region=self.region)
```
lgp171188, line 416: @spokerman12, we can optimize by not running the 2nd `call_command` call if the `num_of_zombies` is `0`. What do you say?

instance/tasks.py
```diff
@@ -339,3 +344,117 @@ def cleanup_old_betatest_users():
...
+    def trigger_warning(self):
+        """
+        Warns when more than `threshold` VMs will be terminated
+        """
+        subject = "Zombie instances are over the current threshold ({})".format(
+            self.threshold
+        )
+        body = (
+            "The number of zombie OpenStack VMs in the {} region "
+            "is over the KILL_ZOMBIES_WARNING_THRESHOLD ({}).\n"
```
lgp171188, line 380: @spokerman12, it will be good to add the actual number of zombies in this text.

instance/tasks.py
```diff
@@ -339,3 +344,117 @@ def cleanup_old_betatest_users():
...
+                "kill_zombies",
+                region=self.region,
+                dry_run=True,
+            )
+            num_of_zombies = self.get_zombie_servers_count(dry_run_output)
+            if num_of_zombies > self.threshold:
+                self.trigger_warning()
+            output = call_command("kill_zombies", region=self.region)
+            logging.info(
+                "Task `kill_zombies_periodically`ran succesfully "
```
lgp171188, line 418:
```suggestion
                "Task `kill_zombies_periodically` ran succesfully "
```

instance/tasks.py
lgp171188: This is what happens if that value is not an integer.

```
 File "/var/www/.pyenv/versions/opencraft/lib/python3.6/site-packages/huey/api.py", line 342, in _execute
    task_value = task.execute()
  File "/var/www/.pyenv/versions/opencraft/lib/python3.6/site-packages/huey/api.py", line 690, in execute
    return func(*args, **kwargs)
  File "/var/www/.pyenv/versions/opencraft/lib/python3.6/site-packages/huey/contrib/djhuey/__init__.py", line 128, in inner
    return fn(*args, **kwargs)
  File "/var/www/opencraft/instance/tasks.py", line 449, in kill_zombies_task
    kill_zombies_periodically()
  File "/var/www/opencraft/instance/tasks.py", line 460, in kill_zombies_periodically
    task_runner.run()
  File "/var/www/opencraft/instance/tasks.py", line 413, in run
    num_of_zombies = self.get_zombie_servers_count(dry_run_output)
  File "/var/www/opencraft/instance/tasks.py", line 397, in get_zombie_servers_count
    num_instances = int(stdout.split("Would have terminated")[1].split()[0])
ValueError: invalid literal for int() with base 10: 'abc'
```

instance/tasks.py
lgp171188: @spokerman12, defining `10` as a default here duplicates the default set in `settings.KILL_ZOMBIES_WARNING_THRESHOLD` and hence is unnecessary. So you should just use `settings.KILL_ZOMBIES_WARNING_THRESHOLD` here.

instance/tasks.py
lgp171188: @spokerman12, if failing silently is a concern, we should notify the failure. Otherwise, it will still be undetected even though it is in the logs.

instance/tests/management/test_kill_zombies.py
spokerman12: Ah yes. Thanks for noticing.

instance/tasks.py
spokerman12: Deleted

instance/tasks.py
spokerman12: Applied the suggestion.

instance/tasks.py
spokerman12: `stdout` is well-defined here https://github.com/open-craft/opencraft/blob/f70334b0b80e12a0f8ddbd0ee981171479774eb5/instance/management/commands/kill_zombies.py#L85 as a well-formed string or None. If those strings were badly-formatted, that's an issue of not having tests for `kill_zombies` in the first place.
Added.

instance/tasks.py
spokerman12: Because I'm not used to having an Env layer that adds the defaults for settings.
Thanks for spotting this. Modified.

instance/tasks.py
spokerman12: Addressed in previous comments.
The first implementation used "environment variables" as it was discussed.

instance/tasks.py
spokerman12: The first implementation used "environment variables" as it was discussed.

instance/tests/management/test_kill_zombies.py
spokerman12: Added

instance/tasks.py
spokerman12: Added

instance/tasks.py
spokerman12: Fixed. Thanks for spotting this.

instance/tasks.py
spokerman12: True. Added.

instance/tasks.py
spokerman12: OK I see what you mean now; I didn't see it from the `run` method being an addition from the reviews.
I'll apply this.

instance/tasks.py
spokerman12: My bad for overlooking this; it's too ingrained in my head that when fetching settings one should always set a default.

spokerman12: Oh, and I also picked it up from here https://github.com/open-craft/opencraft/blob/f043e87a2fd3ac26cf3a7528d411d0ebda3b693b/reports/tasks.py#L26

instance/tasks.py
```diff
@@ -339,3 +344,109 @@ def cleanup_old_betatest_users():
...
+                "kill_zombies",
+                region=self.region,
+                dry_run=True,
+            )
+            num_of_zombies = self.get_zombie_servers_count(dry_run_output)
+            if num_of_zombies > self.threshold:
+                self.trigger_warning(num_of_zombies)
+            elif num_of_zombies != 0:
+                output = call_command("kill_zombies", region=self.region)
+            logging.info(
```
lgp171188, line 418: @spokerman12, shouldn't this log statement be indented to be within the `elif` block so that the log message is printed only when `num_of_zombies != 0`?

instance/tests/management/test_kill_zombies.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_set_env_variables(self, mock_command):
+        """
+        Tests that the task passes env variables to
+        the kill_zombies command call
+        """
+        test_runner = KillZombiesRunner()
+        test_runner.run()
+        first_call = call.mock_command("kill_zombies", region="some-region", dry_run=True)
+        self.assertEqual(mock_command.mock_calls[0], first_call)
+        mock_command.assert_called_with("kill_zombies", region="some-region")
```
lgp171188, line 69: @spokerman12, does this check the last call of the mock command? This is relevant because the previous statement checks the first call and we shouldn't be checking the same again.

lgp171188: If this doesn't test the expected thing, there is a similar assertion below that may have to be modified as well.

spokerman12: Yes, `assert_called_with` only checks the last call (so last call is not a dry_run), that's why I define a `first_call` then use `assertEqual` to compare with the `mock_calls[0]`, which is the first that happened.

instance/tasks.py
spokerman12: It could, but at this point, even if `num_of_zombies != 0`, the task did run successfully and this is the log entry for saying so. If I put it inside the logging call I'd have to write another logging call. Logging `task run successfully and there were no VMs to terminate` is a valid and desirable output to me.
Do you suggest otherwise?

lgp171188: @spokerman12, okay to leave this as-is to avoid another commit and the following CI checks.

lgp171188: @spokerman12, I spoke too soon. Your PR branch is already outdated and has to be updated. So you can go ahead and add the additional log statement.

lgp171188 (approved):
@spokerman12, you can squash and merge this PR after merging in the latest changes from the `master` branch, adding the fix for the logging message review comment and the CI checks are all green after that :+1: 
- [x] I tested the changes in this PR and verified that the periodic task runs on schedule, kills the zombies and sends an email notification to `settings.ADMINS` when the number of zombies to be kill is > the configured threshold.
- [x] I read through the code

PR #723, opened 2021-03-02
refactor: add instance created log message

gabor-boros:
This PR adds a new log record when an instance is created. This is needed to know to be able to have better metrics in the future.

**Dependencies**: None

**Sandbox URL**: N/A

**Merge deadline**: None

**Testing instructions**:

1. Deploy to stage
2. Create a new instance manually or with an instance factory
3. Validate that the line is logged
4. Open Kibana and search for the log line

1. We may need other logs in the future to create visualizations in Kibana, but for now, this is the only case we cannot visualize. somehow

**Reviewers**
- [ ] @0x29a

### DIFF
--- a/instance/models/openedx_instance.py
+++ b/instance/models/openedx_instance.py
@@ -153,6 +153,10 @@ class OpenEdXInstance(
         if self.random_prefix is not None:
             self.mysql_user = self.random_prefix
 
+        # The instance is newly created
+        if not self.pk:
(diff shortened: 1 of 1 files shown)


0x29a (approved):
:+1:

- [ ] I didn't test this because the change is trivial, and we're going to test this on staging.
- [x] Includes documentation

PR #724, opened 2021-03-03
[FAL-873] Prettify ansible output

mtyaka:
**Description**

This tweaks ansible output to make it valid JSON in more cases than the defaults.
It then uses some JS on the frontend to identify and pretty-print JSON fragments in ansible output.

**Screenshot**

Screenshot 2021-03-17 at 09 36 06

**Testing**

This was deployed to stage, so you can check the logs of appservers that were spawned when the code was active on stage. You can use these two:

- https://stage.manage.opencraft.com/instance/7420/edx-appserver/3170/
- https://stage.manage.opencraft.com/instance/7400/edx-appserver/3171/

**Reviewers**

- [ ] @tartansandal

### DIFF
--- a/instance/ansible.py
+++ b/instance/ansible.py
@@ -33,9 +33,16 @@ from django.conf import settings
 
 from instance.utils import poll_streams, create_temp_dir
 
+
 # Logging #####################################################################
 
 logger = logging.getLogger(__name__)
+
+
+# Custom ansible plugins ######################################################
+
+ANSIBLE_CALLBACKS_DIR = os.path.join(os.path.dirname(__file__), 'ansible_callbacks')
+ANSIBLE_STDOUT_CALLBACK = 'prettify'
 
 
 # Functions ###################################################################
@@ -154,6 +161,10 @@ def run_playbook(requirements_path, inventory_str, vars_str, playbook_path, play
         # Disable SSH host key checking by Ansible – since IP addresses are constantly reused,
         # changing host keys are expected.
         env['ANSIBLE_HOST_KEY_CHECKING'] = 'false'
+
+        # Use a custom output callback to make ansible logs more readable.
+        env['ANSIBLE_CALLBACK_PLUGINS'] = ANSIBLE_CALLBACKS_DIR
+        env['ANSIBLE_STDOUT_CALLBACK'] = ANSIBLE_STDOUT_CALLBACK
 
         yield subprocess.Popen(
             cmd,
--- a/instance/static/html/instance/appserver.html
+++ b/instance/static/html/instance/appserver.html
@@ -205,7 +205,7 @@
            ng-repeat="line in appserverLogs.log_error_entries track by $index">
         <span class="timestamp">{{ line.created | date:'yyyy-MM-dd HH:mm:ssZ' }}</span>
         <span class="log-level">{{ line.level }}</span>
-        <pre>{{ line.text | stripLogMeta }}</pre>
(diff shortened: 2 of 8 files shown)


instance/tests/integration/ansible/requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/edx/ansible.git@stable-1.9-plus-edx#egg=ansible==1.9.1-edx.2
```
tartansandal, line 1: Koa requires 2.8.17 so we might consider bumping this higher.

instance/ansible.py
```diff
@@ -33,11 +33,18 @@
...
+
 # Logging #####################################################################
 
 logger = logging.getLogger(__name__)
 
 
+# Custom ansible plugins ######################################################
+
+ANSIBLE_CALLBACKS_DIR = os.path.join(os.path.dirname(__file__), 'ansible_callbacks')
+ANSIBLE_STDOUT_CALLBACK = 'prettify'
```
tartansandal, line 45: @mtyaka Rather than using a custom callback, what about using the built in `yaml` callback (see https://www.shellhacks.com/ansible-human-readable-output-format/). That might work better for web display (fewer double quotes) and would mean less code for us to maintain. Although that would be a major rework to what you have done. Maybe a follow up ticket.

instance/static/html/instance/appserver.html
```diff
@@ -214,7 +214,7 @@ <h6 ng-if="appserver.status == 'failed'">Full Log</h6>
            ng-repeat="line in appserverLogs.log_entries track by $index">
         <span class="timestamp">{{ line.created | date:'yyyy-MM-dd HH:mm:ssZ' }}</span>
         <span class="log-level">{{ line.level }}</span>
-        <pre>{{ line.text | stripLogMeta }}</pre>
+        <pre>{{ line.text | stripLogMeta | prettifyJSON }}</pre>
```
tartansandal, line 217: @mtyaka Would be great if we also had a filter to remove the `stdout` and `stderr` keys which are both unreadable and duplicated by the `stdout_lines` and `stderr_lines` keys. Maybe for a followup ticket?

instance/tests/integration/ansible/requirements.txt
mtyaka: Good idea, I'll do that.

instance/static/html/instance/appserver.html
mtyaka: I didn't notice `stdout`/`stdout_lines` and `stderr`/`stderr_lines` is always duplicated, nice spot! I'll modify the filter to remove `stdout`/`stderr` from the output if the corresponding `*_lines` entry is present.

instance/ansible.py
mtyaka: I tried the built-in `yaml` (and `json`) callbacks before creating a custom one. The `yaml` callback is unfortunately unusable for our purpose. Its output is very similar to the default output and not much easier to pretty print.
The `json` callback is even less useful because it buffers the output of the entire playbook and only outputs a huge dump of json when the playbook finishes running.

tartansandal (approved):
:+1: All good. Be great if you could demo this during your sprint update.  Especially showing off the difference when spinning up an appserver fails an generates a large dump.

- [x] I tested this:
  - [x] checked that tests passed
  - [x] reviewed output on Ocim stage
- [x] I read through the code
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #726, opened 2021-03-04
BB-3777 Add ButtonStyles component

arjunsinghy96:
Adds a ButtonStyles component extracted out of ButtonCustomizationPage component.

**Tickets** [BB-3777](https://tasks.opencraft.com/browse/BB-3777) [Gitlab#710](https://gitlab.com/opencraft/dev/opencraft/-/issues/710)

**Screenshots**: See Gitlab issues

**Testing Instructions**
1. Checkout to this branch
2. Start OCIM backend in vagrant using make run.dev
3. Start frontend using npm run start --prefix=frontend
4. Make sure you have a BetaTestApplication with a test user already created. If not visit admin and create appropriate records. Without these the Frontend will not be fully functional.
5. Visit http://localhost:3000/console/theming/buttons
6. In react devtools, check that `ButtonStyles` component is used to display the color pickers collapsible
7. Update color values and check Button Previews is updated.
8. Check in django admin that theming changes are reflected in the BetaTestApplication

### DIFF
--- a/frontend/src/console/components/ButtonStyles/ButtonStyles.tsx
+++ b/frontend/src/console/components/ButtonStyles/ButtonStyles.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+import { InstanceSettingsModel } from 'console/models';
+import { ThemeSchema } from 'ocim-client';
+import { Col, Container, Row } from 'react-bootstrap';
+import { CollapseEditArea, ColorInputField } from 'ui/components';
+import { WrappedMessage } from 'utils/intl';
+import messages from './displayMessages';
+import './styles.scss';
+
+interface ButtonStylesProp {
+  onChangeColor: Function;
+  buttonName: string;
+  themeData: ThemeSchema;
+  initialExpanded?: boolean;
+  loading: Array<keyof InstanceSettingsModel | 'deployment'>;
+}
+
+export const ButtonStyles: React.FC<ButtonStylesProp> = (
+  props: ButtonStylesProp
+) => {
+  const { themeData } = props;
+  const { loading } = props;
+
+  const styles = {
+    activeStyles: [
+      `btn${props.buttonName}Bg`,
+      `btn${props.buttonName}Color`,
+      `btn${props.buttonName}BorderColor`
+    ],
+    hoverStyles: [
+      `btn${props.buttonName}HoverBg`,
+      `btn${props.buttonName}HoverColor`,
+      `btn${props.buttonName}HoverBorderColor`
+    ]
+  };
+
+  return (
+    <CollapseEditArea initialExpanded={props.initialExpanded || false}>
+      <Container className="theme-button-customization-container">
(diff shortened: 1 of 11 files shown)


frontend/src/console/components/ButtonStyles/ButtonStyles.tsx
```diff
@@ -0,0 +1,74 @@
...
+    ],
+    hoverStyles: [
+      `btn${props.buttonName}HoverBg`,
+      `btn${props.buttonName}HoverColor`,
+      `btn${props.buttonName}HoverBorderColor`
+    ]
+  };
+
+  return (
+    <div>
```
shimulch, line 38: Do we need this outer `div`? Maybe a fragment `<>` or `CollapseEditArea` as root element would be just fine.

shimulch (changes requested):
@arjunsinghy96, The PR seems fine to me. Just address that minor nit.

shimulch (approved):
LGTM :+1:

- [x] I tested this locally. Checked if the new component works correctly and doesn't break anything. 
- [x] I read through the code
- [x] Includes documentation

farhaanbukhsh:
> @arjunsinghy96 @shimulch Has this been reviewed by Cassie or Ali before merging? Remember that this is now a requirement on all the internal accelerated projects.
> 

No this is not reviewed by Cassie or Ali

> @farhaanbukhsh Can you make sure this is done on all the Ocim tasks, including any that has already been recently merged?

I will do that @antoviaque even for the upcoming PR but with respect to this PR I don't think it is required since we have not introduced any changes on the user side(user-facing component), we have just refactored an existing set of the component so that it can transition into the new UI with ease. Let me know if you think otherwise.

And thanks a bunch for checking up on the PRs :smile:

shimulch:
@antoviaque, @farhaanbukhsh, I might have missed that as Cassie or Ali was not assigned as 2nd reviewer for the task. BTW this change has no visual changes, other than just decoupling some components for better reusability.

antoviaque:
@farhaanbukhsh @shimulch OK, thank you for the details. Even in cases where the functionality isn't supposed to be altered, a refactor can be enough to create an unexpected side effect. I would still ask for a review in cases like this, even if it's a quick one.

antoviaque:
@shimulch @farhaanbukhsh 

> I would still ask for a review in cases like this, even if it's a quick one

Was this done?

shimulch:
@antoviaque, Sorry for acting late on this. I've pinged Cassie on the ticket. 

cc @farhaanbukhsh

shimulch:
@antoviaque, Just to let you know that, Cassie [reviewed](https://gitlab.com/opencraft/dev/opencraft/-/issues/710#note_535403655) the changes. Thanks a lot for pointing this out.

cc @farhaanbukhsh

antoviaque:
@shimulch Great! Thank you for taking care of this. :) :+1:

arjunsinghy96:
Thanks for handling this @shimulch

Issue #729, opened 2021-03-08
Installation doc needs to mention provisioning the virtualbox


lsiden:
When I ran "make migration" inside my new OCIM virtual box as per the [installation instructions](https://ocim.opencraft.com/en/latest/installation/), it failed. It complained "coverage: command not found". I installed pip3, then coverage, ran "make migrate" again and this time it was missing "honcho".

In order to get it to just work, I had to quite the virtualbox and run "vagrant provision", then "vagrant reload".  Then, in the virtualbox, "make migration" finished successfully.

The command "vagrant help up" shows that --provision is a default for "vagrant up" so this extra step should not have been required, but that's what I needed to do to get that to work.

I propose to update the Installation doc page to note this.  I will create a PR soon.

lgp171188:
This was found to be an invalid issue. Closing.

PR #730, opened 2021-03-08
Updates installation page in response to Issue 729

lsiden:
Changes 'vagrant up' to 'vagrant --provision up'

https://github.com/open-craft/opencraft/issues/729

### DIFF
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -54,7 +54,7 @@ repository – most notably Ansible:
 
 Now you can run:
 
-    vagrant up
+    vagrant --provision up
 
 This will provision a virtual machine running Ubuntu 16.04, set up local
 Postgres, MySQL, MongoDB and Redis, and install the dependencies.


documentation/installation.md
```diff
@@ -54,7 +54,7 @@ repository – most notably Ansible:
 
 Now you can run:
 
-    vagrant up
+    vagrant --provision up
```
lgp171188, line 57: @lsiden, the last time I checked `vagrant` did run the provisioner unless `--no-provision` was specified. Did that change? If not, can you explain why this is needed?

lsiden: I agree, @lgp171188 . The help for "vagrant up" indicates that "--provision" is set by default, but I couldn't get "make migrate" to finish until I exited the ssh session and ran "vagrant provision; vagrant reload". 

I don't have any other explanation. 

The first time, when I just ran "vagrant up" and then started a ssh shell and ran make in the VM, Make complained that "coverage" was missing. I installed that by hand in the VM. Next Make complained that Honcho was missing.  Then I began to suspect that the VM never got provisioned.  After I exited and ran "vagrant --provision up; reload" and reconnected, Make was able to finish.

It's possible I missed something else, but I'm not sure it's worth the time it would take to recreate the whole process to figure out what else might have gone wrong. If you think it is worth it, then I will create a new ticket for that.

Of course, I could change the mod to say

    If 'make ...' doesn't work, then exit and run 'vagrant provision; vagrant reload' and try again.

Tell me what you think is best.

lgp171188: @lsiden, I can try provisioning the Ocim devstack afresh tomorrow and let you know if I run into the same issue as you did.

lgp171188: @lsiden, I set up a fresh devstack now and I don't see this issue at all. I used Python 3.6.10 for the virtualenv environment on the host where ansible and other requirements are installed from `deploy/requirements.txt`. The provisioning completed fully without any errors.

I was able to then run `pip freeze` inside the VM to check if all the requirements including the ones that you had issues with were installed and they indeed were. I also ran `make migrate` and then `make run.dev` and it worked fine and started the Django dev server without any issues.

Since `vagrant up` does run the provisioner, this looks like an issue with something on or related to your setup and I would recommend against making the change in this PR or even adding this comment

```
If 'make ...' doesn't work, then exit and run 'vagrant provision; vagrant reload' and try again.
```

lsiden: I'll close it, then. You're right, I might have missed something in my setup. When I began setting it up, I installed a virtual environment with Python 3.9, Make failed, so I had to go back and start again with Python 3.8. That's probably why I had to run "vagrant provision" again. My bad. I should have thought of that.

lsiden:
I'll close it, then. You're right, I might have missed something in my setup. When I began setting it up, I installed a virtual environment with Python 3.9, Make failed, so I had to go back and start again with Python 3.8. That's probably why I had to run "vagrant provision" again. My bad. I should have thought of that.

lsiden:
I'll close it, then. You're right, I might have missed something in my setup. When I began setting it up, I installed a virtual environment with Python 3.9, Make failed, so I had to go back and start again with Python 3.8. That's probably why I had to run "vagrant provision" again. My bad. I should have thought of that.

PR #732, opened 2021-03-09
Fix the example code for creating a new OCIM user.

janimo:
Task: BB-3956
Fix the documentation at https://ocim.opencraft.com/en/latest/users/

Make sure to use set_password(), since the plaintext password won't match
the PBKDF2 secured one saved in the DB.


0x29a (approved):
:+1:

PR #733, opened 2021-03-10
prod-vars.yml is no longer used, update the docs.

janimo:


documentation/provisioning-sandboxes.md
```diff
@@ -69,10 +69,7 @@ instance.save()
 ```
 
 If you pass custom `configuration_extra_settings` to `production_instance_factory`,
-they will be merged with the settings in [prod-vars.yml](https://github.com/open-craft/opencraft/blob/master/instance/templates/instance/ansible/prod-vars.yml).
-Settings that you pass in will take precedence over settings in prod-vars.yml,
-that is, if a variable is present in both `configuration_extra_settings` and prod-vars.yml,
-the instance manager will use the value from `configuration_extra_settings` for it.
+they will override variables from the existing hardcoded settings.
```
lgp171188, line 72:
```suggestion
they will override the variables defined in `instance/models/mixins/openedx_config.py`.
```

lgp171188:
@janimo, can you update your PR branch to have all the latest changes in the `master` branch and address my review comment? After that, I will approve and merge this PR.

janimo:
@lgp171188 thanks, I made the change and force pushed a new commit.

PR #736, opened 2021-03-12
Drop Python 2 dependency

janimo:


janimo:
Ocim on stage runs fine with ANSIBLE_PYTHON_PATH=/usr/bin/python3 being used for the virtualenv.
Ansible could be run directly without making a virtualenv for it, now that it runs with Python 3.

PR #737, opened 2021-03-12
Check subdomain name validity for instance factories.

janimo:
Add extra asserts for when calling the instance factory methods
from the Django shell and there's no previous validation step involved.
This prevents domain names that are duplicates and only differ in case,
or domains with invalid characters in them which will cause problems
later in provisioning.


0x29a (approved):
:+1:

- [x] I tested `is_valid_domain_name` doesn't match wrong subdomains.
- [x] I read through the code
- [x] Includes documentation

PR #740, opened 2021-03-16
Simple docker-compose setup

janimo:
Task: BB-3956

This allows running the database services needed for Ocim, including the ones needed for the testsuite. It uses the same versions as the edX devstack when possible. Ocim itself runs as in the traditional local setup in a virtualenv, not dockerized.

The default Postgres user is set as the system user so it works with the existing .env.test setup. Other choices may be better, but will need adjusting existing env files.

A section describing this setup is added to the installation docs.

I saw https://github.com/open-craft/opencraft/compare/kshitij/docker_support and there are probably other similar changes, but it would be good to have them unified and provide a compose setup by default.


0x29a:
This was completed in https://github.com/open-craft/opencraft/pull/795.

PR #741, opened 2021-03-17
WIP: [BB-3647] Implement user account page

giantas:
WIP


registration/tests/test_api.py
```diff
@@ -246,6 +246,76 @@ def test_account_update_password_success(self):
...
+        """
+        self.user.set_password('OldV4l1dPw()')
+        self.user.save()
+
+        self.client.force_login(self.user)
+        url = reverse("api:v2:accounts-password-change", kwargs={'username': self.user.username})
+        response = self.client.patch(
+            url,
+            data={
+                "new_password": "123456789",
```
kaizoku, line 297: I don't see the old password passed here as well. Won't this just be testing that the old password is required?

kaizoku (changes requested):
Looks pretty good @giantas. I have a few small comments, but I didn't run into any functional issues when testing.

frontend/packages/api_client/src/models/OpenEdXAppServer.ts
```diff
@@ -84,6 +84,30 @@ export interface OpenEdXAppServer {
...
+     * @type {string}
+     * @memberof OpenEdXAppServer
+     */
+    ansibleAppserverRequirementsPath?: string;
+    /**
+     * The version of the Ansible playbook repository to checkout.
+     * @type {string}
+     * @memberof OpenEdXAppServer
+     */
+    ansibleAppserverVersion?: string;
```
0x29a, line 110: @giantas, just in case. I just merged a PR that adds the same fields in the client, so you need to regenerate it.

registration/tests/test_api.py
giantas: True, true. Fixing that, too.

frontend/src/console/components/Account/displayMessages.ts
```diff
@@ -0,0 +1,56 @@
+const messages = {
+  explanation: {
```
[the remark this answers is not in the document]
giantas: It is needed since the contents are different from those in `src/console/components/NoticeBoard/displayMessages.ts`. Almost each component has its own set of displayMessages.

frontend/packages/api_client/src/models/OpenEdXAppServer.ts
giantas: @0x29a thanks for the heads-up. There doesn't seem to be a diff currently.

PR #760, opened 2021-04-03
[BB-3801] Fix the course mode defaults in the ecommerce alternative currency howto

lgp171188:
The configuration example now only updates the currency in the default mode (audit)
and doesn't mix it with changing the default mode to honor as it can be
confusing for those who don't want to do that.

Since I didn't want to use `honor` mode as the default mode for the client instance I set up for [BB-3806](https://tasks.opencraft.com/browse/BB-3806), I assumed that the configuration can be skipped and that lead to an outage of sorts :)

### DIFF
--- a/documentation/howtos.md
+++ b/documentation/howtos.md
@@ -59,6 +59,8 @@ ECOMMERCE_REPOS:
     SSH_KEY: '{{ ECOMMERCE_GIT_IDENTITY }}'
 
 # Append to any existing EDXAPP_LMS_ENV_EXTRA:
+# Without this, LMS will error out as the default value
+# doesn't have this currency.
 EDXAPP_LMS_ENV_EXTRA:
   COURSE_MODE_DEFAULTS:
     bulk_sku: !!null
@@ -66,9 +68,9 @@ EDXAPP_LMS_ENV_EXTRA:
     description: !!null
     expiration_datetime: !!null
     min_price: 0
-    name: Honor
+    name: Audit
     sku: !!null
-    slug: honor
+    slug: audit
     suggested_prices: ''
 ```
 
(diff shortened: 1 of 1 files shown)


lgp171188:
@swalladge, can you review this documentation change as the author of the lines that have been changed in this PR?

documentation/howtos.md
```diff
@@ -59,16 +59,18 @@ ECOMMERCE_REPOS:
     SSH_KEY: '{{ ECOMMERCE_GIT_IDENTITY }}'
 
 # Append to any existing EDXAPP_LMS_ENV_EXTRA:
+# Without this, LMS will error out as the default value
+# doesn't have this currency.
```
swalladge, line 63: @lgp171188 the default value of what?

lgp171188: @swalladge, the default value of `COURSE_MODE_DEFAULT` defined [here](https://github.com/open-craft/edx-platform/blob/open-release/juniper.3/lms/envs/common.py#L854-L864). This [line](https://github.com/open-craft/edx-platform/blob/open-release/juniper.3/common/djangoapps/course_modes/models.py#L770) errors out when the value passed to the `min()` function is an empty list, which happens when the default course mode doesn't have the default currency.

lgp171188: @swalladge, I have made this clear in 2761d4f1dc3f254bb837b4350dc443df458820ce.

swalladge: Ah that's perfect, thanks :+1:

swalladge (approved):
@lgp171188 :+1:

- [x] I tested this: checked that the changes make sense
- [x] I read through the code
- [x] Includes documentation

PR #768, opened 2021-04-09
[BB-4015] Cronjob for getting monthly active user count

janimo:
## Description

This adds a task similar to the existing monthly report. It sends an email with the number of logged in users for the past month for each instance.

## Supporting information

BB-4015

## Testing instructions

The management script can be run manually and it will generate a CSV.

The Huey task can be tested by setting it to run more frequently (on stage?) and check whether the results are ok.

## Other information

It may make sense to refactor this, as the new periodic task is very similar of the existing one.
The crontab setting may need to be different so it does not reuse the exact same ones as the other report task.

### DIFF
--- a/instance/management/commands/instance_statistics_csv.py
+++ b/instance/management/commands/instance_statistics_csv.py
@@ -76,6 +76,12 @@ class Command(BaseCommand):
             required=True
         )
         parser.add_argument(
+            '--users',
+            '-u',
+            action='store_true',
+            help='Gather active user statistics for all instances',
+        )
+        parser.add_argument(
             '--start-date',
             default=(
                 datetime.utcnow().date() - timedelta(days=self.DEFAULT_NUM_DAYS)
@@ -131,7 +137,10 @@ class Command(BaseCommand):
             ))
             sys.exit(1)
 
-        self.collect_instance_statistics(out, domains, start_date, end_date)
+        if options['users']:
+            self.collect_active_users(out, domains, start_date, end_date)
+        else:
+            self.collect_instance_statistics(out, domains, start_date, end_date)
 
     def get_instances_from_domain_names(self, domain_names):
         """ Get an instance object for a given domain name """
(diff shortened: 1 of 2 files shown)


instance/management/commands/instance_statistics_csv.py
```diff
@@ -131,7 +137,10 @@ def handle(self, *args, **options):
             ))
             sys.exit(1)
 
-        self.collect_instance_statistics(out, domains, start_date, end_date)
+        if options['users']:
+            self.collect_active_users(out, domains, start_date, end_date)
+        else:
+            self.collect_instance_statistics(out, domains, start_date, end_date)
```
lgp171188, line 143: @janimo, we should add the `active users` data to the existing `collect_instance_statistics` function, i.e., the number of active users should be another column added to the existing report.

instance/management/commands/instance_statistics_csv.py
```diff
@@ -276,6 +285,38 @@ def log_line(line):
...
 
+    def collect_active_users(
+            self,
+            out,
+            domain_names,
+            start_date,
+            end_date
+    ):
+        """Generate the active user statistics CSV."""
+        instances = self.get_instances_from_domain_names(domain_names)
```
lgp171188, line 296: @janimo, we already run a python script on each instance's active appserver(s) to collect some data. Can't the code to do the querying be added to that? That way, the calling code in Ocim need not know about what is being done and how to collect this data?

lgp171188 (changes requested):
@janimo, please check the review comments that I have left.

instance/management/commands/instance_statistics_csv.py
janimo: @lgp171188 the existing task only produces stats for the trial instances created in the last month. Do you mean I should create another CSV which contains not only the active users but all the existing stats for all instances?

instance/management/commands/instance_statistics_csv.py
janimo: I considered that possibility but it meant touching more parts of the whole setup and also looked harder to test. I'll look into that.

instance/management/commands/instance_statistics_csv.py
```diff
@@ -75,6 +75,12 @@ def add_arguments(self, parser):
             help='Comma-separated list of fully qualified domains of the instances to get the data for',
             required=True
         )
+        parser.add_argument(
+            '--users',
+            '-u',
+            action='store_true',
+            help='Gather active user statistics for all instances',
+        )
```
lgp171188, line 83: @janimo, can we enable this for all runs of this command instead of gating it behind a flag?

instance/management/commands/instance_statistics_csv.py
```diff
@@ -131,7 +137,10 @@ def handle(self, *args, **options):
             ))
             sys.exit(1)
 
-        self.collect_instance_statistics(out, domains, start_date, end_date)
+        if options['users']:
+            self.collect_active_users(out, domains, start_date, end_date)
```
lgp171188, line 141: @janimo, as stated [here](https://github.com/open-craft/opencraft/pull/768/files#r611534309), this should be an addition to the existing functionality of the command and not something separate.

instance/management/commands/instance_statistics_csv.py
lgp171188: @janimo, I hope my comments above answers your questions.

The current implementation of the management command produces a report with N columns and the periodic task primarily determines the list of instances to pass to that command based on some criteria (instances created in the last completed month). The desired result for this PR should be to add a new column with the number of active users in the output of the management command - i.e., it should now have N+1 columns.

With this done, the existing periodic task will automatically get this new column.

You will then have to implement another periodic task which just determines the list of instances to pass to the management command using different criteria - all the betatest instances, which includes the trial instances and the P&T plan subscribers. For the sake of simplicity, you can query all the instances that have an associated betatestapplication instance attached.

janimo:
I've opened another PR for a new branch at #781using a different approach.

janimo:
Replaced by another PR.

PR #775, opened 2021-04-21
[BB-3963] Add marketing djangoapp

arjunsinghy96:
### Description
Adds a `marketing` django app with `EmailTemplate`, `Subscriber` and `SentEmail` models.

Adds settings `MARKETING_EMAIL_SENDER` and `MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS`

Adds periodic tasks `send_followup_emails` to dispatch all followup emails for the day and `prune_emails` to delete all old emails (configurable using `settings.MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS`)

### Jira Tickets
[BB-3963](https://tasks.opencraft.com/browse/BB-3963)

#### Common

1. Get the devstack running and checkout to this branch.
2. Run migrations.

#### Checking admin and admin action
1. Start the dev server and visit the admin panel at http://localhost:5000/admin/marketing/
2. Verify that each of the three models `EmailTemplate`, `SentEmail` and `Subscriber` are registered.
2. Create a new inactive `EmailTemplate` object.
3. Verify that `send_sample_emails` action is available in `EmailTemplate` admin.
4. Select the Template created above and trigger the `send_sample_emails` action.
5. Check dev server logs and verfiy that the email was sent.

#### Testing tasks
1. Set `MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS` to `0`. (This will be required when testing `prune_emails`)
2. Restart the dev server.
3. Create a `Subscriber` object and set `trial_started_at` to a appropriate value such that `trial_started_at + EmailTemplate.send_after_days` is equal to now. One example will be to set `send_after_days` to `1` and `trial_started_at` to yesterday's date with time less than current time
4.  With dev server still running open a shell using `make shell` in another terminal.
5. Run following code
```python
from marketing.tasks import send_followup_emails
send_followup_emails()
```
6. In dev server logs verify that the email was sent to subscriber.
7. Trigger the task again (using `send_followup_email`) and verfiy that this time no email is sent.
8. Verify that the `SentEmail` model have the recently sent email (Check in django admin)
9. Verfiy that the saved emails in `SentEmail` are not editable from admin panel.
10. Now run the following code in shell to verfiy the `prune_email` task
```python
from marketing.tasks import prune_emails
prune_emails()
```
10. Verify that the `SentEmail` models has no elements. (Django admin can be used.)

#### Template variables

Following variables are provided in template context:
1. `full_name`: String
2. `username`: String
3. `instance_name`: String
4. `subdomain`: String
5. `application`: Instance of `BetaTestApplication`

Try out these variables using django's `{{ variable }}` template syntax.


marketing/tasks.py
```diff
@@ -0,0 +1,108 @@
...
+            subject=subject,
+            html_body=html_body,
+            plaintext_body=plaintext_body,
+        )
+
+
+# Tasks #######################################################################
+
+@db_periodic_task(crontab(day='*', hour='1', minute='0'))
+def send_followup_emails():
```
giovannicimolin, line 78: The logic between this and `render_email_and_send` doesn't have a clear division of responsibilities, and it also doesn't look like it'll scale well. You can make use of `prefetch` to avoid extra queries here.

marketing/tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        "username": application.user.get_username(),
+        "instance_name": application.instance.name,
+        "subdomain": application.subdomain,
+        "application": application
+    })
+    subject = template.subject_template.render(context)
+    html_body = template.html_body_template.render(context)
+    plaintext_body = template.plaintext_body_template.render(context)
+
+    if not SentEmail.objects.filter(recipient_email=application.user.email, template=template).exists()
```
giovannicimolin, line 58: This is not required if you filter these cases out in the parent method. This is making the template be rendered every time, even if the email is not going to be sent.

giovannicimolin (changes requested):

marketing/signals.py
```diff
@@ -0,0 +1,12 @@
+from django.db.models.signals import post_save
+from django.dispatch import recievers
+
+from registration.models import BetaTestApplication
+
+@recievers(post_save, sender=BetaTestApplication)
```
giovannicimolin, line 6: Instead of adding the user on registration time, why not just add them once the first appserver is provisioned (and add a timestamp in the model)?

There's a method that does the same thing on the registration module: https://github.com/open-craft/opencraft/blob/b7090d2654a4228805819e5fb28631f451c481f1/registration/approval.py#L64

This will help make the filtering/sorting less complex on the `send_followup_emails` much easier. Since we're waiting the first appserver to add a user to the "mailing" list.

marketing/tasks.py
```diff
@@ -0,0 +1,158 @@
...
+        ),
+        Prefetch(
+            'sentemail_set__template'
+        )
+    )
+
+# Tasks #######################################################################
+
+@db_periodic_task(crontab(day='*', hour='1', minute='0'))
+def send_followup_emails():
```
giovannicimolin, line 120: This can be much simpler if we make some assumptions (and the changes I suggested on the signal).

marketing/tasks.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Sends all the configured followup emails.
+    """
+    emails_to_send = []
+    eligible_users = create_eligible_user_queryset()
```
giovannicimolin, line 125: This is not efficient at all (we're potentially fetching all active users + betatest application data + instance data) all in one go.

This can be heavily simplified if we:
1. Have a timestamp of the first activated appserver in the `Subscriber` model.
2. Instead of filtering by users, first retrieve all active templates.
3. Loop over the templates, and use a queryset to filter if we're in the send range instead of doing the calculation in python. 
```
# Something along these lines
for template in active_templates:
    gte = datetime.now() - timedelta(days=template.send_after_days)
    lt = datetime.now() - timedelta(days=template.send_after_days + 1)
    to_send = Subscribers.objects.filter(created_at__gte=gte, created_at__lt=lt)
    # Prefetch stuff 
    ....
    # Render template & send
    ....
```
This way you'll only be querying for users that are actually in the send range for the emails, as opposed of retrieving all instances and doing the filtering in python.

giovannicimolin (changes requested):

marketing/tasks.py
```diff
@@ -0,0 +1,105 @@
...
+    For better understanding following is a timeline containing filter days
+                   +-------------------+
+    <start>--------|<gte>--1-day--<lte>|--------send-after-days----<now>--
+                   +-------------------+
+    Eligible subscribers are those which are created in our one day window
+
+    Uses prefetch to reduce DB queries.
+    """
+    gte = now() - timedelta(days=template.send_after_days + 1)
+    lte = now() - timedelta(days=template.send_after_days)
```
giovannicimolin, line 59: I think it should be `lt` here to avoid overlaps.
```suggestion
    lt = now() - timedelta(days=template.send_after_days)
```

marketing/tests/test_tasks.py
```diff
@@ -0,0 +1,249 @@
...
+        Tests that prune_emails deltes old emails.
+        """
+        self.assertEqual(SentEmail.objects.count(), len(self.day_delta_list))
+        prune_emails()
+        self.assertEqual(SentEmail.objects.count(), 5)
+
+    @override_settings(MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS=5)
+    def test_prune_emails_config_setting(self):
+        """
+        Test that prune emails is configuratable using setting
```
giovannicimolin, line 85:
```suggestion
        Test that prune emails is configurable using setting
```

giovannicimolin (commented):
Can you add testing instructions and update the PR description?

marketing/tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    subject = template.subject_template.render(context)
+    html_body = template.html_body_template.render(context)
+    plaintext_body = template.plaintext_body_template.render(context)
+
+    if not SentEmail.objects.filter(recipient_email=application.user.email, template=template).exists()
+        send_mail(
+            subject=subject,
+            message=plaintext_body,
+            html_message=html_body,
+            from_email=settings.DEFAULT_FROM_EMAIL,
```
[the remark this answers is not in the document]
giovannicimolin: This is still an issue.

marketing/tasks.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Prune the sent emails to control the size of storage.
+
+    We delete all the email sent before 30 days
+    """
+    # TODO: Clean this up
+    delete_delta = settings.EMAIL_DELETE_DELTA_DAYS or 30
+    delete_before = timezone.now() - timedelta(days=delete_delta)
+    delete_count, _ = SentEmail.objects.filter(sent_at__lte=delete_before).delete()
+    logger.info("Purged %s followup email", delete_count)
```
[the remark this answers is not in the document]
giovannicimolin: @arjunsinghy96 ^

marketing/models.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        return Template(self.html_body)
+
+    @cached_property
+    def plaintext_body_template(self):
+        """
+        Get Template instance for plaintext_body.
+        """
+        return Template(self.plaintext_body)
+
```
giovannicimolin, line 87: Minor safeguard to avoid setting up emails after the prune time.
```suggestion

    def save(self, *args, **kwargs):
        # When saving, check if `send_after_days` is lower than MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS
        # otherwise emails will be repeatedly sent since the cleanup job will clear those settings.
        if self.send_after_days > settings.MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS:
            raise Exception(
                 "Configuring follow up emails after %i is not allowed", 
                 settings.MARKETING_DELETE_FOLLOWUP_EMAILS_AFTER_DAYS
            )
        super(Model, self).save(*args, **kwargs)

```

giovannicimolin (changes requested):
Minor changes.

marketing/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+def render_and_dispatch_email(template: EmailTemplate, subscriber: Subscriber):
+    """
+    Given an EmailTemplate and Subscriber, renders the email from
+    template and sends the email.
+    """
+    application = subscriber.user.betatestapplication
+    user = subscriber.user
+    context = Context({
+        "full_name": user.get_full_name(),
```
giovannicimolin, line 46: When registering using the registration form, the user's name is stored in the profile model (`user.profile.full_name`) so this method doesn't return the full name from the correct place (make sure to prefetch it).

giovannicimolin (changes requested):
Found a small bug when retrieving an username.

marketing/models.py
```diff
@@ -0,0 +1,146 @@
...
+
+class Subscriber(models.Model):
+    """
+    Model to store followup email subscribers
+    """
+    user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE)
+    receive_followup = models.BooleanField(
+        default=True,
+        help_text="Should the user recieve followup emails?"
```
lgp171188, line 135: @arjunsinghy96, nit:

```suggestion
        help_text="Should the user receive followup emails?"
```
and to be updated in the corresponding migration as well.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. I've followed the testing instructions.
2. Checked that emails work properly on stage.
3. Checked that emails are not sent twice.
4. Checked that disabling a subscriber disables their emails.
5. Tested HTML templates and also the templating support
- [x] I read through the code

@arjunsinghy96 Good to go :)

PR #781, opened 2021-04-26
[BB-4015] Monitor active users WIP

janimo:
## Description

This adds a task similar to the existing monthly report. It sends an email with the number of logged in users for the past month for each instance.

## Supporting information

BB-4015

## Testing instructions

I have yet to test this on stage.

The management script can be run manually and it will generate a CSV.

The Huey task can be tested by setting it to run more frequently (on stage?) and check whether the results are ok.

## Other information

It may make sense to refactor this, as the new periodic task is very similar of the existing one.
The crontab setting may need to be different so it does not reuse the exact same ones as the other report task.

### DIFF
--- a/instance/management/commands/instance_statistics_csv.py
+++ b/instance/management/commands/instance_statistics_csv.py
@@ -230,7 +230,7 @@ class Command(BaseCommand):
             logger_=log_line,
         )
 
-    def get_instance_usage_data(self, playbook_output_dir, instance_metadata):
+    def get_instance_usage_data(self, playbook_output_dir, instance_metadata, start_date, end_date):
         """ Execute the collect_activity playbook to gather statistics """
         inventory = '[apps]'
         for name_prefix, metadata in instance_metadata.items():
@@ -257,6 +257,7 @@ class Command(BaseCommand):
         log_line.error = log_line
 
         playbook_extra_script_arguments = '--skip-hit-statistics' if settings.INSTANCE_LOGS_SERVER_HOST else ''
+        playbook_extra_script_arguments += f'--start-date {start_date} --end-date {end_date}'
 
         # Launch the collect_activity playbook, which places a file into the `playbook_output_dir`
         # on this host.
@@ -266,7 +267,7 @@ class Command(BaseCommand):
             vars_str=(
                 'local_output_dir: {output_dir}\n'
                 'remote_output_filename: /tmp/activity_report\n'
-                'extra_script_arguments: {extra_script_arguments}'
+                'extra_script_arguments: {extra_script_arguments}\n'
             ).format(
                 output_dir=playbook_output_dir,
                 extra_script_arguments=playbook_extra_script_arguments
@@ -298,7 +299,9 @@ class Command(BaseCommand):
             )
             self.get_instance_usage_data(
                 playbook_output_dir,
-                instance_metadata
+                instance_metadata,
+                start_date,
+                end_date
             )
 
             csv_writer = csv.writer(out, quoting=csv.QUOTE_NONNUMERIC)
@@ -311,6 +314,7 @@ class Command(BaseCommand):
                 'Total Hits',
                 'Total Courses',
                 'Total Users',
+                'Active Users',
                 'Age (Days)'
             ])
 
@@ -334,6 +338,7 @@ class Command(BaseCommand):
                     section.get('total_hits', 'N/A'),
                     section.get('courses', 'N/A'),
                     section.get('users', 'N/A'),
+                    section.get('active_users', 'N/A'),
                     metadata['instance_age']
                 ])
 
--- a/playbooks/collect_activity/stats.py
+++ b/playbooks/collect_activity/stats.py
@@ -3,12 +3,14 @@
 
 from __future__ import print_function
 from argparse import ArgumentParser
+from datetime import datetime, timedelta
 import gzip
 import os
 import re
 import sys
 
 import django
+from django.utils import timezone
 
 from six.moves.configparser import ConfigParser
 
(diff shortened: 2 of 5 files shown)


reports/tasks.py
```diff
@@ -127,3 +127,89 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+@db_periodic_task(
+    crontab(
+        minute=TRIAL_INSTANCES_REPORT_SCHEDULE_MINUTE,
+        hour=TRIAL_INSTANCES_REPORT_SCHEDULE_HOUR + 2,
+        day=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY,
+        month=TRIAL_INSTANCES_REPORT_SCHEDULE_MONTH,
+        day_of_week=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY_OF_WEEK
+    )
+)
+def send_all_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
```
lgp171188, line 142: @janimo, since this function is duplicating the task above, can you use something like the solution documented [here](https://stackoverflow.com/questions/51977211/how-to-create-a-task-function-with-an-argument-using-huey) to refactor the common code and just have a toggle that determines whether to restrict the queryset to the instances created in the last completed month? The code will become simpler with that change.

playbooks/collect_activity/stats.py
```diff
@@ -52,8 +77,22 @@
 
     ms = modulestore()
 
+    start_date = args.start_date
+    end_date = args.end_date
+
+    if start_date is None or end_date is None:
```
lgp171188, line 83: @janimo, this condition will never be true since `--start-date` and `--end-date` are required parameters? I'd recommend having the following logic:
* Both parameters should either be specified together or not at all.
* In the latter case, let the time range default to the last completed month.

Does this make sense?

instance/management/commands/instance_statistics_csv.py
```diff
@@ -230,7 +230,7 @@ def log_line(line):
             logger_=log_line,
         )
 
-    def get_instance_usage_data(self, playbook_output_dir, instance_metadata):
+    def get_instance_usage_data(self, playbook_output_dir, instance_metadata, start_date, end_date):
         """ Execute the collect_activity playbook to gather statistics """
```
lgp171188, line 234: @janimo, can you add/update the existing unit tests for this management command for your changes in this PR?

lgp171188 (changes requested):
@janimo, I have reviewed this PR and left some review comments. Please check and address those. Also fix the failing unit tests before requesting another review.

reports/tasks.py
janimo: @lgp171188 I agree that since much of the code is duplicated refactoring would make sense. Besides the queryset, the email subject and the error messages should be different.

playbooks/collect_activity/stats.py
janimo: @lgp171188 I overlooked the required=True attribute when copying the code, since there was a default=None setting. The logic you describe was my intention, and should work if the required=True is removed which I have done now. The args should not be required, in case the stats.py script is being called from elsewhere, but also for backwards compat while we develop.

lgp171188:
@jani, [this failure](https://app.circleci.com/pipelines/github/open-craft/opencraft/3629/workflows/be6eedf6-1d04-453d-b5ab-11258bde10aa/jobs/32607) is due to the changes made for this task not having the corresponding updates to the integration test case in question. Can you check and fix it?

playbooks/collect_activity/stats.py
```diff
@@ -52,8 +75,22 @@
...
+        beginning_of_this_month = datetime.utcnow().date().replace(day=1)
+        end_of_last_month = beginning_of_this_month - timedelta(days=1)
+        beginning_of_last_month = end_of_last_month.replace(day=1)
+
+        start_date = beginning_of_last_month
+        end_date = end_of_last_month
+
     stats = {
         'users': User.objects.count(),
+        'active_users': User.objects.filter(last_login__range=(start_date, end_date)).count(),
```
lgp171188, line 93: @janimo, this is a native datetime though the time zone support is active. You should use the UTC timezone for these datetime objects.

janimo: @lgp171188 you mean in the active_users filter? That is handled differently from the other date queries (like set of instances created in last month) because they run on the instances database and not on OCIM's own?

Is using utcnow() above not enough? Should it be datetime.now(tz=utc)? From what I see the last_login is stored in UTC, and so is I assume the time argument passed to the script.

reports/tasks.py
```diff
@@ -127,3 +106,83 @@ def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIP
...
+@db_periodic_task(
+    crontab(
+        minute=TRIAL_INSTANCES_REPORT_SCHEDULE_MINUTE,
+        hour=TRIAL_INSTANCES_REPORT_SCHEDULE_HOUR,
+        day=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY,
+        month=TRIAL_INSTANCES_REPORT_SCHEDULE_MONTH,
+        day_of_week=TRIAL_INSTANCES_REPORT_SCHEDULE_DAY_OF_WEEK
+    )
+)
+def send_trial_instances_report(recipients=settings.TRIAL_INSTANCES_REPORT_RECIPIENTS):
```
lgp171188, line 123: @janimo, can you add/update the tests in `reports/tests.py` (`ReportTaskTestCase` class) for the changes made in this PR?

playbooks/collect_activity/stats.py
lgp171188: > The return value will give us the current date in UTC, right?

@janimo, a time zone has a meaning only for the `datetime` objects and not `date` objects.

lgp171188:
@janimo, can you check and update the PR to address my open review comments? I will do another review pass and if everything is okay at that point, we can merge this PR.

reports/tasks.py
janimo: I have now added a few tests in that file.

playbooks/collect_activity/stats.py
lgp171188: @janimo, this is still not addressed.

janimo: @lgp171188 I have fixed it to use tz aware datetimes instead.

lgp171188 (approved):
@janimo, :+1:. You can merge this PR **after** all the CI checks pass. If the integration tests fail due to some issue **not related** to the changes in this PR, please block this PR merge and the task progression on that fix. Otherwise, please fix the issues, if any, before merging.

- [x] I tested this on stage and verified that it works as expected.
- [x] I read through the code
- [x] ~~Includes documentation~~

PR #784, opened 2021-04-29
fix: fixed the style for registration page

farhaanbukhsh:
Fix the registration styling in the page

**JIRA tickets**: [BB-3786](https://tasks.opencraft.com/browse/BB-3786)

**Discussions**: [Mattermost](https://chat.opencraft.com/opencraft/pl/aeq3b9x8i7noxqsugnqd8mj58o)

**Dependencies**: None

**Screenshots**: 

Before:
image

After:
image

**Testing instructions**:

1. Set up the Devstack
2. Go to the registration page it should have the background image

This is a way to fix the current UI

**Reviewers**

- [ ] @0x29a

### DIFF
--- a/frontend/src/registration/components/RegistrationContainer/styles.scss
+++ b/frontend/src/registration/components/RegistrationContainer/styles.scss
@@ -1,0 +1,8 @@
+@import '~styles/theme';
+
+.registration-container {
+  background-image: url("~assets/header-background.jpg");
+  background-color: $primary-1;
+  background-size: 100% 350px;
+  background-repeat: no-repeat;
+}


0x29a (changes requested):
@farhaanbukhsh, fix works fine.

However, I checked what commit is adding breaking change and found that it's https://github.com/open-craft/opencraft/commit/edcb1fa95dacf650a55a35e5b1382a477ebbc1fa.

Given that we already have styles that add exactly the same background, could you check if we can avoid duplication? Probably we can wrap registration container in `ContentPage` component?

If you don't have time to do this, I think it's fine to ask FF for help.

0x29a:
Or, probably we can move styles back to `app-container`.

farhaanbukhsh:
> Or, probably we can move styles back to `app-container`.

In app-container it is breaking the theme-customizer  and in general the background

0x29a (approved):
:+1:

Possibility of reusing existing styles will be checked in [BB-4143](https://tasks.opencraft.com/browse/BB-4143).

- [x] I tested that changes from this PR are fixing header on registration page.
- [x] I read through the changes.

farhaanbukhsh:
@0x29a  merging this in :)

PR #789, opened 2021-05-07
[BB-4141] Add task for daily email report

arjunsinghy96:
Adds the daily report task to `marketing` app

### Related Tickets
[BB-4141](https://tasks.opencraft.com/browse/BB-4141)

#### Testing report email is sent after sending followup emails.
1. Checkout to this branch
2. Run `make migrate` if you haven't already after the addition of marketing app.
4. Run dev server
5. Visit http://localhost:5000/admin/marketing/emailtemplate/add/ and add an active template with send_after_days set to 1.
6. Visit http://localhost:5000/admin/marketing/emailtemplate/add/ and add a subscriber with `trial_started_at` set to yesterday's date and a time less than now.
7. Feel free to create more records of `EmailTemplate` and `Subscriber` with different values
5. In another terminal open the django shell using `make shell`
6. Import the tasks using `from marketing.tasks import send_followup_emails, send_report`
7. Run `send_followup_emails()`
8. In server logs, verifiy that the report was sent for all the triggered emails.
9. Run the `send_followup_emails()` task again and verify that this time the report email says `No emails were sent today to P&T and trail users.`

#### Testing `send_report` standalone

1. In django shell import `send_report` task
2. `send_report` accepts a dictionary with `Email names` as key and `List of emails` as values.
3. Trigger `send_report` task using `send_report({"Welcome Email": ["arjun@example.com", "user2@example.com"]})`
4. In server logs, verify that the email is sent with appropriate template.
5. Change the values (increase keys and emails) and call `send_report` task.
6. Verify that email is sent with all the records passed.

### Author Notes

1. The `send_report` task is scheduled right after the `send_followup_email` with a list of followup emails send. I think it made more sense there because we wouldn't have to make a DB query to get the sent emails and it is the next step right after sending all followup emails.


marketing/tasks.py
```diff
@@ -92,7 +98,38 @@ def send_followup_emails():
...
+
+
+@db_task(retry=3, retry_delay=10)
+def send_report(sent_emails: Dict[str, Set]):
+    """
+    Render and send the report email.
+
+    This task has to be called from send_followup_emails
+    """
+    context = {
```
giovannicimolin, line 115: Two minor nits, don't send if:
1. No emails where sent.
2. No `MARKETING_EMAIL_REPORT_RECIPIENTS` are set up.

marketing/tasks.py
```diff
@@ -92,7 +98,38 @@ def send_followup_emails():
 
     # Try sending all the followup emails
     for args in mails_to_send:
-        render_and_dispatch_email(*args)
+        template, subscriber = args
+        render_and_dispatch_email(template, subscriber)
+        sent_emails[template.name].add(subscriber.user.email)
+
+    send_report(sent_emails)
```
shimulch, line 105: @arjunsinghy96, I understand your reasoning to call this task right after the `send_followup_emails` task. But I don't think they are actually dependent on each other. I have two scenarios in my mind where this can be problematic.

1. Say in the future we want to run the `send_followup_emails` task multiple times in a day. In that case, this approach will not work.
2. If `send_followup_emails` fails (say a timeout or something else) after sending a couple of emails, we will be losing those in our report. Right?

Therefore, I think it would be better to keep them separate. What do you think?
cc @giovannicimolin

shimulch (changes requested):
@arjunsinghy96, Thanks for the amazing work. :)

But I do have some comments on the same thing you've mentioned in Author's Concern section.

marketing/tasks.py
arjunsinghy96: @shimulch
> Say in the future we want to run the `send_followup_emails` task multiple times in a day. In that case, this approach will not work.

I think we should be sending the report after every trigger of `send_followup_emails` as that would provide a better report of what was sent.
If in future we decide to run `send_followup_emails` multiple time, we can either
1. Send report after every run (recommended)
2. Add a separate task to fetch emails from `SentEmail` model and trigger `send_report`

> If send_followup_emails fails (say a timeout or something else) after sending a couple of emails, we will be losing those in our report. Right?

Correct. Thanks for catching this. I think putting it in a `try..finally` block will fix. I will update the PR with it.

marketing/tasks.py
```diff
@@ -91,8 +97,51 @@ def send_followup_emails():
...
+
+    This task has to be called from send_followup_emails.
+
+    The report email will not be sent if either the sent_emails is an
+    empty dictionary or settings.MARKETING_EMAIL_REPORT_RECIPIENTS is
+    not set.
+    """
+    recipients = settings.MARKETING_EMAIL_REPORT_RECIPIENTS
+    if not sent_emails:
+        logger.info("No markerting emails were sent. Skipping report email.")
```
shimulch, line 122: Typo - 

```suggestion
        logger.info("No marketing emails were sent. Skipping report email.")
```

marketing/tasks.py
```diff
@@ -91,8 +97,51 @@ def send_followup_emails():
...
+    try:
+        send_mail(
+            subject="Daily report for marketing emails",
+            message=plaintext_body,
+            html_message=html_body,
+            from_email=settings.MARKETING_EMAIL_SENDER,
+            recipient_list=recipients
+        )
+    except smtplib.SMTPException:
+        logger.error("Failed to send the report for marketing emails!")
```
shimulch, line 144: So that we don't miss out on useful debug info. :)

```suggestion
    except smtplib.SMTPException as e:
        logger.exception("Failed to send the report for marketing emails! %s", e)
```

marketing/templates/email_report.html
```diff
@@ -0,0 +1,23 @@
+<div>
+    <p>Hello from OCIM,</p>
+    {% if sent_emails %}
```
shimulch, line 3: Since we are [checking](https://github.com/open-craft/opencraft/pull/789/files#diff-780cc6cdf2129833893b838950d6686891522ce2a81b4aa51c612e5be5c81c1aR121) if `sent_emails` exists - this might not required

marketing/templates/email_report.txt
```diff
@@ -0,0 +1,10 @@
+Hello from OCIM,
+
+{% if sent_emails %}I've sent today's share of emails to P&T and trail users.
```
shimulch, line 3: Since we are [checking](https://github.com/open-craft/opencraft/pull/789/files#diff-780cc6cdf2129833893b838950d6686891522ce2a81b4aa51c612e5be5c81c1aR121) if `sent_emails` exists - this check might not required

shimulch (changes requested):
@arjunsinghy96, Thanks for the update. It is looking quite good. Could you address following nits as well? :)

marketing/tasks.py
arjunsinghy96: Thanks for  catching this :)

shimulch (approved):
@arjunsinghy96 Thanks for the awesome work. LGTM :+1:

- [x] I tested this locally, checked if the report email is being sent when there are reports, and not if there is none.
- [x] I read through the code
- [x] Includes documentation

shimulch:
@arjunsinghy96, Please rebase and squash your commits with master. :)

PR #795, opened 2021-05-11
feature: Added ability to run OCIM in a container.

kewne:
# Description

This adds some basic infrastructure for developing OCIM inside docker containers.
It already makes it possible to run some of the dev workflow (e.g. tests and OCIM) in Docker and could be used as a starting point for running a dockerized OCIM in production.

Kudos to @janimo for the prior work on #740 

# Testing 

Follow the instructions in [docker.md](https://github.com/open-craft/opencraft/blob/joaocabrita/ocim_in_docker/documentation/development/docker.md)

### DIFF
--- a/Dockerfile.ocim
+++ b/Dockerfile.ocim
@@ -1,0 +1,15 @@
+FROM python:3.6
+
+WORKDIR /usr/src/ocim
+
+COPY requirements.txt ./
+COPY requirements/ requirements/
+RUN pip install -r requirements.txt
+
+COPY Makefile \
+    package.json \
+    ./
+COPY frontend/ frontend/
+COPY static/ static/
+
+CMD ["make", "run.dev"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,58 @@
+# Docker Compose file for the services needed by Ocim, for easy local development setup
+
+version: "3.6"
+
+services:
+
+  postgresql:
+    image: postgres:12.6-alpine
+    environment:
+      POSTGRES_USER: opencraft
+      POSTGRES_DB: opencraft
+      POSTGRES_HOST_AUTH_METHOD: trust
+      POSTGRES_INITDB_ARGS: --encoding utf-8
+    volumes:
+      - /var/lib/postgresql/data
+    ports:
+      - "5432:5432"
+
+  redis:
+    image: postgres:12.6-alpine
+    image: redis:4
+    ports:
+      - "6379:6379"
+    
+  ocim:
+    build:
+      context: .
+      dockerfile: Dockerfile.ocim
+    depends_on: 
+      - postgresql
+      - redis
+    env_file: .env
+    environment: 
+      ALLOWED_HOSTS: '["*"]'
+      DATABASE_URL: '${DATABASE_URL:-postgres://opencraft@postgresql/opencraft}'
+      REDIS_URL: '${REDIS_URL:-redis://redis:6379}'
+    ports:
+      - "127.0.0.1:5000:5000"
+    volumes:
+      - ./:/usr/src/ocim/
+    
+  ocim-frontend:
+    build:
+      context: frontend/
+    depends_on: 
+      - ocim
+    environment: {} 
+    ports:
+      - "127.0.0.1:3000:3000"
+    tty: true
+    volumes:
+      - ./frontend:/usr/src/ocim-frontend/
+      - ./static:/usr/src/static
+
+volumes: {}
+
+networks:
+  default:
--- a/documentation/development/docker.md
+++ b/documentation/development/docker.md
@@ -1,0 +1,59 @@
+# Docker
+
+This describes how to create an environment for development using Docker.
+
+**This is still experimental, using [Vagrant](../installation.md) is much better tested and supported!**
+
+## Running OCIM in a docker container
+
+First, set your `.env` file as you would normally:
+```env
+DEBUG=True
+OPENSTACK_USER='username'
+OPENSTACK_PASSWORD='password'
+OPENSTACK_TENANT='tenant-name'
+OPENSTACK_AUTH_URL='https://auth.cloud.ovh.net/v2.0'
+OPENSTACK_REGION='BHS1'
+OPENSTACK_SANDBOX_SSH_KEYNAME='keypair-name'
+DEFAULT_INSTANCE_BASE_DOMAIN='example.com'
+GANDI_API_KEY='api-key'
+GITHUB_ACCESS_TOKEN='github-token'
+SECRET_KEY='tests'
+DEFAULT_INSTANCE_MYSQL_URL=...
+DEFAULT_RABBITMQ_API_URL=...
+DEFAULT_INSTANCE_RABBITMQ_URL=...
+DEFAULT_MONGO_REPLICA_SET_USER=...
+DEFAULT_MONGO_REPLICA_SET_PASSWORD=...
+DEFAULT_MONGO_REPLICA_SET_NAME=...
+DEFAULT_MONGO_REPLICA_SET_PRIMARY=...
+DEFAULT_MONGO_REPLICA_SET_HOSTS=...
+REDIS_URL=...
+```
+
+Start a one-off container for setup of the backend:
+```sh
+# --rm destroys the container at the end
+docker-compose run --rm ocim bash
+
+# then, inside the container...
+
+# initialize datastores
+make migrate
+# create super user
+make manage createsuperuser
+# exit the container
+exit
+```
+
+Then, start another one-off container for setting up the client-facing frontend:
+```sh
+docker-compose run --rm ocim-frontend bash -c 'npm run build-api-client && npm install'
+```
+
+Then, start OCIM:
+```sh
+docker-compose up
+```
+
+The OCIM UI should be available at http://localhost:5000 .
+The registration UI should be available at http://localhost:3000 .
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -1,7 +1,6 @@
-Installation
-------------
+# Installation
 
-### Vagrant installation
+## Vagrant installation
 
 For development, we recommend using [Vagrant](https://www.vagrantup.com/)
 to automatically provision a development environment in a virtual machine. This
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+frontend/node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,5 @@ media/
 # Generated scripts
 bin/archived_cleanup.sh
 
+# used for customizing docker compose
+docker-compose.override.yml
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ run: clean migrations.check static ## Run Ocim in a production setting with conc
 	honcho start --concurrency "worker=$(WORKERS),worker_low_priority=$(WORKERS_LOW_PRIORITY)"
 
 run.dev: clean migrations.check static_external ## Run the developmental server using `runserver_plus`.
-	honcho start -f Procfile.dev
+	python3 manage.py runserver 0.0.0.0:5000
 
 shell: ## Start the power shell.
 	HUEY_QUEUE_NAME=opencraft_low_priority $(HONCHO_MANAGE) shell_plus
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -1,0 +1,2 @@
+build/
+node_modules/
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:11.15.0
+
+# https://github.com/Automattic/node-canvas#compiling
+RUN apt-get update \
+    && apt-get install -y \
+        build-essential \
+        libcairo2-dev \
+        libpango1.0-dev \
+        libjpeg-dev \
+        libgif-dev \
+        librsvg2-dev
+
+WORKDIR /usr/src/ocim-frontend
+COPY ./ /usr/src/ocim-frontend
+
+# Move node_modules out of the source path
+RUN npm install -g --unsafe-perm
+
+CMD [ "npm", "start" ]
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,3 +19,5 @@ nav:
   - howtos.md
   - periodic_builds.md
   - infrastructure.md
+  - "Development":
+      - "Docker": development/docker.md
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --output-file=requirements/dev.txt requirements/dev.in
 #
--e git+https://github.com/open-craft/django-angular.git@v1-with-django2-support#egg=django_angular
+git+https://github.com/open-craft/django-angular.git@v1-with-django2-support#egg=django_angular
     # via -r requirements/base.in
 aioredis==1.3.1
     # via channels-redis


documentation/installation.md
```diff
@@ -1,7 +1,10 @@
-Installation
-------------
+# Installation
 
+<<<<<<< HEAD
```
Kelketek, line 3: Rebase artifacts in this file.

Kelketek:
@kewne Thank you for doing this. I made a quick pass over it and added some comments.

documentation/development/docker.md
```diff
@@ -0,0 +1,79 @@
+# Docker
+
+This describes how to create an environment for development using Docker.
+
+**This is still experimental, using [Vagrant](../installation.md) is much better tested and supported!**
```
bradenmacdonald, line 5: But if this works well for a few people with no issues, I'd like us to remove all code + documentation for the Vagrant version and standardize on this.

Dockerfile.ocim
```diff
@@ -0,0 +1,24 @@
+FROM python:3.6
+
+WORKDIR /usr/src/ocim
```
bradenmacdonald, line 3: This is totally fine, though I personally prefer shorter paths like `/ocim` for these kind of dev containers.

documentation/development/docker.md
```diff
@@ -0,0 +1,79 @@
...
+
+## 
+
+Add the following entries to your `.env` file:
+```dotenv
+ALLOWED_HOSTS='["*"]'
```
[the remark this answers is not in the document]
bradenmacdonald: +1, or set these as the defaults in the django settings, or find some mechanism so that the docker devstack works out of the box without manual config needed.

documentation/development/docker.md
```diff
@@ -0,0 +1,79 @@
...
+
+The OCIM UI should be available at http://localhost:5000 .
+
+## Local installation with docker-compose services
+
+This quick setup uses Docker Compose to run the database services, while the Ocim instance runs in a regular virtualenv.
+
+Install the prerequisite packages:
+
+    sudo apt install make git libmysqlclient-dev mysql-client postgresql-client python3-dev python3-pip subversion unzip xvfb
```
bradenmacdonald, line 58: I'd prefer to standardize on the "run ocim in docker" approach, and not have this other approach which requires installing a lot of packages onto the host. Is there a compelling advantage to this approach?

bradenmacdonald (commented):
Thanks for this! I haven't tested it, but I'm happy to see it :-)

Let me know if you need an additional reviewer/tester, and I can do that.

docker-compose.yml
```diff
@@ -0,0 +1,84 @@
...
+
+  mongo:
+    # This is similar to what the EdX devstack uses
+    # We use WiredTiger in all environments. In development environments we use small files
+    # to conserve disk space, and disable the journal for a minor performance gain.
+    # See https://docs.mongodb.com/v3.0/reference/program/mongod/#options for complete details.
+    command: mongod --smallfiles --nojournal --storageEngine wiredTiger
+    container_name: opencraft.mongo
+    hostname: mongo
+    image: mongo:4.0.22
```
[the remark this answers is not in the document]
kewne: I think these are the ones OCIM manages and assign OpenEdX instances to and it's part of the prior work done by @janimo.

I've used these instances for running tests locally but most other people probably use other instances running on OVH, so they're not mandatory.
If you agree, I'll leave this as is *for now* and create an issue for only starting them optionally (otherwise I'm afraid this will slip and kill my onboarding budget).

documentation/development/docker.md
kewne: Is it ok to push this to another story? That would:
1. allow more people to try this internally and provide feedback;
2. give use some time to improve the documentation structure a bit first: IMO it's conflating installation with the development setup.

documentation/development/docker.md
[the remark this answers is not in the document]
kewne: I've made some changes in this direction... it's not perfect but it starts OCIM with just
```sh
docker-compose run --rm ocim make migrate
docker-compose run --rm ocim make manage createsuperuser
docker-compose up
```

The main difficulty in getting this to work nicely is achieving a nice integration between `make` and `honcho` both inside and outside the container.

I considered going with an approach similar to the OpenEdX devstack but I don't really like the DX for it: you have container and non-container specific commands mixed together in the `Makefile` and then have much of the `docker-compose` CLI replicated in make targets (but with reduced extensibility).

If you can live with this for a bit, we can forward with this and I'll create specific issues for the parts of the workflow that either don't work or are clunky.

documentation/development/docker.md
kewne: 👍 I've deleted this section.

docker-compose.yml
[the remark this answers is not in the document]
Kelketek: @kewne If you can get it to a working state, then I'd suggest creating a follow-up ticket and placing it under the OCIM account instead, pinging @lgp171188 so it can be prioritized.

Containerization of OCIM's varying components and getting off of Vagrant is something that needs to be done anyway. The M1 is just the impetus for it now. So after you get it up and running in an acceptable fashion, a follow-up in-budget ticket that handles remaining polish would be OK.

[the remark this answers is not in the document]
kewne: @Kelketek @bradenmacdonald ok, let me know if there's anything else you'd like to see fixed right now; otherwise I'll just squash and rebase this unto master so it can be merged 🙂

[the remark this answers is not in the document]
bradenmacdonald: +1, I was going to say, this doesn't need to come out of your onboarding budget, this is work we've wanted for a long time anyways.

It's fine to leave these in for now though.

I guess what I don't understand is: if you want to test Ocim, you need a Mongo/MySQL/Rabbit MQ server _and_ you need to be able to deploy VMs that run Open edX. If you are running everything in the OVH cloud, that makes sense, but if you are running Mongo/MySQL/RabbitMQ locally in docker, how can they even be used, unless you also have some way of deploying Open edX VMs onto your localhost, or allowing VMs in the OVH cloud to access your local Mongo/MySQL/RabbitMQ in your laptop's docker environment?

documentation/development/docker.md
bradenmacdonald: Yes definitely.

documentation/development/docker.md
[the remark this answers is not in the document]
bradenmacdonald: Sounds good!

docker-compose.yml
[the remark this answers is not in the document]
Kelketek: @bradenmacdonald Provisioning servers locally hasn't ever worked well from what I know-- it's completely unsupported. If someone added these, I could think of a couple of reasons why:

1. Maybe they didn't understand what the MySQL servers were for yet when first getting the service to run on Docker
2. They did understand, but their goal wasn't to enable provisioning as much as to make sure that the code didn't crash when assuming these services would be present in passing.

I think there's some pretty good argument for 2 because usually one needs to 'provision' an instance just enough to get it to fail so that there's the relevant local DB entries all set up consistently aside from the fact the provisioning itself fails. That would probably involve creating the DB and setting the resulting created credentials in the instance settings.

Hard to say, though.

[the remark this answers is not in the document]
kewne: In my particular case, [the story I'm working on](https://tasks.opencraft.com/browse/BB-3729) only required me to create some unit tests which don't provision any VMs...
You totally have a point though that Postgres is the only real OCIM dependency and the rest are transitive dependencies of OpenEdX, so I've removed them.

Kelketek:
@kewne I'm getting this error when trying to run migrate:

```
❯ # --rm destroys the container at the end
docker-compose run --rm ocim bash
Creating network "opencraft_default" with the default driver
Creating opencraft_postgresql_1 ... done
Creating opencraft_ocim_run     ... done
root@ab8c60815b51:/usr/src/ocim# make migrate
coverage erase
find -name '*.pyc' -delete
find -name '*~' -delete
find -name '__pycache__' -type d -delete
rm -rf .coverage build
find static/external -type f -not -name 'Makefile' -not -name '.gitignore' -delete
find static/external -type d -empty -delete
honcho run python3 manage.py migrate
  File "manage.py", line 37, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/usr/local/lib/python3.6/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
  File "/usr/local/lib/python3.6/site-packages/django/apps/config.py", line 90, in create
    module = import_module(entry)
  File "/usr/local/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'djng'
make: *** [Makefile:102: migrate] Error 1
root@ab8c60815b51:/usr/src/ocim# exit
```

Also, looks like this needs a rebase, as master has changed.

kewne:
@Kelketek argh, the `django-angular` package gets installed into the `src` directory under the project root, so it gets overwritten by the mount...
That package is installed as editable though, is that really necessary? Removing the `-e` gets it working.

Kelketek:
@kewne Probably not. Chances are it was needed when first implementing the angular interface, but once it was no longer needed, no one bothered to remove the -e. I didn't even realize the package exists until it failed for me. So go ahead and try removing the -e.

We can look at re-adding the -e if we find a need for it and it can be on the person who needs it to make it work with Docker once we have that as a standard feature for OCIM. :)

bradenmacdonald:
@kewne I can review this, but probably not until Tuesday.

MartinKaburu (approved):
I followed the docs and managed to get myself set up on an m1 mac. Thanks.

Kelketek:
@kewne This is working for me now. I notice that the management backend is the UI on 5000-- I think 3000 is where the customer-facing UI is supposed to be-- are there special directions for how to start this up in the docker environment?

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+# Docker Compose file for the services needed by Ocim, for easy local development setup
+
+version: "3.6"
+
+services:
+
```
MartinKaburu, line 6: I also noticed that I needed to set up redis to be able to create new ovh instances. This requires you to set the REDIS_URL env var to "redis://redis:6379/" for this to reflect in my opencraft/settings.py
```suggestion
redis:
    image: redis:latest
    ports:
      - "6379:6379"
```

kewne: Yes, I threw the baby out with the bathwater on this one but just readded it and set it up as the default.
I force pushed though, so you might have to reset your local copy

kewne:
@Kelketek The frontend should be working as well; I added the instructions to the documentation but the workflow is a bit clunky because `node_modules` ends up being shared between the container and the host.

docker-compose.yml
MartinKaburu: We should probably also add redis to ocim's depends_on block

Kelketek:
@kewne Much closer now!

```
Failed to compile.
ocim-frontend_1  | 
ocim-frontend_1  | ./src/console/components/DomainItem/DomainItem.tsx
ocim-frontend_1  | Module not found: Can't resolve 'ocim-client' in '/usr/src/ocim-frontend/src/console/components/DomainItem'
```

I've forced a rebuild of the items and removed node_modules before restarting, but I keep getting this error when beginning from scratch.

Also it would be good to make this container set rootless. https://pythonspeed.com/articles/root-capabilities-docker-security/ since I found I had to use sudo to remove all of the node_modules files (they were owned by root).

There are some other potential solutions to the awkwardness of syncing node_modules-- many available if you Google around, since lots of people have this problem, but I don't think it needs to be in this PR. What you're doing is probably sufficient there.

kewne:
@Kelketek can you try `docker-compose run --rm ocim-frontend bash scripts/build-api-client.sh` **before** the `npm install` ? 

> Also it would be good to make this container set rootless.

I'd like to tackle that as part of (at least) one separate PR/story making containerized OCIM production ready;
my focus here was really to get something that works minimally for development or I'm afraid this'll never end 😓

documentation/development/docker.md
```diff
@@ -0,0 +1,59 @@
...
+
+# initialize datastores
+make migrate
+# create super user
+make manage createsuperuser
+# exit the container
+exit
+```
+
+Then, start another one-off container for setting up the registration backend:
```
Kelketek, line 48: Suggested change: Then, start another one-off container for setting up the client-facing frontend:

Also please add in the line you gave me for compiling the API module. It appears to have worked!

Kelketek:
@kewne Please create that follow-up ticket for the rootless container since I'm afraid it may get overlooked otherwise, and we do intend to containerize OCIM fully, so this work might end up the basis of something much larger.

There should be an OCIM docker epic somewhere you can attach the ticket to. Ping the current owner on it to make them aware of the creation.

I left one comment. Once addressed, this is :+1: from me. No need for me to fiddle with it again.

documentation/development/docker.md
```diff
@@ -0,0 +1,59 @@
...
+```sh
+docker-compose run --rm ocim-frontend npm install
```
bradenmacdonald, line 50: I also had to run

```
docker-compose run --rm ocim-frontend npm run build-api-client
```

... without that, when I ran `docker-compose up`, the frontend failed to build, with the error `Module not found: Can't resolve 'ocim-client' in '/usr/src/ocim-frontend/src/console/components/DomainItem'`

documentation/development/docker.md
```diff
@@ -0,0 +1,59 @@
+# Docker
+
+This describes how to create an environment for development using Docker.
+
+**This is still experimental, using [Vagrant](../installation.md) is much better tested and supported!**
+
+## Running OCIM in a docker container
+
+First, set your `.env` file as you would normally:
```
bradenmacdonald, line 9: Do we not have a working example of this .env file (for dev, not the test or e2e example), private or public somewhere? I thought we did. Doesn't need to block this PR though.

bradenmacdonald (approved):
:+1:  if you fix the frontend build issue I encountered.

- [x] I tested this: started it up locally
- [x] I read through the code
- [x] I checked for accessibility issues: n/a
- [x] Includes documentation: yes, thx

documentation/development/docker.md
kewne: I found [this](https://github.com/open-craft/opencraft/blob/f327ff0c1fd9f406f459ea790f1222b161f7c434/documentation/configuration.md#configuration) in the documentation.

documentation/development/docker.md
kewne: Done!

kewne:
@Kelketek @bradenmacdonald I've created https://tasks.opencraft.com/browse/BB-4246 for the follow up work that came up during the review.

Since I've addressed all of your comments, I'll merge this once it passes all checks; thanks for your patience!

Kelketek:
@kewne Looks like you'll need to do a commit --amend and force push to retrigger the tests. Also please check in the OCIM Mattermost channel to see if this failure is a known one that we're already planning on handling. I think this might be something we already have a ticket to fix and is unrelated to your changes, but I'm not sure.

kewne:
@Kelketek it's failing on master as well: https://app.circleci.com/pipelines/github/open-craft/opencraft/3742/workflows/75ea8813-4744-4c7c-b198-260d5bbc0f16/jobs/33686
I'm asking on Mattermost if this is a know issue that'll be solved soon; I don't mind waiting until then and will rebase when possible.

documentation/development/docker.md
bradenmacdonald: Ah, thanks :)

kewne:
@Kelketek @bradenmacdonald Seems like master is (finally) green again, so I'll be merging this in as soon as I get the flakes to pass.

bradenmacdonald:
@kewne Oh man, I thought this was already merged. Better late than never though. Thanks!!!

PR #796, opened 2021-05-14
feature: Added management command to find orphaned databases

kewne:
# Description

This adds a `find_orphan_dbs` management command that searches for databases that exist in a datastore but are not registered in the OCIM database.

I suspect this may happen because of some failure while deprovisioning databases (by either archiving or deleting the respective instances).

# Testing instructions

## Orphaned MySQL

1. Create a MySQL database (`create database DB_NAME`)
2. `./manage.py find_orphan_dbs mysql --host MYSQL_HOST -u MYSQL_USER -p` (will be prompted for password)
3. Verify that `DB_NAME` is identified as orphan in `stderr`
4. Verify that correct SQL for dropping the database is generated to `stdout`

## Orphaned MongoDB

1. Create a MongoDB database (`use DB_NAME; db.insert({})`)
2. `./manage.py find_orphan_dbs mongo --host MONGO_HOST -u MONGO_HOST -p` (will be prompted for password)
3. Verify that `DB_NAME` is identified as orphan in `stderr`
4. Verify that correct mongo-shell script for dropping the database is generated to `stdout`

### DIFF
--- a/instance/management/commands/find_orphan_dbs.py
+++ b/instance/management/commands/find_orphan_dbs.py
@@ -1,0 +1,208 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2019 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Management command to find orphaned databases
+"""
+import logging
+import traceback
+
+from django.core.management.base import BaseCommand
+import MySQLdb as mysql
+import pymongo
+
+from instance.models.database_server import MongoDBServer, MySQLServer
+from instance.models.openedx_instance import OpenEdXInstance
+
+LOG = logging.getLogger(__name__)
+
+MONGO_SYSTEM_DBS = {"admin", "local", "config"}
+MYSQL_SYSTEM_DBS = {"information_schema", "performance_schema", "sys", "mysql"}
+
+
+class Command(BaseCommand):
+    """
+    find_orphan_dbs management command class
+    """
+
+    help = 'Finds orphaned databases and generates commands to drop them.'
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            "--rm",
+            help="Drop any databases found",
+            action="store_true",
+        )
+
+        subparsers = parser.add_subparsers(
+            title='db type',
+        )
+        mongo_parser = subparsers.add_parser(
+            'mongo', help='Find orphaned MongoDB databases')
+        mongo_parser.add_argument(
+            '--exclude-dbs',
+            help=(
+                'MongoDB databases to ignore when checking for orphans.\n'
+                'System dbs are always excluded.'
+            ),
+            nargs='+',
+            type=str,
+            default=set()
+        )
+        mongo_parser.set_defaults(func=self.find_mongo_orphans)
+
+        mysql_parser = subparsers.add_parser(
+            'mysql', help='Find orphaned MySQL databases')
+        mysql_parser.add_argument(
+            '--exclude-dbs',
+            help=(
+                'MySQL databases to ignore when checking for orphans.\n'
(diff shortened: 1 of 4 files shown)


instance/management/commands/find_orphan_dbs.py
```diff
@@ -0,0 +1,196 @@
...
+                db_name = item[0]
+                if db_name in exclude_dbs:
+                    continue
+                try:
+                    MySQLServer.objects.get(name=db_name, hostname=host)
+                except MySQLServer.DoesNotExist as ex:
+                    self.stderr.write(f"Orphan database '{db_name}' found")
+                    count += 1
+                    self.stdout.write(f"DROP DATABASE {db_name};")
+            return count
```
0x29a, line 195: @kewne, this way to determine whether a database is orphan or not is incorrect. `MySQLServer` represents a server, where MySQL is hosted, and we have only four of them. You can even view them in admin panel.

In this context, a correct definition of `database` would be a database schema, which is created when the first app server for some instance is created.

Take a look at [this function](https://github.com/open-craft/opencraft/blob/80596664aa1de6a684a25ae296b135f04d057ba6/instance/models/mixins/database.py#L217-L248). We call it [when spawning an app server](https://github.com/open-craft/opencraft/blob/80596664aa1de6a684a25ae296b135f04d057ba6/instance/models/openedx_instance.py#L304). As a result, we have 11 new database schemas, with names like:
```
| xxx_opencraft_hosting_analytics_api                          |
| xxx_opencraft_hosting_dashboard                              |
| xxx_opencraft_hosting_discovery                              |
| xxx_opencraft_hosting_ecommerce                              |
| xxx_opencraft_hosting_edx_notes_api                          |
| xxx_opencraft_hosting_edxapp                                 |
| xxx_opencraft_hosting_edxapp_csmh                            |
| xxx_opencraft_hosting_notifier                               |
| xxx_opencraft_hosting_programs                               |
| xxx_opencraft_hosting_reports                                |
| xxx_opencraft_hosting_xqueue                                 |
```

I think you're right with the assumption about broken de-provisioning, though.

instance/management/commands/find_orphan_dbs.py
```diff
@@ -0,0 +1,196 @@
...
+            help="User to authenticate as",
+            type=str
+        )
+        mysql_parser.add_argument(
+            "-p",
+            "--prompt_password",
+            action='store_true',
+            help="Prompt for password",
+        )
+        mysql_parser.set_defaults(func=self.find_mysql_orphans)
```
0x29a, line 119: We [store](https://manage.opencraft.com/admin/instance/mysqlserver/7/change/) all this information in database as fields of `MySQLServer` model. Therefore, probably we could reduce this just to one argument, i.e. `mysqlserver_id`?

instance/management/commands/find_orphan_dbs.py
```diff
@@ -0,0 +1,196 @@
...
+            if db['name'] not in exclude_dbs
+        )
+
+        return self._find_mongo_orphans(host, non_system_db_names)
+    
+    def _find_mongo_orphans(self, host, db_names):
+        count = 0
+        for db_name in db_names:
+            try:
+                MongoDBServer.objects.get(name=db_name, hostname=host)
```
0x29a, line 154: Look similar comment about MySQL.

0x29a (changes requested):

instance/management/commands/find_orphan_dbs.py
kewne: I've considered that but it decided against it because it assumes all `MySQLServer` instances are registered in OCIM...which I now realize, from your other comment, is probably true...
In fact, we probably don't even need `mysqlserver_id` because we can just go through all of them using `MySQLSerevr.objects.get()`, so I'll remove all of these.

kewne:
@0x29a I've made the changes you suggested, can you take another look?

instance/management/commands/find_orphan_dbs.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def find_mongo_orphans(self, exclude_dbs: list, **kwargs) -> int:
+        """
+        Find orphaned databases in MongoDB.
+
+        This searches each MongoDB instance known to OCIM
+        and verifies every database is associated with an
+        OpenEdxInstance.
+        """
+        self.stderr.write("Finding orphaned MongoDB databases...")
```
0x29a, line 95: Should this use standard error descriptor? Why not standard output?

0x29a (changes requested):

instance/management/commands/find_orphan_dbs.py
kewne: Historic reasons 🤣 The original intent was to generate output that would be either a script or something that would help to automate the task of dropping the orphaned databases.
I ended up putting the drop feature directly into the task, so it doesn't make much of a difference; however, from a semantics POV, it makes more sense to put these kinds of messages in `stderr`. IIRC, this still puts the list of orphans in `stdout`.

0x29a (approved):
:+1:

- [x] I tested that this script finds orphaned databases on Ocim staging.
- [x] Includes documentation.

PR #797, opened 2021-05-17
[BB-4226] fix: handle the exceptions when deleting the obsolete appservers

lgp171188:
This will prevent the cleanup task from aborting due to any exceptions
and missing the cleanup of all the remaining instances.


nizarmah (approved):
👍🏼

nizarmah:
The integration tests are failing at the moment on stage and integration, and are being investigated at the moment. This fix is important, however. Therefore, I'm force merging this change.

PR #798, opened 2021-05-17
feature: Enable customization of footer url and image

kewne:
This adds the ability to customize the footer url and image which appears on the bottom right side of LMS pages (not to be confused with the organization url and images, found on the left side).

h1. Related Tickets

* [BB-4135](https://tasks.opencraft.com/browse/BB-4135)

h1. Testing instructions

1. Create a new instance through the registration client (I've already created https://stage.manage.opencraft.com/instance/7970/edx-appserver/3343/);
2. Change the "Footer Logo Image URL" and "Footer Logo Link URL" values in the "Logos" Theme configuration section (the preview should update automatically);
3. Publish the settings and wait until they finish applying;
4. Verify that the logo in the footer and the link changed correctly.


frontend/src/console/components/Logos/displayMessages.ts
```diff
@@ -27,6 +27,14 @@ const messages = {
       "If unset, OpenCraft's logo will be used.",
     description: 'Favicon explanation'
   },
+  footerLogoImageSrc: {
+    defaultMessage: 'Footer Logo Image URL',
+    description: "The URL of the image for the footer's logo"
```
0x29a, line 32: nit: single and double quotes mixing.

registration/tests/test_api.py
```diff
@@ -714,6 +714,28 @@ def test_update_without_changing_subdomain(self, mock_validate):
...
+
+        validation_response = self.client.patch(
+            reverse('api:v2:openedx-instance-config-detail', args=(self.instance_config.pk,)),
+            data=update_data,
+            format="json",
+        )
+
+        self.assertEqual(validation_response.status_code, 200)
+        self.assertDictContainsSubset(update_data, validation_response.json())
+        self.assertFalse(mock_validate.called)
```
0x29a, line 737: nit: do we need this assert statement?

instance/models/mixins/openedx_theme.py
```diff
@@ -83,10 +83,17 @@ def get_common_settings(self, application):
...
+        if application.footer_logo_image:
+            footer_logo_customization["FOOTER_OPENEDX_LOGO_IMAGE"] = application.footer_logo_image
+        if application.footer_logo_url:
+            footer_logo_customization["FOOTER_OPENEDX_URL"] = application.footer_logo_url
         return {
             "SIMPLETHEME_ENABLE_DEPLOY": True,
             "SIMPLETHEME_STATIC_FILES_URLS": self.get_simple_theme_static_files_settings(application),
             "EDXAPP_DEFAULT_SITE_THEME": "simple-theme",
+            "EDXAPP_LMS_ENV_EXTRA": footer_logo_customization,
+            "EDXAPP_CMS_ENV_EXTRA": footer_logo_customization
```
0x29a, line 96: I see a problem with this approach. If someone needs to override `EDXAPP_LMS_ENV_EXTRA`, footer logo customization will be lost. To avoid that, we'll have to implement some "merging" logic for this Ansible variable.

Instead, we should implement this as a new feature of https://github.com/open-craft/edx-simple-theme and generate SASS overrides. Check code of `get_simple_theme_static_files_settings` function and where it's used.

0x29a (changes requested):

frontend/src/console/components/Logos/displayMessages.ts
kewne: I did it to avoid having to escape quotes; it also seems to be the style used by other messages in this file but let me know if you still want it changed.

instance/models/mixins/openedx_theme.py
kewne: My understanding is that all of these variables [are merged recursively already](https://github.com/open-craft/opencraft/blob/master/instance/models/openedx_appserver.py#L452).

I'm not familiar with how theming works but [the OpenEdX docs](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/open-release-koa.master/configuration/changing_appearance/theming/create_theme.html#understanding-which-ui-files-to-customize) state:
> To replace an image, you can override any of the images in the static/images directories for the components you are theming.

However, `FOOTER_OPENEDX_LOGO_IMAGE` defaults to `https://files.edx.org/openedx-logos/open-edx-logo-tag.png`, so I think it can't be overridden by a theme like that; theoretically, we could maybe set this to a default that can be overridden by the theme and then handle it like the other images but I'd prefer to move that into a follow-up story.

0x29a: >My understanding is that all of these variables are merged recursively already.

Oh, then, I think it's fine.

>so I think it can't be overridden by a theme like that

My idea was to override src of `img` element in [this class](https://github.com/open-craft/edx-simple-theme/blob/3d7215a31b72b8b7c3891ec5f61ea874de073fe0/lms/static/sass/theme-overrides.scss#L358-L385
) with a new SASS variable. But, after all, that would be hacky, so I'm fine with the current approach.

However, it looks like `OpenEdXThemeMixin` is not the correct place for adding these variables, since this mixing is only for   generating variables for `simple_theme`. Could you find another place to add `EDXAPP_LMS_ENV_EXTRA` and `EDXAPP_CMS_ENV_EXTRA`?

kewne: I don't think there's a better one that exists already... Each of the configuration mixins is called in `OpenEdXInstance` to populate its respective field in `OpenEdXAppServer` and they all seem like worse options...
```python
CONFIGURATION_EXTRA_FIELDS = [
        'configuration_database_settings',
        'configuration_storage_settings',
        'configuration_theme_settings',
        'configuration_secret_keys',
        'configuration_marketing_links',
        # The extra settings should stay at the end of this list to allow manual overrides of all settings.
        'configuration_extra_settings',
    ]
```

I understand what you mean but, in this case, I don't think it's a huge issue.
I've moved these customizations to separate method and updated the docs to make it easier to understand; I hope this makes it for you.

0x29a (changes requested):
@kewne, thanks for addressing my comments.

I reviewed the code, and it looks fine to me. But, I think the page itself still requires a few enhancements.

image

1. Can we make all headings look similar in the sidebar? Probably, you can place gray text and trade policy link under "Logo Image" and remove "Footer Logo", and change "Logo Image" text to "Footer Logo"?
1. Also, could you add an info circle to the "Footer Logo Link URL" field? Right now, it's not obvious what this field is for.
1. I tried to upload 2Mb logo and see this:
    image
    Could you specify max size, if we have such limitation?
1. I noticed, that `onBlur` is being called every time I click outside of field, even if URL wasn't changed.

Also, I started deploying an app server on staging. I'll leave conditional approval after I test that logo setting works as expected.

kewne:
@0x29a I've tried to address your comments as well as I could.

> Can we make all headings look similar in the sidebar

Unfortunately, these are inside components and it would be hard to make them consistent now...
What I've done is allow the image uploader component to render children, and moved the trademark text and link inside there.

> could you add an info circle to the "Footer Logo Link URL" field

The `TextInputField` doesn't support this and replicating it seems involved, so I've instead used the `helpMessage` in that component, so that message appears below the input; it's not ideal but it's not too bad.

> I tried to upload 2Mb logo and see this:

Argh, this seems to be rejected by NGINX with a 413 Entity Too Large; I'm trying to find a way to handle this but I'm not sure the auto-generated openAPI client allows it.

> I noticed, that onBlur is being called every time I click outside of field, even if URL wasn't changed.

That's a limitation of the implementation: there's no state that I can use to compare the previous value with the new one; the way other components handle this is by copying props to state, which is [a big no-no](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state).

frontend/src/newConsole/components/newLogos/LogoSidebar.tsx
```diff
@@ -88,6 +143,12 @@ export class LogosSideBarComponent extends React.PureComponent<Props, State> {
                 innerPreview={favicon}
               />
             </div>
+            <div className="sidebar-item">
+              <div className="favicon-upload-field">
+                {footerLogoImageSourceInput}
+              </div>
+              <div>{footerLogoUrlInput}</div>
+            </div>
```
0x29a, line 151: Just a nitpick:
```suggestion
            <div className="favicon-upload-field sidebar-item">
              {footerLogoImageSourceInput}
              {footerLogoUrlInput}
            </div>
```

0x29a (approved):
:+1:

- [x] I tested new interface for changing footer logo, and verified that instance has different future logo after redeploying.
- [ ] I checked for accessibility issues.
- [x] Includes documentation

PR #800, opened 2021-05-17
[BB-4226] test: fixes test when terminating obsolete appservers

nizarmah:
Makes the same adjustment that was created in https://github.com/open-craft/opencraft/pull/797 to the unit test which was failing because the mock logger wasn't being called with the updated string


lgp171188 (approved):
@nizarmah, LGTM as this is a trivial change. Okay to merge after the CI checks (unit tests in particular) pass.

PR #801, opened 2021-05-20
Feature: Add instance purpose to instanceReference model

MartinKaburu:
Based on the findings of SE-4147, we need a more reliable way to know from among the following, what an instance's type is supposed to be:

proteacher: a self-service instance created by an end user via web UI
sandbox: an instance created automatically by Ocim in reaction to a PR or a periodic build
development: an instance created manually by an OpenCraft team member for development purposes
institutional: an instance created manually by an OpenCraft team member to serve a particular paying customer

This pull request adds the above instance definitions to our InstanceReference model.


instance/models/instance.py
```diff
@@ -52,14 +52,25 @@ class InstanceReference(TimeStampedModel):
...
     InstanceReference: Holds common fields and provides a list of all Instances
 
-    Has name, created, and modified fields for each Instance.
+    Has name, created, instance_type_id, and modified fields for each Instance.
 
     Instance is an abstract class, so having this common InstanceReference class gives us a
     fully generic way to iterate through all instances and allow instances to be implemented
     using a variety of different python classes and database tables.
     """
+    INSTITUTIONAL = "0"
```
MartinKaburu, line 61: I'm using properties for the types so that I can reference them when adding the default instance purpose with something like Instance(name='name', instance_purpose=InstanceReference.SANDBOX)

instance/models/instance.py
```diff
@@ -52,14 +52,25 @@ class InstanceReference(TimeStampedModel):
...
+    PROTEACHER = "1"
+    SANDBOX = "2"
+    DEVELOPMENT = "3"
+
     name = models.CharField(max_length=250, blank=False, default='Instance')
     instance_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    instance_purpose = models.CharField(max_length=20,
+                                        default=DEVELOPMENT,
```
MartinKaburu, line 69: Should I use a different default?

instance/models/instance.py
```diff
@@ -210,6 +224,11 @@ def modified(self):
         """ Get this instance's modified date, which is stored in the InstanceReference """
         return self.ref.modified
 
+    @property
+    def instance_purpose(self):
+        """ Get this instance's purpose, which is stored in the InstanceReference"""
+        return self.ref.get_instance_purpose_display()
```
MartinKaburu, line 230: The get_FOO_display method returns the human_readable value of the provided choice. [Docs](https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.get_FOO_display).

instance/models/instance.py
```diff
@@ -52,14 +52,25 @@ class InstanceReference(TimeStampedModel):
...
     using a variety of different python classes and database tables.
     """
+    INSTITUTIONAL = "0"
+    PROTEACHER = "1"
+    SANDBOX = "2"
+    DEVELOPMENT = "3"
+
     name = models.CharField(max_length=250, blank=False, default='Instance')
     instance_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    instance_purpose = models.CharField(max_length=20,
```
MartinKaburu, line 68: I wasn't sure what to name this variable, turns out we already have `instance_type` property. I'm fully open to renaming this to something more descriptive.

MartinKaburu (commented):
I have managed to update the models and some parts of the admin console. I'm now trying to understand under which circumstances we create the various types and try setting up default values based on how the instance was created.

MartinKaburu:
I also noticed that there is a lot of "linting noise" on the changes, please ignore those.

arbrandes (changes requested):
@MartinKaburu, I started to review this, but I found myself having to separate your work from what came from other branches - in addition to linting noise, which we want to avoid precisely so that reviewer's lives are easier.  With that in mind, can you please:

1) Rebase this PR on latest master, instead of merging/cherry-picking from it.
2) Drop the Docker PR and any other commits that are not pertaining to the work you're doing here.  You may have to maintain a separate local branch for you own use, but that shouldn't be the reviewer's concern (sorry!)
3) Drop any and all linting changes from this PR.  If you feel strongly about them, submit them in their own PR for separate evaluation.

Once you've done this, I'll come back and review your work.  Thanks!

clemente:
@MartinKaburu In addition, please expand the description and title based on the information from the [pull requests section](https://handbook.opencraft.com/en/latest/processes/#pull-request) of the handbook. The best is to follow a template, as mentioned there. You should include a section with the testing instructions for your reviewer. The title should mention the task ID. Sorry that that there are so many steps. PRs to edX have still more fine requirements, like [the commit text structure](https://github.com/edx/open-edx-proposals/pull/189). It's all mentioned in the first link.

CC @arbrandes

instance/factories.py
```diff
@@ -101,6 +101,7 @@ def instance_factory(**kwargs):
     assert is_valid_domain_name(kwargs.get('sub_domain'))
 
     # Create instance
+    # n = OpenEdXInstance.objects.create(instance_purpose="3")
```
MartinKaburu, line 104: @arbrandes I'm trying to make it possible to instantiate a new OpenEDX instance with a specific instance_purpose but somehow doing this coughs out a TypeError claiming `OpenEdXInstance() got an unexpected keyword argument 'instance_purpose'`. I have defined a property and setter in the Instance class and I expect this value to be available in the sub classes but thats not the case. Note, this only happens when I'm creating a new instance, I can get, view, and set the instance_purpose of an existing OpenEDXInstance but the instance_purpose property doesn't exist yet at creation because it's created when we create a new InstanceReference for this oedxinstance at the Instance.ref() property. Question: How do I propagate this value (instance_purpose) from the OpenEDXInstance, through the ref property in Instance to create an InstanceReference with the desired instance_purpose value at creation time?

PR #802, opened 2021-05-20
BB-4198: Fix issues with package upgrades.

giovannicimolin:
Fixes two issues:
1. From Openstack package upgrade: an exception class was removed in favor of a more generic one.
2. Bug found while testing: Terminate VM is not a background task, which makes the web worker halt if the VM takes too long to shutdown (moved to a huey task).

This is currently deployed to stage.

**Reviewers:**
- [ ] Any FFs: @clemente @symbolist @itsjeyd (high priority review, some Ocim methods are broken on `master` without this)

### DIFF
--- a/instance/api/openedx_appserver.py
+++ b/instance/api/openedx_appserver.py
@@ -37,7 +37,7 @@ from instance.serializers.openedx_appserver import (
     OpenEdXAppServerSerializer,
     SpawnAppServerSerializer,
 )
-from instance.tasks import make_appserver_active
+from instance.tasks import make_appserver_active, terminate_appserver
 from instance.utils import create_new_deployment
 from .filters import IsOrganizationOwnerFilterBackendAppServer, IsOrganizationOwnerFilterBackendInstance
 from ..models.deployment import DeploymentType
@@ -135,12 +135,15 @@ class OpenEdXAppServerViewSet(viewsets.ReadOnlyModelViewSet):
     @action(detail=True, methods=['post'])
     def terminate(self, request, pk):
         """
-        Terminate the VM running the provided AppServer.
+        Trigger worker task to terminate the VM running the provided AppServer.
+
+        Note that this might fail silently without any errors appearing on the Ocim UI,
+        but it's acceptable given the current use case.
         """
         app_server = self.get_object()
         if app_server.is_active:
             return Response({
                 'error': 'Cannot terminate an active app server.'
             }, status=status.HTTP_400_BAD_REQUEST)
-        app_server.terminate_vm()
+        terminate_appserver(app_server.id)
         return Response({'status': 'App server termination initiated.'})
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -364,9 +364,7 @@ class OpenStackServer(Server):
                 # server must have been terminated.
                 self.logger.debug('Server does not exist anymore: %s', self)
                 self._status_to_terminated()
-            except (requests.RequestException,
-                    novaclient.exceptions.ClientException,
-                    novaclient.exceptions.EndpointNotFound) as exc:
+            except (requests.RequestException, novaclient.exceptions.ClientException) as exc:
                 self.logger.debug('Could not reach the OpenStack API due to %s', exc)
                 if self.status != Status.Unknown:
                     self._status_to_unknown()
@@ -397,7 +395,7 @@ class OpenStackServer(Server):
                 key_name=key_name,
                 **kwargs
             )
-        except (novaclient.exceptions.ClientException, novaclient.exceptions.EndpointNotFound) as exc:
+        except novaclient.exceptions.ClientException as exc:
             self.logger.error('Failed to start server: %s', exc)
             self._status_to_build_failed()
         else:
@@ -456,9 +454,7 @@ class OpenStackServer(Server):
         except novaclient.exceptions.NotFound:
             self.logger.error('Error while attempting to terminate server: could not find OS server')
             self._status_to_terminated()
-        except (requests.RequestException,
-                novaclient.exceptions.ClientException,
-                novaclient.exceptions.EndpointNotFound) as exc:
+        except (requests.RequestException, novaclient.exceptions.ClientException) as exc:
             self.logger.error('Unable to reach the OpenStack API due to %s', exc)
             if self.status != Status.Unknown:
                 self._status_to_unknown()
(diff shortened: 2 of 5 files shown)


clemente:
@giovannicimolin  The code seems fine (except small test and logging missing), and the feature works. I can't approve yet since the tests don't pass. This may take some time and deployment hours and I'll be away, so could you check with the firefighter in the next time zone?

- [X] I read through the code
- [X] Includes documentation: Minor issue, please ignore: I see a conflict in [terminate](https://github.com/open-craft/opencraft/blob/giovanni%2Fdeprecated-exception/instance/api/openedx_appserver.py#L138) because it promises it will terminate the VM (`Terminate the VM running the provided AppServer.`) but it ends up confirming `App server termination initiated.`. That's not the same thing as terminating it. In particular: if there's an error in the termination process (now an asynchronous task) it won't be terminated. If we want to make the whole process asynchronous, ~~could we rename the docstring to say that it just initiates the termination?~~  On second though, we can let the documentation and API explain just the main case (terminate the VM), since it's what will happen most of the time. But if you want to differentiate between „terminating VM“ and „requesting VM termination“ you can update the docstrings/APIs
- [ ] currently there's no error control in the background task. Could you at least add logging [same as above](https://github.com/open-craft/opencraft/blob/giovanni%2Fdeprecated-exception/instance/tasks.py#L245)?
- even with logging, there's no guarantee that a server will be terminated when calling `.terminate()` because there could be an error. I think that the current situation is similar (the termination can fail and we can't do anything about it and the process will hang). We're not making the situation worse; we're also not fixing it (i.e. trying to terminate the server repeteadly in case of failure). This is an emergency PR so it's ok to leave further improvements for another PR, if needed.
- should there be a new test in `test_tasks.py` to test the new `terminate_appserver`. I'd say it's not very important since this function is just a wrapper for the real [terminate_vm](https://github.com/open-craft/opencraft/blob/giovanni%2Fdeprecated-exception/instance/models/appserver.py#L252), which is [already tested](https://github.com/open-craft/opencraft/blob/giovanni%2Fdeprecated-exception/instance/tests/models/test_openedx_appserver.py#L676), but since other simple tasks are have tests (e.g. [test_clean_up_task](https://github.com/open-craft/opencraft/blob/giovanni%2Fdeprecated-exception/instance/tests/test_tasks.py#L474)), and since `coverage` may complain, could you include a test?
- the code is indeed removing `novaclient.exceptions.EndpointNotFound`. If tests pass it's good enough
- I couldn't run the tests in my local Ocim due to other configuration errors. I'll fix them; meanwhile I checked CircleCI
- I still see `Can't upgrade a READER transaction to a WRITER mid-transaction` in the [integration tests](https://app.circleci.com/pipelines/github/open-craft/opencraft/3764/workflows/f73c5b5b-4f20-440e-9d57-7bbfc8fbee17/jobs/33911). I see that `keystoneauth1==3.18.0` was recently upgraded to `keystoneauth1==4.3.1`. If upgrading keystone to v4 breaks the API we could leave that upgrade for the next dependencies upgrade (BB-1694). The last keystone upgrade I remember (SE-708) also required rewriting API calls and exceptions and it wasn't trivial.
- there's also [an error](https://app.circleci.com/pipelines/github/open-craft/opencraft/3764/workflows/f73c5b5b-4f20-440e-9d57-7bbfc8fbee17/jobs/33894) in `test_password_changed_on_strong_password`, is this related? If this PR uncovered the next keystoneauth, see the previous suggestion about going to v3
- [X] I tested this: launched new VM, then clicked Terminate. I saw the new task being run:
```
20/May/2021 20:59:52 |     INFO | process=14929 | huey                      | Executing instance.tasks.terminate_appserver: 5f1cac0b-bccb-48f4-905d-2e88f1ae411c
20/May/2021 21:00:46 |     INFO | process=14929 | huey                      | instance.tasks.terminate_appserver: 5f1cac0b-bccb-48f4-905d-2e88f1ae411c executed in 53.534s
```
- (and it took 53 seconds, so it's good to have it in a worker). So the feature works fine, and we just need a test, the small documentation change, and making the tests pass in CircleCI

giovannicimolin:
@clemente 
>I still see Can't upgrade a READER transaction to a WRITER mid-transaction in the integration tests. I see that keystoneauth1==3.18.0 was recently upgraded to keystoneauth1==4.3.1. If upgrading keystone to v4 breaks the API we could leave that upgrade for the next dependencies upgrade (BB-1694). The last keystone upgrade I remember (SE-708) also required rewriting API calls and exceptions and it wasn't trivial.

This is a transient issue caused by OVH. I've tried reverting the version before to no avail. :(

>should there be a new test in test_tasks.py to test the new terminate_appserver. I'd say it's not very important since this function is just a wrapper for the real terminate_vm, which is already tested, but since other simple tasks are have tests (e.g. test_clean_up_task), and since coverage may complain, could you include a test?

Ack, working on it.

>there's also an error in test_password_changed_on_strong_password, is this related? If this PR uncovered the next keystoneauth, see the previous suggestion about going to v3

No, the e2e tests are flaky and a rerun cleared that error (check the latest run).

giovannicimolin:
@clemente Review comments addressed :)

clemente:
@giovannicimolin thanks; :+1: to the code (review comment above: I tested it, read it, and the new tests make sense). I didn't check all the test boilerplate in detail (consul patching etc.), just the general idea.

So when they pass in CircleCI, this can be merged. I see [the bug](https://bugs.launchpad.net/nova/+bug/1746509) you linked to (`Can't upgrade a READER transaction to a WRITER mid-transaction`), but do we have a task to fix that, and do we know how?
This is blocked on that fix, and that fix is urgent.
Of course if we get the chance to merge this one, we can. But fixing the flaky tests will be equally important.

giovannicimolin:
@clemente Pushed two commits reverting the OpenStack upgrade, hopefully, they fix the issue.

clemente:
@giovannicimolin  :+1: ok to merge this, due to:

> some Ocim methods are broken on master without this)

The CI tests are flaky due to https://tasks.opencraft.com/browse/SE-4492 but they have passed in a previous run. Better to merge it (force-merge if needed) so that we can fix production before fixing the CI.

You can deploy it to production.

giovannicimolin:
@clemente We got a lucky CI run after a few retries :)
I'm merging + deploying this now.

For reference, this was approved on comment: https://github.com/open-craft/opencraft/pull/802#issuecomment-849894136 just missing a formal GitHub approval.

PR #805, opened 2021-05-26
[BB-4306] Fix Can't upgrade a READER transaction to a WRITER mid-transaction issue

shimulch:
## Description

This PR adds a fix for `Can't upgrade a READER transaction to a WRITER mid-transaction` issue during integration testing. This issue happens due to random failure thrown by OVH OpenStack API.

The solution provided is to retry on such failures. For example, the error mentioned above throws a 400 bad request error. We can retry all these errors with `Retry` provided in the `urllib3` package that `requests` uses. 

But the error mentioned above is thrown from a `POST` request and by default `Retry` doesn't retry on `POST` requests, to avoid multiple insertions. For our use case, we needed to override that by setting `method_whitelist`. 

## Supporting information

- https://tasks.opencraft.com/browse/BB-4306

## Testing instructions

- Check there is no `Can't upgrade a READER transaction to a WRITER mid-transaction` error in integration test

## Deadline

ASAP

## Other information

**Notes**
- `method_whitelist` exists in [urllib3 v1.25.3](https://github.com/urllib3/urllib3/blob/1.25.3/src/urllib3/util/retry.py#L97) (the version we currently use). But the latest version deprecated it (to be removed in urllib3 v2), instead it has [allowed_methods](https://github.com/urllib3/urllib3/blob/1.26.5/src/urllib3/util/retry.py#L154).

**Reviewers**

- [x] @arjunsinghy96 
- [ ] @clemente

### DIFF
--- a/instance/utils.py
+++ b/instance/utils.py
@@ -44,6 +44,7 @@ from dictdiffer import diff
 if TYPE_CHECKING:
     from registration.models import BetaTestApplication  # pylint: disable=cyclic-import, useless-suppression
 
+
 # Logging #####################################################################
 
 logger = logging.getLogger(__name__)
@@ -81,7 +82,8 @@ def to_json(obj):
 
 
 def get_requests_retry(total=10, connect=10, read=10, redirect=10, backoff_factor=0,
-                       status_forcelist=range(500, 600)):
+                       status_forcelist=settings.OPENSTACK_RETRYABLE_STATUS_CODES,
+                       method_whitelist=settings.OPENSTACK_RETRYABLE_METHOD_WHITELIST):
     """
     Returns a urllib3 `Retry` object, with the default requests retry policy
     """
@@ -92,6 +94,7 @@ def get_requests_retry(total=10, connect=10, read=10, redirect=10, backoff_facto
         redirect=redirect,
         backoff_factor=backoff_factor,
         status_forcelist=status_forcelist,
+        method_whitelist=method_whitelist,
     )
 
 
--- a/opencraft/settings.py
+++ b/opencraft/settings.py
@@ -331,6 +331,20 @@ OPENSTACK_PRODUCTION_INSTANCE_FLAVOR = env.json(
     default={"ram": 8192, "disk": 80}
 )
 
+# Openstack fails to connect randomly, thus we retry on certain status codes.
+# for example Openstack throws 400, and it will work if we keep retrying,
+# more on https://github.com/open-craft/opencraft/pull/805
+OPENSTACK_RETRYABLE_STATUS_CODES = env.list(
+    'OPENSTACK_RETRYABLE_STATUS_CODES',
+    default=list(range(500, 600))
+)
+
+# Set of uppercased HTTP method verbs that we should retry on.
+OPENSTACK_RETRYABLE_METHOD_WHITELIST = env.list(
+    'OPENSTACK_RETRYABLE_METHOD_WHITELIST',
+    default=frozenset(["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"])
+)
+
 # Separate credentials for Swift.  These credentials are currently passed on to each instance
 # when Swift is enabled.
 


instance/utils.py
```diff
@@ -92,6 +95,7 @@ def get_requests_retry(total=10, connect=10, read=10, redirect=10, backoff_facto
         redirect=redirect,
         backoff_factor=backoff_factor,
         status_forcelist=status_forcelist,
+        method_whitelist=method_whitelist,
```
arjunsinghy96, line 98: @shimulch Can you please add the note in PR description (deprecation of `method_whitelist` in favour of `allowed_methods` introduced in v`1.26.0`) here for documentation purposes?

arjunsinghy96 (commented):
@shimulch LGTM 👍🏽 . Just a minor request for adding doc.

instance/utils.py
shimulch: @arjunsinghy96 I've already added that in the `Notes` sections. :)

instance/utils.py
```diff
@@ -44,6 +44,9 @@
 if TYPE_CHECKING:
     from registration.models import BetaTestApplication  # pylint: disable=cyclic-import, useless-suppression
 
+RETRYABLE_STATUS_CODES = frozenset([400] + list(range(500, 600)))
```
clemente, line 47: - This is limited to OpenStack. Could the variables start with `OPENSTACK_` same as other variables?
- These values are useful for OpenCraft's case, but they aren't good defaults:
  - Retrying on a 400 error seems useless in general (400: bad request. If the request is bad, retrying won't help. Unfortunately in OVH's case, it does)
  - Retrying on a POST can be dangerous, and the original default (retrying only on idempotent methods) is saner
  - So it makes more sense if these are set through https://github.com/open-craft/ansible-secrets/ instead (or ansible-playbooks), so that the defaults are sane and the customizations apply to OpenCraft only. The distinction is mainly to keep the code clean and to separate the different parts. In practical terms it doesn't matter much, because I think we're the only ones using Ocim. Ocim is already heavily specific to OpenCraft but we still do. So @arjunsinghy96 you can also mention other solutions as reviewer
- Wherever you add the variables, could you add a comment nearby explaining the reasoning? (why we retry all methods and even 400 errors), since it's not obvious
- Thinking about whether retrying any other POST call will have unintended consequences…: I guess it won't: if N successive calls to do an operation fail and the N+1 th call works, that's still better than failing (failing was the previous behaviour)

clemente (changes requested):

instance/utils.py
shimulch: @clemente, Thanks for your thoughtful review.

> This is limited to OpenStack. Could the variables start with OPENSTACK_ same as other variables?

Done.

> These values are useful for OpenCraft's case, but they aren't good defaults

I've now moved these variables to Django settings to pick up values from env. And by default, it sets sane values for retry.

> So it makes more sense if these are set through https://github.com/open-craft/ansible-secrets/ instead (or ansible-playbooks)

Do we want to retry on `POST` during production? For now, I've set these environment variables in CircleCI so that integration tests retry on `POST`. But do we want the same behavior on production as well? If no, we don't need to set anything in `ansible-secrets`.

> Thinking about whether retrying any other POST call will have unintended consequences…: I guess it won't: if N successive calls to do an operation fail and the N+1 th call works, that's still better than failing (failing was the previous behaviour)

This is why I am a bit suspicious to set retry on `POST` in production.

clemente: @shimulch 

> Do we want to retry on POST during production? For now, I've set these environment variables in CircleCI so that integration tests retry on POST. But do we want the same behavior on production as well? If no, we don't need to set anything in ansible-secrets.

Good idea. No change is required in production. I tried the API calls hundreds of times in Ocim stage and production and they weren't flaky, though the same calls were flaky from outside OVH infrastructure. Probably due to network latency. So I think that it's better to do the non-standard 400 retry just in integration tests, and leave production as it is, just in case.

If you change the setting in CircleCI, please write down the new setting in vault (there's a section).

Then we don't need the 400 in this PR, right? And what about the other methods? (PUT, …). The defaults in this PR should be the same as the defaults before this PR (i.e. no behaviour change); only CircleCI will switch on the special 400 (+POST?) retry.

shimulch: @clemente, I've removed 400 from `OPENSTACK_RETRYABLE_STATUS_CODES`. I've set `OPENSTACK_RETRYABLE_STATUS_CODES` and `OPENSTACK_RETRYABLE_METHOD_WHITELIST` in both CircleCI environment and vault (`OpenCraft IM - Integration test settings`).

`PUT` is a default allowed method from [urllib3](https://github.com/urllib3/urllib3/blob/1.26.5/src/urllib3/util/retry.py#L212). We are only adding `POST` to the list during integration testing.

opencraft/settings.py
```diff
@@ -331,6 +331,20 @@
...
 
+# Openstack fails to connect randomly, thus we retry on certain status codes.
+# for example Openstack throws 400, and it will work if we keep retrying,
+# more on https://github.com/open-craft/opencraft/pull/805
+OPENSTACK_RETRYABLE_STATUS_CODES = env.list(
+    'OPENSTACK_RETRYABLE_STATUS_CODES',
+    default=list(range(500, 600))
+)
+
+# List of HTTP methods that we will retry
```
arjunsinghy96, line 342: @shimulch Could you update this with `Set of uppercased HTTP method verbs that we should retry on.` 

This is taken from the `urllib3` documentation. They use [uppercase HTTP methods](https://github.com/urllib3/urllib3/blob/1.26.5/src/urllib3/util/retry.py#L446) for comparison and I think it would be good to mention that information here.

arjunsinghy96 (changes requested):

clemente:
Thanks, @shimulch, :+1: to merging it after @arjunsinghy96's small change.
Tests run, the new setting isn't intrusive (no behaviour change in production/stage), there's some documentation with a link here. 

The main point of the variable is to fix an issue specific to our OpenStack provider (OVH); other OpenStack providers may not need to redefine the new setting to include the 400. The comments could mention that, but it's very specific. With a link to this PR/task it's good enough.

opencraft/settings.py
shimulch: @arjunsinghy96, Thanks for the suggestion. I've addressed your review comment.

arjunsinghy96 (approved):

arjunsinghy96 (approved):
:+1: @shimulch LGTM

- [x] I tested that the integration tests are passing.
- [x] I read through the code
-  N/A ~~I checked for accessibility issues~~
- [x] Includes documentation
- N/A ~~I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

PR #806, opened 2021-05-28
[BB-4230] feature: Support usage of `EDXAPP_FEATURES_EXTRA`.

kewne:
The user is now allowed to set feature flags through OCIM as he would when using the `configuration` playbooks.

Associated tickets:
    * [BB-4230](https://tasks.opencraft.com/browse/BB-4230)

h1. Testing

1. Launch vanilla instance from this branch
2. Launch vanilla instance from [current `master` branch](https://github.com/open-craft/opencraft/tree/f4d442b97ee32fdadda8325b85aa873365fea225)
3. Verify that feature configurations are identical


kewne:
@viadanna I'm not sure how this can be tested, as I'm not familiar with what most feature flags do...

viadanna (approved):
Good to go 👍

- [x] I tested this on my devstack and instance deployed to Stage.

instance/migrations/0141_features_to_features_extras.py
```diff
@@ -0,0 +1,53 @@
...
+            config.configuration_extra_settings = yaml.safe_dump(settings, default_flow_style=False)
+            config.save()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('instance', '0140_update_openstack_base_image_20210315_1545'),
+    ]
+
+    operations = [
```
giovannicimolin, line 52: FYI @kewne @viadanna There's no operations here, so the forward/backward code doesn't run?

instance/migrations/0141_features_to_features_extras.py
```diff
@@ -0,0 +1,53 @@
...
+        if "EDXAPP_FEATURES" in settings and "EDXAPP_FEATURES_EXTRA" not in settings:
+            settings["EDXAPP_FEATURES_EXTRA"] = settings.pop("EDXAPP_FEATURES")
+            settings["EDXAPP_FEATURES_MIGRATED"] = (
+                "EDXAPP_FEATURES was renamed to EDXAPP_FEATURES_EXTRA to preserve previous behavior."
+            )
+            config.configuration_extra_settings = yaml.safe_dump(settings, default_flow_style=False)
+            config.save()
+        
+
+def bacward(apps, schema_editor):
```
giovannicimolin, line 26: Nit: 
```suggestion
def backward(apps, schema_editor):
```

instance/migrations/0141_features_to_features_extras.py
kewne: Good catch! I readded this after deleting to try another approach and forgot to add the `RunPython`.
It's fixed now.

PR #809, opened 2021-06-03
FAL-1937 Disable learner records by default

swalladge:
[FAL-1937](https://tasks.opencraft.com/browse/FAL-1937)

OCIM doesn't set up the credentials service by default,
and learner records depends on the credentials service.
So, we need to also disable learner records by default.

**Sandbox**:

- LMS: https://fal-1937.stage.opencraft.hosting/
- Studio: https://studio.fal-1937.stage.opencraft.hosting/
- ocim: https://stage.manage.opencraft.com/instance/8070/

**Test instructions**:

- launch a new instance with default config
- verify that the 'view my records' button is hidden on your profile page after logging in to the instance

**Reviewers**:

- [x] @mtyaka

### DIFF
--- a/instance/models/mixins/openedx_site_configuration.py
+++ b/instance/models/mixins/openedx_site_configuration.py
@@ -39,6 +39,8 @@ class OpenEdXSiteConfigurationMixin(models.Model):
         site_configuration_settings = {
             # default site configuration
             'CONTACT_US_CUSTOM_LINK': '/contact',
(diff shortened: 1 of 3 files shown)


mtyaka:
Thanks @swalladge! It looks like you'll have to update one of the tests, but after that this is good to merge :+1:

- [x] I tested this: I verified that the "View My Records" button not visible on the sandbox.
- [x] I read through the code
- [x] Includes documentation (inline code comments)

PR #814, opened 2021-06-15
[BB-4015] fix: add the missing space between the command flags

lgp171188:
This will fix the syntax error when the 'instance_statistics_csv'
management command is invoked from the periodic tasks.

**Testing instructions**:
* Checkout the source branch of this PR in the production environment or on stage after setting the `INSTANCE_LOGS_SERVER_HOST` environment variable in the `.env` file.
* Run the `instance_statistics_csv` management command with one or more instance domains and the `--start-date`, `--end-date` parameters specified.
* Verify that the `collect_activity.py` script runs on the appservers of the given instances without any errors and returns proper values for the fields like `Total users`, `Total active users`, `Total courses`, etc.

### DIFF
--- a/instance/management/commands/instance_statistics_csv.py
+++ b/instance/management/commands/instance_statistics_csv.py
@@ -256,8 +256,9 @@ class Command(BaseCommand):
         log_line.info = log_line
         log_line.error = log_line
 
-        playbook_extra_script_arguments = '--skip-hit-statistics' if settings.INSTANCE_LOGS_SERVER_HOST else ''
-        playbook_extra_script_arguments += f'--start-date {start_date} --end-date {end_date}'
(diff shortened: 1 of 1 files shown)


shimulch (approved):
@lgp171188 Thanks for the fix. LGTM :+1:

- ✅  I tested this on stage. I've set `INSTANCE_LOGS_SERVER_HOST` in Django settings and copied the changes there. Then I've run the `instance_statistics_csv` command with start and end date. It worked and `collect_activity.py` ran correctly.
- ✅  I read through the code
- ❌ I checked for accessibility issues
- ❌ Includes documentation
- ❌ I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #815, opened 2021-06-15
[BB-4378] Fix CI cleanup issues

lgp171188:
This PR fixes the issues that were preventing the cleanup of some unused CI resources like MySQL databases, Gandi DNS records, etc.

It also changes the CircleCI workflow definitions to allow pushing to the `ci-cleanup` branch for running cleanup on-demand (similar to how `stage` is used for the frontend).

# Testing
1. Push this branch to `ci-cleanup`
2. Verify the output of the `cleanup` job

# Links

* [BB-4378 (OpenCraft Private JIRA)](https://tasks.opencraft.com/browse/BB-4378)

### DIFF
--- a/documentation/development/ci.md
+++ b/documentation/development/ci.md
@@ -1,0 +1,15 @@
+# Continuous Integration
+
+This describes aspects of Continuous Integration setup specifically for OpenCraft,
+and is of limited interest for external contributors.
+
+## Cleaning up left-over resources
+
+Test runs (especially failing ones) may occasionally leave behind resources that,
+over time, clutter the underlying infrastructure.
+
+To avoid this, there is [a `scheduled-cleanup` CircleCI workflow](https://github.com/open-craft/opencraft/blob/master/circle.yml)
+that regularly runs the `cleanup` job.
+
+If necessary (usually when testing changes to it), the job may be run on-demand
+by pushing to the `ci-cleanup` branch.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -23,5 +23,6 @@ nav:
   - "Using the Marketing app": marketing_usage.md
   - "Development":
       - "Docker": development/docker.md
+      - "Continuous Integration": development/ci.md
   - "Operations":
       - "Management Tasks": operations/management-tasks.md
--- a/circle.yml
+++ b/circle.yml
@@ -490,7 +490,9 @@ workflows:
               ignore: /.*/
             tags:
               only: /^release\-\w+\-\w+$/
-  cleanup-workflow:
+  
+  # used for periodically cleaning up left-over resources
+  scheduled-cleanup:
     triggers:
       - schedule:
           cron: "0 1 * * *"
@@ -500,3 +502,11 @@ workflows:
                 - master
     jobs:
       - cleanup
+
+  on-demand-cleanup:
+    jobs:
+      - cleanup:
+          filters:
+            branches:
+              only:
+              - ci-cleanup
--- a/cleanup_utils/aws_cleanup.py
+++ b/cleanup_utils/aws_cleanup.py
@@ -34,7 +34,7 @@ DEFAULT_POLICY_NAME = "allow_access_s3_bucket"
 
 # Logging #####################################################################
 
-logger = logging.getLogger('integration_cleanup')
+logger = logging.getLogger(__name__)
 
 
 # Classes #####################################################################
@@ -186,7 +186,7 @@ class AwsCleanupInstance:
         """
         Runs the cleanup of the AWS buckets, IAM users and their policies
         """
-        logger.info("\n --- Starting AWS Cleanup ---")
+        logger.info("--- Starting AWS Cleanup ---")
         if self.dry_run:
             logger.info("Running in DRY_RUN mode, no actions will be taken.")
 
--- a/cleanup_utils/dns_cleanup.py
+++ b/cleanup_utils/dns_cleanup.py
@@ -28,7 +28,7 @@ from instance.gandi import GandiV5API
 
 # Logging #####################################################################
 
-logger = logging.getLogger('integration_cleanup')
+logger = logging.getLogger(__name__)
 
 
 # Constants ###################################################################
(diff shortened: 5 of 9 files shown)


shimulch (approved):
@kewne, @lgp171188 Thanks for the fix. LGTM :+1:

- [x] I tested this: Checked output on CircleCI `cleanup` job. Tried to get a list of empty tables using the query provided.
- [x] I read through the code
- [x] Includes documentation

kewne:
@shimulch It seems that making the `cleanup` step manual has the nasty side-effect of blocking the rerunning of failed jobs... in this specific case, the integration tests group 2 failed but I couldn't select the "rerun from failed" option in circle ci until I had launched the `cleanup` job as well.

This isn't a huge deal but maybe it warrants a follow-up?

lgp171188:
@kewne, I would recommend against changing anything that tweaks how the cleanup job is currently run periodically and separately, outside the typical workflow for the PR CI checks. There is already a way to run the cleanup job on demand - go to CircleCI and re-run the last run of the job. It is expected the automatic runs should do the job and manual run is only needed to debug and fix issues, if any.

This cleanup job has had multiple rewrites/fixes to solve each issue that showed up after some time passed after the previous fix/rewrite. Let's not introduce a lot of changes that tend to increase the surface area for the bugs.

CC @shimulch

documentation/development/cd.md
```diff
@@ -0,0 +1,15 @@
+# Continuous Delivery
+
+This describes aspects of Continuous Delivery setup specifically for OpenCraft,
```
lgp171188, line 3: @kewne, I don't think we are doing `Continuous Delivery` with our current setup for Ocim and the cleanup job has nothing to do with it either. Imho, it is a part of `Continuous Integration`

documentation/development/cd.md
```diff
@@ -0,0 +1,15 @@
...
+and is of limited interested for external contributors.
```
lgp171188, line 4:
```suggestion
and is of limited interest for external contributors.
```

documentation/development/cd.md
```diff
@@ -0,0 +1,15 @@
...
+## Cleaning up left-over resources
+
+Test runs (especially failing ones) may ocasionally leave behind resources that,
+over time, clutter the underlying infrastructure.
+
+To avoid this, there is [a `scheduled-cleanup` CircleCI workflow](https://github.com/open-craft/opencraft/blob/master/circle.yml)
+that regularly runs the `cleanup` job.
+
+If necessary (usually when testing changes to it), the job may be run on-demand
+by manually triggering it in CircleCI.
```
lgp171188, line 15: @kewne, like I mentioned [here](https://github.com/open-craft/opencraft/pull/815#issuecomment-873970787), CircleCI allows re-running jobs without having to write any code/configuration.

mkdocs.yml
```diff
@@ -21,5 +21,6 @@ nav:
   - infrastructure.md
   - "Development":
       - "Docker": development/docker.md
+      - "Continuous Delivery": development/cd.md
```
lgp171188, line 24: @kewne, see [this comment](https://github.com/open-craft/opencraft/pull/815/files#r663796037)

kewne:
@lgp171188
> This is a huge deal because the PR CI runs are blocked by something they shouldn't be blocked by

To clarify, CI runs aren't blocked by the `cleanup` job; what happens is that, **if a job fails in a workflow with another job pending approval**, CircleCI doesn't enable the "re-run from ..." options.

Anyway, I've moved the approval job to a separate workflow, where it shouldn't interfere anymore.

> It is expected the automatic runs should do the job and manual run is only needed to debug and fix issues, if any.

That is the intent of the change: your original PR changed the scheduled job to run on this branch in addition to `master`; that is now unnecessary: if you want to run it on a branch, you can simply use the link in the PR and trigger it right away.
The original schedule that runs from `master` hasn't been changed at all.

kewne:
@shimulch it seems that the pending manual job blocks the github checks; IIRC you can make that specific check optional in the repo settings, which I don't have permissions to do.

shimulch:
@kewne [Looks like](https://discuss.circleci.com/t/workflows-with-optional-approval-jobs-dont-appear-as-successful-in-github-or-circleci/31292) we can't have an optional job that requires approval. So let's just keep the old way of having `ci-cleanup` branch for manual cleanup. And the scheduled job is okay I guess. 

Also, can you temporarily enable scheduled cleanup in the PR branch as well? So that we can test if it runs or not? Maybe just remove `only` section from the scheduled task for now. After PR review we can add that again. :-)

kewne:
@shimulch I meant making the check optional in Github; this is an example of the setting in one of my personal repos:
github_status

In this case, you should be able to remove the `on-demand-cleanup/hold-cleanup` check, to make everything behave as before.

Let me know if you feel like this is the wrong thing to do and I'll revert my change.

shimulch:
@kewne Currently the only required status is the coverage. So the merge is not blocked by that task anyway.

Screenshot 2021-07-06 at 4 49 16 PM

> Let me know if you feel like this is the wrong thing to do and I'll revert my change.

I think we should stay with the "push to the `cleanup-ci` branch" idea instead.

kewne:
@shimulch The cleanup only runs on the `ci-cleanup` branch now. Sorry for all the back and forth on this.

Ironically, I think we've stumbled on a Circle CI bug: the `cleanup` job from the `ci-cleanup` branch appears here when it shouldn't, and some of the other checks also link to jobs on the `ci-cleanup` branch 🤷

shimulch:
@kewne, No worries, we weren't expecting these CI issues. 

I am not sure either why this job is showing up there. Can you try filtering on the workflow instead? `branches` filter in jobs seems [depricated](https://circleci.com/docs/2.0/configuration-reference/#branches-deprecated). I don't know if it will work. Just throwing rocks in dark. Also, check how pushing only on stage branch deploys frontend. 😄

lgp171188:
@kewne, @shimulch, can we drop the changes to how the `cleanup` job is run and revert it to how it was before? We have a way to re-run that job manually and that will avoid all the unnecessary complications that we have run into and are trying to resolve now.

kewne:
@lgp171188 what do you mean by "re-run the job manually"? If I understood correctly the previous configuration only runs the `cleanup` job on a schedule, so the most you can do is search for a previous run and rerun it, and it'll only do so for the `master` branch.
If someone makes some changes to the cleanup job (as we have done here), I don't see how they can do it without changing the Circle CI configuration.

@shimulch The "issue" is that the Github API for setting statuses uses SHA to identify commits: I've tried pushing "my own" status to the commit here, which resulted in this:
image

What's happening here, then, is that I'm pushing the same commit to CircleCI on both this branch and the `ci-cleanup` branch, which results in the status being set twice, but is otherwise harmless.

I think this can be approved now: if you prefer to take @lgp171188's suggestion just put it in the approval and I'll revert my change before merging.

shimulch:
@lgp171188, The current change is fine. We now have the option to run clean up jobs by -

- Scheduled
- On-demand by pushing to the `ci-cleanup` branch

The integration failure CI job is failing for a known unrelated issue.

@kewne can you remove your own status and rerun CI checks in this branch?

documentation/development/ci.md
```diff
@@ -0,0 +1,15 @@
+# Continuous Integration
+
+This describes aspects of Continuous Integration setup specifically for OpenCraft,
+and is of limited interest for external contributors.
+
+## Cleaning up left-over resources
+
+Test runs (especially failing ones) may ocasionally leave behind resources that,
```
shimulch, line 8:
```suggestion
Test runs (especially failing ones) may occasionally leave behind resources that,
```

shimulch (approved):
@kewne, @lgp171188 Thanks a lot for working on improving this. LGTM :+1:

- [x] I tested this: checked the output of Circle CI job
- [x] I read through the code
- [x] I checked for accessibility issues: N/A
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: N/A

@kewne can you re-run the CI and fix the small nit I've mentioned? :) Then this is good to merge.

lgp171188:
@kewne, it looks like there are some issues in the changes made in this PR that are causing the `cleanup` job to fail daily ([example](https://app.circleci.com/pipelines/github/open-craft/opencraft/4114/workflows/5408e40b-9d61-4ca9-b910-b5a09cfedb2e/jobs/37022)). It looks like the base domain for DNS records is not determined correctly and hence the DNS cleanup API calls are failing. Can you take a look and fix it?

CC @shimulch

kewne:
@lgp171188 thanks for noticing! I don't think I have access to properly debug this though, I'd either need to run it locally but don't have Gandi credentials, and to run it in the pipeline I'd have to break the domain name masking.

Any idea what I can do here?

lgp171188:
@kewne, you can re-run the failing job with SSH and that will allow you to log in to the container with your GitHub SSH key(s) and debug the issue.

kewne:
This issue will be fixed in #826

PR #817, opened 2021-06-22
fix: Fix migration when configuration_extra_settings is all whitespace

kewne:
Related tickets:
* [BB-4230](https://tasks.opencraft.com/browse/BB-4230)

# Testing

1. Create a OpenEdXInstance where `configuration_extra_settings = '   '` (or any other string with only whitespace);
2. Apply migration `manage.py migrate instance`
3. Rollback migration `manage.py migrate instance 0140`

This has already been done with [this staging instance](https://stage.manage.opencraft.com/instance/8150/)


viadanna (approved):
Good to go 👍

- [x] I tested this on my local devstack and saw it applied to Stage.

PR #818, opened 2021-06-24
fix: Server archival no longer fails when delete SSH keys

kewne:
This fixes the case where instance archival fails because the VM was already deleted from OpenStack.
Although the code in question already seemed to handle this, deleting the SSH key "covertly" queries the OpenStack API for IP addresses, which fails.

Related tickets:
* [BB-4245](https://tasks.opencraft.com/browse/BB-4245)

# Testing

This includes a new test that verifies the behavior:
```sh
./manage.py test --parallel 4 instance.tests.models.test_server.OpenStackServerTestCase.test_ignores_failure_to_delete_ssh_key
```

### DIFF
--- a/instance/models/server.py
+++ b/instance/models/server.py
@@ -23,7 +23,6 @@ Instance app models - Server
 # Imports #####################################################################
 
 import logging
-import subprocess
 import time
 
 from django.conf import settings
@@ -33,7 +32,7 @@ from django_extensions.db.models import TimeStampedModel
 import novaclient
 import requests
 
-from instance import openstack_utils
+from instance import openstack_utils, ssh
 from instance.logging import ModelLoggerAdapter
 from instance.models.utils import (
     ValidateModelMixin, ResourceState, ModelResourceStateDescriptor, SteadyStateException, default_setting
@@ -452,7 +451,11 @@ class OpenStackServer(Server):
                 self.logger.warning("Server has not reached SHUTOFF state after max wait time; terminating forcefully.")
             os_server.delete()
         except novaclient.exceptions.NotFound:
-            self.logger.error('Error while attempting to terminate server: could not find OS server')
+            self.logger.exception(
+                'Error while attempting to terminate the server "%s":'
+                + ' could not find the corresponding OpenStack server',
+                self.name
+            )
             self._status_to_terminated()
         except (requests.RequestException, novaclient.exceptions.ClientException) as exc:
             self.logger.error('Unable to reach the OpenStack API due to %s', exc)
@@ -486,9 +489,15 @@ class OpenStackServer(Server):
         We can safely ignore the command's return code, because we just need to be sure that the key has been removed
         for non-existing server - we don't care about non-existing keys.
         """
-        self.logger.info('Deleting SSH key of "%s" host.', self.public_ip)
-        command = f'ssh-keygen -R {self.public_ip}'
+        # self.public_ip might fail because it calls OpenStack APIs,
+        # so we ignore those failures if they happen
         try:
-            subprocess.Popen(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        except subprocess.SubprocessError as e:
-            self.logger.error('Failed to delete SSH key of "%s" host: %s', self.public_ip, e)
+            ip = self.public_ip
+        except novaclient.exceptions.ClientException:
+            self.logger.warning(
+                'Failed to get the IP address of the server "%s", skipping the removal of its SSH key.',
+                self.name,
+                exc_info=True
+            )
+            return
+        ssh.remove_known_host_key(ip)
--- a/instance/ssh.py
+++ b/instance/ssh.py
@@ -1,0 +1,23 @@
+"""
+Utilities for dealing with SSH.
+"""
+import logging
+import subprocess
+
+LOGGER = logging.getLogger(__name__)
+
+
+def remove_known_host_key(ip: str):
+    """
+    Removes the known_hosts key entry for the given IP.
+
+    :param ip: the IP of the entry to delete
+
+    This doesn't check whether the entry existed.
+    """
+    LOGGER.info('Deleting the SSH host key of the server with the IP address "%s" from the SSH known hosts file.', ip)
+    command = f'ssh-keygen -R {ip}'
+    try:
+        subprocess.Popen(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    except subprocess.SubprocessError:
+        LOGGER.exception('Failed to delete the SSH host key of the server with the IP address "%s"', ip)
--- a/instance/tests/models/test_server.py
+++ b/instance/tests/models/test_server.py
@@ -337,15 +337,21 @@ class OpenStackServerTestCase(TestCase):
         ('ready-server-id', ServerStatus.Ready),
     )
     @unpack
-    def test_terminate_server_vm_created(self, openstack_id, server_status):
+    @patch('instance.ssh.remove_known_host_key')
+    def test_terminate_server_vm_created(self, openstack_id, server_status, remove_key):
         """
         Terminate a server with a VM
         """
-        server = OpenStackServerFactory(openstack_id=openstack_id, status=server_status)
+        server = OpenStackServerFactory(
+            openstack_id=openstack_id,
+            status=server_status,
+            # fake a public ip so SSH key can be removed
+            _public_ip='127.0.0.1'
+        )
         server.terminate()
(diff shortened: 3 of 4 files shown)


instance/models/server.py
```diff
@@ -452,7 +451,7 @@ def terminate(self):
                 self.logger.warning("Server has not reached SHUTOFF state after max wait time; terminating forcefully.")
             os_server.delete()
         except novaclient.exceptions.NotFound:
-            self.logger.error('Error while attempting to terminate server: could not find OS server')
+            self.logger.warning('Error while attempting to terminate server: could not find OS server')
```
lgp171188, line 454: @kewne, nit: Imho, this should be an error message as this is an error condition. We just don't want to abort due to this error. 😛

instance/models/server.py
```diff
@@ -486,9 +485,15 @@ def _delete_ssh_key(self) -> None:
...
+        # self.public_ip might fail because it calls OpenStack APIs,
+        # so we ignore those failures if they happen
         try:
-            subprocess.Popen(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        except subprocess.SubprocessError as e:
-            self.logger.error('Failed to delete SSH key of "%s" host: %s', self.public_ip, e)
+            ip = self.public_ip
+        except novaclient.exceptions.ClientException:
+            self.logger.warning(
+                'Failed to get IP for server "%s", skipping removal of SSH key...',
```
lgp171188, line 494:
```suggestion
                'Failed to get IP for server "%s", skipping the removal of SSH key.',
```

instance/models/server.py
```diff
@@ -486,9 +485,15 @@ def _delete_ssh_key(self) -> None:
...
-        command = f'ssh-keygen -R {self.public_ip}'
+        # self.public_ip might fail because it calls OpenStack APIs,
+        # so we ignore those failures if they happen
         try:
-            subprocess.Popen(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        except subprocess.SubprocessError as e:
-            self.logger.error('Failed to delete SSH key of "%s" host: %s', self.public_ip, e)
+            ip = self.public_ip
+        except novaclient.exceptions.ClientException:
+            self.logger.warning(
```
lgp171188, line 493: `self.logger.exception` will be cleaner and do the same job, no?

instance/ssh.py
```diff
@@ -0,0 +1,23 @@
...
+
+def remove_known_host_key(ip: str):
+    """
+    Removes the known_hosts key entry for the given IP.
+
+    :param ip: the IP of the entry to delete
+
+    This doesn't check whether the entry existed.
+    """
+    LOGGER.info('Deleting key of "%s" from SSH known hosts.', ip)
```
lgp171188, line 18:
```suggestion
    LOGGER.info('Deleting the SSH host key of the server with IP address "%s" from the SSH known hosts file.', ip)
```

instance/ssh.py
```diff
@@ -0,0 +1,23 @@
...
+    command = f'ssh-keygen -R {ip}'
+    try:
+        subprocess.Popen(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    except subprocess.SubprocessError as e:
+        LOGGER.error('Failed to delete SSH key of "%s" host: %s', ip, e)
```
lgp171188, line 23: Again, `logger.exception` would be cleaner here.

```suggestion
        LOGGER.exception('Failed to delete the SSH host key of the server with the IP address "%s", ip)
```

instance/models/server.py
kewne: For me, that's precisely what distinguishes logging something as an error vs logging a warning 😃 

The classic example here is what happens if there is a temporary network issue connecting to the database: the server is deleted from OpenStack but not marked as terminated.
The natural thing, then, would be to retry the operation, which would raise the not found exception; IMO it's incorrect to then log an error, as it incorrectly leads the user to think there is one.

Would you agree with slightly rewording the message to make this clearer, then?
Something like `Failed to terminate server: could not find OS server. It may have been deleted by a previous attempt to terminate.`?

instance/models/server.py
kewne: The rationale is the same as above, since `logger.exception` implicitly logs at `ERROR` level.

instance/models/server.py
lgp171188: >  The natural thing, then, would be to retry the operation, which would raise the not found exception; IMO it's incorrect to then log an error, as it incorrectly leads the user to think there is one.

@kewne, there is an error here since Ocim thinks that there is an OpenStack server and tries to delete it. But there is none. Unless Ocim then cleans up its internal state to reflect that the server no longer exists, this should be an error message imho.

instance/ssh.py
```diff
@@ -0,0 +1,23 @@
+"""
+Utilities for dealing with SSH.
```
lgp171188, line 2: @kewne, since this is just one utility function, wouldn't it be simpler to add it to `instance/utils.py`

instance/tests/models/test_server.py
```diff
@@ -337,15 +337,21 @@ def test_vm_not_created(self, server_status):
...
+    def test_terminate_server_vm_created(self, openstack_id, server_status, remove_key):
         """
         Terminate a server with a VM
         """
-        server = OpenStackServerFactory(openstack_id=openstack_id, status=server_status)
+        server = OpenStackServerFactory(
+            openstack_id=openstack_id,
+            status=server_status,
+            # fake a public ip so SSH key can be removed
+            _public_ip='127.0.0.1'
```
lgp171188, line 349: @kewne, can you add this to the factory object creation code itself instead of repeating it multiple times in similar tests. Afaik, that wouldn't cause any undesirable side effects.

instance/models/server.py
```diff
@@ -486,9 +485,15 @@ def _delete_ssh_key(self) -> None:
...
+                'Failed to get IP for server "%s", skipping the removal of its SSH key.',
```
lgp171188, line 494:
```suggestion
                'Failed to get the IP address for the server "%s", skipping the removal of its SSH key.',
```

lgp171188 (changes requested):
@kewne, please check and address the open review comments.

instance/tests/models/test_server.py
kewne: @lgp171188 I considered that but decided to do this instead because it's less surprising behavior: if you move it into the factory, code that uses `public_ip` will surprisingly never call through to OS and, to get that behavior, you have to use the even less obvious hack of setting `_public_ip = None`.

Another option is creating a specific Factory for this but that doesn't seem like a good solution either; tbh I think the root issue here is the way `public_ip` is handled.

instance/ssh.py
kewne: @lgp171188 That module contains a mix of totally unrelated functions, so I think that would just add to the confusion.

I personally think that, ironically, `utils` isn't a useful name (except when someone doesn't want to think too much about a name and uses it as a fallback 😛 ).

lgp171188: @kewne, I am not convinced here that this should be in a separate file since Ocim never uses SSH directly except for this functionality. So let us agree to disagree here. But since there is no issue in having it here, I am resolving this thread in the interest of wrapping this up before the sprint ends.

instance/tests/models/test_server.py
lgp171188: @kewne, my suggestion was to set a public IP as a default when there is none set. So what would be the issue then? I am not sure if I understand this

> code that uses public_ip will surprisingly never call through to OS and, to get that behavior, you have to use the even less obvious hack of setting _public_ip = None.

instance/models/server.py
kewne: Fixed.

instance/ssh.py
```diff
@@ -0,0 +1,23 @@
...
+
+def remove_known_host_key(ip: str):
+    """
+    Removes the known_hosts key entry for the given IP.
+
+    :param ip: the IP of the entry to delete
+
+    This doesn't check whether the entry existed.
+    """
+    LOGGER.info('Deleting the SSH host key of the server with the IP address "%s" from the SSH known hosts file.', ip)
```
lgp171188, line 18: @kewne, when testing this, I see a lot of log lines that say

```
Deleting the SSH host key of the server with the IP address "None" from the SSH known hosts file.
```

Can't we just not run code in this function when the IP is `None`?

kewne: I'm resolving this because the issue already existed. I'll create a follow-up story, it seems like a particularly newcomer friendly issue.

PR #819, opened 2021-06-30
FAL-2031 docker dev tweaks

samuelallan72:
[FAL-2031](https://tasks.opencraft.com/browse/FAL-2031)

This makes some tweaks to the docs and compose file for the docker devstack for ocim, for better usability.

**Test instructions**:

- follow the edited instructions in the docker.md doc
- verify that the docker devstack launches and runs ocim successfully

**Reviewers**:

- [ ] @gabor-boros 
- [x] @kewne

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,30 +21,25 @@ services:
     image: redis:4
     ports:
       - "6379:6379"
-    
+
   ocim:
     build:
       context: .
       dockerfile: Dockerfile.ocim
-    depends_on: 
+    depends_on:
       - postgresql
       - redis
     env_file: .env
-    environment: 
-      ALLOWED_HOSTS: '["*"]'
-      DATABASE_URL: '${DATABASE_URL:-postgres://opencraft@postgresql/opencraft}'
-      REDIS_URL: '${REDIS_URL:-redis://redis:6379}'
     ports:
       - "127.0.0.1:5000:5000"
     volumes:
       - ./:/usr/src/ocim/
-    
+
   ocim-frontend:
     build:
       context: frontend/
-    depends_on: 
+    depends_on:
       - ocim
-    environment: {} 
     ports:
       - "127.0.0.1:3000:3000"
     tty: true
--- a/documentation/development/docker.md
+++ b/documentation/development/docker.md
@@ -6,54 +6,40 @@ This describes how to create an environment for development using Docker.
 
 ## Running OCIM in a docker container
 
-First, set your `.env` file as you would normally:
-```env
-DEBUG=True
-OPENSTACK_USER='username'
-OPENSTACK_PASSWORD='password'
-OPENSTACK_TENANT='tenant-name'
-OPENSTACK_AUTH_URL='https://auth.cloud.ovh.net/v2.0'
-OPENSTACK_REGION='BHS1'
-OPENSTACK_SANDBOX_SSH_KEYNAME='keypair-name'
-DEFAULT_INSTANCE_BASE_DOMAIN='example.com'
-GANDI_API_KEY='api-key'
-GITHUB_ACCESS_TOKEN='github-token'
-SECRET_KEY='tests'
-DEFAULT_INSTANCE_MYSQL_URL=...
-DEFAULT_RABBITMQ_API_URL=...
-DEFAULT_INSTANCE_RABBITMQ_URL=...
-DEFAULT_MONGO_REPLICA_SET_USER=...
-DEFAULT_MONGO_REPLICA_SET_PASSWORD=...
-DEFAULT_MONGO_REPLICA_SET_NAME=...
-DEFAULT_MONGO_REPLICA_SET_PRIMARY=...
-DEFAULT_MONGO_REPLICA_SET_HOSTS=...
-REDIS_URL=...
-```
+First, set your `.env` file as you would normally,
+but set the following vars:
 
-Start a one-off container for setup of the backend:
 ```sh
-# --rm destroys the container at the end
-docker-compose run --rm ocim bash
-
-# then, inside the container...
-
-# initialize datastores
-make migrate
-# create super user
-make manage createsuperuser
-# exit the container
-exit
+ALLOWED_HOSTS='["*"]'
+DATABASE_URL='postgres://opencraft@postgresql/opencraft'
+REDIS_URL='redis://redis:6379'
 ```
 
-Then, start another one-off container for setting up the client-facing frontend:
+Run some initial setup tasks in the backend:
+
+```sh
+docker-compose run --rm ocim make migrate
+docker-compose run --rm ocim make manage createsuperuser
+```
+
+And on the frontend:
+
 ```sh
 docker-compose run --rm ocim-frontend bash -c 'npm run build-api-client && npm install'
 ```
 
-Then, start OCIM:
+Then, launch the stack (detached):
+
 ```sh
-docker-compose up
+docker-compose up -d
 ```
 
-The OCIM UI should be available at http://localhost:5000 .
-The registration UI should be available at http://localhost:3000 .
+The OCIM UI should be available at <http://localhost:5000>.
+The registration UI should be available at <http://localhost:3000>.
+
+To view or follow the logs for any of the running services, use `docker-compose logs`.
+For example:
+
+```sh
+docker-compose logs -f ocim
+```


docker-compose.yml
```diff
@@ -21,30 +21,25 @@ services:
...
       dockerfile: Dockerfile.ocim
-    depends_on: 
+    depends_on:
       - postgresql
       - redis
     env_file: .env
-    environment: 
-      ALLOWED_HOSTS: '["*"]'
-      DATABASE_URL: '${DATABASE_URL:-postgres://opencraft@postgresql/opencraft}'
-      REDIS_URL: '${REDIS_URL:-redis://redis:6379}'
```
kewne, line 36: The idea of having these variables here is that you don't have to set them in `.env`, especially since they are specific to the Docker setup.

`ALLOWED_HOSTS` is particularly problematic because, without the wildcard, you won't be able to access the Django app from outside the container.

Looking at the docs though, it would make sense to document that these don't need to be set.

documentation/development/docker.md
```diff
@@ -6,54 +6,33 @@ This describes how to create an environment for development using Docker.
...
+
 ```sh
 docker-compose run --rm ocim-frontend bash -c 'npm run build-api-client && npm install'
 ```
 
 Then, start OCIM:
+
 ```sh
-docker-compose up
+docker-compose up -d
```
kewne, line 34: Out of curiosity, what's the rationale for running detached here?

To clarify, I can see why someone would want to run it detached; however, I prefer to assume the reader might not know `docker-compose`  well and thus be confused when he "starts OCIM" but doesn't see any output.

docker-compose.yml
swalladge: Thanks!  This is only a WIP dump of what I was working on btw; I'll refine it when I come back to it tomorrow :)

documentation/development/docker.md
gabor-boros: > be confused when he "starts OCIM" but doesn't see any output.

@kewne Actually, it would see the message that the containers are started, though it is also true that the logs wouldn't be shown. We could add `docker-compose up -d && docker-compose logs` as a separate command in the docs to have the detached behavior with showing the logs too.

arbrandes: @kewne,

> what's the rationale for running detached here?

That is the default behavior of the edx devstack and others I've used in the past, with separate commands for viewing logs when you need them.  I second @swalladge's implicit opinion that this is a generally better default than the alternative.

kewne: @gabor-boros 

My point isn't that plain `docker-compose up` is the best command for every or even most workflows but rather that it is the least surprising behavior, especially considering that someone reading this might not be an expert in Docker and Docker Compose;
conversely, someone who *is* an expert doesn't need additional documentation to know he *can* use `-d`.
The intention here isn't to impose or suggest a particular workflow, it's literally "this is the **simplest** sequence of commands that gets you started": I find it hard to argue that `docker-compose up -d && docker-compose logs` is superior to `docker-compose` in that regard.

> That is the default behavior of the edx devstack and others I've used in the past

@arbrandes I'm a bit surprised by your reply: I've always considered starting stuff in the foreground to be the "correct" behavior, so much so that it's the default for pretty much everything (`docker`, `./manage runserver`, etc).
Besides, this is actually something I dislike about the way devstack works: it attempts to hide its use of `docker-compose` but, in doing so, creates a bespoke interface that won't allow me, an experience user, to properly customize commands.

For the sake of "meeting in the middle", perhaps the best would be to state this uses Docker Compose, and link to its documentation.
Optionally, we could document "useful beginner commands", in which case I feel `-d` and `logs` would be very appropriate.

arbrandes: > I've always considered starting stuff in the foreground to be the "correct" behavior, so much so that it's the default for pretty much everything

As far as Unix-like systems go, you're right, I agree, and it's why `-d` is in fact an optional flag.  No argument there.

It's just that for development stacks that feature more than one process, I find launching them in the background and looking for individual logs when I need them to be the more efficient experience.  In Ocim's case it's harder to press this point, as it is evidently a single process.  So I won't. :)

swalladge: whoa whoa this is a work in progress dump of notes I was making while trying to get everything working! :sweat_smile: Let's not bikeshed here just yet. :stuck_out_tongue: 


>  In Ocim's case it's harder to press this point, as it is evidently a single process. 

Ocim is no longer a single process; there's a separate frontend app too.  And the docker compose launches postgres, redis, etc.

I'll come back to this discussion once I've had a chance to work with it a bit longer.

docker-compose.yml
swalladge: @kewne so I didn't like this because there is no longer a single source of truth for the environment, and overriding it requires either overriding the vars somewhere else or directly editing the docker-compose.yml.  Also, I couldn't get it working with the default config there anyway.

docker-compose.yml
```diff
@@ -55,4 +50,4 @@ services:
 volumes: {}
 
 networks:
-  default:
\ No newline at end of file
+  default:
```
swalladge, line 53: It's 2021; what editors are still not properly adding newlines to the end of files. :|

documentation/development/docker.md
swalladge: Ok, I updated the docs so it's clear what's happening.  Also added a note about viewing the logs.

docker-compose.yml
swalladge: I see this a lot, so I'm going to leave this stackoverflow answer as a reference for why that last newline is important: https://stackoverflow.com/a/729795

documentation/development/docker.md
swalladge: > perhaps the best would be to state this uses Docker Compose, and link to its documentation.

The docs literally show use of the `docker-compose` command, so nothing is being hidden. :thinking:

docker-compose.yml
kewne: VSCode! I've enabled the option to add the line, so it shouldn't happen again.

Thanks for the link as well!

docker-compose.yml
kewne: > I didn't like this because there is no longer a single source of truth for the environment

I agree with the sentiment; my reasoning is, once again, from the POV of someone trying to set this up for the first time: if the instructions tell me to set up a `.env` file with certain values, I'd expect them to be specific to my local environment (i.e. my OpenStack credentials, my MySQL credentials, etc); however, most of the time these three specific variables need to match the `docker-compose.yml` file and shouldn't be changed;
while you *can* change them, I think it's acceptable to do it differently from the other config because it's not the common use case.

The ideal thing here would really be to have a way to "stack" configuration (e.g. like devstack' `private.py` does), I just don't know how it could be achieved because neither Docker Compose or Honcho support it: the only thing I can think of is having a custom entrypoint script but that has other issues.

> I couldn't get it working with the default config there anyway

Can you describe what's not working? That would help me to better understand the issue and maybe find a better solution.

swalladge: > Can you describe what's not working? 

I had a lot of trouble setting the database_url variable.  It kept wanting to use the database_url set from the env file I reckon.  That's why I'm pushing for a single source of truth.

> from the POV of someone trying to set this up for the first time: if the instructions tell me to set up a .env file with certain values

I think this is fine - we're telling them in the docs exactly what variables are required for the docker devstack, and why.  This makes it clear, while also keeping the single source of truth and allowing flexibility to set whatever you want for different use cases.

kewne: > I had a lot of trouble setting the database_url variable. It kept wanting to use the database_url set from the env file I reckon.

That shouldn't be possible, according to [the docker compose docs](https://docs.docker.com/compose/compose-file/compose-file-v3/#env_file):
> Environment variables declared in the environment section override these values – this holds true even if those values are empty or undefined.

I imagine it could be caused by having `DATABASE_URL` set in your environment though (which would use `.env` if  running something like `honcho run bash`).

kewne (approved):
:+1:

- [x] I tested this: Followed the instructions
- [x] I read through the code
- [x] I checked for accessibility issues: N/A
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.: N/A

docker-compose.yml
swalladge: @kewne 

bq. That shouldn't be possible, according to the docker compose docs:

Ah right, the var was defined as:

```
'${DATABASE_URL:-postgres://opencraft@postgresql/opencraft}'
```

So it was never hitting the default, because DATABASE_URL was set to something else previously.

We _could_ remove the default fallback and hardcode the desired postgres url, but then it's impossible for the user to override it if they wish without editing the docker-compose file.

PR #820, opened 2021-06-30
[BB-4378] Fix CI cleanup issues

kewne:
This PR fixes the issues that were preventing the cleanup of some unused CI resources like MySQL databases, Gandi DNS records, etc.

It also changes the CircleCI workflow definitions to allow pushing to the branches with names starting with `ci-cleanupl` for running cleanup on-demand (similar to how `stage` is used for the frontend).

# Testing
1. Verify that `cleanup` job for this branch passes correctly

# Links

* [BB-4378 (OpenCraft Private JIRA)](https://tasks.opencraft.com/browse/BB-4378)
* [@lgp171188's original PR](https://github.com/open-craft/opencraft/pull/815)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -490,7 +490,9 @@ workflows:
               ignore: /.*/
             tags:
               only: /^release\-\w+\-\w+$/
-  cleanup-workflow:
+  
+  # used for periodically cleaning up left-over resources
+  scheduled-cleanup:
     triggers:
       - schedule:
           cron: "0 1 * * *"
@@ -498,6 +500,15 @@ workflows:
             branches:
               only:
                 - master
-                - guruprasad/BB-4378-fix-CI-cleanup-issues
     jobs:
       - cleanup
+
+  # used for running cleanup on-demand
+  # by pushing to `ci-cleanup` branch
+  ondemand-cleanup:
+    jobs:
+      - cleanup:
+          filters:
+            branches:
+              only:
+                - ci-cleanup
--- a/circle.yml
+++ b/circle.yml
@@ -504,11 +504,11 @@ workflows:
       - cleanup
 
   # used for running cleanup on-demand
-  # by pushing to `ci-cleanup` branch
+  # by pushing to branches that start with `ci-cleanup/`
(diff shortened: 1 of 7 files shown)


circle.yml
```diff
@@ -504,11 +504,11 @@ workflows:
...
-  # by pushing to `ci-cleanup` branch
+  # by pushing to branches that start with `ci-cleanup/`
   ondemand-cleanup:
     jobs:
       - cleanup:
           filters:
             branches:
               only:
-                - ci-cleanup
+                - /ci-cleanup\/.*/
```
shimulch, line 514: @kewne, Can you provide an example of when this would be useful? I think sticking with a single branch here is a better idea. In that way, we can restrict pushes to `ci-cleanup` branch if needed. But changes of this commit mean, any push to `ci-cleanup/*` branches will run the job.

shimulch (changes requested):

circle.yml
kewne: @shimulch The idea is that this allows you to create "feature" branches that run the cleanup job, which is useful for reviewing because it shows directly in the PR, instead of having to `git push -f origin HEAD:ci-cleanup` and having to grab the build link every time you add a commit.
IMO having multiple branches isn't an issue because anyone with push permissions can always add their specific branch to the `.circle.yml` file anyway (which would have to be reverted before merging to `master`).

Ideally, there would be a way to add metadata to branches (or commits) that tells CircleCI that we want to run the cleanup job but I haven't found anything suitable; I'll gladly try other suggestions.

PS: what if I remove the `only:` part and make the deployment manual?

shimulch: @kewne, Yes, I thin removing `only` and making the deployment manual might be a better approach. Then we would not need to push empty commits to run a clean-up.

https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval

kewne: @shimulch 👍 Since the specific branch name is no longer necessary, I'm closing this in favor of the original #815

PR #821, opened 2021-07-01
[BB-4387] When provisioning an App Server and security group checks fail, state is left in New

kewne:
This fixes an issue where, if something failed while checking security group definitions on OpenStack, the machine state would be left in New; the user would only be able to tell the process failed in the UI by looking through the logs.

Now, the machine state is set to error.

# Related issues
* [BB-4387 (OpenCraft Private JIRA)](https://tasks.opencraft.com/browse/BB-4387)

# Testing
```sh
./manage.py test instance.tests.models.test_openedx_appserver.OpenEdXAppServerTestCase.test_provision_security_group_check_failed
```

### DIFF
--- a/instance/models/openedx_appserver.py
+++ b/instance/models/openedx_appserver.py
@@ -53,6 +53,21 @@ OPENEDX_APPSERVER_SECURITY_GROUP_RULES = [
 ]
 
 
+class ProvisioningError(Exception):
+    """
+    Raised when provisioning fails.
+    """
+
+    def __init__(self, message, log=None):
+        """
+        Arguments:
+            message: a message describing the error
+            log (optional): a log from an underlying process
+        """
+        super().__init__(message)
+        self.message = message
+        self.log = log
+
 # Models ######################################################################
 
 class OpenEdXAppConfiguration(models.Model):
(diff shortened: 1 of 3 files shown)


instance/models/openedx_appserver.py
```diff
@@ -53,6 +53,21 @@
 ]
 
 
+class ProvisioningError(Exception):
+    """
+    Raised when an provisioning fails.
+    """
+
+    def __init__(self, message, log=None):
```
shimulch, line 61: @kewne Argument `log` has not been used anywhere. Can we just skip it and simplify the error class to -

```python
class ProvisioningError(Exception):
    """
    Raised when an provisioning fails.
    """
    pass
```

instance/models/openedx_appserver.py
```diff
@@ -603,88 +618,116 @@ def provision(self):
 
         Returns True on success or False on failure
         """
-        # pylint: disable=cyclic-import, useless-suppression
-        from instance.models.openedx_deployment import OpenEdXDeployment
+        try:
+            return self._do_provision()
+        except ProvisioningError as pe:
+            self.logger.exception(pe.message)
+            self.provision_failed_email(pe.message, pe.log)
```
shimulch, line 625: Since there will be no `log` in `ProvisioningError`, we can skip sending it. This will simplify and reduce changes in some tests as well.
```suggestion
            self.provision_failed_email(pe.message)
```

shimulch (changes requested):
@kewne Thanks for the PR. I've left some comments. Please address those.

instance/models/openedx_appserver.py
kewne: Ah, you're right: my intent was to also use it [here](https://github.com/open-craft/opencraft/blob/419e0aa32072a36eb5c83b6d80154fb625c4fefd/instance/models/openedx_appserver.py#L718) so it would become something like
```python
if exit_code != 0:
    self._status_to_configuration_failed()
    self.manage_instance_services(active=False)
    raise ProvisionError("AppServer deploy failed: Ansible play exited with non-zero exit code", log)
```

What do you think? I've also realized it'd be best to document the intended use of `ProvisionError` in these methods.

shimulch: @kewne, Ah! okay, I didn't see that use case. Yes, I think we can use the new `ProvisionError` there and document it properly.

instance/models/openedx_appserver.py
kewne: Resolving this based on [response here](https://github.com/open-craft/opencraft/pull/821#discussion_r663508376)

shimulch (approved):
@kewne Thanks for the fix here. Both the fix and the refactor LGTM :+1:

- [x] I tested this: I've tested locally if failing on security group checking moves the appserver to the failed state.
- [x] I read through the code
- [x] Includes documentation

PR #822, opened 2021-07-01
[FAL-2031] Add Redis support for Ocim provisioned instances

gabor-boros:
This Pull Request adds Redis cache support to Ocim. The changes won't take effect on already provisioned servers, but new ones or those that explicitly set to use Redis over RabbitMQ. This way we can slowly roll out changes for every server while we are not generating more work when new instances are created.

**~Dependencies~ Related PRs**:

* https://github.com/celery/kombu/pull/1349
* https://github.com/celery/kombu/pull/1351
* https://github.com/edx/edx-platform/pull/28020

**Screenshots**: 

_Admin site integration of `RedisServer`_

Screenshot 2021-07-01 at 16 10 00

_Instance admin configuration_

Screenshot 2021-07-01 at 14 49 00

_Redis cluster showing the Ocim created user and its ACL_

Screenshot 2021-07-01 at 15 35 27

**Sandbox URL**: Upon request, I can create an ngrok URL for my local Ocim

**Merge deadline**: None

**Testing instructions**:

1. Checkout the changes on Ocim stage 
2. Create a Redis user and ACL for Ocim if no user was created (or use the default user) and double-check the connection to the Redis server
3. Set the necessary environment variables in the `.env` file (ie. `DEFAULT_INSTANCE_REDIS_URL` )
4. Restart Ocim
5. Provision a new app server
6. Login with edX user
7. Go to instructor/download data tab
8. Download a grades report (as that uses celery, that would use Redis now)
9. Validate the instructor task is finished successfully

A note about staging env: The instance creation nowadays is hectic -- it seems for some instances the LB URL registration did not work.

N/A

**Reviewers**
- [ ] @pomegranited 
- [ ] @swalladge 
- [ ] @gabor-boros


instance/models/redis.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class RedisUser(ValidateModelMixin, TimeStampedModel):
+    """
+    A model representing a single Redis user identified by ACLs.
+    """
+
+    # TODO: The user can have multiple passwords set by ACLs. Create signals to
+    # handle ACL setting on user creation, ACL update on password change, and
+    # ACL deletion on user deletion.
```
swalladge, line 36: I feel like a safer method would be to add a methods `create_redis_user`, `delete_redis_user`, `set_password`, which would include the required acl setup/teardown steps.  Then we don't need to rely on signals and all the logic is in one place.

swalladge: (Similar to the `provision_rabbitmq`, `deprovision_rabbitmq` methods)

documentation/configuration.md
```diff
@@ -210,6 +210,10 @@ Required settings:
 * `DEFAULT_INSTANCE_RABBITMQ_URL`: The RabbitMQ AMQPS URI to be used by instances. E.g.,
   `amqps://rabbitmq.example.com:5671`
 
+### Redis settings
+* `DEFAULT_INSTANCE_REDIS_URL` The Redis URI (including the protocol, port, db, and basic auth)
+  to be used by instances. E.g., `rediss://admin:admin_password@redis.example.com:6397/0`
```
swalladge, line 215: @gabor-boros can we add a comment that this is processed into the database with `RedisServerManager._create_default` - so we can see the link between this and the `RedisServer` model?

instance/models/mixins/openedx_database.py
```diff
@@ -490,6 +541,11 @@ def get_database_settings(self):
             new_settings.update(self._get_mongo_settings())
 
         # RabbitMQ:
-        new_settings.update(self._get_rabbitmq_settings())
+        # TODO: Use a real condition here
+        if False:
```
swalladge, line 545: @gabor-boros yeah we could define a boolean field on the `RedisInstanceMixin` like `uses_redis`, then:

```
if self.uses_redis:
  new_settings.update(self._get_redis_settings())
else:
  new_settings.update(self._get_rabbitmq_settings())
```

instance/models/mixins/redis.py
```diff
@@ -0,0 +1,169 @@
...
+    Return the primary key of a new Redis user.
+    """
+
+    if max_tries == 0:
+        raise IntegrityError("Cannot create unique redis user")
+
+    username = get_random_string(length=32, allowed_chars=string.ascii_lowercase)
+
+    if RedisUser.objects.filter(username=username).count() > 0:
+        return new_redis_user(max_tries - 1)
```
swalladge, line 52: @gabor-boros I think we can skip this retry code; the chance of a 32 character cryptographically random string colliding is extremely low.  The simpler the better. :)

instance/models/mixins/redis.py
```diff
@@ -0,0 +1,169 @@
...
+        username=username,
+        password=get_random_string(length=64)
+    )
+
+    return user.pk
+
+
+def select_random_redis_server():
+    """
+    Helper for the field default of `redis_server`.
```
swalladge, line 64: Could you expand the docstring here to explain why this is required?

instance/models/mixins/redis.py
```diff
@@ -0,0 +1,169 @@
...
+
+# Classes #####################################################################
+
+class RedisAPIError(Exception):
+    """Exception indicating that a call to the Redis API failed."""
+
+
+class RedisInstanceMixin(models.Model):
+    """
+    An instance that uses a Redis vhost with a set of users.
```
swalladge, line 77: s/vhost/redis terminology/

instance/models/mixins/redis.py
gabor-boros: I had the same assumption too, though I wanted to follow somewhat the rabbitmq provisioning does. I'm more than happy to remove the retry logic, I don't like it either 😁

instance/models/redis_server.py
```diff
@@ -0,0 +1,120 @@
...
+                    "Found DEFAULT_INSTANCE_REDIS_URL, but it must contain "
+                    "basic auth credentials, hostname, port and db."
+                )
+
+            logger.info("Creating RedisServer %s", defaults['instance_host'])
+
+            defaults['name'] = defaults['instance_host']
+            defaults['instance_db'] = int(defaults['instance_db'])
+
+            server, created = self.get_or_create(defaults=defaults)
```
swalladge, line 83: can this get_or_create based on the hostname or something, so it doesn't need that random pk function (which I'm guessing that's what it's for)?

instance/models/mixins/redis.py
gabor-boros: Oh, renaming went wrong. This remained here. Going to replace vhost with db (which is the redis terminology)

instance/models/mixins/redis.py
gabor-boros: I don't think it is required, wanted to write a todo comment here, but it seems I forgot. Actually that's how the rabbitmq server selection worked though I think that was a bad choice back in time to do this. Let's not follow that pattern. Will remove it.

instance/models/redis.py
swalladge: @gabor-boros actually, why do we need this model at all?  Can't it all be handled by the `RedisInstanceMixin`?  Or is there a possibility of multiple redis users defined per instance? :thinking: After reading the code of both models, it seems that we're splitting logic across two models.

instance/models/mixins/redis.py
gabor-boros: I checked the logic behind it. The select random selects a random server that accepts new connections and if the default instance redis URL is set, create a redis server based on it. Actually, we don't need the "random selection" logic, but the create default server logic is handy.

instance/models/redis_server.py
gabor-boros: The `get_or_create` is based on the connection parameters only (a.k.a the parsed URI/defaults). Since changing any parameter of the connections parameters should be represented as a new redis server (even if the db param is changed).

instance/models/redis.py
gabor-boros: > Can't it all be handled by the RedisInstanceMixin?

I spent some time thinking about this. Having the user info in a separate model is not a bad idea since the `RedisUser` model encapsulates the user management logic (create/delete). If that would be explicitly handled by the `RedisInstanceMixin` -- not just calling the user creation but actually know how to create a user -- it could be a mess soon if we need to do more Redis user-related calls.

I'd say that the rule of thumb here could be
* `RedisUser` model should have all Redis user-related logic
* `RedisInstanceMixin` should have all provisioning logic
* `RedisServer` model should have all Redis logic that is **not** user-related

Based on the above, getting the Redis client should that is passed around should be handled by the `RedisServer` model as it is responsible for the connection parameters including credentials. What do you think about this?

instance/models/redis_server.py
swalladge: Ok sure.  As long as the behaviour is documented - otherwise the relationship between `DEFAULT_INSTANCE_REDIS_URL` and an actual redis configuration in the database is confusing.

instance/models/redis.py
swalladge: Ok, using signals makes sense if we're adding/removing RedisUsers in django.  But we're tied to one redis 'user' per openedxinstance right?  Then we don't need to worry about creating/deleting RedisUsers, because the provisiong/deprovisioning of redis users/acls will be managed by provisioning/archiving openedx instances.

> RedisUser model should have all Redis user-related logic

Just be careful that we aren't making unecessary database models simply as a way to namespace functions. ;)

swalladge: In the RedisInstanceMixin, we can have the `redis_user` and `redis_password`, and move the `delete_acl` and `create_acl` methods to RedisInstanceMixin.provision_redis/deprovision_redis.  Then we have half as many methods and models to maintain.

swalladge: For example if we did something like this:

```diff
diff --git a/instance/models/mixins/redis.py b/instance/models/mixins/redis.py
index 65cefaae..fd065197 100644
--- a/instance/models/mixins/redis.py
+++ b/instance/models/mixins/redis.py
@@ -38,16 +38,6 @@ from instance.models.redis_server import RedisServer
 
 # Functions ###################################################################
 
-def new_redis_user() -> RedisUser:
-    """
-    Return a new Redis user.
-    """
-    return RedisUser.objects.create(
-        username=get_random_string(length=32, allowed_chars=string.ascii_lowercase),
-        password=get_random_string(length=64),
-    )
-
-
 def select_random_redis_server():
     """
     Helper for the field default of `redis_server`.
@@ -73,14 +63,8 @@ class RedisInstanceMixin(models.Model):
         on_delete=models.PROTECT,
     )
 
-    redis_consumer_user = models.ForeignKey(
-        RedisUser,
-        related_name='consumer_instance',
-        on_delete=models.CASCADE,
-        blank=True,
-        null=True,
-        default=new_redis_user,
-    )
+    redis_username = models.CharField(max_length=32, unique=True, default=lambda:get_random_string(length=32, allowed_chars=string.ascii_lowercase))
+    redis_password = models.CharField(max_length=64, default=lambda:get_random_string(length=64))
 
     redis_provisioned = models.BooleanField(default=False)
 
@@ -119,11 +103,19 @@ class RedisInstanceMixin(models.Model):
         self.logger.info('Provisioning Redis started.')
 
         client = self._redis_client()
-        self.redis_consumer_user.create_acl(client)
+        client.acl_setuser(
+            username=self.redis_username,
+            passwords=[f"+{self.redis_password}"],
+            keys=[self.get_key_prefix()],
+            enabled=True
+        )
 
         self.redis_provisioned = True
         self.save()
 
+    def get_key_prefix(self) -> str:
+        return self.redis_username + _
+
     def deprovision_redis(self, ignore_errors=False):
         """
         Deletes the Redis vhost and users.
@@ -141,7 +133,7 @@ class RedisInstanceMixin(models.Model):
 
         try:
             client = self._redis_client()
-            self.redis_consumer_user.delete_acl(client)
+            client.acl_deluser(self.redis_username)
         except Exception as exc:
             self.logger.exception(
                 'Cannot delete redis user: %s. %s',
```

instance/models/mixins/redis.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    redis_server = models.ForeignKey(
+        RedisServer,
+        null=True,
+        blank=True,
+        default=select_random_redis_server,
```
swalladge, line 72: @gabor-boros oh _this_ is where the select random server is used.  Would we be able to rework that so it selects the redis server defined by `DEFAULT_INSTANCE_REDIS_URL` by default?

documentation/configuration.md
gabor-boros: Fixed in 53688046

instance/models/redis.py
gabor-boros: @swalladge I had some time to think about this:

I did apply the suggestion related to `RedisUser` and committed the changes already. Regarding my previous statement about signals, I still believe that we could have signals, though I would leave that as an improvement idea.

I mean, implementing the above-mentioned signals would make sense, though I cannot remember for a single time when I had to manually change the credentials of a rabbit mq vhost, so I believe that won't be often for redis either.

How about creating a ticket in the backlog about this improvement idea to not forget it, but skip the implementation for that now?

instance/models/mixins/redis.py
gabor-boros: What are your thoughts on this @swalladge ?

swalladge: Fine by me, as long as it's documented. :)

instance/models/mixins/openedx_database.py
```diff
@@ -475,6 +489,56 @@ def _get_rabbitmq_settings(self):
...
+            "XQUEUE_REDIS_USER": self.redis_username,
+            "XQUEUE_REDIS_PASS": self.redis_password,
+            "XQUEUE_REDIS_VHOST": str(self.redis_server.instance_db),
+            "XQUEUE_REDIS_HOSTNAME": self.redis_server.instance_host,
+            "XQUEUE_REDIS_PORT": self.redis_server.instance_port,
+            "XQUEUE_REDIS_TLS": self.redis_server.use_ssl_connections,
+            "XQUEUE_SESSION_ENGINE": "django.contrib.sessions.backends.cache",
+            "XQUEUE_CACHES": {
+                "default": {
+                    # FIXME: What should be this? Is it needed at all?
```
swalladge, line 519: @gabor-boros I have no idea about this sorry. :/

instance/models/mixins/openedx_database.py
```diff
@@ -475,6 +489,56 @@ def _get_rabbitmq_settings(self):
...
+            self.redis_server.instance_port
+        )
+
+        redis_transport_options = {
+            "CELERY_BROKER_TRANSPORT_OPTIONS": {
+                "global_keyprefix": self.redis_key_prefix
+            }
+        }
+
+        # TODO: Validate the key names as I just copy pasted them from RabbitMQ
```
swalladge, line 508: Flagging this TODO @gabor-boros

instance/models/redis.py
swalladge: @gabor-boros looks good :+1:

bq. I mean, implementing the above-mentioned signals would make sense...

When would a signal be required?  If we needed to update the credentials, wouldn't the flow be:

- change credentials
- launch new appserver
  - which will run the redis_provision method, updating the acls

swalladge:
@gabor-boros somewhere we will need to set the acl policy for the key prefix for an instance.  I'm not sure if it's been done already?  I couldn't find it.

instance/models/redis.py
gabor-boros: @swalladge Actually the workflow should be:
1. change credentials in ocim
2. delete ACL in Redis
3. uncheck `redis_provisioned`
4. launch new app server

Without step 2, the Redis instance would be still accessible with old credentials. Again, I cannot remember a single time when we had to change credentials so it is really rare.

gabor-boros:
@swalladge 

We are setting the [key prefix](https://github.com/open-craft/opencraft/pull/822/files#diff-0e3dc6233348de21e4c96e5aeaf4eb5a39aa6db9a87f7482695970fcd1c2830cR103) based on the Redis username, and [creating the ACL](https://github.com/open-craft/opencraft/pull/822/files#diff-0e3dc6233348de21e4c96e5aeaf4eb5a39aa6db9a87f7482695970fcd1c2830cR118) using the prefix. When the key prefix exists (we have a Redis username) and an ACL is created for that, the only thing left to do is setting the `global_keyprefix` in the Redis transport options, which is done [here](https://github.com/open-craft/opencraft/pull/822/files#diff-ebbcd7e81e8c37b3db2714f5fc136134fcc206ab27f651706cd470d9228e7f27R504).

instance/models/redis.py
swalladge: @gabor-boros 

> Without step 2, the Redis instance would be still accessible with old credentials.

Are you sure about that?  According to [the docs for redis.acl_setuser](https://redis-py.readthedocs.io/en/stable/#redis.Redis.acl_setuser):

> Create or update the ACL for username. If the user already exists, the existing ACL is completely overwritten and replaced with the specified values.

So there shouldn't be old acls hanging around.

swalladge:
@gabor-boros ah got it, thanks :+1:

instance/models/redis.py
gabor-boros: Oh, then I misread something! Sorry about that. In this case we are fine here.

swalladge: Great!  I also had in my mind that we would need manual processing to delete old ACLs, but maybe it's just that the python library handles this for us, while the redis client doesn't.  CC @pomegranited in case I missed something.

gabor-boros: As this seems to be resolved, I'm resolving the thread too.

instance/models/mixins/redis.py
gabor-boros: Added more documentation as part of `cc739563`

instance/models/redis_server.py
gabor-boros: This has been documented in a previous commit.

instance/models/mixins/redis.py
gabor-boros: @swalladge That would affect all the classes that are inherited from `SharedServerManager` (`LoadBalancingServerManager`, `RabbitMQServerManager`, `RedisServerManager` and `DatabaseServerManager`). I'm happy to rework, but I want to double-check: Are you sure we want to refactor that logic as part of this PR?

instance/models/mixins/openedx_database.py
gabor-boros: @swalladge Well, I believe this is not needed at all. Based on the [changelog](https://github.com/edx/configuration/blob/b2bf69e0dad429792693f4365827e498811b3991/CHANGELOG.md), XQueue configuration was removed on 2017-02-27 and the remaining config got removed on 2018-04-13.

I believe we could remove all XQueue config 😅 For Redis settings, I removed as part of 56aa44b9.

Do you agree with me?

instance/models/mixins/openedx_database.py
gabor-boros: Removed in 56aa44b9.

instance/models/mixins/redis.py
swalladge: Yeah ok, good idea to leave it for now then :+1:

instance/models/mixins/openedx_database.py
swalladge: Lol ok cool.  I agree. :)

swalladge (approved):
@gabor-boros :+1:

- [x] I tested this: tested async tasks on stage appserver
- [x] I read through the code
- [NA] I checked for accessibility issues
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding

instance/models/openedx_instance.py
```diff
@@ -310,8 +310,23 @@ def _spawn_appserver(self, deployment_id=None):
         elif self.storage_type == self.S3_STORAGE:
             self.logger.info('Provisioning S3 bucket...')
             self.provision_s3()
-        self.logger.info('Provisioning RabbitMQ vhost...')
-        self.provision_rabbitmq()
+
+        # Provision redis server if stated explicitly or no app servers were created before (ie. this is the first one)
+        if self.cache_db == OpenEdXDatabaseMixin.REDIS or self.appserver_set.count() == 0:
```
gabor-boros, line 315: @swalladge please see the related conversation: https://chat.opencraft.com/opencraft/pl/ppf7ycskapnjfc9g861eeoqfxe

instance/tests/models/test_redis_mixin.py
```diff
@@ -0,0 +1,275 @@
...
+from instance.tests.base import TestCase
+from instance.tests.models.factories.openedx_appserver import make_test_appserver
+from instance.tests.models.factories.openedx_instance import OpenEdXInstanceFactory
+from registration.models import BetaTestApplication
+
+
+# Tests #######################################################################
+
+@ddt.ddt
+class OpenEdXThemeMixinTestCase(TestCase):
```
pomegranited, line 42: @swalladge @gabor-boros -- these tests look unrelated to this PR? Do we need to merge latest master or something?

pomegranited:
Fantastic work here @gabor-boros  and @swalladge !

I'm happy with what's running on stage and the recent fixes, so this is ready to go live :)

instance/tests/models/test_redis_mixin.py
swalladge: @pomegranited ah yes this looks like a copy/paste issue.  :P  I can remove this file altogether.

PR #825, opened 2021-07-12
[FAL-2137] Ocim: don't run periodic builds for archived instances

ChrisChV:
"Is archived" filter added in launch_periodic_builds function

**Jira ticket:** [FAL-2137](https://tasks.opencraft.com/browse/FAL-2137)

- On test_tasks.py of periodic_build, I added a filter to avoid get the archived instances
- Test updated to demonstrate that change in the function behavior

**How to test:**

- Create a test instance with the periodic build enabled.
- Test if the periodic build works normally.
- Archive the instance.
- Test if the periodic build is no longer performed.

### DIFF
--- a/periodic_builds/tasks.py
+++ b/periodic_builds/tasks.py
@@ -47,7 +47,10 @@ def launch_periodic_builds():
     """
     Automatically deploy new servers for all Open edX instances configured for periodic builds.
     """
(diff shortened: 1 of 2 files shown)


swalladge (approved):
@ChrisChV :+1:

- [x] I tested this: verified the instances filter works as expected
- [x] I read through the code
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding

PR #827, opened 2021-07-15
[FAL-2123] Ocim Add playbook and block structure caching configuration

alfredchavez:
Added a course block structure caching configuration, and a playbook to run the commands to enable it(specified here https://tasks.opencraft.com/browse/FAL-1918)
The think that the playbook could be run on appserver spawn, as the commands that are executed willl not affect in a bad way the server if they are runned on each appserver deploy.

**Jira ticket**: [FAL-2123](https://tasks.opencraft.com/browse/FAL-2123)
- On openedx_config I added the EDXAPP_BLOCK_STRUCTURES_SETTINGS required for the caching
- Created a new playbook enable_course_block_structure_caching, to execute the commands established on [FAL-1918](https://tasks.opencraft.com/browse/FAL-1918)
- The playbook is added to `get_playbooks` method in `openedx_appserver` model, si it is called on the provisioning

**How to test**
- Once deployed on ocim stage, we should redeploy the instances
- Check that the S3 buckets related to the redeployed instances have been populated
- Check caching is working for the Course Outline pages of the existing courses


instance/models/mixins/openedx_config.py
```diff
@@ -66,6 +66,29 @@ def _get_configuration_variables(self):
...
+                # DEPRECATED: this should be a default behavior in Lilac, along with the Waffle Switches mentioned in
+                # SE-4088. Delete previous file versions.
+                "PRUNING_ACTIVE": True,
+                # S3 tiering cache for speeding up course-outline
+                "STORAGE_CLASS": "{{ EDXAPP_DEFAULT_FILE_STORAGE }}",
+                "STORAGE_KWARGS": {
+                    "AWS_ACCESS_KEY_ID": "{{ EDXAPP_AWS_ACCESS_KEY_ID }}",
+                    "AWS_SECRET_ACCESS_KEY": "{{ EDXAPP_AWS_SECRET_ACCESS_KEY }}",
+                    "AWS_STORAGE_BUCKET_NAME": "{{ EDXAPP_AWS_STORAGE_BUCKET_NAME }}",
+                    "DIRECTORY_PREFIX": "courses/"
```
swalladge, line 88: @alfredchavez , `DIRECTORY_PREFIX` is a key on `EDXAPP_BLOCK_STRUCTURES_SETTINGS`, not `STORAGE_KWARGS`.

swalladge (changes requested):
Just one tweak (see comment about the config structure), then I reckon we can test this on stage.  Nice work!

@mtyaka what do you think about running this playbook on every deploy?  @alfredchavez says it will be fine, and it does simplify things (only requires a single playbook, instead of a once-off/temporary one + a permanent one).   Unless we can upstream any of the playbook to edx/configuration?

mtyaka:
> what do you think about running this playbook on every deploy? @alfredchavez says it will be fine, and it does simplify things (only requires a single playbook, instead of a once-off/temporary one + a permanent one). Unless we can upstream any of the playbook to edx/configuration?

I would prefer to only run it on first deploy unless it makes things too complicated (I just find it a bit cleaner, although I agree running it on every deploy wouldn't really be much of a problem in practice).

Running it on every deploy would simplify the setting up block structure caching on existing instances, but other than that I don't think it would simplify things? If that's the case, we could temporarily deploy the version that runs on every appserver deployment, re-deploy all existing instances, and after all existing instances are successfully redeployed with block structure caching enabled, deploy an update to Ocim to make the playbook only run on first deploy.

instance/models/mixins/openedx_config.py
alfredchavez: @swalladge ok, corrected it

swalladge: @alfredchavez awesome, looks good to test on stage now. :+1:

swalladge (approved):
@alfredchavez :+1:

- [x] I tested this: discussed on the ticket, checked the sandbox instance
- [x] I read through the code
- [x] Includes documentation

PR #828, opened 2021-07-18
[BB-4408] Allow creating instances with domain like "subdomain.sandbox.opencraft.hosting"

dliberat:
This PR addresses [BB-4408](https://tasks.opencraft.com/browse/BB-4408).

**Issues addressed in this PR:**
Provisioning sandbox servers with domain names of the form `subdomain.sandbox.opencraft.hosting` is a common use case for developers and in automated sandbox builds. However, PR #737 introduced validation that renders this impossible. 

**Affected components:**
This PR affects the instance factory methods `instance_factory` and `production_instance_factory`.

**Test instructions:**
Create `OpenEdXInstance`s as describes [here](https://ocim.opencraft.com/en/latest/provisioning-sandboxes/#manual-provisioning).

1. Create an instance using `instance_factory` with an arbitrary subdomain. This should succeed.
2. Create an instance using `production_instance_factory` with uppercase letters in the subdomain. This should succeed, but the subdomain should be coerced to lowercase.
3. Create an instance using `production_instance_factory` that has the same subdomain as in step 2, but all lowercase. This should fail due to another instance already having the same subdomain.


instance/factories.py
```diff
@@ -168,6 +168,25 @@ def production_instance_factory(**kwargs):
...
+    once created, and then performs a case-insensitive check to see if
+    any other instances have the same domain name.
+    """
+    if 'internal_lms_domain' in instance_kwargs:
+        internal_lms_domain = instance_kwargs['internal_lms_domain']
+    else:
+        sub_domain = instance_kwargs['sub_domain']
+        internal_lms_domain = generate_internal_lms_domain(sub_domain)
+
+    count = OpenEdXInstance.objects.filter(internal_lms_domain__iexact=internal_lms_domain).count()
```
xitij2000, line 191: You can just use:
```
    OpenEdXInstance.objects.filter(internal_lms_domain__iexact=internal_lms_domain).exists()
```

dliberat: Removed in #8be0f2c94009d88c2ba1cf4adacbce7c335a73db

instance/factories.py
```diff
@@ -129,7 +116,9 @@ def production_instance_factory(**kwargs):
 
     # Ensure caller provided required arguments
     assert "sub_domain" in kwargs
-    assert is_valid_domain_name(kwargs.get('sub_domain'))
```
xitij2000, line 132: Sorry for the confusion. Please don't remove this check from the **production** instance factory. Just the normal one used for sandboxes.

xitij2000 (approved):
I think the general approach is fine, but I think I've probably been unclear in explaining what was needed. Sorry about that. 

We don't want or need to change the `produciton_instance_factory`, but just `instance_factory`, which is for sandbox instances. 

So please undo the change to `production_instance_factory` and remove assert from `instance_factory`. You can add the `.lower` subdomain name code to both.

instance/factories.py
dliberat: If we keep this check, then converting to lowercase will be redundant (an `AssertionError` will be raised if any uppercase letters appear in the subdomain). I feel like it could be removed since it's an unnecessary line of code, but there's an argument to be made that it makes our intent clear. You prefer that I keep the `.lower()` line in there?

xitij2000 (approved):
:+1:
- [x] I tested this: tested on stage
- [x] I read through the code
- [na] I checked for accessibility issues
- [na] Includes documentation

PR #830, opened 2021-07-20
[BB-4447] Handle 413 Entity Too Large responses when uploading images

kewne:
This adds handling for 413 Entity Too Large responses when uploading images.

# Related Tickets:
* [BB-4447](https://tasks.opencraft.com/browse/BB-4447)

# Testing

1. Apply the following configuration to NGINX (required to avoid CORS errors):
```
<html><head></head><body>

# internal redirect when request exceeds max_client_body_size
--
error_page 413 = @entity_too_large;
 
# Add CORS headers, we might be able to trim this
location @entity_too_large {
add_header 'Access-Control-Allow-Origin' $http_origin always;
add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS' always;
add_header 'Access-Control-Allow-Headers' 'Authorization,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range' always;
add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range' always;
return 413;
}
 

</body></html>
```
2. Upload a large image (> 1 MB) using any of the console's configuration forms (e.g. `/console/theming/logos`)
3. *Image size is too large* error message should display next to control.

### DIFF
--- a/documentation/development/console.md
+++ b/documentation/development/console.md
@@ -1,0 +1,17 @@
+# Developing the console frontend
+
+## Using a different backend
+
+If you only want to make changes to the console frontend
+(or don't have credentials for running a functional backend),
+you can point your devserver at a deployed environment (usually `stage`):
+
+    REACT_APP_OCIM_API_BASE=https://stage.manage.opencraft.com npm start
+
+However, if you then try to access the frontend at `localhost`, you'll
+run into CORS issues.
+To avoid those, add an entry to your `/etc/hosts` file:
+
+    127.0.0.1 localhost.opencraft.com
+
+You can then access the devserver at http://localhost.opencraft.com:3000.
--- a/frontend/src/console/actions.ts
+++ b/frontend/src/console/actions.ts
@@ -16,6 +16,16 @@ import {
 import { NOTIFICATIONS_LIMIT, ROUTES } from 'global/constants';
 import { sanitizeErrorFeedback } from 'utils/string_utils';
 
+import { defineMessages, IntlShape } from 'react-intl';
+
+const messages = defineMessages({
+  imageTooLarge: {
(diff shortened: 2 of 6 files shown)


opencraft/settings.py
```diff
@@ -951,6 +951,7 @@
 # CORS Settings - https://github.com/adamchainz/django-cors-headers
 CORS_ORIGIN_REGEX_WHITELIST = [
     r"(http|https)://(.*).opencraft.com",
+    r'http://localhost:3000'
```
kewne, line 954: I added this to allow running a local console but there should be a better way, maybe we should try parsing from a JSON env var?

0x29a: @kewne, I can see this below:
```python
if DEBUG:
    CORS_ORIGIN_ALLOW_ALL = True
```

Shouldn't it make local console work if `DEBUG` is `True`?

kewne: @0x29a I did see that but wasn't sure if we wanted to enable it in `stage` as it might enable insecure features.

0x29a: @kewne oh, I thought you were facing CORS issues with local Ocim. I think you can use local frontend with staging this way:
1. Add `127.0.0.1 local.opencraft.com` to `/etc/hosts`.
2. Run frontend this way: `HOST=local.opencraft.com PORT=80 npm start`.

What do you think?

kewne: @0x29a This works great, I've reverted my change and added instructions for this to the documentation.

0x29a (approved):
:+1:

- [x] I've tested, that when the backend is responding with 413 status code, we see an explicit error: 
image
- [x] Includes documentation

PR #832, opened 2021-07-29
Fix statistics playbook.

mtyaka:
This updates the settings to access the Elasticsearch API on the logs server. The Elasticsearch configuration changed recently and we need to update the credentials we use to access it from Ocim when running the `instance_statistics_csv` command.

**Test instructions**:

1. SSH into `stage.ocim.net.opencraft.hosting`.
2. Switch to the `www-data` user, activate the Ocim virtual environment and go to the `/var/www/opencraft` folder (or just attach yourself to the ocim screen session).
3. Run the instance statistics command (example: `make manage -- "instance_statistics_csv --domains=campus-xblocks-test.stage.opencraft.hosting --start-date=2021-06-01 --end-date=2021-06-30"`).
4. Verify that the script completes without any failures (note you might see some N/A in the final result table, but that should be because we don't have much data on stage, not because we weren't able to access elasticsearch).


ChrisChV:
@mtyaka Looks good :+1: 

- [X] I tested the command in `stage` multiple times with different values
- [X] I read through the code.

ChrisChV (approved):

mtyaka:
Thanks @ChrisChV!

playbooks/collect_instance_statistics/collect_elasticsearch_data.yml
```diff
@@ -19,6 +19,11 @@
         src: logs_server_requirements.txt
         dest: /tmp/logs_server_requirements.txt
 
+    - name: Copy Elasticsearch CA certificate
```
lgp171188, line 22: @mtyaka, see [this comment](https://github.com/open-craft/ansible-secrets/pull/295/files#r685361497).

playbooks/collect_instance_statistics/query_elasticsearch.py
```diff
@@ -45,14 +45,52 @@ def valid_date(s):
         help='The last day on which statistics should be gathered.',
         required=True
     )
+    parser.add_argument(
+        '--host',
+        default='localhost',
```
lgp171188, line 50: @mtyaka, since the host is configurable here, and we now have SSL and proper authentication, can't we reconfigure Elasticsearch to make its API available on the public IP address of the server? If the ELK VM and the Ocim prod VM are in the same OVH account and region, we can even restrict the access via a firewall rule.

CC @giovannicimolin @toxinu as the devops specialists.

mtyaka: Yes, if Elasticsearch was accessible directly from the Ocim VM, this script would be much cleaner.

I didn't feel confident doing that in this ticket since I don't know much about our Elasticsearch setup and I'm not sure what security implications opening access to it would have, but if we can restrict access via firewall, and if @giovannicimolin and @toxinu agree, I can create a follow-up ticket to do that.

PR #834, opened 2021-08-18
feat: [BB-4532] Add new column to report containing external lms domain, and change default report recipient

alfredchavez:
### Description:
This PR adds a new column to the report generated by instance_statistics_csv, which is called by `send_all_instances_report` to generate a report on all instances and then send it to the recipients defined by
`settings.TRIAL_INSTANCES_REPORT_RECIPIENTS`.

Also this PR changes the default recipient for this report  to `contact@opencraft.com`

### Jira Issue: 
https://tasks.opencraft.com/browse/BB-4532

### Test Steps:
- Deploy this branch on stage
- Set external_lms_domain in some instances
- Run `instance_statistics_csv` command, and check that the values in the generated `csv` are correct

### DIFF
--- a/instance/management/commands/instance_statistics_csv.py
+++ b/instance/management/commands/instance_statistics_csv.py
@@ -161,6 +161,7 @@ class Command(BaseCommand):
             name_prefix = appserver.server_name_prefix
             domain = instance.domain
             ref_name = instance.ref.name
+            external_lms_domain = instance.external_lms_domain or 'N/A'
 
             try:
                 betatestapplication = instance.betatestapplication_set.get()
@@ -177,7 +178,8 @@ class Command(BaseCommand):
                 'public_ip': appserver.server.public_ip,
                 'instance_age': datetime.now(instance.created.tzinfo) - instance.created,
                 'email': email,
-                'status': status
+                'status': status,
+                'external_lms_domain': external_lms_domain,
             }
 
         return instance_metadata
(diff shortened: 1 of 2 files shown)


instance/management/commands/instance_statistics_csv.py
```diff
@@ -161,6 +161,7 @@ def get_instance_metadata(self, instances):
             name_prefix = appserver.server_name_prefix
             domain = instance.domain
             ref_name = instance.ref.name
+            external_lms_domain = instance.external_lms_domain
```
lgp171188, line 164: @alfredchavez, instead of using the `''` which is the default value for this model field, can you use something like `'N/A'` like it is done for the other fields?

lgp171188 (approved):
@alfredchavez, :+1: You can merge (use the `Squash and merge` button in GitHub if there are multiple commits in the PR after adding the fix) this PR after addressing the open review comment and the CI checks pass.

- [x] I tested this and verified that a new field for the external domain is added to the report and its value is present in the report when an instance has the domain configured.
- [x] I read through the code
- [x] ~~Includes documentation~~

instance/management/commands/instance_statistics_csv.py
```diff
@@ -161,6 +161,7 @@ def get_instance_metadata(self, instances):
             name_prefix = appserver.server_name_prefix
             domain = instance.domain
             ref_name = instance.ref.name
+            external_lms_domain = instance.external_lms_domain if instance.external_lms_domain else 'N/A'
```
lgp171188, line 164:
```suggestion
            external_lms_domain = instance.external_lms_domain or 'N/A'
```

PR #838, opened 2021-08-26
[BB-4653] feat: implement a page in Ocim to enter the conversion data to send it to Matomo

lgp171188:
This PR implements the changes to add a page in Ocim, using which the superusers can enter the conversion data for the active betatest instances to be sent to Matomo.

This also adds a fix for an issue breaking the `make run.dev` command on the vagrant devstack.

**Testing instructions**:
* On a working Ocim devstack, checkout the source branch of this PR.
* Configure the `MATOMO_URL`, `MATOMO_SITE_ID`, `MATOMO_SITE_ID`, `MATOMO_CONVERSION_GOAL_ID` variables with appropriate values in the `.env` file.
* Try accessing `/marketing/conversion/` and verify that it requires logging in.
* Also verify that the logged-in user has to be a superuser to access the page.
* Log in as a superuser user and open `/marketing/conversion/`.
* Verify that the page renders properly with a form containing fields for instance, revenue and custom matomo tracking data.
* Verify that the queryset for the instance field is restricted to instances which have an associated betatest application and has been successfully provisioned. In the devstack, this can be done by creating a betatest application, creating an associated Open edX instance, and manually setting its `successfully_provisioned` field to `True`.
* Verify that the valid data submitted through this page is captured and available in Matomo.
* Try any other tests to try and break the functionality implemented in this PR. :)

### DIFF
--- a/marketing/forms.py
+++ b/marketing/forms.py
@@ -19,8 +19,10 @@
 """
 Forms for the `marketing` app
 """
+from decimal import Decimal
+
 from django.contrib.postgres.forms.jsonb import JSONField
-from django.forms import DecimalField, Form, ModelChoiceField
+from django.forms import DecimalField, Form, ModelChoiceField, ValidationError
 
 from instance.models.openedx_instance import OpenEdXInstance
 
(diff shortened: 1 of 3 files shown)


alfredchavez:
Hi @lgp171188 it looks good, and I have some questions about some things I've seen while testing this PR:
- In the custom matomo tracking data field:
    - The default value is null(writable as text, it seems confusing), I think maybe the default value should be an empty json(`{}`)
    - It accepts simple numbers(integers and floats) as valid json
- In the Revenue field:
   - Not sure if we should allow negative values as revenue, 

Apart from that, I think that after I get the credentials and see that data is being send correctly to matomo, it should be ready to get merged

lgp171188:
@alfredchavez,

> The default value is null(writable as text, it seems confusing), I think maybe the default value should be an empty json({})

I will check and try to fix this.

> It is accepting simple numbers(integers and floats) as valid json

These are valid JSON values and that is why the field is accepting it. However, the code in `views.py` is checking whether the submitted JSON data is a Python dictionary and only then adds it to the request data. I will try to move this to the form validation as well.

> Not sure if we should allow negative values as revenue

I didn't restrict it to just zero and positive values because there could be some loss from a customer. :)

lgp171188:
@alfredchavez,

> Hi @lgp171188 it looks good, and I have some questions/suggestions about some things I've seen while testing this PR:

Btw, it will be better to leave the comments in the relevant lines of the code so that it is easy to track the changes made for addressing the comments.

alfredchavez:
@lgp171188 I am rerunning the integration tests group that have failed, and I am approving this PR after it passes the checks 👍🏽

alfredchavez (approved):

PR #839, opened 2021-08-31
FAL-2270 Make production extra configuration configurable from config

samuelallan72:
[FAL-2270](https://tasks.opencraft.com/browse/FAL-2270)

This is so we don't need to update ocim code when the required configuration changes.  

The motivation is that we needed to update the config to add the sysadmin app, and thought why not make the whole config configurable instead of hardcoded.

**Test instructions**:

- deploy https://github.com/open-craft/ansible-secrets/pull/310
- launch a production instance using the prod instance factory
- verify that the sysadmin app is installed

**Reviewers**:

- [x] @pomegranited

### DIFF
--- a/instance/factories.py
+++ b/instance/factories.py
@@ -140,21 +140,16 @@ def production_instance_factory(**kwargs):
         logger.warning("Environment not ready. Please fix the problems above, then try again. Aborting.")
         return None
 
-    # Gather settings
-    production_settings = {
-        # Don't create default users on production instances
-        "DEMO_CREATE_STAFF_USER": False,
-        "demo_test_users": [],
-        # Disable certificates process to reduce load on RabbitMQ and MySQL
-        "SANDBOX_ENABLE_CERTIFICATES": False,
-    }
     configuration_extra_settings = kwargs.pop("configuration_extra_settings", "")
     if configuration_extra_settings:
         configuration_extra_settings = yaml.load(configuration_extra_settings, Loader=yaml.SafeLoader)
     else:
         configuration_extra_settings = {}
     extra_settings = yaml.dump(
-        ansible.dict_merge(production_settings, configuration_extra_settings),
+        ansible.dict_merge(
+            settings.PRODUCTION_INSTANCE_EXTRA_CONFIGURATION,
+            configuration_extra_settings
+        ),
         default_flow_style=False
     )
     instance_kwargs = dict(
--- a/instance/tests/test_factories.py
+++ b/instance/tests/test_factories.py
@@ -41,13 +41,6 @@ class FactoriesTestCase(TestCase):
     Test cases for functions in the factories module
     """
 
-    CONFIGURATION_EXTRA_SETTINGS = (
-        "{"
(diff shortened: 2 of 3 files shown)


instance/factories.py
```diff
@@ -147,6 +147,9 @@ def production_instance_factory(**kwargs):
         "demo_test_users": [],
         # Disable certificates process to reduce load on RabbitMQ and MySQL
         "SANDBOX_ENABLE_CERTIFICATES": False,
+        "EDXAPP_EXTRA_REQUIREMENTS": [
+            { "name": "git+https://github.com/mitodl/edx-sysadmin.git@v0.1.3#egg=edx-sysadmin==v0.1.3" },
```
pomegranited, line 151: Can you add an env variable for the version, something like `EDXAPP_EDX_SYSADMIN_VERSION`? Then we can update it when we upgrade to maple etc., without having to update the code.

pomegranited (changes requested):

instance/factories.py
swalladge: @pomegranited do you mean a variable for ocim, or a variable to go in the ansible vars?

pomegranited: Hmm.. I was thinking a variable for Ocim, but maybe it would be better to move this whole block into config, and read it into the Ocim settings?

```yaml
PRODUCTION_INSTANCE_EXTRA_CONFIGURATION:
  # Don't create default users on production instances
  DEMO_CREATE_STAFF_USER: false
  demo_test_users: [],
  # Disable certificates process to reduce load on RabbitMQ and MySQL
  SANDBOX_ENABLE_CERTIFICATES: false
  EDXAPP_EXTRA_REQUIREMENTS
  - name: git+https://github.com/mitodl/edx-sysadmin.git@v0.1.3#egg=edx-sysadmin==v0.1.3 
```

So you can just do this:
```
production_settings = settings.PRODUCTION_INSTANCE_EXTRA_CONFIGURATION
```

That way, if/when we have other apps that we need to install on production instances, we can just add them to the configuration, including any extra env variables or whatever settings they need.

swalladge: Nice :ok_hand:

pomegranited (approved):
:+1: -- though please fix the failing unit tests before merging.

- [x] I tested this by following the https://github.com/open-craft/ansible-secrets/pull/310's instructions to test this change on ocim stage.
   Created https://stage.manage.opencraft.com/instance/8490/, which configured as expected.
- [x] I read through the code
- [x] <strike>Includes documentation</strike> N/A
- [x] <strike>I made sure any change in configuration variables is reflected in the corresponding
    client's `configuration-secure` repository.</strike> N/A

swalladge:
Note: do not merge until [open-craft/ansible-secrets#312|https://github.com/open-craft/ansible-secrets/pull/312] is approved

swalladge:
Merging because the failing integration test is unrelated.  See [FAL-2409](https://tasks.opencraft.com/browse/FAL-2409)

PR #848, opened 2021-10-28
feat: Deprovision the s3 buckets of specific archived ocim instances

joidegn:
## Description

Provide a management command to deprovision S3 buckets. Can either be run without arguments and it will deprovision S3 buckets of sandbox instances or can be run with a space separated list of instance_ref_ids which limits the operation to these ids but it checks first if the instances have been archived more than 3 months ago.

[SE-4912](https://tasks.opencraft.com/browse/SE-4912)

## Supporting information

[Forum post](https://forum.opencraft.com/t/deprovisioning-s3-buckets-of-archived-instances-should-we-keep-for-any-clients/1154)

## Testing instructions
- [ ] Create instance with S3 bucket, archive instance and run management task with instance_ref_ids set
- [ ] Run management command without argument to deprovision S3 buckets of sandbox instances. Confirm successfully deprovisioned buckets from logs.

## Deadline
Before buckets run out

### DIFF
--- a/instance/management/commands/deprovision_buckets.py
+++ b/instance/management/commands/deprovision_buckets.py
@@ -43,11 +43,20 @@ class Command(BaseCommand):
         'Deprovision S3 buckets for archived instances'
     )
 
+    @staticmethod
+    def _check_positive_int(arg):
+        """
+        Ensure that num-days-archived is a positive (integer) number
+        """
+        arg = int(arg)
+        if arg < 0:
+            raise ValueError("the value of num_days_archived must be greater or equal to zero")
+
     def add_arguments(self, parser):
         parser.add_argument(
             '--instance-ref-ids',
             nargs='*',
-            type=int,
+            type=self._check_positive_int,
             help='list of instance ref ids to deprovision for'
         )
         parser.add_argument(
--- a/instance/management/commands/deprovision_buckets.py
+++ b/instance/management/commands/deprovision_buckets.py
@@ -52,7 +52,6 @@ class Command(BaseCommand):
             type=int,
             help='list of instance ref ids to deprovision for'
         )
-        # TODO: add argument to change number of months to retain archived instances for
         parser.add_argument(
             '--num_days_archived',
             nargs='?',
--- a/instance/management/commands/deprovision_buckets.py
+++ b/instance/management/commands/deprovision_buckets.py
@@ -17,8 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 """
-This management command will reprovision the S3 bucket for each available
-instance to have its configuration updated on AWS.
+This management command will deprovision the S3 bucket for archived instances
 """
 
 from datetime import timedelta
--- a/instance/management/commands/deprovision_buckets.py
+++ b/instance/management/commands/deprovision_buckets.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2019 OpenCraft <contact@opencraft.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This management command will reprovision the S3 bucket for each available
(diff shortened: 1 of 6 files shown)


instance/management/commands/deprovision_buckets.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def handle(self, *args, **options):
+        """
+        Deprovisions the S3 bucket for the instance provided by
+        instance_ref_ids or if instance_ref_ids is not given for all sandbox
+        instances
+        """
+
+        instance_ref_ids = options['instance_ref_ids']
+        num_days_archived = options['num_days_archived']
```
gabor-boros, line 71: What happens if `num_days_archived` is less than `0` (zero)? In that case, if `timedelta` is not raising an exception, the condition in line 90 won't evaluate as expected.

How about introducing the following?

```suggestion
        num_days_archived = options['num_days_archived']
        
        if num_days_archived < 0:
            raise ValueError("the value of num_days_archived must be greater or equal to zero")
```

instance/management/commands/deprovision_buckets.py
```diff
@@ -0,0 +1,95 @@
...
+    It accepts a list of instance ref ids in case buckets ofnon-sandbox
+    instances should get archived.
+    """
+    help = (
+        'Deprovision S3 buckets for archived instances'
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--instance_ref_ids',
```
gabor-boros, line 50: To be a bit more POSIX compliant, how about using dashes instead of underscores in the flag name?

```suggestion
            '--instance-ref-ids',
            dest='instance_ref_ids',
```

instance/management/commands/deprovision_buckets.py
```diff
@@ -0,0 +1,95 @@
...
+            nargs='*',
+            type=int,
+            help='list of instance ref ids to deprovision for'
+        )
+        parser.add_argument(
+            '--num_days_archived',
```
gabor-boros, line 56: Same question.

```suggestion
            '--num-days-archived',
            dest='num_days_archived',
```

instance/models/openedx_instance.py
```diff
@@ -281,6 +282,19 @@ def first_activated(self):
         except models.ObjectDoesNotExist:
             return None
 
+    @property
+    def most_recently_archived(self):
+        """
+        Returns the date the instance has been most recently archived
```
gabor-boros, line 288:
```suggestion
        Returns the date the instance has been most recently archived.
```

instance/models/openedx_instance.py
```diff
@@ -281,6 +282,19 @@ def first_activated(self):
...
+        :return: Union[None, datetime]
+        """
+        relevant_log_regex = r'Archiving instance finished.'
```
gabor-boros, line 291: Is the raw-string indicator needed?

```suggestion
        relevant_log_regex = 'Archiving instance finished.'
```

instance/models/openedx_instance.py
```diff
@@ -281,6 +282,19 @@ def first_activated(self):
         except models.ObjectDoesNotExist:
             return None
 
+    @property
+    def most_recently_archived(self):
```
gabor-boros, line 286: I think the naming of this property is not the most fortunate. How about renaming it to something like `latest_archiving_date`?

```suggestion
    def latest_archiving_date(self):
```

if you agree, please make sure to update the management command as well.

gabor-boros (changes requested):
Please fix the quality issues indicated by https://app.circleci.com/pipelines/github/open-craft/opencraft/4374/workflows/766d29e2-4420-4153-957b-8c44a5201807/jobs/38800

Also, rebasing the branch will be needed.

instance/management/commands/deprovision_buckets.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            archived_instances = OpenEdXInstance.objects.filter(
+                ref_set__is_archived=True,
+                s3_bucket_name__isnull=False,
+                ref_set__id__in=instance_ref_ids
+            )
+        archived_instances = [
+            i for i in archived_instances if
+                i.most_recently_archived
+                and i.most_recently_archived + timedelta(days=num_days_archived) < timezone.now()
```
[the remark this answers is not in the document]
joidegn: No but it should not have to have the same zoneinfo. As long as both are timezone aware datetimes Python should know how to do the comparison.

instance/models/openedx_instance.py
joidegn: I personally don't love it when names are unspecific with regards to types (i.e. "date" vs "datetime") but I am following your suggestion because you know the teams naming conventions.

instance/models/openedx_instance.py
joidegn: No but since people commonly use raw strings for regexes to deal with escaping I like to do that even if it is not strictly speaking needed. That way if someone later on changes the regex string they will not need to change the type.

gabor-boros:
👍  🎉  The script is tested in Staging, so there is no need for an additional test.

- [x] I read through the code
- ~~I checked for accessibility issues~~
- ~~Includes documentation~~
- ~~I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

PR #850, opened 2021-11-10
[BB-5077] fix: End of trial email sent to paying customer

kaustavb12:
## Description

This PR is fix for a bug identified in OCIM application, where a paying client would get an end of trial reminder email after they launched a new appserver

## Supporting information

[BB-5077](https://tasks.opencraft.com/browse/BB-5077)

## Testing instructions

1. Register a new trial instance in OCIM
2. Once the first app server is ready, delete the corresponding entry created in Subscribers table from the Django admin page
3. Launch a new appserver from console
4. After the new appserver is up and ready, check that there should be no new entry in the Subscribers table

## Deadline

"None"

### DIFF
--- a/marketing/signals.py
+++ b/marketing/signals.py
@@ -35,5 +35,4 @@ def register_subscriber(sender, **kwargs):
 
     application = kwargs['application']
 
-
     Subscriber.objects.get_or_create(user_id=application.user.id)
--- a/marketing/signals.py
+++ b/marketing/signals.py
@@ -38,5 +38,5 @@ def register_subscriber(sender, **kwargs):
     # Ignore if the appserver was not spawned for a beta tester.
     if not application:
         return
-    
+
     Subscriber.objects.get_or_create(user_id=application.user.id)
--- a/marketing/signals.py
+++ b/marketing/signals.py
@@ -21,27 +21,22 @@ Signals related to marketing.
 """
 from django.dispatch import receiver
 
-from instance.signals import appserver_spawned
+from registration.signals import betapplication_accepted
 from marketing.models import Subscriber
 
 
-@receiver(appserver_spawned)
+@receiver(betapplication_accepted)
 def register_subscriber(sender, **kwargs):
     """
     Registers a marketing email subscriber for Beta testers
     on first successful appserver spawn. Only registers if user is not
     already registered as subscriber.
     """
-    instance = kwargs['instance']
-    appserver = kwargs['appserver']
-    application = instance.betatestapplication_set.first()  # There should only be one
+
+    application = kwargs['application']
 
     # Ignore if the appserver was not spawned for a beta tester.
     if not application:
         return
-    # Ignore if appserver didn't provision
-    elif appserver is None:
-        return
-    else:
-        # Register the user as followup email subscriber
-        Subscriber.objects.get_or_create(user_id=application.user.id)
+    
+    Subscriber.objects.get_or_create(user_id=application.user.id)
--- a/marketing/signals.py
+++ b/marketing/signals.py
@@ -35,8 +35,5 @@ def register_subscriber(sender, **kwargs):
 
     application = kwargs['application']
 
-    # Ignore if the appserver was not spawned for a beta tester.
-    if not application:
-        return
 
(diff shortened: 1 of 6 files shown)


marketing/signals.py
```diff
@@ -21,27 +21,22 @@
...
     """
-    instance = kwargs['instance']
-    appserver = kwargs['appserver']
-    application = instance.betatestapplication_set.first()  # There should only be one
+
+    application = kwargs['application']
 
     # Ignore if the appserver was not spawned for a beta tester.
     if not application:
         return
```
giovannicimolin, line 40: This signal is only triggered if there's a betatest application, so we don't need to re-check this here.
Ref: https://github.com/open-craft/opencraft/pull/850/files#diff-6137d8550b33801d66d4ffcb337e4d2bc3bdb9f1fe2c48911f8589fe277517c5L72
```suggestion
```

marketing/signals.py
```diff
@@ -21,27 +21,22 @@
 """
 from django.dispatch import receiver
 
-from instance.signals import appserver_spawned
+from registration.signals import betapplication_accepted
```
giovannicimolin, line 24: Rename the signal `betatestpplication_accepted` for consistency with model names and other references in the code.
```suggestion
from registration.signals import betatestpplication_accepted
```

registration/approval.py
```diff
@@ -56,6 +57,8 @@ def accept_application(application, appserver):
 
     send_welcome_email(application)
 
+    betapplication_accepted.send(sender=None, application=application)
+
     application.status = BetaTestApplication.ACCEPTED
     application.save()
```
giovannicimolin, line 63: To avoid having issues with receivers that can possibly break and raise issues either move the signal emission to the end of the function or use `send_robust` (I prefer the first option, it doesn't risk not saving the application, but propagates errors to make debugging easy).

```suggestion
    application.status = BetaTestApplication.ACCEPTED
    application.save()
    
    # Send signal after saving application to avoid potential issues in receivers
    betapplication_accepted.send(sender=None, application=application)
```

registration/tests/test_approval.py
```diff
@@ -126,9 +128,21 @@ def test_first_appserver_is_active(self):
...
+
+        UserProfile.objects.create(
+            user=user,
+            full_name="Test user 1",
+            accepted_privacy_policy=datetime.now(),
+            accept_domain_condition=True,
+            subscribe_to_updates=True,
+        )
+
+        user.refresh_from_db()
```
giovannicimolin, line 141: This code is being re-used 3 times. There's two options:
1. Can you make a `BetaTestUserFactory` (or similar) on `test/utils.py` and use that instead?
2. Use the https://github.com/open-craft/opencraft/blob/master/registration/tests/utils.py#L181 test helper (with a few minor changes to also add the user profile).

I prefer option 1, but I want to know which one do you think is cleaner?

giovannicimolin (changes requested):
@kaustavb12 Looks good, but needs a few minor changes. See my comments.

marketing/signals.py
kaustavb12: Done.

registration/approval.py
kaustavb12: Done

registration/tests/test_approval.py
kaustavb12: I went ahead with option 1 because of three reasons:

1. UserMixin is currently being used in test cases which deal exclusively with tests involving user model (ex. user login, authentication, etc.). In our case user model is not our main test subject, it is just needed to run the code.
2. Using BetaTestUserFactory would be easier to read and follow for a developer without any context at a later stage and can be reused more flexibly in future test cases.
3. The exact same User/UserProfile creation logic seems to be pretty useful as it is in use across multiple test cases throughout the application. The BetaTestUserFactory can be potentialy be used to easily eliminate redundant code in existing test cases.

giovannicimolin (approved):
:+1:

- [x] I tested this: 
1. CI checks are green (my devstack is broken - but the tests make sense and are passing)
- [x] I read through the code

@kaustavb12 Good to go!

PR #851, opened 2021-11-11
[SE-5006] fix: deprovision s3 buckets with better filtering

gabor-boros:
## Description

When deprovisioning S3 bucket for an instance, we need to check for both
the instance and bucket names. In case of any of them meet the
conditions, we must deprovision the related bucket.

## Supporting information

Related PRs:

* https://github.com/open-craft/opencraft/pull/848

### Dependencies

N/A

## Testing instructions

* Run the command on stage and prod using the `--dry-run` flag
* Check for the output for any suspicious instance name
* Run the commands without the `--dry-run` flag
* Check that the AWS buckets are removed

## Deadline

ASAP

## Other information

A new `--dry-run` option is added beside the fix to ensure we can run the command without any harm.

### DIFF
--- a/instance/management/commands/deprovision_buckets.py
+++ b/instance/management/commands/deprovision_buckets.py
@@ -24,11 +24,13 @@ from datetime import timedelta
 import logging
 
 from django.core.management.base import BaseCommand
+from django.db.models import Q
 from django.utils import timezone
 
 from instance.models.openedx_instance import OpenEdXInstance
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


mtyaka:
Thanks for these changes @gabor-boros, they look good :+1:
It looks like there are still some issues with the tests to be sorted out before merging.

- [x] I tested this: I ran the command on Ocim prod and verified that it found some instances for which buckets could be purged and the instances looked good (were all sandboxes).
- [x] I read through the code
- [x] Includes documentation

mtyaka (approved):

PR #852, opened 2021-11-22
[FAL-2259] koa.3 -> lilac.2

michaelbukachi:
This PR updates Ocim to use opencraft-release/lilac.2 branch for new instances and also opencraft-release/lilac.2 branch for integration tests.

**JIRA tickets**:

- [FAL-2259](https://tasks.opencraft.com/browse/FAL-2259)

**Dependencies**:

- https://github.com/open-craft/ansible-secrets/pull/312/files. Needs to be reverted

**Testing instructions**:

Ensure that PR passed all CI pipeline checks.

**Reviewers**

- [ ] @mtyaka


instance/tests/integration/factories/instance.py
```diff
@@ -79,8 +79,8 @@ def create(cls, *args, **kwargs):  # pylint: disable=arguments-differ
     # or a release candidate tag will work.  We point both the edx-platform and the configuration
     # versions to the branch "integration" in our own forks.  These branches are based on the
     # corresponding openedx_release versions from upstream, but can contain custom modifications.
-    openedx_release = 'open-release/koa.3'
+    openedx_release = 'open-release/lilac.2'
     configuration_source_repo_url = 'https://github.com/open-craft/configuration.git'
-    configuration_version = 'integration-koa.3'
+    configuration_version = 'opencraft-release/lilac.2'
```
mtyaka, line 84: This will have to be changed to `integration-lilac.2`, once https://tasks.opencraft.com/browse/BB-4996 is ready.

michaelbukachi: @mtyaka Would you consider this an external blocker for this ticket? Or can this be merged and a follow-up PR done?

mtyaka: I'm not sure how the code in the integration branch differs from the standard release branch. If integration tests work with the standard branch, then I guess we don't need to wait for the integration branch to be ready. I guess we'll see if tests on this PR work :)

instance/tests/models/test_load_balanced_mixin.py
```diff
@@ -55,13 +55,15 @@ class LoadBalancedInstanceTestCase(TestCase):
         'preview.test.dns',
         'studio.test.dns',
         'ecommerce.test.dns',
-        'discovery.test.dns'
+        'discovery.test.dns',
+        'app.test.dns'
```
mtyaka, line 59: Nit: add trailing comma

instance/tests/models/test_load_balanced_mixin.py
```diff
@@ -55,13 +55,15 @@ class LoadBalancedInstanceTestCase(TestCase):
...
     ]
     dns_domains_with_prefix_domains = dns_domains + [
         'studio-test.dns',
         'preview-test.dns',
         'discovery-test.dns',
         'ecommerce-test.dns',
+        'app-test.dns'
```
mtyaka, line 66: Nit: add trailing comma

mtyaka (commented):
Nice work @michaelbukachi! The changes look good to me.
I think the two tests are failing because we added a new prefix domain, and [this line in tests](https://github.com/open-craft/opencraft/blob/691bb2667e117e02d627e05c66ca67b79322cfa6/instance/tests/models/test_openedx_instance.py#L747) assumes that there are exactly 4 prefix domains. Try changing that to 5 and see if the tests pass. It might make sense to use something like `len(DOMAIN_PREFIXES)` instead of hardcoding that number.

After the tests are green, we'll try to deploy this to stage and spawn a few instances to see if any other adjustments need to be made to the lilac config.

michaelbukachi:
@mtyaka  Integration tests are failing. On investigating, I've found the error to be:
screenshot_20211123175021
Might this be because we are not using an `integration` branch? :thinking:

mtyaka:
@michaelbukachi Ah, yes... the `opencraft_integration.yml` is a custom playbook that is only available in the integration branch. We can consider this task blocked until https://github.com/open-craft/configuration/pull/178 is ready.

instance/models/mixins/domain_names.py
```diff
@@ -256,7 +264,7 @@ def lms_extended_heartbeat_url(self):
         """
         return u'{}heartbeat?extended'.format(self.url)
 
-    def save(self, **kwargs):  # pylint: disable=arguments-differ, too-many-branches, useless-suppression
+    def save(self, **kwargs):  # pylint: disable=arguments-differ, too-many-branches, useless-suppression; # noqa: MC0001
```
xitij2000, line 267: Just a drive-by comment, but I think we should prefer using a human-readable name here instead of MC0001

michaelbukachi: @xirdneh I'm unable to find the human-readable version of `MC0001` in the [repo](https://github.com/PyCQA/mccabe).

xitij2000: @michaelbukachi I think you pinged the wrong person there :-) 

Maybe we can break this method down to avoid the issue? Or add a short comment about what the error is and why its being suppressed. 

Like:
```suggestion
    # Suppress McCabe complexity check MC0001 temporarily to avoid refactoring right now.
    def save(self, **kwargs):  # pylint: disable=arguments-differ, too-many-branches, useless-suppression; # noqa: MC0001
```

michaelbukachi: @xitij2000 Thanks for this.

mtyaka:
@michaelbukachi It's not easy to get integration tests to run locally. What you can do instead, is to use the "Re-run with SSH" feature of CircleCI, although I'm not sure if you have the required Github permissions to do that.
Debugging Ocim integration tests is always a pain though, because they take such a long time to run.

The particular favicon related failure is very confusing, but I think I found out why the test is failing. There was a commit that touched the favicon `<link>` tag [here](https://github.com/edx/edx-platform/commit/64e26407f2b3e032f17b6806ac1f9b984fdcb8b2#diff-1adbd21958e452c19656801dc282c38d458be2b2afeecad134e77c1de607a74d). If you look closely, you'll see that there was a space before the closing `/>`, but that commit removed it, so the regex in the test is failing.

I think you'll have to revert your change from https://github.com/open-craft/opencraft/pull/852/commits/c7387b5bceed6b1af29ff7b197fcc544b9200d16 and update the regex to remove the space.

michaelbukachi:
Hello@mtyaka,
Any more feedback on this? I'm struggling with debugging :cold_sweat: . I've tried [this](https://github.com/open-craft/opencraft/pull/852/commits/e86eee033da615aeff424f35aa814479a55f57ed) and [this](https://github.com/open-craft/opencraft/pull/852/commits/b6d1ecce2f0527525488479814ec47cf3e39e1e9) but still got the same error.

mtyaka:
@michaelbukachi It fails with a different error now:

```
FAIL: test_spawn_appserver (instance.tests.integration.integration_instance.InstanceIntegrationTestCase)
  File "/usr/local/lib/python3.6/unittest/mock.py", line 1183, in patched
    return func(*args, **keywargs)
  File "/home/********/project/instance/tests/integration/integration_instance.py", line 442, in test_spawn_appserver
    self.assert_load_balanced_domains(instance)
  File "/home/********/project/instance/tests/integration/integration_instance.py", line 345, in assert_load_balanced_domains
    expected_domain_names
AssertionError: Lists differ: ['3fc[192 chars]t', 'app.3fc39a.integration.**********', 'cust[71 chars]net'] != ['3fc[192 chars]t', 'custom1.3fc39a.integration.**********', '[34 chars]net']

First differing element 5:
'app.3fc39a.integration.**********'
'custom1.3fc39a.integration.**********'

First list contains 1 additional elements.
First extra element 7:
'custom2.3fc39a.integration.**********'

  ['3fc39a.integration.**********',
   'preview.3fc39a.integration.**********',
   'studio.3fc39a.integration.**********',
   'discovery.3fc39a.integration.**********',
   'ecommerce.3fc39a.integration.**********',
-  'app.3fc39a.integration.**********',
   'custom1.3fc39a.integration.**********',
   'custom2.3fc39a.integration.**********']
```


It looks like you need to add the MFE `app` subdomain to [this hardcoded list of expected domains](https://github.com/open-craft/opencraft/blob/e01c75d251327b983f5df859970793266c416ff7/instance/tests/integration/integration_instance.py#L329-L342).

michaelbukachi:
@mtyaka Thanks for pointing this out. I had clicked the wrong status link so I was seeing results from the previous ci tests.
Anyway, this is ready for a final review.

mtyaka:
Thanks @michaelbukachi! The code looks great, I'm going to deploy it to stage and test it out.

mtyaka (approved):

PR #854, opened 2021-12-01
[BB-4996] fix favicon regex that was still failing

felipetrz:


0x29a (approved):
:+1:

- [x] I checked that tests are passing.
- [x] I read through the diff.

PR #855, opened 2021-12-03
Limit size of mongo index names

saksham115:
Ticket: [FAL-2543](https://tasks.opencraft.com/browse/FAL-2543)

We want to avoid any errors while provisioning an instance on OCIM due to long subdomains that might cause long index names to be generated in mongo and therefore break the provisioning.

In this PR we are trying to truncate the database name if it crosses a particular threshold (40 characters, found it by trial and error, refer to the above ticket).

**Testing Instructions**
- Provision a server with subdomain name more than 22 characters.
- Check if the server was provisioned successfully.


instance/models/openedx_instance.py
```diff
@@ -134,7 +134,7 @@ def database_name(self):
         # for different services (e.g. xqueue) we don't want to use the maximum length here.
         allowed = string.ascii_letters + string.digits + '_'
         escaped = ''.join(char for char in name if char in allowed)
-        return truncate_name(escaped, length=50)
+        return truncate_name(escaped, length=40)
```
mtyaka, line 137: Hmm... interesting, I didn't know we use `truncate_name` here. If we do this, do we even need to limit the subdomain name to 22 characters? Is there any other place where the length of the subdomain can be problematic, other than the database name?

saksham115: Yes, Sounds fair. I have removed the limitation on the sub-domain name.

instance/models/openedx_instance.py
```diff
@@ -77,6 +77,7 @@ class OpenEdXInstance(
     # Most settings/fields are inherited from mixins
 
     successfully_provisioned = models.BooleanField(default=False)
+    current_database_name = models.CharField(blank=True, max_length=60, default='')
```
mtyaka, line 80: Please rename the field to `database_name`, and the current `database_name` property to `generate_database_name()` (and make it a regular method instead of a property).

instance/models/openedx_instance.py
```diff
@@ -127,14 +128,22 @@ def database_name(self):
...
-        # Escape all non-ascii characters and truncate to 50 chars.
-        # The maximum length for the name of a MySQL database is 64 characters.
-        # But since we add suffixes to database_name to generate unique database names
-        # for different services (e.g. xqueue) we don't want to use the maximum length here.
-        allowed = string.ascii_letters + string.digits + '_'
-        escaped = ''.join(char for char in name if char in allowed)
-        return truncate_name(escaped, length=50)
+        current_database_name = self.current_database_name
+
+        if current_database_name is not None and current_database_name == '':
```
mtyaka, line 133: While this would work, we use a similar pattern elsewhere in the code where we set the values inside the `save()` method, see [here](https://github.com/open-craft/opencraft/blob/8b9634ff96c97e48c384353a2baec0ffe90d08d4/instance/models/openedx_instance.py#L154-L161) for example.
This also lets us set the field to `blank=False` which ensures that no new instance is created with the particular field left empty.

instance/models/openedx_instance.py
```diff
@@ -127,14 +128,22 @@ def database_name(self):
...
+            name = self.internal_lms_domain.replace('.', '_')
+            # Escape all non-ascii characters and truncate to 50 chars.
+            # The maximum length for the name of a MySQL database is 64 characters.
+            # But since we add suffixes to database_name to generate unique database names
+            # for different services (e.g. xqueue) we don't want to use the maximum length here.
+            allowed = string.ascii_letters + string.digits + '_'
+            escaped = ''.join(char for char in name if char in allowed)
+            current_database_name = truncate_name(escaped, length=40)
```
mtyaka, line 141: We'll have to do the `length=50` -> `length=40` change in a separate PR, so that we can first set the new `database_name` field to the current value (capped at 50 chars) for the existing instances.

mtyaka (changes requested):
We'll have to split this change into two PRs:

- First one to create the new database field and populate it with the current value of database_name. Once the field is populated, we can also change it to `unique=True` and `blank=False`.
- Second one to change the truncate length from `50` to `40`.
 
This will ensure that nothing changes for existing instances (even those that currently have a database_name longer than 40 characters) after the two PRs are deployed, but any instance created after that, will have their database_name capped at 40 characters.

See https://github.com/open-craft/opencraft/pull/852/files for an example of a similar migration that populates a new DB field.

saksham115:
@mtyaka made the changes by setting the base brach in the PR as `saksham115/FAL-2543_base`. Please review.

instance/models/openedx_instance.py
```diff
@@ -77,7 +77,7 @@ class OpenEdXInstance(
     # Most settings/fields are inherited from mixins
 
     successfully_provisioned = models.BooleanField(default=False)
-    database_name = models.CharField(blank=True, max_length=60, default='')
+    database_name = models.CharField(blank=False, unique=True, max_length=60, default='')
```
mtyaka, line 80: Nit: we can remove the `default=''` now that every new instance is required to have a unique database_name.

mtyaka (commented):
Code looks good! Can you deploy to stage and let me test it out?

mtyaka:
Thanks @saksham115, the code looks good!

I started deploying a couple of appservers. Once they get provisioned successfully, I'll run a few tests and if everything looks good, this will be ready to go.

mtyaka (approved):
:+1:

- [x] I tested this: I verified this works as expected on Ocim stage.
- [x] I read through the code

PR #857, opened 2021-12-24
Changes to support truncate for database name

saksham115:
Ticket: [FAL-2543](https://tasks.opencraft.com/browse/FAL-2543)

This PR is in support to the comment: https://github.com/open-craft/opencraft/pull/855#pullrequestreview-839088396 and acts as a part 1 for the PR: https://github.com/open-craft/opencraft/pull/855

Testing Instructions

1. Run the migration
2. Check database_name field is populated for all the instances


instance/models/openedx_instance.py
```diff
@@ -77,6 +77,7 @@ class OpenEdXInstance(
     # Most settings/fields are inherited from mixins
 
     successfully_provisioned = models.BooleanField(default=False)
+    database_name = models.CharField(blank=True, max_length=60, default='')
```
saksham115, line 80: will be updating this to `blank=False, unique=True` in the subsequent [PR](https://github.com/open-craft/opencraft/pull/855) as mentioned here: https://github.com/open-craft/opencraft/pull/855#pullrequestreview-839088396

instance/models/openedx_instance.py
```diff
@@ -154,6 +154,9 @@ def save(self, **kwargs):
         if self.random_prefix is not None:
             self.mysql_user = self.random_prefix
 
+        if self.database_name is None or self.database_name == '':
```
mtyaka, line 157: nit: you could shorten this to `if not self.database_name:`

instance/models/mixins/openedx_database.py
```diff
@@ -70,7 +70,7 @@ def mysql_database_name(self):
         """
         The mysql database name for this instance
         """
-        return self.database_name
+        return self.generate_database_name()
```
mtyaka, line 73: The only place we call `generate_database_name()` should be in OpenEdXInstance's `save()` method where we initially set the value of the `database_name` field. Everywhere else, we should keep referencing `self.database_name`.

That will ensure nothing will change for existing instances after we change the details of `generate_database_name` in the follow-up PR.

mtyaka (changes requested):
Thanks for this @saksham115!

The migration and the updated `save()` method looks good, but we shouldn't replace `self.database_name` with `self.generate_database_name()` elsewhere in the code -- that will ensure that nothing changes for existing instances after we deploy the follow-up PR which will change the implementation of `generate_database_name()`, since they'll keep referencing the DB-stored value of `self.database_name`.

instance/models/openedx_instance.py
saksham115: Got it

instance/models/mixins/openedx_database.py
saksham115: Oh yes! My bad.

saksham115:
@mtyaka  made the changes. Back to you.

mtyaka:
Looks good @saksham115!

Please deploy this to stage. The usual process looks like this:
1. Log into the screen session where Ocim is running, and stop it with `Ctrl+C`.
2. Run the migrations with `make migrate`.
3. Start Ocim back up with `make run`.

saksham115:
@mtyaka verified on staging. Working 🥳 
Screenshot from 2021-12-24 15-48-01

mtyaka:
Great, thanks! :+1:

- [x] I tested this: I verified that the migration ran successfully on stage and that database names were stored to the DB field.
- [x] I read through the code

mtyaka (approved):

PR #859, opened 2022-01-12
[SE-5070] fix: apply workaround for CloudFlare proxied domain check

gabor-boros:
## Description

This PR ensures that users registered with CloudFlare (and using DNS proxying) can still set custom domain names without assistance.

## Supporting information

Ocim is validating domain/subdomain setup by looking into the DNS records of the given domain. This probably works for most of the domain names (as we experience it), though in some cases (like CF proxied domains) it fails and will never work.

What are proxied domains? Basically a DoT (DNS over TLS) setup for the given DNS record (in this case a CNAME) that gives possibility for the end users to be saved from DDOS attacks as the attack would hit CF name servers instead of the server the record pointing to (a.k.a OpenCraft edX app server).

Couldn't the users simply turn of proxying? They could and the DNS resolution should start working almost immediately. Do we or the user want to disable it? Obviously no. This is good for us and them too – we are not hit directly in case of a DDOS.

### Dependencies

N/A

## Testing instructions

1. Register as a client
1. Go to setting custom domains on the client console
1. Try the domain mentioned in the related ticket

## Deadline

ASAP

### DIFF
--- a/registration/tests/test_domain_utils.py
+++ b/registration/tests/test_domain_utils.py
@@ -21,7 +21,7 @@ Test the domain verification utils
 """
 
 # Imports #####################################################################
-from unittest.mock import patch
+from unittest.mock import patch, call
 
 import dns.resolver
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


mtyaka (approved):
This looks good :+1:

- [x] I tested this: I didn't test myself since the patch is no longer on stage, but I watched Gabor test it, and the changes are trivial enough
- [x] I read through the code
- [ ] I checked for accessibility issues ~~N/A~~
- [x] Includes documentation (inline comments)

gabor-boros:
Thank you for the approve @mtyaka, I'm going to merge and deploy this.

PR #888, opened 2022-03-22
fix: [BB-6000]adds correct regex to configure CORS

navinkarkera:
**Description**
Setup correct value of NGINX_PROXY_ORIGIN_REGEX during instance spinup which is used by ansible-playbooks to configure CORS for nginx generated errors 

**Ticket link**: https://tasks.opencraft.com/browse/BB-6000

**Testing instructions**
Looking into the steps required to test the change on staging.

### DIFF
--- a/instance/models/mixins/openedx_config.py
+++ b/instance/models/mixins/openedx_config.py
@@ -153,7 +153,9 @@ class OpenEdXConfigMixin(ConfigMixinBase):
 
             # Nginx
             "NGINX_SET_X_FORWARDED_HEADERS": False,
-            "NGINX_PROXY_ORIGIN_REGEX": r'^(http|https)://.*\.{}'.format(self.instance.domain),
+            "NGINX_PROXY_ORIGIN_REGEX": r'^(http|https)://.*{}'.format(
+                str(self.instance.domain).replace(".", r"\.")
+            ),
 
             # SSL is handled on the load balancer, and the appservers are HTTP only.
             "NGINX_ENABLE_SSL": False,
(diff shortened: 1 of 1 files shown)


instance/models/mixins/openedx_config.py
```diff
@@ -153,6 +153,7 @@ def _get_configuration_variables(self):
 
             # Nginx
             "NGINX_SET_X_FORWARDED_HEADERS": False,
+            "NGINX_PROXY_ORIGIN_REGEX": r'^(http|https)://.*\.{}'.format(self.instance.domain),
```
navinkarkera, line 156: As per my understanding `self.instance.domain` will be set to base domain for eg. `navin.opencraft.hosting`.
So the regex will match any url starting with http/https and ending with the `navin.opencraft.hosting`.

instance/models/mixins/openedx_config.py
```diff
@@ -153,6 +153,9 @@ def _get_configuration_variables(self):
 
             # Nginx
             "NGINX_SET_X_FORWARDED_HEADERS": False,
+            "NGINX_PROXY_ORIGIN_REGEX": r'^(http|https)://.*{}'.format(
+                str(self.instance.domain).replace(".", r"\.")
+            ),
```
navinkarkera, line 158: As per my understanding self.instance.domain will be set to base domain for eg. navin.opencraft.hosting.
So the regex will match any url starting with http/https and ending with the navin.opencraft.hosting.

navinkarkera:
Closing PR as the config is not needed in openedx. ([explanation](https://tasks.opencraft.com/browse/BB-6000?focusedCommentId=243780&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-243780))

PR #880, opened 2022-03-03
[SE-5391] refactor: activate app servers before setting dns

gabor-boros:
## Description

This PR implements the monitoring changes introduced by NewRelic and causes provisioning issues.

## Testing instructions

* check out the branch on Ocim stage
* run `OpenEdXInstance.objects.get(id=965).enable_monitoring()` in shell
* validate the monitoring items are set correctly in NewRelic
* validate that no critical or warning messages shown

PR #889, opened 2022-03-25
[SE-5426] reconfigure NewRelic monitoring

gabor-boros:
## Description

The current monitoring is set for the legacy NewRelic infrastructure, though NewRelic improved its event aggregation and monitoring. As part of this patch, the aggregation method, window and delay has been changed in order to provide more reliable alerting and reduce the number of false positive alerts.

The default values of settings are updated, since we didn't set the desired behaviour previously. We wanted to be alerted if an issue exists for 10 minutes but calculate with some event delays, so the alert time was set to 11 minutes. At that time, we had no better option as the `aggregation_delay` and `event_flow` aggregation type did not exist. Now that we have these extra settings, we can be alerted if the issue persists for 10 minutes plus we wait for some time (20% of the `aggregation_window`) to receive missing data points. If the data point arrives, no alert will be triggered. Otherwise, we got alerted.

The 20% for delay may seems too much, but since the **maximum** aggregation window allowed by NewRelic is 15 minutes, the maximum delay is 3 minutes. That's not that bad, especially that load balancers may need some time to update their configuration within the current infrastructure. We are using 10 minutes with 20% delay, so if we have no data points for 12 minutes, NewRelic will alert us.

## Supporting information

* [Available event aggregation methods](https://docs.newrelic.com/whats-new/2021/10/wn-new-aggregation-methods)
* [Example configurations for NRQL alerting](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_alert_condition#additional-examples)
* [Related PR that influenced this change](https://github.com/open-craft/opencraft/pull/880)
* [Mattermost discussion about the issue](https://chat.opencraft.com/opencraft/pl/ajm4qbn9xpf6by39k9i6en6qcc)

### Dependencies

N/A

## Sandbox

No alerts were updated yet to use the new alert conditions, as I wanted to have a review first.

## Testing instructions

1. First of all, read through ALL the linked resources
2. Validate that this PR description meets the code changes
3. Deploy to Ocim stage, disable monitoring for one (test) server using `disable_monitoring` function, then re-enable it using `enable_monitoring` -- this will delete and re-register NRQL alert policies
4. Check the NewRelic UI that the configuration is created as expected

## Deadline

As soon as possible, since we are getting false alerts.

## Other information

We won't need to update any playbooks as Ocim is using the default values, set in `settings.py` for NewRelic alert configuration. That's true for CI/CD as well.

### DIFF
--- a/instance/newrelic.py
+++ b/instance/newrelic.py
@@ -195,15 +195,26 @@ def add_alert_nrql_condition(policy_id, monitor_url, name):
                     'query': query,
                 },
                 'signal': {
-                    'aggregation_window': f"{settings.NEWRELIC_NRQL_ALERT_SIGNAL_EXPIRATION}",
-                    'aggregation_method': 'CADENCE',
-                    'aggregation_delay': 120,
+                    #
+                    'aggregation_window': settings.NEWRELIC_NRQL_SIGNAL_AGGREGATION_WINDOW,
+                    # Event Flow aggregation is best designed to be used for situations when data come in frequently
+                    # and with low event spread, or mostly "in-order". The original "cadence" aggregation method used
+                    # to trigger false positive alerts.
+                    'aggregation_method': 'EVENT_FLOW',
+                    # How long we wait for data that belongs in each aggregation window. A longer delay may increase
+                    # accuracy but delay notifications. Here we set 20% for the delay in seconds of the aggregation
+                    # window. This means if the window is set to 10 minutes, we will wait an additional 2 minute before
+                    # we send an alert.
+                    'aggregation_delay': int(settings.NEWRELIC_NRQL_SIGNAL_AGGREGATION_WINDOW * 0.2),
                     'fill_option': 'static',
                     'fill_value': '0.0',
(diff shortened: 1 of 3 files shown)


mtyaka (approved):
@gabor-boros I must say I find it quite hard to understand the New Relic documentation around event aggregation, but your reasoning looks good to me. Also the inline comments help, thanks a lot for those.

It's hard to be 100% if this will fix the race condition, but it's worth giving it a shot.

:+1:

- [x] I tested this: I did not test anything yet. We'll first deploy this to stage and see how it works.
- [x] I read through the code
- [x] Includes documentation (inline comments)

gabor-boros:
@mtyaka 

> I find it quite hard to understand the New Relic documentation around event aggregation

Honestly, I find their whole documentation hard to understand.

FTR: The changes are already tested by be on stage. May I go ahead and deploy to production?

mtyaka:
> May I go ahead and deploy to production?

Let's do it 🚀

PR #892, opened 2022-03-30
fix: [BB-6078] docker development environment

navinkarkera:
### Description
This PR fixes docker development environment.

### Related tickets:
- [BB-6078](https://tasks.opencraft.com/browse/BB-6078)
- [BB-6079](https://tasks.opencraft.com/browse/BB-6079)

### Testing instructions
- Clone repository locally
- Copy .env.test -> .env
- Follow [instructions](https://ocim.opencraft.com/en/latest/development/docker/) to setup local docker setup
- Run `docker-compose run --rm ocim make test.unit`


docker-compose.yml
```diff
@@ -16,8 +16,22 @@ services:
...
+  mongodb:
+    image: mongo:5.0.6
+    ports:
+      - "27017:27017"
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: opencraft
+      MONGO_INITDB_ROOT_PASSWORD: opencraft
+
+  mysqldb:
+    image: mysql:8.0.28-debian
```
0x29a, line 28: We're still using MySQL 5.6 on the production. Shouldn't we use the same version here?

0x29a (changes requested):

0x29a:
@navinkarkera, one more note. Since you're removing the Vagrant-related docs, do you think we should remove `Vagrantfile` as well? From the original ticket:
```
Completion criteria
    * Vagrant related files are removed
```

navinkarkera:
> @navinkarkera, one more note. Since you're removing the Vagrant-related docs, do you think we should remove `Vagrantfile` as well? From the original ticket:
> 
> ```
> Completion criteria
>     * Vagrant related files are removed
> ```

It is removed in this [commit](https://github.com/open-craft/opencraft/pull/892/commits/25d5b382634944efdd54dd545c6c9c213bb71996).

docker-compose.yml
navinkarkera: Updated to use 5.6

0x29a (approved):
:+1:

- [x] I followed the instructions and tested that a docker-based setup is working.
- [x] I read through the changes.
- [x] Includes documentation.

PR #904, opened 2022-05-28
[SE-5560] Fix smaller issues

gabor-boros:
## Description

This PR fixes some of the issues emerged in production. The angular module cannot use empty brackets at modul init, otherwise it will not load the page.

Squasing migrations was required because of a faulty migration (0022) accidentally ended up in master branch. The staging server had no issues with this migration as the migration was partially applied manually during a debugging session. Production server on the other hand had no migrations like that and couldn't run the migration successfully. Squashing the migrations eliminates the issue though, since the not squashed migrations are replaced at runtime by django.

Since the squashed migraiton is already applied to bring the server back online, we cannot rename it (at least easily).

## Supporting information

N/A

### Dependencies

N/A

## Testing instructions

* Proofread the changes

## Deadline

ASAP


0x29a (approved):
:+1:

- [x] I read the PR description.
- [x] I read the diff and ensured that it's matching the description.

